PR #5, opened 2020-09-26
Add lint-test Github action

dmitryax:


bogdandrutu (approved):

ct.yaml
```diff
@@ -0,0 +1,7 @@
+# See https://github.com/helm/chart-testing#configuration
+remote: origin
+chart-dirs:
+  - charts
+chart-repos:
+  - bitnami=https://charts.bitnami.com/bitnami
```
naseemkullah, line 6: I believe this can be removed as we do not need dependency charts from bitnami

dmitryax: Thanks for catching it. Submitted another PR https://github.com/open-telemetry/opentelemetry-helm-charts/pull/9

Issue #14, opened 2020-10-07
Q: What does config look like to export traces to two backends?


naseemkullah:
Maybe a snippet to export to jaeger + zipkin at the same time would be nice to have in README.md as a reference.

naseemkullah:
In particular I would have expected this to be a list https://github.com/open-telemetry/opentelemetry-helm-charts/blob/8d5bcf0f7e30a7d6d187f149ff7527735263c5f4/charts/opentelemetry-collector/values.yaml#L19 if multiple exporters are supported, @dmitryax do you know if its possible to have multiple exporters configured via this chart?

dmitryax:
@naseemkullah yes, multiple exporters can be set in `agent(standalone)Collector.configOverride`. anything set there will be merged to the default config. For example to have two exporters, we'd need:
```
configOverride:
  exporters:
    jaeger:
      ...
    zipkin:
      ...
  service:
    pipelines:
      traces:
        exporters: [zipkin, jaeger]
```

I'll add a few examples

naseemkullah:
Great, thanks!

PR #15, opened 2020-10-08
Add more configuration examples

dmitryax:
+ fix a zipkin port in examples

Issue: https://github.com/open-telemetry/opentelemetry-helm-charts/issues/14

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.1.0
+version: 0.1.1
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-collector/README.md
+++ b/charts/opentelemetry-collector/README.md
@@ -29,7 +29,7 @@ By default this chart will deploy an OpenTelemetry Collector as daemonset with t
 and logging exporter enabled by default. Besides daemonset (agent), it can be also installed as standalone deployment.
 Both modes can be enabled together, in that case metrics and traces will be flowing from agents to standalone collectors.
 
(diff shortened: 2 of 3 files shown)


naseemkullah (approved):

naseemkullah:
lgtm @dmitryax but one thing I still don't fully understand, why is it in the values.yaml file exporter is singular? Could we not instead have it be `exporters` and take a list of exporters?

On that note, perhaps having the `telemetry` value is a leaky abstraction? Can we just put some uncommented logging exporters  in `configOverride` ?

dmitryax:
@naseemkullah I was thinking that the `telemetry` abstraction can cover most of the cases. 

It also provide an easy way to disable telemetry like 
```
telemetry:
  metrics:
    enabled: false
```

But it must be possible to do with
```
configOverride:
  service:
    pipelines:
      metrics: null
```
I'd need to verify it ^

So you may be right , it looks like unnecessary abstraction. We can remove it in `0.2.0` release. It's still early days of the chart.

naseemkullah:
Sounds good @dmitryax, thanks!

naseemkullah:
FWIW I'd hold off on these README example and remove the abstraction, if possible, and always encourage use of configOverride, i.e. I'd add examples along with `0.2.0`

tigrannajaryan:
@naseemkullah Can I merge this or there are still concerns?

naseemkullah:
We may want to go straight with `0.2.0` and only use `configOverride` and rid the chart of the potentially unnecessary abstraction value `telemetry`... @dmitryax does that make sense? I can open a PR for that if so.

dmitryax:
@naseemkullah I agree, that makes sense. I'm OOO these days. please feel free to open the PR if you have a chance. otherwise I can do it probably next week

naseemkullah (changes requested):
Will work on a `0.2.0` release, at which point the examples can be adapted to use the `configOverride` value exclusively

PR #17, opened 2020-10-21
use conf override in favor of telemetry abstraction

naseemkullah:
Replaces previous `.Values.telemetry` abstraction with a top level `configOverride` to configure the collector.


charts/opentelemetry-collector/templates/config/_agent-config-override.tpl
```diff
@@ -11,13 +11,17 @@ exporters:
...
-    {{- if and .Values.telemetry.metrics.enabled .Values.standaloneCollector.enabled }}
+    {{- if .Values.standaloneCollector.enabled }}
+
+    {{- if "opentelemetry-collector.metrics.enabled" }}
     metrics:
       exporters: [otlp]
     {{- end }}
 
-    {{- if and .Values.telemetry.traces.enabled .Values.standaloneCollector.enabled }}
+    {{- if "opentelemetry-collector.traces.enabled" }}
```
dmitryax, line 21: I think it should be possible to avoid `traces/metrics.enabled` option completely. If user wants to disable traces or metrics pipeline, it should be possible to set `configOverride.service.pipelines.metrics: null` and that should disable the pipeline after merging into default config.

naseemkullah: This is indeed an alias for if it is defined in root,agent or standalone confit override, does that make sense?

charts/opentelemetry-collector/templates/_helpers.tpl
```diff
@@ -62,16 +62,30 @@ Create the name of the service account to use
...
-    {{- $config := include (printf "opentelemetry-collector.%sPipelineConfig" $telemetryName) $ | fromYaml | mustMergeOverwrite $config }}
-  {{- end }}
+{{- define "opentelemetry-collector.metrics.enabled" -}}
+{{- or .Values.configOverride.service.pipelines.metrics  .Values.agentCollectorconfigOverride.service.pipelines.metrics .Values.standaloneCollector.configOverride.service.pipelines.metrics }}
 {{- end }}
-{{- $config | toYaml }}
+
+{{/*
+*/}}
+{{- define "opentelemetry-collector.traces.enabled" -}}
```
naseemkullah, line 72: @dmitryax see here

charts/opentelemetry-collector/templates/_helpers.tpl
```diff
@@ -62,16 +62,30 @@ Create the name of the service account to use
...
+{{- or .Values.configOverride.service.pipelines.traces  .Values.agentCollectorconfigOverride.service.pipelines.traces .Values.standaloneCollector.configOverride.service.pipelines.traces }}
```
dmitryax, line 73: `agentCollectorconfigOverride` -> `agentCollector.configOverride`

charts/opentelemetry-collector/templates/_helpers.tpl
```diff
@@ -62,16 +62,30 @@ Create the name of the service account to use
...
+{{- end }}
+
+{{/*
+Build base config with top level override
+*/}}
+{{- define "opentelemetry-collector.overridenConfig" -}}
+{{- $source := .Values.configOverride  -}}
+{{- $config := include "opentelemetry-collector.baseConfig" . | fromYaml  -}}
+{{- if "opentelemetry-collector.metrics.enabled" }}
+{{- $config := include (printf "opentelemetry-collector.metricsPipelineConfig" ) $ | fromYaml | mustMergeOverwrite $config }}
```
dmitryax, line 83: `printf` not needed anymore

charts/opentelemetry-collector/templates/_helpers.tpl
```diff
@@ -62,16 +62,30 @@ Create the name of the service account to use
...
+{{- end }}
+{{- if "opentelemetry-collector.traces.enabled" }}
+{{- $config := include (printf "opentelemetry-collector.tracesPipelineConfig" ) $ | fromYaml | mustMergeOverwrite $config }}
```
dmitryax, line 86: same here

charts/opentelemetry-collector/templates/_helpers.tpl
```diff
@@ -62,16 +62,30 @@ Create the name of the service account to use
...
+{{- end }}
+{{- $source | mergeOverwrite $config  | toYaml }}
```
dmitryax, line 88: seems like `$source` defined only once, why don't set it right here?

also can we use `mustMergeOverwrite` to catch errors and for consistency?

charts/opentelemetry-collector/templates/config/_agent-config-override.tpl
dmitryax: I was thinking that we could avoid having this alias and conditions like `if "opentelemetry-collector.traces.enabled"` altogether.

According to the docs https://helm.sh/docs/chart_template_guide/values_files/#deleting-a-default-key, if a dict key is set to `null` in user's values.yaml, then that key should be removed as a result. That would work perfectly for disabling a default pipeline by setting `configOverride.service.pipelines.metrics/traces: null` even if some definitions of unused metrics/traces related components will stay in exporters/receivers section (they will not be initialized anyway). I haven't test it yet tho. 

I can try that as the next step after this PR is merged. And for now we can keep the existing approach.

naseemkullah:
Thanks for the feedback @dmitryax 

I've been thinking, while the baseConfig generation is clever, it might be nice and simple to have, within the values file, a top level `.Values.baseConfig` laid out in its entirety, no templating tricks, and then `.Values.(agent/standalone)Collector.configOverride` values. WDYT?

dmitryax:
Having the templates for yaml configs, makes it clean and easy to introduce tricks like default memory_limiter processor configuration based on provided k8s resources memory limits. With `.Values.baseConfig`, it would be not that straightforward, I tried that approach, but I'm open to suggestions if you can manage to do it in a cleaner way.

charts/opentelemetry-collector/templates/config/_agent-config-override.tpl
naseemkullah: OK thanks for explaining. I've adjusted accordingly, PTAL, if it's good I will update existing README examples.

charts/opentelemetry-collector/templates/config/_agent-config-override.tpl
```diff
@@ -11,12 +11,9 @@ exporters:
 
 service:
   pipelines:
-    {{- if and .Values.telemetry.metrics.enabled .Values.standaloneCollector.enabled }}
+    {{- if .Values.standaloneCollector.enabled }}
```
naseemkullah, line 14:
```suggestion
    {{- if .Values.agentCollector.enabled }}
```

Should that be this?

dmitryax: yes, sounds good

dmitryax (commented):
Looks good. Thanks!

Should we update the examples in README along with this change?

naseemkullah:
> Looks good. Thanks!
> 
> Should we update the examples in README along with this change?

Yep, done! PTAL

charts/opentelemetry-collector/templates/config/_agent-config-override.tpl
```diff
@@ -3,20 +3,17 @@ Default config override for agent collector deamonset
 */}}
 {{- define "opentelemetry-collector.agentConfigOverride" -}}
 exporters:
-  {{- if .Values.standaloneCollector.enabled }}
+  {{- if .Values.agentCollector.enabled }}
```
dmitryax, line 6: this should be `.Values.standaloneCollector.enabled`. if a standalone collector is enabled, we send the telemetry to it.

charts/opentelemetry-collector/templates/config/_agent-config-override.tpl
```diff
@@ -3,20 +3,17 @@ Default config override for agent collector deamonset
...
   otlp:
     endpoint: {{ include "opentelemetry-collector.fullname" . }}:55680
     insecure: true
   {{- end }}
 
 service:
   pipelines:
-    {{- if and .Values.telemetry.metrics.enabled .Values.standaloneCollector.enabled }}
+    {{- if .Values.agentCollector.enabled }}
```
dmitryax, line 14: same here `.Values.standaloneCollector.enabled`

charts/opentelemetry-collector/README.md
```diff
@@ -46,15 +48,18 @@ There are two ways to configure collector pipelines, which can be used together
...
-      type: zipkin
-      config:
-        endpoint: zipkin-all-in-one:14250
+configOverride:
+  exporters:
+    zipkin:
+      endpoint: zipkin-all-in-one:14250
+  service:
+    pipelines:
+      metrics: null
```
dmitryax, line 59: I've just tried this example, and it appeared to be not working with `mustMergeOverwrite`. `metrics` pipeline still has `receivers` and and `processors`, and collector cannot start. 

I'll think how to resolve it later, for now you can bring back the approach you had before.

dmitryax (commented):
Sorry for the delay

naseemkullah:
Thanks @dmitryax , PTAL

charts/opentelemetry-collector/templates/config/_metrics.tpl
```diff
@@ -1,3 +1,10 @@
+{{/*
+Returns true if any config override defines a metrics pipeline.
+*/}}
+{{- define "opentelemetry-collector.metrics.enabled" -}}
+{{- or .Values.configOverride.service.pipelines.metrics  .Values.agentCollector.configOverride.service.pipelines.metrics .Values.standaloneCollector.configOverride.service.pipelines.metrics }}
```
dmitryax, line 5: This would require all of the `configOverride` sources to be defined up to `metrics` key, otherwise it causes `nil pointer evaluating interface {}.pipelines`.

We could use `.Values.configOverride.service.pipelines.metrics` only to disable a pipeline.

Another option is to have all the parts of the pipelines (receivers+processors+exporters, not only `exporters`) pre-defined in `values.yaml` ->`.Values.configOverride` instead of `_metrics.tpl`. In that case it will be fully overridable, and a pipeline can be easily disabled that way.

charts/opentelemetry-collector/templates/_helpers.tpl
```diff
@@ -62,16 +62,17 @@ Create the name of the service account to use
...
 */}}
-{{- define "opentelemetry-collector.pipelinesConfig" -}}
-{{- $config := include "opentelemetry-collector.baseConfig" . | fromYaml -}}
-{{- range $telemetryName, $telemetryParams := .Values.telemetry }}
-  {{- if index $telemetryParams "enabled" }}
-    {{- $config := include (printf "opentelemetry-collector.%sPipelineConfig" $telemetryName) $ | fromYaml | mustMergeOverwrite $config }}
-  {{- end }}
+{{- define "opentelemetry-collector.overridenConfig" -}}
+{{- $config := include "opentelemetry-collector.baseConfig" . | fromYaml  -}}
+{{- if not "opentelemetry-collector.metrics.enabled" }}
```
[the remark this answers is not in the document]
naseemkullah: Ugh sorry

charts/opentelemetry-collector/templates/config/_metrics.tpl
naseemkullah: Good point and I'll try the latter approach of moving more to the values file

naseemkullah: Alright I've moved the metrics and traces bases straight into values, and have renamed configOverride to config, although it is technically an override, what it overrides is pretty minimal, WDYT?

naseemkullah: And I just converged all config related helpers into `_config.tpl` if that's alright?

naseemkullah: resulting out of the box config is

```yaml
data:
  relay: |
    exporters:
      logging: {}
    extensions:
      health_check: {}
    processors:
      batch: {}
      memory_limiter:
        ballast_size_mib: 204
        check_interval: 5s
        limit_mib: 409
        spike_limit_mib: 128
      queued_retry: {}
    receivers:
      jaeger:
        protocols:
          grpc:
            endpoint: 0.0.0.0:14250
          thrift_http:
            endpoint: 0.0.0.0:14268
      otlp:
        protocols:
          grpc: null
          http: null
      prometheus:
        config:
          scrape_configs:
          - job_name: opentelemetry-collector
            scrape_interval: 10s
            static_configs:
            - targets:
              - ${MY_POD_IP}:8888
      zipkin:
        endpoint: 0.0.0.0:9411
    service:
      extensions:
      - health_check
      pipelines:
        metrics:
          exporters:
          - logging
          processors:
          - memory_limiter
          - batch
          - queued_retry]
          receivers:
          - otlp
          - prometheus
        traces:
          exporters:
          - logging
          processors:
          - memory_limiter
          - batch
          - queued_retry
          receivers:
          - otlp
          - jaeger
          - zipkin
```


And setting metrics and traces to null yields:

```yaml
    exporters:
      logging: {}
    extensions:
      health_check: {}
    processors:
      batch: {}
      memory_limiter:
        ballast_size_mib: 204
        check_interval: 5s
        limit_mib: 409
        spike_limit_mib: 128
      queued_retry: {}
    receivers:
      jaeger:
        protocols:
          grpc:
            endpoint: 0.0.0.0:14250
          thrift_http:
            endpoint: 0.0.0.0:14268
      otlp:
        protocols:
          grpc: null
          http: null
      prometheus:
        config:
          scrape_configs:
          - job_name: opentelemetry-collector
            scrape_interval: 10s
            static_configs:
            - targets:
              - ${MY_POD_IP}:8888
      zipkin:
        endpoint: 0.0.0.0:9411
    service:
      extensions:
      - health_check
      pipelines:
        metrics: null
        traces: null
```

charts/opentelemetry-collector/values.yaml
```diff
@@ -5,26 +5,49 @@
...
+  service:
+    pipelines:
+      metrics:
+        receivers:
+          - otlp
+          - prometheus
+        processors:
+          - memory_limiter
+          - batch
+          - queued_retry]
```
dmitryax, line 37: `]` typo

charts/opentelemetry-collector/templates/config/_metrics.tpl
dmitryax: > And setting metrics and traces to null yields:

interesting. disabling a pipeline with
```
config:
  service:
    pipelines:
      metrics: null
```
should remove the key from pipelines, not setting it to `null`. That's exactly what I got when testing it on helm 3.4.0

dmitryax: > Alright I've moved the metrics and traces bases straight into values, and have renamed configOverride to config, although it is technically an override, what it overrides is pretty minimal, WDYT?

`config` sounds good to me

charts/opentelemetry-collector/values.yaml
naseemkullah: thanks

charts/opentelemetry-collector/templates/config/_metrics.tpl
naseemkullah: Yeah, that sounds like a good idea.

dmitryax (approved):
Looks good to me. Just one nit. Thanks for rewriting it!

charts/opentelemetry-collector/README.md
```diff
@@ -46,15 +48,18 @@ There are two ways to configure collector pipelines, which can be used together
 ### Basic top level configuration with `telemetry` property
```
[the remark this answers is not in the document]
naseemkullah: > Looks good to me. Just one nit. Thanks for rewriting it!

Fixed! Thanks for bouncing back ideas with me. I like how this turned out!

naseemkullah:
> Looks good to me. Just one nit. Thanks for rewriting it!

My pleasure and thanks for bouncing back ideas, I like how it turned out!

ferrarimarco:
Thanks for this PR! I had to dig a bit to understand how configuration worked in the previous format. Now it's much more clearer 👍

naseemkullah:
Great to hear @ferrarimarco!

PR #20, opened 2020-11-20
change service annotation template to support multiple annotations

rabunkosar-dd:
There were some issues with multiple annotations, this tries to change the template to support them.

### DIFF
--- a/charts/opentelemetry-collector/templates/service.yaml
+++ b/charts/opentelemetry-collector/templates/service.yaml
@@ -7,7 +7,7 @@ metadata:
     {{- include "opentelemetry-collector.labels" . | nindent 4 }}
   {{- if .Values.service.annotations }}
   annotations:
-{{ toYaml .Values.service.annotations | trim | indent 4 }}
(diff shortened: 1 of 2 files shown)


charts/opentelemetry-collector/templates/service.yaml
```diff
@@ -5,9 +5,9 @@ metadata:
   name: {{ include "opentelemetry-collector.fullname" . }}
   labels:
     {{- include "opentelemetry-collector.labels" . | nindent 4 }}
-  {{- with .Values.service.annotations }}
+  {{- if .Values.service.annotations }}
   annotations:
-    {{ toYaml . | trim | indent 4 }}
+{{ toYaml .Values.service.annotations | trim | indent 4 }}
```
naseemkullah, line 10:
```suggestion
    {{ toYaml .Values.service.annotations | nindent 4 }}
```

nit: little more readable imho

naseemkullah (approved):
lgtm, left a nit

PR #24, opened 2021-01-21
Use iana port

rabunkosar-dd:
Breaking change to default to 4317 for otlp (https://github.com/open-telemetry/opentelemetry-specification/issues/1148). Not a good one but it's better to make the change early.


naseemkullah (approved):
Thanks! Lgtm

naseemkullah:
@open-telemetry/helm-maintainers PTAL 🙏

PR #32, opened 2021-02-10
fix: agent collector using the legacy gRPC port

kobim:
When enabling the standalone collector, the chart wires the agent's OTLP exporter to the standalone collector at the [legacy port](https://github.com/open-telemetry/opentelemetry-collector/blob/88ab1a6fd923010b1397edb38fbd01375531e3bb/receiver/otlpreceiver/otlp.go#L129-L140).

It didn't work because the `service` of the standalone collector [only listens](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/3057edf2d973b0268ec82ffe92857a6b9926f8bb/charts/opentelemetry-collector/values.yaml#L101-L106) on the correct port (4317).

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.4.1
+version: 0.4.2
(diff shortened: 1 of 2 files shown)


tigrannajaryan:
@naseemkullah can you review please?

naseemkullah (approved):
Lgtm, thanks!

PR #36, opened 2021-03-11
Add k8s container logs collection

astencel-sumo:
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/2536

### DIFF
--- a/charts/opentelemetry-collector/README.md
+++ b/charts/opentelemetry-collector/README.md
@@ -141,7 +141,10 @@ The container logs pipeline uses the `logging` console exporter by default.
 Paired with the default `filelog` receiver that receives all containers' console output,
 it is easy to accidentally feed the exported logs back into the receiver.
 
-To prevent this, the default configuration of the receiver excludes logs from the collector's containers.
+Also note that using the `--log-level=debug` option for the `logging` exporter causes it to output
+multiple lines per single received log, which when looped, would amplify the logs exponentially.
+
+To prevent the looping, the default configuration of the receiver excludes logs from the collector's containers.
 
 If you want to include the collector's logs, make sure to replace the `logging` exporter
 with an exporter that does not send logs to collector's standard output.
--- a/charts/opentelemetry-collector/README.md
+++ b/charts/opentelemetry-collector/README.md
@@ -42,8 +42,7 @@ By default collector has the following receivers enabled:
 
 - **metrics**: OTLP and prometheus. Prometheus is configured only for scraping collector's own metrics.
 - **traces**: OTLP, zipkin and jaeger (thrift and grpc).
-- **logs**: OTLP and logs from Kubernetes containers (see [Configuration for Kubernetes container logs](#configuration-for-kubernetes-container-logs)
-  below).
+- **logs**: OTLP (to enable container logs, see [Configuration for Kubernetes container logs](#configuration-for-kubernetes-container-logs)).
 
 There are two ways to configure collector pipelines, which can be used together as well.
 
@@ -111,27 +110,55 @@ agentCollector:
 
 ### Configuration for Kubernetes container logs
 
-The collector is preconfigured to receive logs sent to standard output by Kubernetes containers.
+The collector can be used to collect logs sent to standard output by Kubernetes containers.
+This feature is disabled by default. It has the following requirements:
 
-This feature needs agent collector to be deployed, which means it will not work if only standalone collector is enabled.
+- It needs agent collector to be deployed, which means it will not work if only standalone collector is enabled.
+- It requires the [contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib) version
+of the collector image.
 
-The feature can be disabled by setting the `agentCollector.containerLogs.enabled` property to `false`.
-
-The preconfigured container logs pipeline uses `logging` console exporter by default.
-To prevent looping the logs exported from collector back into the logs receiver,
-the default configuration of the receiver excludes logs from the collector containers.
-To change that, replace the exporter on the logs pipeline and remove the exclude on the logs receiver.
-
-Here's an example `values.yaml` file that replaces the default `logging` exporter on the `logs` pipeline
-with a `file` exporter that writes the container logs to `/var/log/container-logs/container-logs.json`.
-It also sets the `exclude` property of `filelog` receiver to empty array, for collector logs not to be excluded.
+To enable this feature, set the  `agentCollector.containerLogs.enabled` property to `true` and replace the collector image.
+Here is an example `values.yaml`:
 
 ```yaml
 agentCollector:
+  containerLogs:
+    enabled: true
+
+image:
+  repository: otel/opentelemetry-collector-contrib
+
+command:
+  name: otelcontribcol
+```
+
+The way this feature works is it adds a `filelog` receiver on the `logs` pipeline. This receiver is preconfigured
+to read the files where Kubernetes container runtime writes all containers' console output to.
+
+#### :warning: Warning: Risk of looping the exported logs back into the receiver, causing "log explosion"
+
+The container logs pipeline uses the `logging` console exporter by default.
+Paired with the default `filelog` receiver that receives all containers' console output,
+it is easy to accidentally feed the exported logs back into the receiver.
+
+To prevent this, the default configuration of the receiver excludes logs from the collector's containers.
+
+If you want to include the collector's logs, make sure to replace the `logging` exporter
+with an exporter that does not send logs to collector's standard output.
+
+Here's an example `values.yaml` file that replaces the default `logging` exporter on the `logs` pipeline
+with an `otlphttp` exporter that sends the container logs to `https://example.com:55681` endpoint.
+It also clears the `filelog` receiver's `exclude` property, for collector logs to be included in the pipeline.
+
+```yaml
+agentCollector:
+  containerLogs:
+    enabled: true
+
   configOverride:
     exporters:
-      file:
-        path: /var/log/container-logs/container-logs.json
+      otlphttp:
+        endpoint: https://example.com:55681
     receivers:
       filelog:
         exclude: []
@@ -139,12 +166,13 @@ agentCollector:
       pipelines:
         logs:
           exporters:
-            - file
-  extraHostPathMounts:
-    - name: varlogotclogs
-      hostPath: /var/log/container-logs
-      mountPath: /var/log/container-logs
-      readOnly: false
+            - otlphttp
+
+image:
+  repository: otel/opentelemetry-collector-contrib
+
+command:
+  name: otelcontribcol
 ```
 
 ### Other configuration options
--- a/charts/opentelemetry-collector/README.md
+++ b/charts/opentelemetry-collector/README.md
@@ -25,9 +25,9 @@ helm install my-opentelemetry-collector open-telemetry/opentelemetry-collector
 
 ### Default configuration
 
-By default this chart will deploy an OpenTelemetry Collector as daemonset with two pipelines (traces and metrics)
+By default this chart will deploy an OpenTelemetry Collector as daemonset with three pipelines (logs, metrics and traces)
 and logging exporter enabled by default. Besides daemonset (agent), it can be also installed as standalone deployment.
-Both modes can be enabled together, in that case metrics and traces will be flowing from agents to standalone collectors.
+Both modes can be enabled together, in that case logs, metrics and traces will be flowing from agents to standalone collectors.
 
 *Example*: Install collector as a standalone deployment, and do not run it as an agent.
 
@@ -42,6 +42,8 @@ By default collector has the following receivers enabled:
 
 - **metrics**: OTLP and prometheus. Prometheus is configured only for scraping collector's own metrics.
 - **traces**: OTLP, zipkin and jaeger (thrift and grpc).
+- **logs**: OTLP and logs from Kubernetes containers (see [Configuration for Kubernetes container logs](#configuration-for-kubernetes-container-logs)
+  below).
 
 There are two ways to configure collector pipelines, which can be used together as well.
 
@@ -105,6 +107,44 @@ agentCollector:
     mountPath: /hostfs
     readOnly: true
     mountPropagation: HostToContainer
+```
+
+### Configuration for Kubernetes container logs
+
+The collector is preconfigured to receive logs sent to standard output by Kubernetes containers.
+
+This feature needs agent collector to be deployed, which means it will not work if only standalone collector is enabled.
+
+The feature can be disabled by setting the `agentCollector.containerLogs.enabled` property to `false`.
+
+The preconfigured container logs pipeline uses `logging` console exporter by default.
+To prevent looping the logs exported from collector back into the logs receiver,
+the default configuration of the receiver excludes logs from the collector containers.
+To change that, replace the exporter on the logs pipeline and remove the exclude on the logs receiver.
+
+Here's an example `values.yaml` file that replaces the default `logging` exporter on the `logs` pipeline
+with a `file` exporter that writes the container logs to `/var/log/container-logs/container-logs.json`.
+It also sets the `exclude` property of `filelog` receiver to empty array, for collector logs not to be excluded.
+
+```yaml
+agentCollector:
+  configOverride:
+    exporters:
+      file:
+        path: /var/log/container-logs/container-logs.json
+    receivers:
+      filelog:
+        exclude: []
+    service:
+      pipelines:
+        logs:
+          exporters:
+            - file
+  extraHostPathMounts:
+    - name: varlogotclogs
+      hostPath: /var/log/container-logs
+      mountPath: /var/log/container-logs
+      readOnly: false
 ```
 
 ### Other configuration options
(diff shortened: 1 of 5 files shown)


charts/opentelemetry-collector/README.md
```diff
@@ -107,6 +109,44 @@ agentCollector:
     mountPropagation: HostToContainer
 ```
 
+### Configuration for Kubernetes container logs
+
+The collector is preconfigured to receive logs sent to standard output by Kubernetes containers.
+
+This feature needs agent collector to be deployed, which means it will not work if only standalone collector is enabled.
+
+The feature can be disabled by setting the `agentCollector.containerLogs.enabled` property to `false`.
```
naseemkullah, line 118: For metrics and traces pipelines, we set them to null to disable, would it be possible to keep this pattern?

```yaml
config:
  service:
    pipelines:
      logs: null
```

astencel-sumo: It seems to me these are two different things: disabling the whole logs pipeline vs. disabling container logs collection. You might want to disable container logs collection but keep the logs pipeline for other uses?

astencel-sumo: Unless the container logs collection should be in its own pipeline? As opposed to a general "logs" pipeline.

naseemkullah: a `logs` pipeline does seem like the right thing

charts/opentelemetry-collector/values.yaml
```diff
@@ -63,17 +71,17 @@ config:
 # Can be overridden here or for any component independently using the same keys.
 
 image:
-  repository: otel/opentelemetry-collector
-  # repository: otel/opentelemetry-collector-contrib
+  # repository: otel/opentelemetry-collector
+  repository: otel/opentelemetry-collector-contrib
```
naseemkullah, line 75: will logs eventually land in core? / is the contrib image a better default?

charts/opentelemetry-collector/values.yaml
```diff
@@ -39,6 +39,14 @@ config:
     extensions:
       - health_check
     pipelines:
+      logs:
+        exporters:
+          logging
```
naseemkullah, line 44:
```suggestion
          - logging
```

charts/opentelemetry-collector/values.yaml
sumo-drosiek: filelog receiver which is going to be used for scrapping logs from kubernetes env is under development in contrib. I think that it is not going to be included in core soon if anytime

naseemkullah: I see, thanks

naseemkullah:
PTAL @dmitryax

charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -115,9 +116,89 @@ exporters:
...
       exporters: [otlp]
 {{- end }}
 {{- end }}
+
+{{- define "opentelemetry-collector.agent.containerLogsConfig" -}}
+{{- if .Values.agentCollector.containerLogs.enabled }}
+receivers:
+  filelog:
+    include: [ /var/log/pods/*/*/*.log ]
+    exclude: [ /var/log/pods/{{ .Release.Namespace }}_{{ include "opentelemetry-collector.fullname" . }}*_*/{{ .Chart.Name }}/*.log ]
```
tigrannajaryan, line 133: Please add a comment to explain what this does. I believe we need to make this warning very clear since it can be dangerous to collect own logs.

charts/opentelemetry-collector/values.yaml
```diff
@@ -63,17 +71,17 @@ config:
 # Can be overridden here or for any component independently using the same keys.
 
 image:
-  repository: otel/opentelemetry-collector
-  # repository: otel/opentelemetry-collector-contrib
+  # repository: otel/opentelemetry-collector
```
tigrannajaryan, line 74: Remove commented line.

charts/opentelemetry-collector/values.yaml
tigrannajaryan: @dmitryax any thoughts on what the implications of these change could be? Contrib is a much larger image, do we care about it?

charts/opentelemetry-collector/values.yaml
```diff
@@ -63,17 +71,17 @@ config:
...
   pullPolicy: IfNotPresent
   # Overrides the image tag whose default is the chart appVersion.
   tag: ""
 imagePullSecrets: []
 
 # OpenTelemetry Collector executable
 command:
-  name: otelcol
-  # name: otelcontribcol
+  # name: otelcol
```
tigrannajaryan, line 83: Remove commented line.

astencel-sumo: Done

charts/opentelemetry-collector/values.yaml
astencel-sumo: Done

charts/opentelemetry-collector/README.md
```diff
@@ -107,6 +109,44 @@ agentCollector:
...
+
+The preconfigured container logs pipeline uses `logging` console exporter by default.
+To prevent looping the logs exported from collector back into the logs receiver,
+the default configuration of the receiver excludes logs from the collector containers.
+To change that, replace the exporter on the logs pipeline and remove the exclude on the logs receiver.
+
+Here's an example `values.yaml` file that replaces the default `logging` exporter on the `logs` pipeline
+with a `file` exporter that writes the container logs to `/var/log/container-logs/container-logs.json`.
```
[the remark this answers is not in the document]
astencel-sumo: > I believe using file exporter AND collecting own logs will most likely result in amplification and exploding of logs.

This is not accurate. Explosion of logs will happen only when using `logging` exporter that sends the logs to console (why is it not named `console` exporter is a mystery to me btw. 🤔).

Anyhow, I will rephrase this paragraph to warn more loudly about the logs explosion. 👍

dmitryax:
> @dmitryax any thoughts on what the implications of these change could be? Contrib is a much larger image, do we care about it?
The original idea for this chart was is to support only core collector functionality by default and provide users with an option to use the contrib collector for other use cases. I suppose that log collection components can be moved to the core collector build later, so I'm good with temporarily switching it to contrib.

naseemkullah:
> > @dmitryax any thoughts on what the implications of these change could be? Contrib is a much larger image, do we care about it?
> 
> The original idea for this chart was is to support only the core collector functionality by default and provide users with an option to use the contrib collector for other use cases. I suppose that log collection components can be moved to the core collector build later, so I'm good with temporarily switching it to contrib.

According to https://github.com/open-telemetry/opentelemetry-helm-charts/pull/36#discussion_r592389265 the `filelog` receiver will not be making it to core anytime soon.

I have a slight preference for sticking to the core image by default and just adding a sample values.yaml in the readme as how to achieve the container logging, but don't let that be a blocker!

charts/opentelemetry-collector/values.yaml
dmitryax: The original idea for this chart was is to support only the core collector (like "essential") functionality by default and provide users with an option to use the contrib collector for other use cases. If we can move `filelog` receiver to the core collector eventually, I'm good with temporarily switching it to contrib.

dmitryax: If the logging pipeline is unstable, I'd recommend to set `containerLogs.enabled` to `false` by default, and use contrib build only if user sets it to `true`. @tigrannajaryan @naseemkullah thoughts?

naseemkullah: sgtm, provided the user can still override the image in either case

charts/opentelemetry-collector/README.md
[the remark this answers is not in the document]
rockb1017: if the file exporter would write to a path that it tails, only then, it will explode.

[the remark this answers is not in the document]
tigrannajaryan: >This is not accurate. Explosion of logs will happen only when using logging exporter that sends the logs to console

You are right.

>if the file exporter would write to a path that it tails, only then, it will explode.

Correct. This is more difficult to do accidentally but also a danger. So, it would be good to warn about this too.

charts/opentelemetry-collector/values.yaml
tigrannajaryan: > If the logging pipeline is unstable, I'd recommend to set `containerLogs.enabled` to `false` by default, and use contrib build only if user sets it to `true`. @tigrannajaryan @naseemkullah thoughts?

@dmitryax Is it possible use different builds/images based on a chart setting? I think that's not a bad idea to do temporarily until filelog is moved to the core.

dmitryax: > @dmitryax Is it possible use different builds/images based on a chart setting?

It can be a bit tricky, but must be possible by introducing another template for the image and command arguments.

>  I think that's not a bad idea to do temporarily until filelog is moved to the core.

If we will eventually move the `filelog` receiver to the core, I'm good with temporarily bringing contrib image. Let's put a comment saying that if user don't need the logs pipeline they can set the image back to core.

tigrannajaryan: The exact set of receivers we will have in the core is currently being discussed. We are not yet sure the full `filelog` will be in the core. Let's assume that maybe it will not happen in the near future. We need a reasonable solution for now that uses `filelog` from contrib.

dmitryax: let's see if we can easily image the image (core vs contrib) dependant on `containerLogs.enabled`. if it's becomes too involved, we can keep contrib image by default?

perk-sumo: > I'd recommend to set `containerLogs.enabled` to `false` by default, and use `contrib` build only if user sets it to true

Hi!
The `containerLogs.enabled=false` by default make sure at this point.
But I think this should change pretty soon, as soon as we confirm this is working.
That's because gathering logs from the k8s environment is one of the most obvious things a user would want from this chart IMO.

naseemkullah: While that may be true for other users, today as a user of this chart in production I would disable k8s container logging as that is covered by my providers managed service, I use the collector exclusively for trace data collection.

dmitryax: I agree that we shouldn't enable any functionality that isn't part of the stable core functionality

charts/opentelemetry-collector/templates/_config.tpl
astencel-sumo: I have added a comment in https://github.com/open-telemetry/opentelemetry-helm-charts/pull/36/commits/4df897e8729399eaa3e4fde019ca7a3b31d9a316 and more documentation in https://github.com/open-telemetry/opentelemetry-helm-charts/pull/36/commits/f005dde14bfdd2971aad69a3c0adbe3233a78a97. Please take a look @tigrannajaryan.

charts/opentelemetry-collector/README.md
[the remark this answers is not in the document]
astencel-sumo: I have added a "warning" section to the docs in https://github.com/open-telemetry/opentelemetry-helm-charts/pull/36/commits/f005dde14bfdd2971aad69a3c0adbe3233a78a97, please take a look.

charts/opentelemetry-collector/values.yaml
astencel-sumo: I have disabled container logs by default and reverted to using core collector image in https://github.com/open-telemetry/opentelemetry-helm-charts/pull/36/commits/f005dde14bfdd2971aad69a3c0adbe3233a78a97.

charts/opentelemetry-collector/README.md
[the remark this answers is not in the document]
astencel-sumo: > Let's not advertise the `file` exporter yet. It is unstable and is typically used for local troubleshooting only. Can we show the example using some other exporter instead?

Changed to `otlphttp` exporter, please let me know if this is any better :)

astencel-sumo:
Folks, I have applied all the code review remarks:

- Switched to setting `agentCollector.containerLogs.enabled` to `false` by default,
- Switched back to using core collector image in the chart,
- Added docs on how to enable container log collection,
- Added docs on the warning of looping logs from exporter back into receiver,
- Changed `file` exporter in the example to `otlphttp` exporter, according to code review remarks.

Please take a look.

naseemkullah (approved):

charts/opentelemetry-collector/README.md
```diff
@@ -107,6 +108,73 @@ agentCollector:
...
+#### :warning: Warning: Risk of looping the exported logs back into the receiver, causing "log explosion"
+
+The container logs pipeline uses the `logging` console exporter by default.
+Paired with the default `filelog` receiver that receives all containers' console output,
+it is easy to accidentally feed the exported logs back into the receiver.
+
+To prevent this, the default configuration of the receiver excludes logs from the collector's containers.
+
+If you want to include the collector's logs, make sure to replace the `logging` exporter
+with an exporter that does not send logs to collector's standard output.
```
tigrannajaryan, line 147: Please also add warning that `--log-level=debug` command line option should be avoided when collecting Collector's own logs since at debug level the Collector may produce logs per each collected input log item, which again can result in amplified loop.

charts/opentelemetry-collector/README.md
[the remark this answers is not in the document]
tigrannajaryan: LGTM.

charts/opentelemetry-collector/values.yaml
```diff
@@ -73,7 +80,6 @@ imagePullSecrets: []
 # OpenTelemetry Collector executable
 command:
   name: otelcol
-  # name: otelcontribcol
```
dmitryax, line 76: Can we bring back the comments? they seem to be useful

dmitryax (commented):
One nit, otherwise LGTM

charts/opentelemetry-collector/values.yaml
tigrannajaryan: If we want comments let's make sure they are actual readable sentences and don't look like a commented code that we forgot to delete. 

A comment like the one that was on line 76 is not very helpful, it is confusing. I don't know if it is supposed to be uncommented or deleted. How do I know it should replace the previous line?

tigrannajaryan: @dmitryax how do you want to move forward? Address this in this PR or a separate PR?

dmitryax: Adding a comment would be helpful. I'm good with either way: addressing it here or in separate PR.

dmitryax (approved):

charts/opentelemetry-collector/values.yaml
tigrannajaryan: @astencel-sumo up to you. I can merge now or merge for you to address it.

charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -115,9 +116,90 @@ exporters:
...
       exporters: [otlp]
 {{- end }}
 {{- end }}
+
+{{- define "opentelemetry-collector.agent.containerLogsConfig" -}}
+{{- if .Values.agentCollector.containerLogs.enabled }}
+receivers:
+  filelog:
+    include: [ /var/log/pods/*/*/*.log ]
+    # Exclude collector container's logs. The file format is /var/log/pods/<namespace_name>_<pod_name>_<pod_uid>/<container_name>/<run_id>.log
```
rockb1017, line 133: I think it should be configurable from values.yaml. 
k8s admin would may like to ingest otel agents' logs to monitor them after they disable `logging` exporter.

charts/opentelemetry-collector/values.yaml
astencel-sumo: I'll add a comment to this.

charts/opentelemetry-collector/README.md
astencel-sumo: Added note in https://github.com/astencel-sumo/opentelemetry-helm-charts/commit/8a123a1ec05bf4197fa2f8edc89bba4dd1b810e2.

charts/opentelemetry-collector/values.yaml
astencel-sumo: Added comment in https://github.com/open-telemetry/opentelemetry-helm-charts/pull/36/commits/04c8f4734cf4a5a48fbab40bc1ed8b6ba85da618.

charts/opentelemetry-collector/templates/_config.tpl
astencel-sumo: Let's make it a separate PR if that's OK?

We probably need to prevent a scenario when user changes the property and inadvertently causes the looping of logs from exporter back into the receiver. It needs to be assured that changing of such property alone does not cause the loop.

astencel-sumo:
I have hopefully addressed all the comments. Are we ready to merge @tigrannajaryan?

charts/opentelemetry-collector/templates/_config.tpl
tigrannajaryan: Sounds good, let's do in a separate PR.

tigrannajaryan:
Thanks @astencel-sumo

PR #38, opened 2021-03-13
Helm logging support (#1)

rockb1017:
Improving the helm chart:
- k8s tagger to enrich with k8s metadata from container logs
- remove hostport from default values.yaml
- Optimize container log parsing by specifying the container runtime
- make it configurable to collect otel collector log
- create cluster role and role binding required for k8s metadata enrichment


charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -121,3 +122,116 @@ service:
...
+          - remove: logtag
+          - remove: stream
+          - remove: container_name
+          - remove: namespace
+          - remove: pod_name
+          - remove: run_id
+          - remove: uid
+{{- if .Values.agentCollector.containerLogs.enrichK8sMetadata }}
+processors:
+  k8s_tagger:
```
sumo-drosiek, line 197: `groupbyattrs` processot is required to use before `k8s_tagger` as `k8s_tagger` uses resource labels

```
  groupbyattrs:
    keys:
      - k8s.container.name
      - k8s.namespace.name
      - k8s.pod.name
      - run_id
      - k8s.pod.uid
```

tigrannajaryan: `groupbyattrs` will have a performance impact. Probably we have to do it now, but for the future I think it would be better if `filelog` allowed placing some attributes directly into the resource and we use that capability in k8s collection. I opened an issue https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/2725

tigrannajaryan:
@rockb1017 please sign the CLA.

rockb1017:
> @rockb1017 please sign the CLA.

Do I do a individual CLA?

charts/opentelemetry-collector/templates/_config.tpl
rockb1017: I am putting it in resource label, no? 
```
      - type: metadata
        attributes:
          stream: 'EXPR($.stream)'
          k8s.container.name: 'EXPR($.container_name)'
          k8s.namespace.name: 'EXPR($.namespace)'
          k8s.pod.name: 'EXPR($.pod_name)'
          run_id: 'EXPR($.run_id)'
          k8s.pod.uid: 'EXPR($.uid)'
        resource:
          k8s.pod.uid: 'EXPR($.uid)'
```

rockb1017: I confirm enrichment works without using groupbyattrs. 
```
2021-03-20T00:05:40.172Z        INFO    loggingexporter/logging_exporter.go:471 LogsExporter    {"#logs": 1}
2021-03-20T00:05:40.173Z        DEBUG   loggingexporter/logging_exporter.go:498 ResourceLog #0
Resource labels:
     -> k8spoduid: STRING(eda0a956-99ec-49cb-a954-d733e2f881dd)
     -> k8s.pod.annotations.splunk.com/index: STRING(test)
     -> k8s.pod.name: STRING(loggen-zjklf)
     -> k8s.namespace.name: STRING(default)
     -> k8s.pod.startTime: STRING(2021-03-20 00:03:32 +0000 UTC)
     -> k8s.pod.uid: STRING(eda0a956-99ec-49cb-a954-d733e2f881dd)
     -> k8s.node.name: STRING(ip-192-168-82-91.us-west-1.compute.internal)
     -> k8s.pod.labels.hello: STRING(loggen)
     -> k8s.pod.labels.app: STRING(loggen)
InstrumentationLibraryLogs #0
InstrumentationLibrary  
LogRecord #0
Timestamp: 1616198740006040006
Severity: Undefined
ShortName: 
Body: {
     -> log: STRING(num: 127 | loggen-zjklf | 2021-03-20T00:05:40.005870 | rrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrr
)
}
Attributes:
     -> k8s.pod.uid: STRING(eda0a956-99ec-49cb-a954-d733e2f881dd)
     -> run_id: STRING(0)
     -> stream: STRING(stdout)
     -> k8s.container.name: STRING(loggen)
     -> k8s.namespace.name: STRING(default)
     -> k8s.pod.name: STRING(loggen-zjklf)
```

@tigrannajaryan 
However, when i ingest this log to splunk, I only see metadata in `Attributes`, not in `Resources`
image
What should I do to ingest data in `Attributes` ?

tigrannajaryan: @rockb1017 see if @atoulme can help, I have never used the HEC exporter. My guess is that HEC exporter does not handle Resource attributes, but I may be wrong. If that's the case then it is a bug in HEC exporter.

tigrannajaryan:
@rockb1017 is this PR still needed?

rockb1017:
@tigrannajaryan 
I rebased and updated my commit. 
It includes 
* k8s tagger to enrich with k8s metadata from container logs
* remove hostport from default values.yaml
* Optimize container log parsing by specifying the container runtime
* make it configurable to collect otel collector log
* create cluster role and role binding required for k8s metadata enrichment
* disable liveness and readiness because it is not working atm

charts/opentelemetry-collector/templates/clusterRole.yaml
```diff
@@ -0,0 +1,24 @@
...
+    - ""
+  resources:
+    - "namespaces"
+    - "pods"
+  verbs:
+    - "list"
+    - "get"
+    - "watch"
+{{- end -}}
+{{- end -}}
```
sumo-drosiek, line 24: please add newline at the EOF

charts/opentelemetry-collector/templates/clusterRoleBinding.yaml
```diff
@@ -0,0 +1,18 @@
...
+roleRef:
+  kind: ClusterRole
+  name: {{ template "opentelemetry-collector.fullname" . }}
+  apiGroup: rbac.authorization.k8s.io
+subjects:
+- kind: ServiceAccount
+  name: {{ include "opentelemetry-collector.serviceAccountName" . }}
+  namespace: {{ .Release.Namespace }}
+{{- end -}}
+{{- end -}}
```
sumo-drosiek, line 18: please add newline at the EOF

tigrannajaryan:
EasyCLA is stuck. Let me try to close and reopen.

tigrannajaryan:
@rockb1017 please update PR description / commit message to reflect what you listed here https://github.com/open-telemetry/opentelemetry-helm-charts/pull/38#issuecomment-816338812

charts/opentelemetry-collector/values.yaml
```diff
@@ -110,25 +110,21 @@ ports:
     enabled: true
     containerPort: 4317
     servicePort: 4317
-    hostPort: 4317
```
naseemkullah, line 113: Can you please keep the hostPorts enabled conditionally if being run as a daemonset (agent) ?

charts/opentelemetry-collector/templates/_pod.tpl
```diff
@@ -37,17 +37,22 @@ containers:
...
-        path: /
-        port: 13133
+    # livenessProbe: # TODO: does not work. it keeps killing pod. 
+    #   httpGet:
+    #     path: /
+    #     port: 13133
+    # readinessProbe:
+    #   httpGet:
+    #     path: /
+    #     port: 13133
```
naseemkullah, line 55: This works prior to this commit, do you know what might have affected it? Would be nice to sort out before merging.

charts/opentelemetry-collector/values.yaml
rockb1017: @naseemkullah Could you explain why you would want it by default? 
without using hostPort, it can be accessed through using k8s service. 
in my AWS EKS cluster, it fails to be deployed when trying to use hostPort. it seems that it requires additional cluster configuration to map it to host port. It would make it a bit more complicated to deploy it.

charts/opentelemetry-collector/templates/_pod.tpl
rockb1017: Hmm I just tested it and it is working again. I will uncomment this part. Thanks!

charts/opentelemetry-collector/templates/clusterRole.yaml
rockb1017: thanks

charts/opentelemetry-collector/values.yaml
naseemkullah: This how the instrumented workloads communicate with the collector when it runs as a daemon set in agent mode, I agree that in standalone collector mode they should be disabled and the k8s service used, hence enabling them under the condition that agent mode is enabled.

rockb1017: so pods that are running apps that has been instrumented to send trace information are sending it to its node ip it is running on?

naseemkullah: thats right! with something like this in their env:

```
      - name: OTEL_EXPORTER_JAEGER_AGENT_HOST
        valueFrom:
          fieldRef:
            fieldPath: status.hostIP
 ```

Comment (naseemkullah) [reply charts/opentelemetry-collector/values.yaml:113]: the alternative is running the otel-collector agent as a sidecar container, which is outside the scope of this helm chart

File: charts/opentelemetry-collector/values.yaml [hunk 6]
```diff
@@ -137,6 +133,17 @@ agentCollector:
 
   containerLogs:
     enabled: false
+    # Container runtime. One of `docker`, `cri-o`, or `containerd`
+    containerRunTime: ""
+    includeAgentLogs: false
+    enrichK8sMetadata: true
+    listOfAnnotations: []
+    listOfLabels: []
```
dmitryax, line 141: K8s metadata applicable not only to logs, but widely used for traces and metrics as well.

All the default configuration can be easily overridden (by merging with defaults) in user's values.yaml. So I don't see a lot of benefit in providing extra configuration ways. Especially `listOfAnnotations` and `listOfLabels` can be easily replaced by the existing configuration method.
```
config:
  processors:
    k8s_tagger:
      annotations: ...
      labels: ...
```
it's just one line more to print than
```
agentCollector:
  containerLogs:
    listOfAnnotations:
    listOfLabels:
```
but less configuration options -> less confusion.

Speaking of `enrichK8sMetadata`, I would recommend to move the option one level up to be able to apply it to any of the pipelines. Like:
```
agentCollector:
  pipelinesWithK8sMetadata: [metrics, traces, logs]
```

Let's keep `agentCollector.k8sMetadataPipelines: []` by default.

@tigrannajaryan @naseemkullah thoughts?

charts/opentelemetry-collector/values.yaml
dmitryax: I agree, let's keep host ports. The only stable pipeline of the otel collector currently is tracing. And running agent with no ports exposed doesn't provide any value for traces collection.

dmitryax: > in my AWS EKS cluster, it fails to be deployed when trying to use hostPort. it seems that it requires additional cluster configuration to map it to host port. It would make it a bit more complicated to deploy it.

@rockb1017 Could you please share the errors you see? I never saw any issues with exposed ports on this deamonset.

charts/opentelemetry-collector/templates/_pod.tpl
```diff
@@ -37,10 +37,15 @@ containers:
...
+      - name: KUBE_NODE_NAME
+        valueFrom:
+          fieldRef:
+            apiVersion: v1
+            fieldPath: spec.nodeName
       {{- with .Values.extraEnvs }}
       {{- . | toYaml | nindent 6 }}
       {{- end }}
-    livenessProbe:
+    livenessProbe:
```
dmitryax, line 48: nit: trailing space

charts/opentelemetry-collector/values.yaml
naseemkullah: > K8s metadata applicable not only to logs, but widely used for traces and metrics as well.
> 
> All the default configuration can be easily overridden (by merging with defaults) in user's values.yaml. So I don't see a lot of benefit in providing extra configuration ways. Especially `listOfAnnotations` and `listOfLabels` can be easily replaced by the existing configuration method.
> 
> ```
> config:
>   processors:
>     k8s_tagger:
>       annotations: ...
>       labels: ...
> ```
> 
> it's just one line more to print than
> 
> ```
> agentCollector:
>   containerLogs:
>     listOfAnnotations:
>     listOfLabels:
> ```
> 
> but less configuration options -> less confusion.

Good catch 💯 

> Speaking of `enrichK8sMetadata`, I would recommend to move the option one level up to be able to apply it to any of the pipelines. Like:
> 
> ```
> agentCollector:
>   pipelinesWithK8sMetadata: [metrics, traces, logs]
> ```
> 
> Let's keep `agentCollector.k8sMetadataPipelines: []` by default.
> 

I just learned about this. so we would have both `agentCollector.pipelines` and `agentCollector.pipelinesWithK8sMetadata`?

Seeing that this is a helm chart maybe we just add the `k8s_tagger` and so on directly in the pipelines out of the box if that makes sense?

dmitryax: > I just learned about this. so we would have both agentCollector.pipelines and agentCollector.pipelinesWithK8sMetadata?

right

> Seeing that this is a helm chart maybe we just add the k8s_tagger and so on directly in the pipelines out of the box if that makes sense?

Sounds good, but having `k8s_tagger` enabled by default requires installing the cluster role. At least we need a way to opt out from that.

tigrannajaryan: > but less configuration options -> less confusion.

I agree.

charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -131,22 +131,15 @@ service:
...
-      - type: router
-        id: get-format
-        routes:
-          - output: parser-docker
-            expr: '$$record matches "^\\{"'
-          - output: parser-crio
-            expr: '$$record matches "^[^ Z]+ "'
-          - output: parser-containerd
-            expr: '$$record matches "^[^ Z]+Z"'
+      {{- if eq .Values.agentCollector.containerLogs.containerRunTime "cri-o" }}
```
tigrannajaryan, line 142: I do not fully understand what this change does. Did this change delete the ability to automatically detect the format used by kubernetes and requires the user to specify the format?

sumo-drosiek: It looks like that

tigrannajaryan: @rockb1017 please clarify why want to change this. It appears to me automatic detection is valuable since there is one less thing for users to worry about.

rockb1017: yes, instead of collector detecting the format, we are allowing users to provide their container runtime. this is how it is done in SCK. Though I am not saying this is definitely better way, but at least it will definitely be better for performance. I would say asking users to specify the container runtime once and have more efficient pipeline is better.

charts/opentelemetry-collector/values.yaml
rockb1017: I think by having less config, you are making the users to study more on documentation to properly set up configOverride. 
For example, 
```
      k8s_tagger:
        passthrough: false
        auth_type: "kubeConfig"
        pod_association:
          - from: resource_attribute
            name: k8s.pod.uid
        extract:
          metadata:
            - deployment
            - cluster
            - namespace
            - node
            - startTime
          annotations: []
          labels: []
        filter:
          node_from_env_var: KUBE_NODE_NAME
```
this is required config for enriching it with k8s metadata. 
most of it is static. Users doesn't need to know what `passthrough` value they need to pick, nor search though this plugin documentation to properly choose what `pod_association` I need to use. 
So I think in this case, "less configuration options -> less confusion" doesn't quite fit here. it is more like "more guided helm chart -> less confusion". 
We are helping users by presenting the configuration they need to look and hiding what they don't need to waste their minds on.

charts/opentelemetry-collector/templates/_config.tpl
tigrannajaryan: It is a tradeoff between ease of use and performance. Both are valid approach and different users likely will have different preferences. I think what we can do is by default have automatic detection of the format and in addition to that allow the user to specify the format, in which case the detection will be disabled and we will have better performance.

dmitryax:
Closing this PR since it's too specific for a particular logs collection use case. Feel free to open it again if you want to address the comments

Issue #34, opened 2021-03-09
Error creating processor "memory_limiter" on clean install


astencel-sumo:
Helm chart: [opentelemetry-collector-0.4.3](https://github.com/open-telemetry/opentelemetry-helm-charts/releases/tag/opentelemetry-collector-0.4.3) (latest at the time of writing)

When installing the chart with:

```sh
helm install otc open-telemetry/opentelemetry-collector
```

the installation succeeds, but the OTC pod crashes with the following error:

```console
Error: cannot setup pipelines: cannot build pipelines: error creating processor "memory_limiter" in pipeline "metrics": checkInterval must be greater than zero
```

<details>
<summary>See full OTC pod logs</summary>

```console
2021-03-09T12:37:54.144Z	info	service/service.go:411	Starting OpenTelemetry Collector...	{"Version": "v0.20.0", "GitHash": "a10a1a7a", "NumCPU": 8}
2021-03-09T12:37:54.147Z	info	service/service.go:592	Using memory ballast	{"MiBs": 204}
2021-03-09T12:37:54.147Z	info	service/service.go:255	Setting up own telemetry...
2021-03-09T12:37:54.148Z	info	service/telemetry.go:102	Serving Prometheus metrics	{"address": "0.0.0.0:8888", "level": 0, "service.instance.id": "49b6c5c0-07a7-41c2-9a92-b5e01a14b86b"}
2021-03-09T12:37:54.149Z	info	service/service.go:292	Loading configuration...
2021-03-09T12:37:54.149Z	info	service/service.go:303	Applying configuration...
2021-03-09T12:37:54.149Z	info	service/service.go:324	Starting extensions...
2021-03-09T12:37:54.149Z	info	builder/extensions_builder.go:53	Extension is starting...	{"component_kind": "extension", "component_type": "health_check", "component_name": "health_check"}
2021-03-09T12:37:54.149Z	info	healthcheckextension/healthcheckextension.go:40	Starting health_check extension	{"component_kind": "extension", "component_type": "health_check", "component_name": "health_check", "config": {"TypeVal":"health_check","NameVal":"health_check","Port":13133}}
2021-03-09T12:37:54.149Z	info	builder/extensions_builder.go:59	Extension started.	{"component_kind": "extension", "component_type": "health_check", "component_name": "health_check"}
[...]

PR #39, opened 2021-03-22
Fix default memory_limiter configuration for helm version 3.5.x

dmitryax:
Starting from helm 3.5.0, merging rules were changed: now if null value merged into a dict, null value is preserved.

In order to make it cleaner and working across all helm versions, this commit replaces the merge operation with a conditional set.

If user wants to override the default memory_limiter configuration, they should override all the fields. That behavior seems more practical than partial override for the this processor.

Resolves https://github.com/open-telemetry/opentelemetry-helm-charts/issues/34


dmitryax:
@naseemkullah PTAL

tigrannajaryan (approved):

PR #41, opened 2021-03-30
implement secret mounting

jdgeisler:
Adds the ability for people to mount secrets. This can be used for referencing certificates in the config, like so:

```
secretMounts:
- name: jaeger-cert
  secretName: jaeger-cert
  mountPath: /etc/tls
  readOnly: true
  subPath: jaeger # subPath is optional
```

```
exporters:
    jaeger:
      endpoint: localhost
      cert_file: /etc/tls/tls.crt
      key_file: /etc/tls/tls.key
```

### DIFF
--- a/charts/opentelemetry-collector/templates/_pod.tpl
+++ b/charts/opentelemetry-collector/templates/_pod.tpl
@@ -61,6 +61,10 @@ containers:
         mountPropagation: {{ .mountPropagation }}
         {{- end }}
       {{- end }}
+      {{- range .Values.secretMounts }}
+      - name: {{ .name }}
+        mountPath: {{ .path }}
+      {{- end }}
       {{- if and $.isAgent .Values.agentCollector.containerLogs.enabled }}
       - name: varlogpods
         mountPath: /var/log/pods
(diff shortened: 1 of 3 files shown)


charts/opentelemetry-collector/templates/_pod.tpl
```diff
@@ -61,6 +61,10 @@ containers:
         mountPropagation: {{ .mountPropagation }}
         {{- end }}
       {{- end }}
+      {{- range .Values.secretMounts }}
+      - name: {{ .name }}
+        mountPath: {{ .path }}
```
naseemkullah, line 66:
```suggestion
        mountPath: {{ .mountPath }}
```

nit: leak the abstraction (in a good way)

naseemkullah (changes requested):
Thanks! Would it make sense to also add:

```yaml
subPath: {{ .subPath }}
readOnly: {{ .readOnly }}
```
 ?

jdgeisler:
> Thanks! Would it make sense to also add:
> 
> ```yaml
> subPath: {{ .subPath }}
> readOnly: {{ .readOnly }}
> ```
> 
> ?

Yes that is a good idea, I can make the changes. And in this case, subPath can be set as a conditional so it is optional.

naseemkullah (approved):

PR #49, opened 2021-06-17
Added priorityClassName to pod template and values file

Carlos1711:
This will add the ability to set a priorityClassName as part of the deployment.

The Deployment for the agents can be set to a higher priority to make sure the scheduler deploys the agents first before any apps are deployed.


dmitryax (approved):
LGTM

Issue #57, opened 2021-07-12
Migrate the KUTTL smoke tests of the operator Helm chart to Github CI workflow


Saber-W:
**Is your feature request related to a problem? Please describe.**

Currently, during every release of the operator Helm chart, maintainers need to clone the [operator repo](https://github.com/open-telemetry/opentelemetry-operator) and run the [KUTTL tests](https://github.com/open-telemetry/opentelemetry-operator/blob/main/kuttl-test.yaml) locally. We want to migrate all the tests to Github CI workflow so that the tests will run automatically when a new PR is submitted.

**Describe the solution you'd like**

Add KUTTL tests automation to current Github actions.

cc @alolita

Saber-W:
Hi @dmitryax, please assign this to me. Thank you!

PR #62, opened 2021-07-27
Migrate Operator Helm chart KUTTL tests to Github workflow

Saber-W:
**Description**

Currently, during every release of the Operator Helm chart, maintainers need to clone the operator repo and run the KUTTL tests locally. This PR migrates the KUTTL tests to Github workflow so that the tests will run automatically when a new PR is submitted.

**Related Issue**
#57

cc @alolita @dmitryax @tigrannajaryan

### DIFF
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -53,3 +53,17 @@ jobs:
 
       - name: Run chart-testing (install)
         run: ct install --target-branch main
+
+      - name: Run KUTTL smoke tests
+        run: |
+          sudo curl -Lo /usr/local/bin/kubectl-kuttl https://github.com/kudobuilder/kuttl/releases/download/v0.11.0/kubectl-kuttl_0.11.0_linux_x86_64
+          sudo chmod +x /usr/local/bin/kubectl-kuttl
+          export PATH=$PATH:/usr/local/bin
+          helm repo add open-telemetry https://open-telemetry.github.io/opentelemetry-helm-charts
+          helm repo update
+          sleep 5s
+          helm install my-opentelemetry-operator open-telemetry/opentelemetry-operator
+          kubectl wait --timeout=5m --for=condition=available deployment opentelemetry-operator-controller-manager -n opentelemetry-operator-system
+          git clone https://github.com/open-telemetry/opentelemetry-operator.git
+          kubectl kuttl test ./opentelemetry-operator/tests/e2e --config $GITHUB_WORKSPACE/charts/opentelemetry-operator/release/kuttl-test.yaml
+        if: steps.list-changed.outputs.operator-changed == 'true'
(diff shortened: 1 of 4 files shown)


.github/workflows/lint-test.yaml
```diff
@@ -53,3 +53,15 @@ jobs:
 
       - name: Run chart-testing (install)
         run: ct install --target-branch main
+
+      - name: Run KUTTL smoke tests
+        run: |
+          sleep 10s
```
dmitryax, line 59: Why sleep needed?

.github/workflows/lint-test.yaml
```diff
@@ -53,3 +53,15 @@ jobs:
...
+          sudo curl -Lo /usr/local/bin/kubectl-kuttl https://github.com/kudobuilder/kuttl/releases/download/v0.11.0/kubectl-kuttl_0.11.0_linux_x86_64
+          sudo chmod +x /usr/local/bin/kubectl-kuttl
+          export PATH=$PATH:/usr/local/bin
```
dmitryax, line 62: isn't `/usr/local/bin` already in $PATH?

.github/workflows/lint-test.yaml
Saber-W: It's because chart-testing deletion needs some time to finish, which we need to wait until the namespace is deleted. Otherwise, we will get an error.

.github/workflows/lint-test.yaml
Saber-W: Yes, I will remove this export.

.github/workflows/lint-test.yaml
dmitryax: Which chart-testing deletion do you mean? I don't see any deletes prior to this

Saber-W: This one: `ct install --target-branch main` in line 55. It will delete the chart after the testing.

dmitryax: Can we use `kubectl wait --for=delete` instead?

Saber-W: I think we can. I will make the changes.

Saber-W: I tried to use `kubectl wait` command to wait until the ns is deleted. But it sometimes won't work and will cause time out error because the ns could be deleted before this command runs. @dmitryax Please let me know if I miss something. Thanks!

dmitryax: I guess you can do something like this 
```
until kubectl get ns opentelemetry-operator-system 2>&1 >/dev/null; do sleep 1; done
```
It'll be more descriptive them just sleep for 10 seconds

Saber-W: Using `until` command makes more sense. I have made the changes. Thanks.

dmitryax (approved):

Issue #58, opened 2021-07-12
Add functions to current operator Helm chart release script to validate which template files need to be updated


Saber-W:
**Is your feature request related to a problem? Please describe.**

Currently, in operator Helm chart, we provide a release helper (main.go) for maintainers to update CRDs, values.yaml and Chart.yaml automatically during every release. We want to add one more function to this, which would compare all the template files under `templates` directory with the latest OTEL Operator manifest, and tell maintainers which files need to be updated.

**Describe the solution you'd like**

- Retrieve the latest operator manifest
- Unmarshal it and compare each YAML file with the template file under `templates` directory

cc @alolita

Saber-W:
Hi @dmitryax, please assign this to me. Thank you!

PR #65, opened 2021-08-09
Implement the validation that checks if every template file is up-to-date

Saber-W:
**Description**
This is the second OpenTelemetry Operator Helm chart release automation PR, which adds the ability to validate which template files need to be updated. This PR resolves the issue that the Helm chart maintainers need to manually compare the template files and the upstream manifest in order to determine which template files are outdated.

Example output of the release code:
```console
2021/08/09 12:26:09 The OpenTelemetry Collector CRD update finished. It's up-to-date now.
2021/08/09 12:26:09 ATTENTION: opentelemetry-operator-leader-election-rolebinding configuration doesn't exist. Please create it in the file: rolebinding.yaml
2021/08/09 12:26:09 The values.yaml and Chart.yaml update finished. They are up-to-date now.
2021/08/09 12:26:09 ATTENTION: opentelemetry-operator-controller-manager configuration doesn't exist. Please create it in the file: deployment.yaml
2021/08/09 12:26:09 ATTENTION: validatingwebhookconfiguration.yaml file needs to be updated
```

**Related Issue**
#58 

**Testing**
This PR adds unit tests for each release function.

cc @alolita @dmitryax @tigrannajaryan


charts/opentelemetry-operator/release/release.go
```diff
@@ -0,0 +1,144 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// updateImageTags retrieve the latest image tags and update the values.yaml and Chart.yaml respectively.
+func updateImageTags(object K8sObject, valuesYAMLPath string, chartYAMLPath string) error {
+	// Retrieve the latest image repository and tag of the two container images.
+	managerImage := strings.Split(object.Spec["template"].(map[string]interface{})["spec"].(map[string]interface{})["containers"].([]interface{})[0].(map[string]interface{})["image"].(string), ":")
```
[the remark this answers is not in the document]
Saber-W: It's because Operator contains two images: one is `manager`, the other is `kube-rbac-proxy`. Also, `manager` is the container's name. So I think `managerImage` is more informative.

charts/opentelemetry-operator/release/release.go
```diff
@@ -0,0 +1,144 @@
...
+	kubeRBACProxyImage := strings.Split(object.Spec["template"].(map[string]interface{})["spec"].(map[string]interface{})["containers"].([]interface{})[1].(map[string]interface{})["image"].(string), ":")
```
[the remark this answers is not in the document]
Saber-W: Yeah, this makes sense. I will change this code.

charts/opentelemetry-operator/release/release.go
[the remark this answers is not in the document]
dmitryax: ok, makes sense

charts/opentelemetry-operator/release/release.go
```diff
@@ -0,0 +1,153 @@
...
+// updateImageTags retrieves the latest image tags and update the values.yaml and Chart.yaml respectively.
+func updateImageTags(object K8sObject, valuesYAMLPath string, chartYAMLPath string) error {
+	// Retrieve the latest image repository and tag of the two container images.
+	containers := object.Spec["template"].(map[string]interface{})["spec"].(map[string]interface{})["containers"].([]interface{})
+	var managerImage, kubeRBACProxyImage []string
+	for i := range containers {
+		image := containers[i].(map[string]interface{})["image"].(string)
+		switch containers[i].(map[string]interface{})["name"].(string) {
+		case "manager":
+			managerImage = strings.Split(image, ":")
```
dmitryax, line 69: Is image name `managerImage[0]` actually being used after that? If not, let's use `managerImageTag string`  instead. Otherwise, if image name is used, managerImage should be a struct with `name` and `tag` fields. It's hard to track the logic with `[]string` data type

charts/opentelemetry-operator/release/release.go
```diff
@@ -0,0 +1,153 @@
...
+		case "kube-rbac-proxy":
+			kubeRBACProxyImage = strings.Split(image, ":")
+		}
+	}
+
```
dmitryax, line 74: Let's have a check here that both image tags are actually found, and fail if not?

charts/opentelemetry-operator/release/release.go
Saber-W: No, it isn't. Okay, I will make that clear.

charts/opentelemetry-operator/release/release.go
Saber-W: For sure. Let me add this check.

charts/opentelemetry-operator/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.1.1
+version: 0.2.0
```
dmitryax, line 3: This change is no-op for end user. Let's call it `0.1.2`

Saber-W: Okay, makes sense.

dmitryax (approved):

PR #80, opened 2021-09-22
chore: update otel-collector to 0.36.0

bbeattie-phxlabs:
Collector bumped to 0.36.0

Additional compatibility changes:
* `insecure` relocated as per https://github.com/open-telemetry/opentelemetry-collector/pull/4063
* `mem-ballast-size-mib` removed as per open-telemetry/opentelemetry-collector#3626

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.5.10
+version: 0.5.11
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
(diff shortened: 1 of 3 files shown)


charts/opentelemetry-collector/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.5.10
+version: 0.5.11
```
dmitryax, line 3: let's move to `0.6.0` given the config change

dmitryax:
@bbeattie-phxlabs thanks for updating it, but I believe there must be more interface changes:
- `ballast_size_mib` memory_memory option should be moved to ballast extension
- may be required to switch base image to contrib because of jaeger and zipkin receivers added by default

Also please sign the CLA

dmitryax:
Also looks like 0.36.0 wasn't pushed to dockerhub yet. So probably we should upgrade to 0.35.0 for now

PR #82, opened 2021-09-29
Added clusterrole and clusterrolebinding

etiennejournet:
- Made them feature-flippable based on the same clusterRole yaml block
- Values and examples provided
- Bumped minor chart version


dmitryax:
LGTM. Thanks for adding this. please bump the chart version

charts/opentelemetry-collector/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.5.11
+version: 0.6.11
```
dmitryax, line 3: Let's use `0.6.0` instead

PR #97, opened 2021-11-08
feat(collector): add additional ports and explicitly define OTLP endpoints

### LINKED ISSUE #94: Default configuration not deployable due to misconfigured oltp protocol
I am new to the open telemetry, and was starting to set up an Operator and Collector via helm charts. It is not really a bug, but the default helm chart configuration for Collector **does** specify the protocol "otlp" but provides **no** configuration to it ([here](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/582acccee0dfb91f77b760f7474756b3815738e6/charts/opentelemetry-collector/values.yaml#L25)) resulting in following error when running the container:
```
receiver "otlp" has invalid configuration: must specify at least one protocol when using the OTLP receiver
```

The error is easy to understand and by adding the endpoint values to the "oltp" protocol, everything launches smoothly.
```yaml
  config:
    receivers:
      otlp:
        protocols:
          grpc:
            endpoint: 0.0.0.0:4317 # <= Missing
          http:
            endpoint: 0.0.0.0:4318 # <= Missing
```

Adding this default values or removing the "oltp" driver no require this values improves the onboarding and learning process  for newbies like me without having to debug the helm charts first.

secustor:
Fixes #94

This PR: 
- explicitly defines the OTLP endpoints in the collectors config 
- enables Jaeger compact thrift protocol to serve as drop in replacement for the Jaeger Agent

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.7.1
+version: 0.8.0
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-collector/values.yaml
+++ b/charts/opentelemetry-collector/values.yaml
@@ -148,6 +148,12 @@ ports:
     servicePort: 4317
     hostPort: 4317
     protocol: TCP
+  otlp-http:
+    enabled: true
+    containerPort: 4318
+    servicePort: 4318
+    hostPort: 4318
+    protocol: TCP
   jaeger-thrift-compact:
     enabled: true
     containerPort: 6831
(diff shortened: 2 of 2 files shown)


dmitryax:
Hey @secustor, thank you for fixing it! Can you please keep only OTLP fix in this PR and bump version to `0.7.2` to keep atomic updates?

About thrift compact. It doesn't seem to be a popular format. To be honest I'm not convinced that we should add it by default. Do you use if yourself?

secustor:
@dmitryax I have made a break out PR for the endpoint fixes. 

Regarding thrift compact. This is the protocol which Jaeger instrumentations use to communicate with their node agent.     
As Jaeger is deprecating their own Agent in favor of the OTL Collector it makes sense IMO to enable this by default, to give such user a better migration experience.  https://www.jaegertracing.io/docs/1.27/deployment/#agent 

That way their integrations simply "work" instead of looking up the implementation details of the collector and jaeger

dmitryax:
@secustor Ok, sounds good. Let's add it by default. Could you please update this PR?

dmitryax (approved):

PR #98, opened 2021-11-10
fix(collector): explicitly add otlp endpoints and expose otlp-http

### LINKED ISSUE #94: Default configuration not deployable due to misconfigured oltp protocol

secustor:
This is a breakout PR of #97 

Fixes #94

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.7.1
+version: 0.7.2
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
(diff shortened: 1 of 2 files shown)


dmitryax (approved):
Thank you!

PR #99, opened 2021-11-11
declare metrics port by default

raffis:
## Current situation
The metrics port is not declared, not sure what the reason is for that.

## Proposal
Declare metrics port by default.

### DIFF
--- a/charts/opentelemetry-collector/values.yaml
+++ b/charts/opentelemetry-collector/values.yaml
@@ -183,6 +183,13 @@ ports:
     servicePort: 9411
     hostPort: 9411
     protocol: TCP
+  metrics:
+    # The metrics port is disabled by default. However you need to enable the port
+    # in order to use the ServiceMonitor (serviceMonitor.enabled) or PodMonitor (podMonitor.enabled).
+    enabled: false
+    containerPort: 8888
+    servicePort: 8888
+    protocol: TCP
 
 # Configuration for agent OpenTelemetry Collector daemonset, enabled by default
 agentCollector:
@@ -309,19 +316,24 @@ ingress:
 #        - collector.example.com
 
 podMonitor:
+  # The pod monitor by default scrapes the metrics port.
+  # The metrics port needs to be enabled as well.
   enabled: false
-  metricsEndpoints: {}
-  # - port: prometheus
+  metricsEndpoints:
+  - port: metrics
+    # interval: 15s
 
   # additional labels for the PodMonitor
   extraLabels: {}
   #   release: kube-prometheus-stack
 
 serviceMonitor:
+  # The service monitor by default scrapes the metrics port.
+  # The metrics port needs to be enabled as well.
   enabled: false
-  metricsEndpoints: {}
-  # - port: metrics
-  #   interval: 15s
+  metricsEndpoints:
+  - port: metrics
+    # interval: 15s
 
   # additional labels for the ServiceMonitor
   extraLabels: {}
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.9.1
+version: 0.9.2
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/


dmitryax:
Why do you want this port to be exposed? this port should be used by collector itself, no need to expose it to outside world

raffis:
> Why do you want this port to be exposed? this port should be used by collector itself, no need to expose it to outside world

To enable the serviceMonitor to scrape the collector metrics, the port is even mentioned there: https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/values.yaml#L305-L306

charts/opentelemetry-collector/values.yaml
```diff
@@ -180,6 +180,12 @@ ports:
     servicePort: 9411
     hostPort: 9411
     protocol: TCP
+  metrics:
+    enabled: true
+    containerPort: 8888
+    servicePort: 8888
+    hostPort: 8888
```
FuyuanChu, line 187: Suppose we don't need the `hostPort` for metrics.

raffis: Agreed, added it because of consistency but probably better to remove it.

EJKinger:
Is this blocked for any reason?

dmitryax:
I don't think it should be enabled by default. It's needed only when serviceMonitor or any other external scraper is configured to fetch the collector's metrics. Since none of them are enabled by default this port shouldn't be exposed as well.

We can add this port commented out with a comment that this is required for serviceMonitor.

Another option would be to add this port explicitly in the template under condition `{{ if serviceMonitor.enabled }}`.

raffis:
> I don't think it should be enabled by default. It's needed only when serviceMonitor or any other external scraper is configured to fetch the collector's metrics. Since none of them are enabled by default this port shouldn't be exposed as well.
> 
> We can add this port commented out with a comment saying that this is required for serviceMonitor.
> 
> Another option would be to add this port explicitly in the template under condition `{{ if serviceMonitor.enabled }}`.

Can we agree on the way to go before I make the change?
From a pov of a chart user I would expect to enable the serviceMonitor/podMonitor and it works out of the box and I don't have to bother with port declarations. Meaning the port could be defined on the service (as well as deployment) if the serviceMonitor is enabled. And if the podMonitor is enabled the port will only be defined on the deployment.

From my perspective even the endpoints in the monitors should have defaults. It should be enough to just toggle the enabled flags.

dmitryax:
Makes sense. Feel free to adjust the templating to make sure that necessary ports are enabled if serviceMonitor/podMonitor are enabled. If not, 8888 should not be exposed anywhere

dmitryax:
Also please sign CLA

raffis:
> Makes sense. Feel free to adjust the templating to make sure that necessary ports are enabled if serviceMonitor/podMonitor are enabled. If not, 8888 should not be exposed anywhere

Done. Only concern about this pr is that this could break backwards compatibility if somebody declared that port already manually. Either bump major version or leave it like that? Alternatively we could add a condition in the declared ports if port==8888 skip it.

dmitryax:
> Alternatively we could add a condition in the declared ports if port==8888 skip it.

What if users don't want to expose 8888 (due to a conflict for example) but expose metrics under different port like 8889 and want to scrape them with the podMonitor? I don't think this solution covers that use case.

I'm still not sure that these parts should be coupled. I would rather just add a comment to podMonitor/serviceMonitor saying that user should enable the metrics port.

raffis:
> What if users don't want to expose 8888 (due to a conflict for example) but export metrics under different port like 8889 and want to scrape them with the podMonitor? I don't think this solution covers that use case.
> 
> I'm still not sure that these parts should be coupled. I would rather just add a comment to podMonitor/serviceMonitor saying that user should enable the metrics port.

It's not, but the port could be configurable as well while the default is 8888 (`.Values.metricsPort`).
If I see it correctly with the current version of the chart if one wants to enable podMonitor the port needs to be declared in .ports right? That also means the port is in the service which would not be required for a podMonitor.

In the end it might also be what the pr did more or less in the first place, declare the metrics port but disabled. Set default endpoints for pod/service monitors. Then a user needs to enable a monitor as well as the metrics port. This would also not break compatibility. (Port would still be on the service if only podMonitor is needed though).

dmitryax:
> In the end it might also be what the pr did more or less in the first place, declare the metrics port but disabled. Set default endpoints for pod/service monitors. Then a user needs to enable a monitor as well as the metrics port.

This sounds good to me. Please add a comment to the disabled port describing when it can be needed

dmitryax (approved):
LGTM

PR #100, opened 2021-11-22
typo fix

nickytd:
fix a typo in the helm chart values


dmitryax:
@nickytd thanks for fixing it. Any change requires a version bump, even small ones like this. Please update your PRs accordingly

PR #103, opened 2021-11-29
Reorder opentelemetry-collector HPA metrics

danielgblanco:
A current issue between ArgoCD (https://github.com/argoproj/argo-cd/issues/1079) and Kubernetes (https://github.com/kubernetes/kubernetes/issues/74099) makes the HPA stay constantly out of sync. This affects Gitops pipelines when both memory and CPU metrics are used to autoscale the deployment.

This change reorders HPA metrics to make sure that they remain in the order that the Kubernetes HPA controller currently expects them to be.

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.8.0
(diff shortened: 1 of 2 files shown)


dmitryax (approved):

dmitryax:
@danielgblanco, thanks for the contribution!

PR #117, opened 2022-01-28
fixed: helm chart broken with otel/opentelemetry-collector-contrib >=…

### LINKED ISSUE #110: Release opentelemetry-collector Helm chart for latest app version (0.42.0)
As of writing, the latest opentelemetry-collector (0.8.1) chart supports appVersion 0.37.1. Attempts to use this Helm chart version with an the image for the app version 0.42.0 produces this error in the pod:

`Error: failed to start container "opentelemetry-collector": Error response from daemon: OCI runtime create failed: container_linux.go:380: starting container process caused: exec: "/otelcontribcol": stat /otelcontribcol: no such file or directory: unknown`

paulgrav:
These changes attempt to make the helm chart compatible with the otel/opentelemetry-collector-contrib container >= 0.42.0. It resolves #110

It looks like not only the container’s entrypoint changed but so too has its args.

### DIFF
--- a/charts/opentelemetry-collector/templates/_pod.tpl
+++ b/charts/opentelemetry-collector/templates/_pod.tpl
@@ -11,7 +11,6 @@ containers:
     command:
       - /{{ .Values.command.name }}
       - --config=/conf/relay.yaml
-      - --metrics-addr=0.0.0.0:8888
       {{- range .Values.command.extraArgs }}
       - {{ . }}
       {{- end }}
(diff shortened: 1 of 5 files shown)


charts/opentelemetry-collector/templates/_pod.tpl
```diff
@@ -11,7 +11,6 @@ containers:
     command:
       - /{{ .Values.command.name }}
       - --config=/conf/relay.yaml
-      - --metrics-addr=0.0.0.0:8888
```
dmitryax, line 14: This should be defined in the default config service -> telemetry section. Can you please update the config accordingly?

paulgrav: I’ve added the missing services. Checked that my prom instance scraping works as before.

dmitryax (approved):

PR #120, opened 2022-02-02
Add option for PodDisruptionBudget

Mahpara93:
Add option for PodDisruptionBudget(pdb) to opentelemetry-collector

### DIFF
--- a/charts/opentelemetry-collector/templates/pdb.yaml
+++ b/charts/opentelemetry-collector/templates/pdb.yaml
@@ -1,0 +1,13 @@
+{{- if .Values.pdb.enabled .Values.standaloneCollector.enabled }}
+apiVersion: policy/v1beta1
+kind: PodDisruptionBudget
+metadata:
+  name: {{include "opentelemetry-collector.fullname" .}}
+  labels:
+    {{- include "opentelemetry-collector.labels" . | nindent 4 }}
+spec:
+  minAvailable: {{.Values.pdb.minAvailable}}
+  selector:
+    matchLabels:
+      {{- include "opentelemetry-collector.selectorLabels" . | nindent 6}}
+{{- end }}
--- a/charts/opentelemetry-collector/templates/pdb.yaml
+++ b/charts/opentelemetry-collector/templates/pdb.yaml
@@ -2,13 +2,13 @@
 apiVersion: policy/v1beta1
 kind: PodDisruptionBudget
 metadata:
-  name: {{include "opentelemetry-collector.fullname" .}}
+  name: {{ include "opentelemetry-collector.fullname" . }}
   labels:
     {{- include "opentelemetry-collector.labels" . | nindent 4 }}
 spec:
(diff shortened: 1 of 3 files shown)


charts/opentelemetry-collector/values.yaml
```diff
@@ -312,6 +312,11 @@ serviceMonitor:
   extraLabels: {}
   #  release: kube-prometheus-stack
 
+# PodDisruptionBudget (pdb) is used only if standaloneCollector enabled
+pdb:
```
dmitryax, line 316: pdb may be too ambiguous in scope of values.yaml. Let's use full name

charts/opentelemetry-collector/values.yaml
```diff
@@ -312,6 +312,11 @@ serviceMonitor:
...
+  enabled: false
+  minAvailable: 1
```
dmitryax, line 318: let's add `maxUnavailable`

charts/opentelemetry-collector/values.yaml
Mahpara93: Good idea! I changed it to PodDisruptionBudget.

charts/opentelemetry-collector/values.yaml
Mahpara93: I added maxUnavailable as well.

charts/opentelemetry-collector/values.yaml
```diff
@@ -312,6 +312,12 @@ serviceMonitor:
   extraLabels: {}
   #  release: kube-prometheus-stack
 
+# PodDisruptionBudget is used only if standaloneCollector enabled
+PodDisruptionBudget:
```
dmitryax, line 316: Should not be capitalized `podDisruptionBudget` according to helm style guides

Mahpara93: fixed

charts/opentelemetry-collector/templates/pdb.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if .Values.podDisruptionBudget.enabled .Values.standaloneCollector.enabled }}
```
dmitryax, line 1: Hm.. I think `and` is missed here

charts/opentelemetry-collector/templates/pdb.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if .Values.podDisruptionBudget.enabled and .Values.standaloneCollector.enabled }}
```
dmitryax, line 1: This doesn't work :) Should be:
```suggestion
{{- if and .Values.podDisruptionBudget.enabled .Values.standaloneCollector.enabled }}
```

Mahpara93: Thanks! Appreciate your feedback :)

dmitryax (approved):
Thanks for the contribution!

PR #121, opened 2022-02-03
Update podDisruptionBudget default values

Mahpara93:
By default, you can only have one or the other. If both are set then we get the following error: 

```
EXIT STATUS
  1
 
COMBINED OUTPUT:
  Error: UPGRADE FAILED: release opentelemetry-collector failed, and has been rolled back due to atomic being set: failed to create resource: PodDisruptionBudget.policy "opentelemetry-collector" is invalid: spec: Invalid value: policy.PodDisruptionBudgetSpec{MinAvailable:(*intstr.IntOrString)(0xc045b8be40), Selector:(*v1.LabelSelector)(0xc045b8be60), MaxUnavailable:(*intstr.IntOrString)(0xc045b8be20)}: minAvailable and maxUnavailable cannot be both set
```

### DIFF
--- a/charts/opentelemetry-collector/values.yaml
+++ b/charts/opentelemetry-collector/values.yaml
@@ -328,8 +328,8 @@ serviceMonitor:
   #  release: kube-prometheus-stack
 
 # PodDisruptionBudget is used only if standaloneCollector enabled
-podDisruptionBudget: {}
-#   minAvailable: 2
+podDisruptionBudget:
+  minAvailable: 2
 #   maxUnavailable: 1
 
 # autoscaling is used only if standaloneCollector enabled
--- a/charts/opentelemetry-collector/values.yaml
+++ b/charts/opentelemetry-collector/values.yaml
@@ -330,7 +330,7 @@ serviceMonitor:
 # PodDisruptionBudget is used only if standaloneCollector enabled
 podDisruptionBudget:
   enabled: false
-  minAvailable: 2
+#   minAvailable: 2
 #   maxUnavailable: 1
 
 # autoscaling is used only if standaloneCollector enabled
--- a/charts/opentelemetry-collector/values.yaml
+++ b/charts/opentelemetry-collector/values.yaml
@@ -329,6 +329,7 @@ serviceMonitor:
 
 # PodDisruptionBudget is used only if standaloneCollector enabled
 podDisruptionBudget:
+  enabled: false
   minAvailable: 2
 #   maxUnavailable: 1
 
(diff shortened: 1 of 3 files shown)


charts/opentelemetry-collector/values.yaml
```diff
@@ -315,8 +315,9 @@ serviceMonitor:
 # PodDisruptionBudget is used only if standaloneCollector enabled
 podDisruptionBudget:
   enabled: false
+  # Use minAvailable or maxUnavailable option
   minAvailable: 2
-  maxUnavailable: 1
+  # maxUnavailable: 1
```
dmitryax, line 320: You also need to wrap usage of these vars in a condition like 

```
{{- with podDisruptionBudget.minAvailable }}
minAvailable: {{ . }}
{{- end }}
``` 


Or, as another option, you can use the whole map as is:
```
{{ podDisruptionBudget | nindent 2 }}
```

but in this case `enabled` must be removed. Instead of `enabled` we can check if `podDisruptionBudget` has any fields defined in it. values.yaml would be the following by default:
```
podDisruptionBudget: {}
```
And PodDisruptionBudget template would start with
```
{{- if and .podDisruptionBudget standaloneCollector.enabled }}
```

charts/opentelemetry-collector/values.yaml
```diff
@@ -313,10 +313,7 @@ serviceMonitor:
   #  release: kube-prometheus-stack
 
 # PodDisruptionBudget is used only if standaloneCollector enabled
-podDisruptionBudget:
-  enabled: false
-  minAvailable: 2
-  maxUnavailable: 1
```
dmitryax, line 319: please add these two as commented lines

charts/opentelemetry-collector/values.yaml
Mahpara93: Where would this `{{ podDisruptionBudget | nindent 2 }}` be placed?

dmitryax: in the template

dmitryax: `{{ .Values.podDisruptionBudget | toYaml | trim | indent 2 }}` to be precise

dmitryax:
Please test it locally

dmitryax:
Hey @Mahpara93, tests are failing.

Mahpara93:
Hi @dmitryax, I had a pretty busy week last week. I am going to take a look today.

dmitryax (approved):

PR #122, opened 2022-02-04
Allow user-defined install namespace for operator

julienvincent:
As mentioned in #73 it is rather unusual for a chart to both 1) Automatically create the install namespace; and 2) Not allow changing the install namespace.

In my experience it is more idiomatic in Helm to control the install namespace by specifying the `--namespace` helm flag.

This PR:

- Removes the namespace auto-create functionality
- Removes the hardcoded namespace name and;
- Uses `{{ .Release.Namespace }}` for `metadata.namespace` and related references

Fixes #73

PR #125, opened 2022-02-22
Fix: use --collector-image instead of --otelcol-image

julienvincent:
It looks like the underlying operator image was updated to accept the flag `--collector-image` instead of `--otelcol-image` (which we currently set).

See https://github.com/open-telemetry/opentelemetry-operator/blob/7101339a6b346ec929e81cacfdbd270f04824356/main.go#L87

This updates the Chart to set the correct flag


julienvincent:
@dmitryax Looks like the tests are broken upstream again, similar to what I was seeing in #122. Can I get some help resolving that?

PR #129, opened 2022-03-02
Update operator e2e tests to always run against appVersion

### LINKED ISSUE #126: Make opentelemetry-operator tests to not fail once new operator version is released
opentelemetry-operator helm chart tests was written in a way that the validation is happening against the latest version of the operator. It breaks the opentelemetry-operator helm chart tests every time a new version of the operator is released.

For example, currently this PR is blocked https://github.com/open-telemetry/opentelemetry-helm-charts/pull/125

TylerHelmuth:
Fixes #126 

Today the lint-test workflow runs the operator e2e tests against latest.  This can cause problem if the operator latest is ahead of the chart's target version.  This change updates the lint-test workflow to always run the operator e2e tests against the operator version that matches appVersion.


dmitryax (approved):
Thanks for fixing it!

Issue #135, opened 2022-03-04
Introduce values schema for the collector chart


dmitryax:
Introduce [values.schema.json](https://helm.sh/docs/topics/charts/#schema-files) for values.yaml input validation in opentelemetry-collector helm chart. All the objects in values.yaml with predefined set of properties should not allow unknown properties.

This is a preliminary step for any big change to the helm chart configuration, e.g. https://github.com/open-telemetry/opentelemetry-helm-charts/issues/91

The following schema can be used as a reference: https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/values.schema.json

TylerHelmuth:
I can look at this.

TylerHelmuth:
There is a lot of schema to add.  Will probably submit PRs in  small chunks to make review easier since each bit of validation added is valid in itself.

dmitryax:
> There is a lot of schema to add. Will probably submit PRs in small chunks to make review easier since each bit of validation added is valid in itself.

Sounds good. Thanks!

PR #139, opened 2022-03-04
Improve values schema

TylerHelmuth:
Works towards #135 

This PR defines all the properties for opentelemetry-collector's values.schema.yaml.  Actual validations to follow.


TylerHelmuth:
The amount of changes kinda got away from me.  Let me know what you think and if I should go an add actual validations to this PR as well

charts/opentelemetry-collector/values.schema.json
```diff
@@ -2,30 +2,446 @@
...
+            "Never"
+          ]
+        }
+      }
+    },
+    "imagePullSecrets": {
+      "type": "array",
+      "items": {
+        "type": "string"
+      }
```
dmitryax, line 44: It's not a list of strings in this helm chart. Can be changed later, but currently it's a list of objects like {name: myregistrykey}

charts/opentelemetry-collector/values.schema.json
```diff
@@ -2,30 +2,446 @@
...
+        "annotations": {
+          "type": "object"
+        },
+        "name": {
+          "type": "string"
+        },
+        "rules": {
+          "type": "array",
+          "items": {
+            "type": "string"
```
dmitryax, line 96: rule is an object

dmitryax:
> Let me know what you think and if I should go an add actual validations to this PR as well or if I should make it smaller instead.

What actual validation do you mean? Isn't it enforced already?

charts/opentelemetry-collector/values.schema.json
TylerHelmuth: Updated type to be object.

charts/opentelemetry-collector/values.schema.json
TylerHelmuth: Updated type to be object.

TylerHelmuth:
Looking through the only other validation I could come up with was something fancy around standaloneCollector.enabled and podDisruptionBudget/autoscaling, but I think all that matters is that enabled is required by the properties.  The actual value of enabled doesn't matter. So I think this PR gets the schema into a good state to validate the current values.yaml.

dmitryax:
> Looking through the only other validation I could come up with was something fancy around standaloneCollector.enabled and podDisruptionBudget/autoscaling, but I think all that matters is that enabled is required by the properties. The actual value of enabled doesn't matter. So I think this PR gets the schema into a good state to validate the current values.yaml.

I don't the fancy things are that important to have. My idea is that we need to define all the properties with `"additionalProperties": false` in order to make sure that any replacements of the existing configuration parameters blocks the chart upgrade. It make any breaking change at least not silently breaking.

dmitryax (approved):
LGTM

PR #142, opened 2022-03-07
feat: support environment variables in manager

mat-rumian:
This PR adds support for configuration of additional environment variables needed  to configure Opentelemetry-Operator Manager. Thanks to that `WATCH_NAMESPACE` variable can be configured during chart deployment.

Functionality:

- If `values.manager.env` is empty then `env` is not added to the manager deployment (default)
- If `values.manager.env` contains map (KEY: value) then it's added to the manager deployment as k8s env obj

### DIFF
--- a/charts/opentelemetry-operator/templates/_helpers.tpl
+++ b/charts/opentelemetry-operator/templates/_helpers.tpl
@@ -24,8 +24,8 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 {{/*
 Manager add environment variables from map to env obj.
 */}}
-{{- define "opentelemetry-operator.envs" -}}
-{{- range $name, $value := .Values.manager.envs }}
+{{- define "opentelemetry-operator.env" -}}
+{{- range $name, $value := .Values.manager.env }}
 - name: {{ $name }}
   value: {{ $value | quote -}}
 {{- end }}
--- a/charts/opentelemetry-operator/templates/_helpers.tpl
+++ b/charts/opentelemetry-operator/templates/_helpers.tpl
@@ -19,3 +19,14 @@ Selector labels
 app.kubernetes.io/name: {{ include "opentelemetry-operator.name" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
 {{- end }}
+
+
+{{/*
+Manager add environment variables from map to env obj.
+*/}}
+{{- define "opentelemetry-operator.envs" -}}
+{{- range $name, $value := .Values.manager.envs }}
+- name: {{ $name }}
+  value: {{ $value | quote -}}
+{{- end }}
+{{- end }}
--- a/charts/opentelemetry-operator/templates/_helpers.tpl
+++ b/charts/opentelemetry-operator/templates/_helpers.tpl
@@ -19,14 +19,3 @@ Selector labels
 app.kubernetes.io/name: {{ include "opentelemetry-operator.name" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
 {{- end }}
-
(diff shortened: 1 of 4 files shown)


charts/opentelemetry-operator/templates/_helpers.tpl
```diff
@@ -19,3 +19,14 @@ Selector labels
 app.kubernetes.io/name: {{ include "opentelemetry-operator.name" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
 {{- end }}
+
+
+{{/*
+Manager add environment variables from map to env obj.
+*/}}
+{{- define "opentelemetry-operator.env" -}}
+{{- range $name, $value := .Values.manager.env }}
```
TylerHelmuth, line 28: if the values.yaml env section contains entries like:
`- testing: a-value`
it will create name value entries that look something like:
```
- name: 0
  value: "map[testing:a-value]"
```

Thats definitely user error, but it'll pass check and install ok which might not be a great user experience.  Is there anything we can do to enforce a map in the values.yaml?

TylerHelmuth: @dmitryax another use case for expanding the values.schema.json for operator charts?

mat-rumian: @TylerHelmuth We can try do something like:

```
{{/*
Manager add environment variables from map to env obj.
*/}}
{{- define "opentelemetry-operator.env" -}}
{{ $envs := .Values.manager.env }}
{{- range $name, $value := .Values.manager.env }}
{{- if kindIs "int" $name }}
{{ fail "Value for .Values.manager.env is invalid. .Values.manager.env is a map of "KEY: value" pairs e.g. KEY: value" }}
{{ else }}
- name: {{ $name }}
  value: {{ $value | quote -}}
{{- end }}
{{- end }}
{{- end }}
```

In case of `- key: value` it will return execution error with message.

TylerHelmuth: That would work but I like the simplicity of your original solution. I think we can probably enforce it with the schema and until that's in place there is your example in the values.yaml.

dmitryax:
> @dmitryax another use case for expanding the values.schema.json for operator charts?

Yes, indeed

TylerHelmuth (approved):
LGTM

charts/opentelemetry-operator/templates/_helpers.tpl
```diff
@@ -19,3 +19,14 @@ Selector labels
 app.kubernetes.io/name: {{ include "opentelemetry-operator.name" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
 {{- end }}
+
+
+{{/*
+Manager add environment variables from map to env obj.
+*/}}
+{{- define "opentelemetry-operator.env" -}}
```
dmitryax, line 27: Why do you think the helper is needed? It seems more complicated than putting the range right in the template IMO

mat-rumian: I think it's a good practice to have deployment.yaml simple as possible (readable) and all extra functions in the helpers.

TylerHelmuth: Good catch, I missed that leak. Doesn't feel right to have the helper wrap the value but then use it directly anyways.

mat-rumian: Got it! Thanks a lot for your tips. Going to provide changes.

dmitryax (approved):

PR #143, opened 2022-03-07
collector chart 0.9.4 cannot be deployed with some helm versions

bender316:
The helm chart with 0.9.4 set `additionalProperties` to false in `values.schema.json`.

Depending on the helm version used the chart fails to be deployed with the following error message:
 \- (root): Additional property global is not allowed

There is also an issue already in the helm repository:
https://github.com/helm/helm/issues/10392

Current established workaround is to add global to the allowed properties or set additionalProperties to true. I choose the first possibility

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.9.4
+version: 0.9.5
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-collector/values.schema.json
+++ b/charts/opentelemetry-collector/values.schema.json
@@ -4,6 +4,9 @@
   "title": "Values",
   "additionalProperties": false,
   "properties": {
+    "global": {
+      "type": "object"
+    },
     "nameOverride": {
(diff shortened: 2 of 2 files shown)


dmitryax:
Thanks @bender316, can you please sign the CLA?

dmitryax:
@TylerHelmuth Yes, we have it already, it's failing

TylerHelmuth:
> @TylerHelmuth Yes, we have it already, it's failing

😬

dmitryax:
@TylerHelmuth Oh I was thinking you are talking about CLA :) 

Catching issues like the one fixed in this PR makes sense. Probably we can have a matrix of popular helm versions in the [lint-test job](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/.github/workflows/lint-test.yaml).

@bender316 which helm version do you use when it's failing?

TylerHelmuth:
> Catching issues like the one fixed in this PR makes sense. Probably we can have a matrix of popular helm versions in the [lint-test job](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/.github/workflows/lint-test.yaml).

I like this idea. Maybe test against the latest helm verion and a couple previous Latest Stable Versions.

dmitryax:
> Maybe test against the latest helm version and a couple previous Latest Stable Versions.

Sounds good

bender316:
> Thanks @bender316, can you please sign the CLA?

Done.

PR #147, opened 2022-03-14
Add more ci scenarios

### LINKED ISSUE #140: Add more comprehensive CI scenarios
Currently we validate only standalone collector mode with default configuration https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/ci/standalone-collector-values.yaml

We need more test cases to cover as many configuration options as possible

TylerHelmuth:
Resolves #140. 

Adds the following scenarios for the `ct install` step of the lint-test workflow:

- Agent only collector
- Agent and Standalone collector
- Metrics collection enabled
- ClusterRole
- Agent container logs
- Disabling specific protocols
- config override (hostmetrics example)

All the extra scenarios have increased the time it takes the lint-test workflow to run from ~3 mins to ~5 mins, so I'm not against combining some of the scenarios if workflow performance is factor.


charts/opentelemetry-collector/ci/metrics-values.yaml
```diff
@@ -0,0 +1,8 @@
+ports:
+  metrics:
+    enabled: true
```
dmitryax, line 3: Not sure we need this one. This port needed for PodMonitor and ServiceMonitor only which we decided to not cover here

TylerHelmuth: Fair, only testing that the metric port can be enabled isn't very substantial.  Removed.

dmitryax (approved):

PR #148, opened 2022-03-15
Allow collector to use use hostNetwork

carlosjgp:
In the same fashion as the operator allows agents to be attached to the host network interface

This is particularly useful when configuring the address for the agents using 
```
- name: HOST_IP
  valueFrom:
    fieldRef:
      fieldPath: status.hostIP
  name: http://${HOST_IP}:4318
```

### DIFF
--- a/charts/opentelemetry-collector/templates/_pod.tpl
+++ b/charts/opentelemetry-collector/templates/_pod.tpl
@@ -128,5 +128,4 @@ affinity:
 tolerations:
   {{- toYaml . | nindent 2 }}
 {{- end }}
-hostNetwork: {{ .Values.hostNetwork }}
 {{- end }}
--- a/charts/opentelemetry-collector/templates/_pod.tpl
+++ b/charts/opentelemetry-collector/templates/_pod.tpl
@@ -128,4 +128,5 @@ affinity:
 tolerations:
   {{- toYaml . | nindent 2 }}
 {{- end }}
+hostNetwork: {{ .Values.hostNetwork }}
 {{- end }}
--- a/charts/opentelemetry-collector/values.yaml
+++ b/charts/opentelemetry-collector/values.yaml
@@ -130,9 +130,6 @@ clusterRole:
 
 podSecurityContext: {}
 securityContext: {}
-# Host networking requested for this pod. Use the host's network namespace.
-# If this option is set, the ports that will be used must be specified.
-hostNetwork: false
 
 nodeSelector: {}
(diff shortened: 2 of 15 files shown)


charts/opentelemetry-collector/templates/_pod.tpl
```diff
@@ -128,4 +128,5 @@ affinity:
 tolerations:
   {{- toYaml . | nindent 2 }}
 {{- end }}
+hostNetwork: {{ .Values.hostNetwork }}
```
TylerHelmuth, line 131: Since hostNetwork is added in the _pod.tpl, which is used in both the daemonset and deployment yamls, is there a risk that if the chart is installed with both agent and standalone collectors that the agent and deployment will experience port conflict on any nodes they are both running?

carlosjgp:
> @carlosjgp thanks for the contribution. Can you also create an Issue detailing out the new feature this PR implements?

I created this issue
https://github.com/open-telemetry/opentelemetry-helm-charts/issues/149

charts/opentelemetry-collector/values.yaml
```diff
@@ -207,6 +207,10 @@ agentCollector:
 
   podLabels: {}
 
+  # Host networking requested for this pod. Use the host's network namespace.
+  # If this option is set, the ports that will be used must be specified.
```
TylerHelmuth, line 211: Can you clarify what you mean by "ports that will be used must be specified."? I thought if hostNetwork was true that every port exposed on the pod gets exposed on the hostNetwork.  Is there extra specificity that has to happen?

TylerHelmuth (approved):
Overall looks good.  1 nit on a comment in values.yaml.

charts/opentelemetry-collector/values.yaml
carlosjgp: The `ports` blocks for `containers` on a `Pod` are optional when forwarding traffic from a `Service` but required when using `hostNetwork: true`

I copied the text from the Kubernetes API docs... Maybe we need less documentation here in this case? Since we are always setting the ports on the containers?

What do you think?

TylerHelmuth: Ya that works.  If the second comment is removed I think the setting will still be understood.

TylerHelmuth (approved):

dmitryax (approved):

dmitryax:
@carlosjgp thanks for the contribution!

PR #151, opened 2022-03-17
Added more flexible volume variables

### LINKED ISSUE #133: Make volume mounts configuration more flexible
## Problem

Currently there are 3 config parameters for k8s volumes:
- `extraConfigMapMounts`
- `extraHostPathMounts`
- `secretMounts`

They all applied to both `volumes` and `volumeMounts` in the templates. Existing configuration approach is limited and doesn't provide a way to configure many other use cases, for example:
- Ability to mount a secret to env var (https://github.com/open-telemetry/opentelemetry-helm-charts/issues/31)
- Ability to mount other type of volumes like `csi`, `emptyDir`, `persistentVolumeClaim` etc

## Proposal

Introduce two other config parameters instead:
- `extraVolumes`
- `extraVolumesMounts`

Content of these config parameters should be added as is the pod and container manifests. This give all the flexibility to be able to mount any type of volumes.

### Breaking change approach

Since this is a breaking change, we should be careful replacing the existing config parameters. The new parameters should be added along with the existing ones. Existing parameters should be removed from the values.yaml while kept for a while in the templates. And we should raise a deprecation warning if the old parameters provided by the user. The following example can be used as a reference: https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/templates/NOTES.txt#L40

TylerHelmuth:
This PR adds new `extraVolumes` and `extraVolumeMounts` variable and deprecates `extraConfigMapMounts`, `extraHostPathMounts`, and `secretMounts`.  

Going forward, `extraVolumes` and `extraVolumeMounts` will allow more flexible `volumes` and `volumeMounts` by putting the configuration in the hands of the user.

Fixes #133

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.9.6
+version: 0.10.0
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.10.0
+version: 0.11.0
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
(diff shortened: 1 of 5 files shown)


TylerHelmuth:
I've realized that we have no automated test that check that the changes to the values.yaml are actually valid. I tested locally but would be good to run it through the workflow with real values. 

Should one of our CI scenarios actually fill out every possible values.yaml setting?

dmitryax:
> Should one of our CI scenarios actually fill out every possible values.yaml setting?

That would be useful. But they have be different from the default values. Not sure it's possible to cover all options within one scenario

dmitryax (approved):
LGTM. Please update the version. The previous PR got merged sooner

TylerHelmuth:
> LGTM. Please update the version. The previous PR got merged sooner

Done

dmitryax (approved):

Issue #141, opened 2022-03-04
Add rendered folder with example renders of the charts.


TylerHelmuth:
It could be nice to have an example of the rendered charts for customers to visualize the output of running `helm template` or `helm install` without having to do it themselves.  Could also be a nice way for us to visualize the outcome of changes to charts during Pull Requests. Could use 1 or 2 of the example values.yamls from #138 .

Would want to build in something to the Github workflows to ensure that the rendered files are updated if the charts are updated so that things don't fall behind.

PR #156, opened 2022-03-21
Added examples folder with common exampes

### LINKED ISSUE #138: Add example folder with common values.yaml scenarios
It would be nice to have some example values.yaml for common scenarios customers might have.  Some scenarios could be:

- agent only configuration
- standalone only configuration
- agent and standalone
- Traces only
- Metrics only
- Logs only
- Common receivers (hostmetrics?)
- Common processors (resourceprocessor or k8sattributeprocessor?)
- Proxy configuration

Could be worth combining certain examples so we don't have too many values.yaml to maintain on breaking changes.

TylerHelmuth:
The PR adds an example folder with our 3 most common use cases.  Each use case contains the values.yaml of that use case and the generated configuration files the charts generate with the values.yaml.

Resolves #138 and #141


dmitryax (commented):
LGTM. But we also need a mechanism to keep it in sync, right?

TylerHelmuth:
What do you think would be better, to update the lint-test workflow to check if the manifest has changed when the corresponding values.yaml has change, or to add an action that will automatically re-render the manifests of all folders in `examples/` on either PR create or merge?

dmitryax:
I'm not sure if we can find a sane way to commit rendered code changes from CI so I think we should start with forcing user to do that

TylerHelmuth:
Ya my research isnt coming up with anything good for committing during the CI, looks like we'd have to write in a whole git workflow to do it.

Should be easy enough to fail the lint-test workflow if any values.yaml in examples is updated but the corresponding manifest folder isn't modified.  I'll work on that.

TylerHelmuth:
Added new workflow step to ensure every example in the examples/ folder is up-to-date.  Instead of fiddling with git and trying to identify diffs vs file adds vs file removes, I went with using helm to render each example values.yaml and compare to the committed manifest.  This will hopefully be more robust and handle example additions and deletions better.

.github/workflows/lint-test.yaml
```diff
@@ -39,6 +39,20 @@ jobs:
...
+        run: |
+          for d in */ ; do
+            echo "Checking example: ${d}"
+            helm template example ../../charts/opentelemetry-collector --values "${d}values.yaml" --output-dir "./test/${d}"
+            if diff "${d}manifest" "./test/${d}opentelemetry-collector/templates" >/dev/null; then
+                echo "Passed ${d}";
+            else
+                echo "Failed ${d}. Check that manifest/ has been rendered with the latest ${d}values.yaml";
+                exit 1                
+            fi
```
dmitryax, line 54: I think this should be replaced with a Makefile target. Which user can easily run before submitting a PR. Error message here should say to use that target as well

TylerHelmuth: Are you thinking each example would be a target that would render the manifest and flatten the result? Or a make file that would render each example?

dmitryax: I think one make target should render all the examples

TylerHelmuth: Will do

TylerHelmuth:
@dmitryax added a makefile with capabilities to generate examples and check examples.  I've not worked in makefiles very much, so please pick it apart.

Makefile
```diff
@@ -0,0 +1,29 @@
+OPENTELEMETRY_CHART = ./charts/opentelemetry-collector
+EXAMPLES_DIRECTORY = ./examples/opentelemetry-collector
+EXAMPLES = $(notdir $(shell find $(EXAMPLES_DIRECTORY) -type d -maxdepth 1 -mindepth 1))
+TMP_DIRECTORY = ./tmp
+
+.PHONY: generate-examples
+generate-examples:
+	for example in $(EXAMPLES); do \
```
dmitryax, line 8: We should add similar examples to the operator chart later. So it would be nice if this target is not restricted to `opentelemetry-collector` chart, but can generates manifests for both charts potentially.

Makefile
```diff
@@ -0,0 +1,29 @@
+OPENTELEMETRY_CHART = ./charts/opentelemetry-collector
+EXAMPLES_DIRECTORY = ./examples/opentelemetry-collector
```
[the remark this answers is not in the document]
TylerHelmuth: I like the word `examples` as the outer folder since I think it makes them easier to find.  How about `./charts/opentelemetry-collector/examples` but in each subfolder instead of `manifest` we use `rendered` ?

Makefile
TylerHelmuth: Good thinking.  I'll update it look for examples in operator also.

Makefile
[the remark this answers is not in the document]
dmitryax: Sounds good to me

Makefile
```diff
@@ -0,0 +1,75 @@
...
+generate-collector-examples: EXAMPLES=${COLLECTOR_EXAMPLES}
+generate-collector-examples: CHART_NAME=opentelemetry-collector
+generate-collector-examples: _generate-examples-helper
+
+.PHONY: generate-operator-examples
+generate-operator-examples: CHART=${OPERATOR_CHART}
+generate-operator-examples: EXAMPLES_DIR=${OPERATOR_EXAMPLES_DIR}
+generate-operator-examples: EXAMPLES=${OPERATOR_EXAMPLES}
+generate-operator-examples: CHART_NAME=opentelemetry-operator
+generate-operator-examples: _generate-examples-helper
```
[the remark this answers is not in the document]
TylerHelmuth: It's the little things like ?= and defining variables via commandline that I don't know about that make life so much easier lol. Your idea is better, I will update the makefile.

dmitryax (approved):
LGTM. Thanks!

PR #159, opened 2022-04-04
Remove two-deployment mode

### LINKED ISSUE #157: Remove two-deployment mode
As agreed in https://github.com/open-telemetry/opentelemetry-helm-charts/issues/91, we need remove the two-deployment mode and simplify the configuration.

Instead of `agentCollector` and `standaloneCollector`, we should have another config option `mode` which can have one of the following values: `daemonset` or `deployment`. All the nested configuration options from `agentCollector` and `standaloneCollector` should be moved to the root level. Some of them can be no-op based on the selected `mode`.

Ideally we need to provide backward compatibility with existing interface for now. It means that we translate `agentCollector` and `standaloneCollector` configs into the new config and show deprecation warning if they are used.

If user has both modes enabled (`agentCollector.enabled=true` and `standaloneCollector.enabled=true`) the upgrade should fail. This can be achieved by updating `schema.values.json`. For this use case we need to provide a detailed upgrade guideline on how to replace it with two separate helm installations.

TylerHelmuth:
Fixes #157 

Removes the capability to install both the agent and deployment collector at the same time.  values.yaml containing both `agentCollector.enabled=true` and `standaloneCollector.enabled=true` will fail `helm install` and `helm upgrade` commands.

The charts now support a simplified `mode=daemonset` or mode=`deployment`.  The top-level config section in the values.yaml can now be modified and will be merged into the default .Values.Config.  If there is a section of the default config you don't need you can remove it with `null`.

Single usage of `agentCollector.enabled=true` and `standaloneCollector.enabled=true` are still supported for backwards compatibility.  If `agentCollector.enabled=true` or `standaloneCollector.enabled=true` are used, their settings will take precedence over `mode`.


TylerHelmuth:
@dmitryax I _think_ this branch is in a pretty good spot with this refactor but there are so many changes I need an extra set of eyes.  Key things I tackled:

- added new `mode` and root-level configs from the agentCollector and standaloneCollector.  root-level configs are never merged with the more specific version; if agentCollector or standaloneCollector is defined, their configs are always used in place of the new root-level ones.
- updated values.schema.json.  Should prevent both mode but allow single mode and the new configurations.
- updated examples and makefile to support the new daemonset-and-deployment example that takes 2 values.yamls
- updated CI examples.  I removed the both examples bc I couldn't find a way to use CT to test both values.yamls at the same time. **don't love this**

Still need to add upgrade.md and readme changes, but wanted more eyes before I wrote documentation in case something big changes.

charts/opentelemetry-collector/templates/daemonset.yaml
```diff
@@ -14,18 +14,22 @@ spec:
...
       {{- $podData := dict "Values" $podValues "configmapSuffix" "-agent" "isAgent" true }}
       {{- include "opentelemetry-collector.pod" ($podData | mustMergeOverwrite (deepCopy .)) | nindent 6 }}
-      hostNetwork: {{ .Values.agentCollector.hostNetwork }}
+{{- $hostNetwork := .Values.hostNetwork -}}
+{{- if .Values.agentCollector.hostNetwork }}
+{{- $hostNetwork = .Values.agentCollector.hostNetwork -}}
 {{- end }}
+      hostNetwork: {{ $hostNetwork }}
+{{- end }}
+
```
dmitryax, line 35:
```suggestion

```

charts/opentelemetry-collector/templates/_helpers.tpl
```diff
@@ -110,3 +110,35 @@ Return if ingress is stable.
 {{- define "ingress.isStable" -}}
   {{- eq (include "ingress.apiVersion" .) "networking.k8s.io/v1" -}}
 {{- end -}}
+
+
+{{- define "opentelemetry-collector.podAnnotations" -}}
```
dmitryax, line 115: I don't think we can do this. It can lead to subtle bugs. We should have two helpers instead `opentelemetry-collector.deployment-podAnnotations` and `opentelemetry-collector.daemonset-podAnnotations`. Same for `opentelemetry-collector.podLabels`

charts/opentelemetry-collector/templates/configmap-agent.yaml
```diff
@@ -7,5 +7,9 @@ metadata:
     {{- include "opentelemetry-collector.labels" . | nindent 4 }}
 data:
   relay: |
+    {{- if .Values.agentCollector.enabled }}
     {{- include "opentelemetry-collector.agentCollectorConfig" . | nindent 4 -}}
+    {{- else }}
+    {{- include "opentelemetry-collector.baseConfig" . | nindent 4 -}}
+    {{- end }}
```
dmitryax, line 14: The logic spread in two places now making it hard to read. I would keep these lines untouched and make changes only in `opentelemetry-collector.agentCollectorConfig` instead

charts/opentelemetry-collector/templates/_helpers.tpl
TylerHelmuth: I can split them into deployment/daemonset specific version, but would it end up looking like

```
{{- define "opentelemetry-collector.daemonset-podAnnotations" -}}
{{- if .Values.podAnnotations }}
{{- toYaml .Values.podAnnotations }}
{{- else if .Values.agentCollector.podAnnotations }}
{{- toYaml .Values.agentCollector.podAnnotations }}
{{- end }}
{{- end }}
```

The logic is almost the same as before, would the subtle bugs you're referencing still persist?  We could merge `.Values.podAnnotations` and `.Values.agentCollector.podAnnotations`, but my philosophy so far with the refactor is to let the legacy settings completely override and of the new root-level settings.

charts/opentelemetry-collector/templates/configmap-agent.yaml
TylerHelmuth: Ok I will take another look.

charts/opentelemetry-collector/templates/_helpers.tpl
dmitryax: By subtle bugs I mean that `agentCollector.podAnnotations` potentially can be applied to daemon set and vise versa

dmitryax:
> The way I've written this refactor so far will not use their resource.limits since the new default value in the values.yaml for agentCollector.enabled is false. I think I need to set the default value back to true, but allow mode: deployment to take precedence over agentCollector

> If I do that I think it'll be tricky to deal with customers who want to use mode: daemonset and the new root-level values instead of the agentCollector values. Would probably need to have customers explicitly set agentCollector.enabled=false and but that in the UPGRADE.md.

I agree that `deployment` mode is better default than `daemonset`, but we can handle it separately. For now let's keep mode=daemonset as default with a comment that it will be replaced with `deployment` in future so we encourage users setting it explicitly.

At some point in future we can make 2-step upgrade.
1. Remove default completely and fail if `mode` is unset.
2. Few versions later we set `deployment` as default.

How does it sound?

TylerHelmuth:
> > The way I've written this refactor so far will not use their resource.limits since the new default value in the values.yaml for agentCollector.enabled is false. I think I need to set the default value back to true, but allow mode: deployment to take precedence over agentCollector
> 
> > If I do that I think it'll be tricky to deal with customers who want to use mode: daemonset and the new root-level values instead of the agentCollector values. Would probably need to have customers explicitly set agentCollector.enabled=false and but that in the UPGRADE.md.
> 
> I agree that `deployment` mode is better default than `daemonset`, but we can handle it separately. For now let's keep mode=daemonset as default with a comment that it will be replaced with `deployment` in future so we encourage users to set it explicitly.
> 
> At some point in future we can make 2-step upgrade.
> 
> 1. Remove default completely and fail if `mode` is unset.
> 2. Few versions later we set `deployment` as default.
> 
> How does it sound?

I can actually see us always leaving mode unset and forcing it to be set so that we force intentionality.  I'm going back and forth on if we should ever default to `deployment` after forcing `mode` to be set, I'm torn.

All that said, I think the more pressing matter is the comment I just left in https://github.com/open-telemetry/opentelemetry-helm-charts/issues/157.  Having to leave `agentCollector.enabled=true` is resulting in some gross values.yamls.  Checkout the examples/CI folder in this PR so see what I mean.

TylerHelmuth:
@dmitryax Updated charts to have `mode` not set by default, and to always take precedence if it is set.

TylerHelmuth:
@dmitryax ready for official review.  Lots of files, but 70% of them are examples/ or ci/.

charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -34,28 +34,40 @@ Merge user supplied top-level (not particular to standalone or agent) config int
...
 {{- end }}
 
-
 {{/*
-Build config file for agent OpenTelemetry Collector
+Build config file for daemonset OpenTelemetry Collector
 */}}
-{{- define "opentelemetry-collector.agentCollectorConfig" -}}
-{{- $values := deepCopy .Values.agentCollector | mustMergeOverwrite (deepCopy .Values)  }}
+{{- define "opentelemetry-collector.daemonset" -}}
```
dmitryax, line 40:
```suggestion
{{- define "opentelemetry-collector.daemonsetConfig" -}}
```

charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -34,28 +34,40 @@ Merge user supplied top-level (not particular to standalone or agent) config int
...
+{{- end}}
 {{- end }}
 
 {{/*
-Build config file for standalone OpenTelemetry Collector
+Build config file for deployment OpenTelemetry Collector
 */}}
-{{- define "opentelemetry-collector.standaloneCollectorConfig" -}}
-{{- $values := deepCopy .Values.standaloneCollector | mustMergeOverwrite (deepCopy .Values)  }}
+{{- define "opentelemetry-collector.deployment" -}}
```
dmitryax, line 59:
```suggestion
{{- define "opentelemetry-collector.deploymentConfig" -}}
```

charts/opentelemetry-collector/templates/_helpers.tpl
```diff
@@ -1,3 +1,7 @@
+{{- define "opentelemetry-collector.var_dump" -}}
```
dmitryax, line 1: Debug?

charts/opentelemetry-collector/templates/_helpers.tpl
```diff
@@ -110,3 +114,47 @@ Return if ingress is stable.
...
+{{- else if .Values.standaloneCollector.podLabels }}
+{{- toYaml .Values.standaloneCollector.podLabels }}
+{{- end }}
+{{- end }}
+
+{{- define "opentelemetry-collector.annotations" -}}
+{{- if or .Values.annotations .Values.standaloneCollector.annotations}}
+annotations:
+{{- if .Values.annotations }}
+  {{ toYaml .Values.annotations }}
```
dmitryax, line 155: Shouldn't it be?
```suggestion
{{- toYaml .Values.annotations | nindent 2 }}
```

Otherwise I think only first line will be aligned correctly

charts/opentelemetry-collector/templates/configmap-agent.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.agentCollector.enabled -}}
+{{- if and (not .Values.standaloneCollector.enabled) (not (eq .Values.mode "deployment")) (or .Values.agentCollector.enabled (eq .Values.mode "daemonset")) -}}
```
dmitryax, line 1: Why we need too many conditions here? This shouldn't be a place to validate deployment mode. 

```suggestion
{{- if or .Values.agentCollector.enabled (eq .Values.mode "daemonset") -}}
```

charts/opentelemetry-collector/templates/daemonset.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.agentCollector.enabled -}}
+{{- if and (not .Values.standaloneCollector.enabled) (not (eq .Values.mode "deployment")) (or .Values.agentCollector.enabled (eq .Values.mode "daemonset")) -}}
```
dmitryax, line 1: Same here

```suggestion
{{- if or .Values.agentCollector.enabled (eq .Values.mode "daemonset") -}}
```

charts/opentelemetry-collector/values.schema.json
```diff
@@ -438,28 +501,94 @@
...
+          }
+        }
+      },
+      "then": {
+        "properties": {
+          "standaloneCollector": {
+            "properties": {
+              "enabled": {
+                "const": false
+              }
```
dmitryax, line 547: Maybe we can use some templates instead of this rule and show meaningful error instead? `{{ fail "" }}` should work I believe. We can put it in NOTES.txt

charts/opentelemetry-collector/templates/NOTES.txt
```diff
@@ -8,4 +8,8 @@
 
 {{- if not (eq (toString .Values.secretMounts) "<nil>") }}
 [WARNING] "secretMounts" parameter is deprecated, please use "extraVolumes" or "extraVolumesMounts" instead.
+{{ end }}
+
+{{- if not (eq (toString .Values.mode) "<nil>") }}
+[WARNING] "standaloneCollector" and "agentCollector" are deprecated, please use "mode". See UPGRADING.md for instructions.
```
dmitryax, line 14:
```suggestion
[WARNING] "standaloneCollector" and "agentCollector" are deprecated, please use "mode". See https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/UPGRADING.md for instructions.
```

charts/opentelemetry-collector/templates/_helpers.tpl
TylerHelmuth: I think it was working bc I didn't have the leading `-` that would remove the whitespace, but explicitly using `nindent` is better.

charts/opentelemetry-collector/templates/configmap-agent.yaml
TylerHelmuth: Since `.Values.agentCollector.enabled` is true by default and we want `mode: deployment` to override it I believe we need to check that `mode != deployment`, but I believe I can remove the check against `.Values.standaloneCollector.enabled`.  That was there because mode was being set to daemonset by default.

charts/opentelemetry-collector/templates/daemonset.yaml
TylerHelmuth: Same as before

charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -34,28 +34,40 @@ Merge user supplied top-level (not particular to standalone or agent) config int
...
 {{/*
-Build config file for agent OpenTelemetry Collector
+Build config file for daemonset OpenTelemetry Collector
 */}}
-{{- define "opentelemetry-collector.agentCollectorConfig" -}}
-{{- $values := deepCopy .Values.agentCollector | mustMergeOverwrite (deepCopy .Values)  }}
+{{- define "opentelemetry-collector.daemonsetConfig" -}}
+{{- $values := deepCopy .Values.agentCollector | mustMergeOverwrite (deepCopy .Values) }}
+{{- if eq .Values.mode "daemonset" }}
+{{- $values =  deepCopy .Values }}
```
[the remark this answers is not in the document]
TylerHelmuth: I thought this too, but I was running into a strange issue where if I used := inside the if statement block it wouldn't actually modify the value of the variable outside the if statement block. Using = would change the variable so that the new value was used outside the if statement block.

charts/opentelemetry-collector/templates/_helpers.tpl
```diff
@@ -110,3 +110,47 @@ Return if ingress is stable.
...
+{{- define "opentelemetry-collector.annotations" -}}
+{{- if or .Values.annotations .Values.standaloneCollector.annotations}}
+annotations:
+{{- if .Values.annotations }}
+{{- toYaml .Values.annotations | nindent 2 }}
+{{- else if .Values.standaloneCollector.annotations }}
+{{- toYaml .Values.standaloneCollector.annotations | nindent 2 }}
+{{- end }}
+{{- end }}
+{{- end }}
```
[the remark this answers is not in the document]
TylerHelmuth: I can add that check like in the other places.

charts/opentelemetry-collector/templates/_config.tpl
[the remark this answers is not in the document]
TylerHelmuth: I've not run into this kind of issue before but this is also the first time I've done this type of "programming-like" variable override.

charts/opentelemetry-collector/templates/_helpers.tpl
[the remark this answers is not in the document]
dmitryax: Thanks. I didn't see it missing in other places, but at least these helpers should be updated to use `mode`

charts/opentelemetry-collector/templates/_config.tpl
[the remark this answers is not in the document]
dmitryax: Ok sounds good. Please verify that it's working as expected in few modern helm versions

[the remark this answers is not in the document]
TylerHelmuth: Sure, I'll check the current and last 3 stable versions. I think this was the SO post I found to help through the issue. https://stackoverflow.com/questions/57600772/is-it-possible-to-define-variables-use-if-else-condition-in-helm-chart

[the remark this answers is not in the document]
dmitryax: Nice. Thanks for posting this

charts/opentelemetry-collector/templates/_helpers.tpl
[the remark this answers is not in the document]
TylerHelmuth: Updated.  I think it's a little messier now, but these templates now also adhere to the strategy the `mode` always takes precedence.

charts/opentelemetry-collector/templates/_config.tpl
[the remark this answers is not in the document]
TylerHelmuth: Testing using `generate-charts` with versions 3.7.0, 3.7.1, 3.7.1, 3.8.0, and 3.8.1.  `if` conditions worked as expected.

charts/opentelemetry-collector/examples/daemonset-collector-logs/values.yaml
```diff
@@ -0,0 +1,9 @@
+mode: daemonset
+
+containerLogs:
+  enabled: true
+
+config:
+  receivers:
+    filelog:
+      exclude: []
```
dmitryax, line 9: nit

```suggestion
      exclude: []

```

charts/opentelemetry-collector/templates/_helpers.tpl
```diff
@@ -110,3 +110,56 @@ Return if ingress is stable.
 {{- define "ingress.isStable" -}}
   {{- eq (include "ingress.apiVersion" .) "networking.k8s.io/v1" -}}
 {{- end -}}
+
+
+{{- define "opentelemetry-collector.daemonset-podAnnotations" -}}
+{{- if eq .Values.mode "daemonset" }}
+  {{- if .Values.podAnnotations }}
```
dmitryax, line 117: Why do we have this if condition here but don't have it below after `{- else if .Values.agentCollector.podAnnotations }}`?

dmitryax: We probably don't need it since `.Values.podAnnotations` and `.Values.agentCollector.podAnnotations` both have the same `{}` default?

TylerHelmuth: The goal was that if mode was set to daemonset then all agentCollector settings would be ignored, even if they were set and even if .Values.podAnnotations was not set. Hence the funky outer if statement checking for daemonset.

TylerHelmuth: Oh I see what you mean. I would agree, we can send the empty {} to toYaml safely. Good thinking, I'll update that tomorrow.

TylerHelmuth: Since `podLabels` and `podAnnotations` are being merge with an existing label (`component`) and an existing annotation (`checksum/config`) I don’t think we can safely pipe {} to toYaml in those helpers.  It results in something like
```yaml
labels:
  {}
```
which then blows up in helm.

TylerHelmuth: Could move all the default podLabels/podAnnotations into the template and then merge them together before sending it to `toYaml`, but I think that template would get messy bc the defaults would have to be put in a dict instead of just written in the file like they are now.

dmitryax: Actually I was wrong initially. For some reason I read `{{- else if .Values.agentCollector.podAnnotations }}` as `else`. Sorry for bringing the confusion. It's all done correctly.

dmitryax (approved):

dmitryax:
Thanks @TylerHelmuth. Nicely done!

PR #165, opened 2022-04-12
Add rolling Update daemonset

juliocleal13:
Resolves https://github.com/open-telemetry/opentelemetry-helm-charts/issues/166


TylerHelmuth:
@juliocleal13 Can you open a issue that details what this PR adds/fixes?

charts/opentelemetry-collector/values.yaml
```diff
@@ -205,6 +205,8 @@ agentCollector:
   podAnnotations: {}
 
   podLabels: {}
+  rollingUpdate:
+    create: false
```
dmitryax, line 209: Why `create`? It sounds a bit confusing to me.

Given that we are moving away from agent specific configs, we should allow setting other update strategies.

I'd define interface with the following default:

```yaml
updateStrategy: RollingUpdate
rollingUpdate: {}
```

Reconfiguring rollingUpdate would be:
```yaml
rollingUpdate:
  maxSurge: 1
  maxUnavailable: 2
```

Also please update values.schema.json

juliocleal13: Hi @dmitryax , thank your for you review!
you was right. It got a bit confused use true or false in rollingUpdate (always need configure)
I changed it.
in maxSurge and maxUnavailable I used that values because are default values in daemonset, but I changed too.
I update values.schema.json.
Thank you.

charts/opentelemetry-collector/templates/daemonset.yaml
```diff
@@ -10,6 +10,10 @@ spec:
     matchLabels:
       {{- include "opentelemetry-collector.selectorLabels" . | nindent 6 }}
       component: agent-collector
+  updateStrategy:
+    rollingUpdate:
+      {{- include "opentelemetry-collector.rollingUpdate" . | nindent 6 }}
```
dmitryax, line 15: I wouldn't recommend overriding k8s's defaults

```suggestion
    {{- if or .Values.agentCollector.rollingUpdate.maxSurge .Values.agentCollector.rollingUpdate.maxUnavailable }}
    rollingUpdate:
      {{- if .Values.agentCollector.rollingUpdate.maxSurge }}
      maxSurge: {{ .Values.agentCollector.rollingUpdate.maxSurge }}
      {{- end }}
      {{- if .Values.agentCollector.rollingUpdate.maxUnavailable }}
      maxSurge: {{ .Values.agentCollector.rollingUpdate.maxUnavailable }}
      {{- end }}
    {{- end }}
```

charts/opentelemetry-collector/values.schema.json
```diff
@@ -262,6 +262,12 @@
         "annotations": {
           "type": "object"
         },
+        "updateStrategy": {
+          "type": "string"
+        },
+        "rollingUpdate": {
+          "type": "object"
```
dmitryax, line 269: Please define all `rollingUpdate` properties with `"additionalProperties": false`

juliocleal13: Adjusted

charts/opentelemetry-collector/templates/daemonset.yaml
juliocleal13: Adjusted, but, to keep default values required, i did a little different

charts/opentelemetry-collector/templates/daemonset.yaml
```diff
@@ -10,6 +10,17 @@ spec:
...
+    rollingUpdate:
+      {{- if .Values.agentCollector.rollingUpdate.maxSurge }}
+      maxSurge: {{ .Values.agentCollector.rollingUpdate.maxSurge }}
+      {{- end }}
+      {{- if .Values.agentCollector.rollingUpdate.maxUnavailable }}
+      maxUnavailable: {{ .Values.agentCollector.rollingUpdate.maxUnavailable }}
+      {{- else }}
+      maxUnavailable: 1
+      {{- end }}
+    type: {{ .Values.agentCollector.updateStrategy }}
```
TylerHelmuth, line 23: If someone set `.Values.agentCollector.updateStrategy` to `OnDelete`, what would the above `rollingUpdate` section do?  My gut says that we should only render a `rollingUpdate` section if the `updateStrategy` is set to `RollingUpdate`.

charts/opentelemetry-collector/values.schema.json
```diff
@@ -262,6 +262,28 @@
         "annotations": {
           "type": "object"
         },
+        "updateStrategy": {
+          "type": "string",
+          "additionalProperties": false
```
TylerHelmuth, line 267:
```suggestion
          "enum": [
            "RollingUpdate",
            "OnDelete"
          ]
```

TylerHelmuth:
@juliocleal13 also run `make generate-examples` so that all the examples are updated with the new chart version.

charts/opentelemetry-collector/templates/daemonset.yaml
TylerHelmuth, line 23: What happens if `updateStrategy` is set to `OnDelete`? My gut says that the `rollingUpdate` section should only be rendered if `updateStrategy` is set to `RollingUpdate`

charts/opentelemetry-collector/values.schema.json
TylerHelmuth, line 267:
```suggestion
          "enum": [
              "RollingUpdate",
              "OnDelete"
           ]
```

charts/opentelemetry-collector/templates/daemonset.yaml
juliocleal13: You're right @TylerHelmuth . Thank you, fixed.

charts/opentelemetry-collector/values.schema.json
juliocleal13: Adjusted.

charts/opentelemetry-collector/templates/daemonset.yaml
```diff
@@ -10,6 +10,19 @@ spec:
     matchLabels:
       {{- include "opentelemetry-collector.selectorLabels" . | nindent 6 }}
       component: agent-collector
+  updateStrategy:
+    {{- if not (eq .Values.agentCollector.updateStrategy "OnDelete") }}
```
TylerHelmuth, line 14:
```suggestion
    {{- if eq .Values.agentCollector.updateStrategy "RollingUpdate" }}
```
Since this section is only valid when the updateStrategy is `RollingUpdate`, lets explicitly check for it to be set (even tho its the default).

dmitryax: Agreed. Also `deployment` mode will bring another option after https://github.com/open-telemetry/opentelemetry-helm-charts/pull/159. So we need to be precise here

dmitryax:
@juliocleal13 please update your PR after changes made in https://github.com/open-telemetry/opentelemetry-helm-charts/pull/159

The additional configuration now should be moved from `agentCollector` to the root level.

PR #174, opened 2022-04-15
Update NOTES.txt to properly display warnings

### LINKED ISSUE #173: NOTES.txt is reporting "Deprecated" warning even when mode is used.
I saw in our tests that there is a "Deprecated" warning for agentCollector and standaloneCollector even when `mode` is used.  Looks like the [entry in NOTES.txt](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/fb487b09f06ff80b5a8d1249a12d1fc653937135/charts/opentelemetry-collector/templates/NOTES.txt#L13) is incorrect.

TylerHelmuth:
Fixes #173

### DIFF
--- a/charts/opentelemetry-collector/templates/NOTES.txt
+++ b/charts/opentelemetry-collector/templates/NOTES.txt
@@ -15,5 +15,5 @@
 {{ end }}
 
 {{- if and .Values.standaloneCollector.enabled .Values.agentCollector.enabled  }}
-{{ fail "[ERROR] Using 'standaloneCollector' and 'agentCollector' together is deprecated, please use 'mode'. See https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/UPGRADING.md for instructions." }}
+{{ fail "[ERROR] Using 'standaloneCollector' and 'agentCollector' together is not supported anymore. See https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/UPGRADING.md for instructions." }}
 {{ end }}
--- a/charts/opentelemetry-collector/templates/NOTES.txt
+++ b/charts/opentelemetry-collector/templates/NOTES.txt
@@ -10,7 +10,7 @@
 [WARNING] "secretMounts" parameter is deprecated, please use "extraVolumes" or "extraVolumesMounts" instead.
 {{ end }}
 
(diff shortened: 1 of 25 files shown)


charts/opentelemetry-collector/templates/NOTES.txt
```diff
@@ -10,10 +10,10 @@
...
 {{ end }}
 
-{{- if not (eq (toString .Values.mode) "<nil>") }}
+{{- if not .Values.mode }}
 [WARNING] "standaloneCollector" and "agentCollector" are deprecated, please use "mode". See https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/UPGRADING.md for instructions.
 {{ end }}
 
 {{- if and .Values.standaloneCollector.enabled .Values.agentCollector.enabled  }}
-{{ fail "[ERROR] 'standaloneCollector' and 'agentCollector' are deprecated, please use 'mode'. See https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/UPGRADING.md for instructions." }}
+{{ fail "[ERROR] Using 'standaloneCollector' and 'agentCollector' together is deprecated, please use 'mode'. See https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/UPGRADING.md for instructions." }}
```
dmitryax, line 18:
```suggestion
{{ fail "[ERROR] Using 'standaloneCollector' and 'agentCollector' together is not supported anymore. See https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/UPGRADING.md for instructions." }}
```

charts/opentelemetry-collector/templates/NOTES.txt
```diff
@@ -10,10 +10,10 @@
 [WARNING] "secretMounts" parameter is deprecated, please use "extraVolumes" or "extraVolumesMounts" instead.
 {{ end }}
 
-{{- if not (eq (toString .Values.mode) "<nil>") }}
+{{- if not .Values.mode }}
```
dmitryax, line 13: Makes sense. I thought we didn't define default mode in values.yaml at all

TylerHelmuth: values schema didn't like the null as a possible string enum, so its default is "", which is also false.

dmitryax (approved):

PR #178, opened 2022-04-18
Add test scenario for daemonset and deployment install

### LINKED ISSUE #175: Add test to lint-test workflow that checks installing the daemonset and deployment versions together.
As part of the #157 refactor we lost the ability to test two-deployment mode since it now has to be down with 2 distinct installs.  I don't see a way to do this using [chart testing](https://github.com/helm/chart-testing) so I think we will need to add something to the workflow ourselves that will install the [daemonset-and-deployment example](https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-collector/examples/daemonset-and-deployment) and confirm that it works.

TylerHelmuth:
Updates the lint-test workflow to install the daemonset-and-deployment example to test the agent -> deployment scenario that was lost as part of the work in #159.

Also updated the resources for CI scenarios to ensure no tests fail due to lack of resources.

Fixes #175

### DIFF
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -57,6 +57,12 @@ jobs:
       - name: Run chart-testing (install)
         run: ct install --target-branch main
 
+      - name: Run daemonset and deployment install test
+        run: |
+          kubectl apply -f ./charts/opentelemetry-collector/examples/daemonset-and-deployment/rendered
+          kubectl rollout status daemonset example-opentelemetry-collector-agent --timeout=2m
+          kubectl rollout status deployment example-opentelemetry-collector --timeout=2m
+
       - name: Run KUTTL smoke tests
         run: |
           until kubectl get ns opentelemetry-operator-system 2>&1 | grep "namespaces \"opentelemetry-operator-system\" not found"; do sleep 1; done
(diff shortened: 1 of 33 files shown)


.github/workflows/lint-test.yaml
```diff
@@ -57,6 +57,12 @@ jobs:
       - name: Run chart-testing (install)
         run: ct install --target-branch main
 
+      - name: Run daemonset and deployment install test
+        run: |
+          kubectl apply -f ./charts/opentelemetry-collector/examples/daemonset-and-deployment/rendered
+          kubectl rollout status daemonset example-opentelemetry-collector-agent --timeout=30s
+          kubectl rollout status deployment example-opentelemetry-collector --timeout=30s
```
dmitryax, line 64: I think we probably need to extract chart specific steps to separate CI jobs

dmitryax (approved):

.github/workflows/lint-test.yaml
dmitryax: https://github.com/open-telemetry/opentelemetry-helm-charts/issues/179

dmitryax:
Please rebase

.github/workflows/lint-test.yaml
TylerHelmuth: Want that to be part of this PR, or can this be merged and the. The breakout happen in a different PR

dmitryax: Yes, that's what I assumed

dmitryax (approved):

PR #181, opened 2022-04-20
Break out collector and operator specific tests

### LINKED ISSUE #179: Extract chart specific jobs from lint-test.yaml
Currently https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/.github/workflows/lint-test.yaml has steps related to one chart or another, and conditions are not always correct. We should consider extracting those steps into other jobs.

TylerHelmuth:
This PR breaks our collector and operator specific tests into their own workflows.  Setup steps are moved to a composite action so they can be used across the 3 different workflows.  The collector and operator workflows are only run when files in the chart have been changed.

After this change the integration tests will run regardless of whether or not the chart passes linting, but all existing lint checks are still applied via the lint workflow.  This will allow the charts to be tested even if they aren't passing the lint test, but also the PR will not be completable until linting also passes.

Fixes #179

### DIFF
--- a/.github/actions/setup/action.yaml
+++ b/.github/actions/setup/action.yaml
@@ -8,11 +8,6 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - name: Checkout
-      uses: actions/checkout@v2
-      with:
-        fetch-depth: 0
-
     - name: Set up Helm
       uses: azure/setup-helm@v1
       with:
--- a/.github/actions/setup/action.yaml
+++ b/.github/actions/setup/action.yaml
@@ -1,0 +1,30 @@
+name: Setup
+description: sets up helm lint and testing environment
+inputs:
+  create-kind-cluster:  # id of input
+    description: 'Whether or not to create a kind cluster during setup'
+    required: true
+    default: "false"
+runs:
+  using: "composite"
+  steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - name: Set up Helm
+      uses: azure/setup-helm@v1
+      with:
+        version: v3.4.1
+
+    - uses: actions/setup-python@v2
+      with:
+        python-version: 3.7
+
+    - name: Set up chart-testing
+      uses: helm/chart-testing-action@v2.0.1
+
+    - name: Create kind cluster
+      uses: helm/kind-action@v1.2.0
+      if: ${{ inputs.create-kind-cluster == 'true' }}
--- a/.github/workflows/collector-test.yaml
+++ b/.github/workflows/collector-test.yaml
@@ -11,7 +11,10 @@ jobs:
   collector-test:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
       - name: Setup
         uses: ./.github/actions/setup
         with:
--- a/.github/workflows/collector-test.yaml
+++ b/.github/workflows/collector-test.yaml
@@ -1,0 +1,30 @@
+name: Test Collector Chart
+
+on:
+  pull_request:
+    paths: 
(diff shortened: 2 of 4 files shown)


.github/workflows/collector-test.yaml
```diff
@@ -0,0 +1,30 @@
...
+      - name: Setup
+        uses: ./.github/actions/setup
+        with:
+          create-kind-cluster: "true"
+
+      - name: Run make check-examples
+        run: make check-examples
+
+      - name: Run chart-testing (install)
+        run: ct install --charts charts/opentelemetry-collector
```
dmitryax, line 24: Why splitting this instead of keeping it in lint-test?

TylerHelmuth: It could totally stay in lint-test, but it felt strange to do all the linting and some testing in one workflow, and then do other testing in different workflows. I didn't love how the tests (CI and custom) were split between different workflows. 

So I made lint-test into just lint, and move the test into their own workflows. Now if we need to add tests for collector or operator there is 1 specific workflow for each chart that focuses solely on tests.

dmitryax: What about another workflow for test, so we have:
- lint
- test
- collector-extra-test
- operator-extra-test
?

I just don't like copy pasting the it. If we want to add another chart, it has to get the same steps

TylerHelmuth: That's a good consideration. I can make that change tomorrow.

TylerHelmuth: Ah, I forgot the other reason I change it was because the operator tests need the cert manager installed. By moving it into their own tests I could eliminate the step that checked for changed files and conditionally added the cert manager. 

If we move it into its own workflow I think we'd need to add all the conditional statements back, which I thought we were trying to remove.

TylerHelmuth: I could move the cert creation step into the setup action and always create it in the test.yaml workflow, even if it is the collector chart being tested.

TylerHelmuth:
Just noticed I have duplicate checkout actions, I will fix that tomorrow

.github/workflows/collector-test.yaml
dmitryax: Ok, if cert-manager is really needed for running `ct install` on the operator, let's keep it as you've done

TylerHelmuth: Agreed. I will confirm that dependency and update the PR if it isn't needed. Hopefully it's only needed for the smoke tests 🤞

TylerHelmuth: Can confirm, the ct install tests need cert manager:

image

dmitryax: Ok sounds good. Let's keep this approach

.github/workflows/collector-test.yaml
```diff
@@ -0,0 +1,33 @@
...
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Setup
+        uses: ./.github/actions/setup
+        with:
+          create-kind-cluster: "true"
+
+      - name: Run make check-examples
+        run: make check-examples
```
dmitryax, line 24: Shouldn't this be part of lint workflow?

TylerHelmuth: Ya it should be. Forgot it wasn't collector specific.

.github/actions/setup/action.yaml
```diff
@@ -0,0 +1,25 @@
...
+    - uses: actions/setup-python@v2
+      with:
+        python-version: 3.7
+
+    - name: Set up chart-testing
+      uses: helm/chart-testing-action@v2.0.1
+
+    - name: Create kind cluster
+      uses: helm/kind-action@v1.2.0
+      if: ${{ inputs.create-kind-cluster == 'true' }}
```
dmitryax, line 25: Nice!

dmitryax (approved):

PR #183, opened 2022-04-25
Update sidecar example in Operator README

erichsueh3:
This PR updates the sidecar example in the Operator README, as it does not work. It changes the `jaegerreceiver` protocol to `thrift_compact` instead of `grpc`. This is also reflected in a PR on the `opentelemetry-operator` repository, [837](https://github.com/open-telemetry/opentelemetry-operator/pull/837).

### DIFF
--- a/charts/opentelemetry-operator/Chart.yaml
+++ b/charts/opentelemetry-operator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.6.7
+version: 0.6.8
 description: OpenTelemetry Operator Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-operator/README.md
+++ b/charts/opentelemetry-operator/README.md
@@ -257,7 +257,7 @@ spec:
     receivers:
       jaeger:
         protocols:
-          grpc:
+          thrift_compact:
     processors:
 
     exporters:


TylerHelmuth (dismissed):
Let's make the example correct by making it be OTLP instead of jaeger.  

```yaml
otlp:
  protocols:
    grpc:
    http:
...
service:
      pipelines:
        traces:
          receivers: [otlp]
          processors: []
          exporters: [logging]
```
I suggest doing the same in the operator repo's README

TylerHelmuth:
@erichsueh3 please bump the operator chart version.

erichsueh3:
> Let's make the example correct by making it be OTLP instead of jaeger.
> 
> ```yaml
> otlp:
>   protocols:
>     grpc:
>     http:
> ...
> service:
>       pipelines:
>         traces:
>           receivers: [otlp]
>           processors: []
>           exporters: [logging]
> ```
> 
> I suggest doing the same in the operator repo's README

@TylerHelmuth The example does not work with the otlp receiver, I think because the `myapp` pod uses `image: jaegertracing/vertx-create-span:operator-e2e-tests`.

TylerHelmuth:
> The example does not work with the otlp receiver, I think because the myapp pod uses image: jaegertracing/vertx-create-span:operator-e2e-tests.

I should've read down further to see the image.  Thanks for the clarification.

TylerHelmuth (commented):
Do you know why `grpc` isn't working?  [It is the default protocol according to the receiver.](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/jaegerreceiver).

erichsueh3:
> Do you know why `grpc` isn't working? [It is the default protocol according to the receiver.](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/jaegerreceiver)

What do you mean by default protocol? A protocol must be specified in the `jaegerreceiver` configuration, and in the Getting Started section of that README it says "By default, the Jaeger receiver will not serve any protocol".

TylerHelmuth:
> What do you mean by default protocol? A protocol must be specified in the `jaegerreceiver` configuration, and in the Getting Started section of that README it says "By default, the Jaeger receiver will not serve any protocol".

Sorry, "a protocol".  I would expect grpc as a protocol to work, unless the `jaegertracing/vertx-create-span:operator-e2e-tests` image doesn't know how to send over grpc.  My guess is it only knows how to send over `thrift_compact`?

TylerHelmuth (approved):
Assuming `jaegertracing/vertx-create-span:operator-e2e-tests` can't send over GRPC and can send over `thrift_compact` then this looks good to me.  I would hold off on merging until the operator PR is approved, to ensure this example stays up-to-date with any feedback from that PR.

erichsueh3:
> Assuming `jaegertracing/vertx-create-span:operator-e2e-tests` can't send over GRPC and can send over `thrift_compact` then this looks good to me. I would hold off on merging until the operator PR is approved, to ensure this example stays up-to-date with any feedback from that PR.

Yes, after testing it myself, it seems like it could only send over `thrift_compact` so it's most likely that `jaegertracing/vertx-create-span:operator-e2e-tests` can't send over GRPC.

PR #185, opened 2022-04-29
Refactoring labels and additional labels support

nlamirault:


charts/opentelemetry-collector/templates/clusterrole.yaml
```diff
@@ -5,6 +5,11 @@ metadata:
   name: {{ include "opentelemetry-collector.clusterRoleName" . }}
   labels:
     {{- include "opentelemetry-collector.labels" . | nindent 4 }}
+    {{- if or .Values.standaloneCollector.enabled (eq .Values.mode "deployment") -}}
+    app.kubernetes.io/component: standalone-collector
+    {{- else if or (and .Values.agentCollector.enabled (not (eq .Values.mode "deployment"))) (eq .Values.mode "daemonset") -}}
+    app.kubernetes.io/component: agent-collector
```
dmitryax, line 11: The terms `standalone-collector` and `agent-collector` are discouraged after we moved to the mode setting. I would say that `app.kubernetes.io/component: opentelemetry-collector` and mode potentiall can be reflected in another label if there is a real need.

dmitryax (commented):
Please provide reasoning why new labels are needed in the PR description or create an issue

charts/opentelemetry-collector/templates/clusterrole.yaml
TylerHelmuth: Yes, `app.kubernetes.io/component : opentelemetry-collector` would be preferred in all resources.  If the mode (currently `deployment` and `daemonset`) also needs reflected in a label that would be a different label.

nlamirault:
@dmitryax i make some new commits

PR #197, opened 2022-05-06
Support pod lifecycle hooks

### LINKED ISSUE #193: Allow setting lifecycle policies to collector pods
When scaling down nodes in a Kubernetes cluster, Python services using Opentelemetry libraries tend to hang on shutdown, because the Otel collector pod (daemonset managed) gets terminated before the Python service manages to flush its final traces to the collector. 

To make the error less likely, it would be great, if the collector helm chart would allow using Kubernetes lifecycle hooks, so that I could set a `preStop` commad to wait for some time before terminating the pod.

Looks like the container image is based on `scratch` and has no tools installed, so support for init containers to copy a statically linked `sleep` command to the main container would also be needed to support this use case.

juissi-t:
Fixes #193 

This PR adds support for Pod lifecycle hooks and Init containers. Use case in mind is using the init container to copy some binary to a shared volume, which the main container can use as part of a preStop hook to e.g. delay shutdown for some time. When running in daemonset mode, it allows pods in the same worker to finish their works and send in any outstanding traces to the collector before the collector process is terminated.

Tested with these values:

```yaml
initContainers:
  - name: test
    command:
      name: cp
      extraArgs:
      - /bin/sleep
      - /test/sleep
    image:
      repository: busybox
      pullPolicy: IfNotPresent
      tag: latest

extraVolumes:
  - name: test
    emptyDir: {}

extraVolumeMounts:
  - name: test
    mountPath: /test

lifecycleHooks:
  postStart:
    exec:
      command:
      - /test/sleep
      - "5"
```

There's some documentation on how to use the new values in `values.yaml`, as well as an example in the `examples` directory.


charts/opentelemetry-collector/values.schema.json
```diff
@@ -257,6 +263,9 @@
         }
       }
     },
+    "lifecycleHooks": {
```
TylerHelmuth, line 266: Let's restrict this to only allowing the `PostStart` and `PreStop` properties.

charts/opentelemetry-collector/values.schema.json
```diff
@@ -145,6 +145,12 @@
     "priorityClassName": {
       "type": "string"
     },
+    "initContainers": {
+      "type": "array",
+      "items": {
+        "type": "object"
```
TylerHelmuth, line 151: Is there ever a time an object in this array would contain more properties than:

```yaml
  - name: test
    command:
      name: cp
      extraArgs:
      - /bin/sleep
      - /test/sleep
    image:
      repository: busybox
      pullPolicy: IfNotPresent
      tag: latest
```

If not, I'd opt to enforce this object's properties.

TylerHelmuth (commented):
Great stuff.  Please also bump the chart's patch version, and run `make generate-examples`

charts/opentelemetry-collector/values.schema.json
juissi-t: I added `exec` and `httpGet` type handlers (only one can be active) with their possible properties, as defined here: https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#LifecycleHandler

charts/opentelemetry-collector/values.schema.json
juissi-t: I added the properties to the schema. It's entirely possible that someone might require env variables or perhaps mounting secrets to the init container, but in that case the pod construct would need to be changed to add support for them, so I would say this is enough for now.

juissi-t: The current version of this PR matches my use case, can't speak for anyone else. Secrets and env vars are just the things that I've ran into when dealing with init containers previously, but obviously even those won't handle 100% of the possible use cases there are.

I can add support for container schema. What levels should I add: just the top level properties, or all they way down?

dmitryax: @TylerHelmuth, why `initContainers.enabled` is needed? can't we can check if `initContainers` list is empty instead?

TylerHelmuth: We could, and if this is the only place it would get used (not part of a preset or used in some other decision making somewhere) then that would be best.  It is in line with some other objects in the values.yaml.

juissi-t: Thanks for the comments, I will update this pull request tomorrow with the schema changes.

juissi-t: Understood. I'll make the relevant changes there as well.

juissi-t: I updated the `initContainers` schema to be a list of objects, which only require `name` property (which seems to be the only required property for a container) but allow any other ones. The objects are rendered in `_pod.tlp` using `toYaml` if the `initContainers` list is not empty.

TylerHelmuth (approved):
Please bump the chart version and run `make generate-examples`

juissi-t:
> Please bump the chart version and run `make generate-examples`

Both done and PR updated.

dmitryax (approved):

Issue #158, opened 2022-03-30
Replace default config with config presets


dmitryax:
Instead of providing default configuration, we should have presets of configurations that are applied to the collector. Suggested presets configuration:

```yaml
presets:
  traces:
    enabled: true
  metrics:
    enabled: false
    hostMetrics:
      enabled: false
    selfScraping:
      enabled: true
  logs:
    enabled: false
    containerLogs: 
      enabled: false
```

User also must be able to provide their own config that must be applied after all the presets.

```yaml
config: {}
# receivers: ...
# processors: ...
# exporters: ...
# service: ...
```

Each of the enabled presets and user custom config can be provided to the collector as a configuration argument, no need to merge them in the helm chart.

TylerHelmuth:
The charts now support removing the default config [using null](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/ci/config-override-values.yaml).  User are able to wipe remove all default configuration and replace it with their own.  Does that meet the needs of this issue?

dmitryax:
Yes, I think we still need to move to the presets as a cleaner way than setting nulls. We already have a preset for logs collection, I think we need to apply similar approach to other use cases.

Some presets can require rendering of other resources not configmap only, e.g. `hostMetrics`

TylerHelmuth:
> Some presets can require rendering of other resources not configmap only, e.g. hostMetrics

Like using the presets to determine if we need to render extra permissions or something?

> no need to merge them in the helm chart.

What do you mean by this? 

Finally, if a customer enabled host metrics (or something else also preset) but also defined that component in their config then which wins? I would say their version of the hostMetrics config be what renders in the configmap.

dmitryax:
> Like using the presets to determine if we need to render extra permissions or something?

hostMetrics requires host volume mounts 

> What do you mean by this?

Collector can accept many config files https://github.com/open-telemetry/opentelemetry-collector/blob/150c1ede2f7fb5607035d3c7a10bfcab61c39afe/service/flags.go#L47-L48. I think we should supply each preset as another configuration file and let collector do the merging instead of using helm capabilities. Let me know if it makes sense to you

> Finally, if a customer enabled host metrics (or something else also preset) but also defined that component in their config then which wins? I would say their version of the hostMetrics config be what renders in the configmap.

I believe what user provides in `config` should have precedence over all the presets

TylerHelmuth:
[...]

PR #200, opened 2022-05-06
Fix collector-logs example

TylerHelmuth:
Currently the collector logs example configures the file log receiver, but doesn't add it to the pipeline.  This PR updates the example's values.yaml to list filelog in the log pipeline's receivers.

This issue arose from the `mode` refactor, where `.Values.config` needed to be applied after all other configuration.  In doing so, the default logs pipeline in `.Values.config` is [overwriting the list from the containerLog preset](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/f811817298ea434074d0457d68d7b4dfd4b50f7a/charts/opentelemetry-collector/templates/_config.tpl#L194).  

This is the same issue we've been discussing in #158 around presets and custom config.  Luckily, the proposal in #158 will do away with any default values in `.Values.config`.  If there is no `service.pipelines.logs.receivers` entry in `.Values.config`, then the `containerLogs` preset would work as expected.  Adding a filelog entry to  `.Values.Config.service.pipelines.logs.receivers` would only be necessary if there is some other logs receiver being configured in `.Values.config` while the `containerLogs` preset is being used.

### DIFF
--- a/charts/opentelemetry-collector/examples/daemonset-collector-logs/values.yaml
+++ b/charts/opentelemetry-collector/examples/daemonset-collector-logs/values.yaml
@@ -7,3 +7,9 @@ config:
   receivers:
     filelog:
       exclude: []
+  service:
+    pipelines:
+      logs:
+        receivers:
+          - otlp
+          - filelog
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.16.0
+version: 0.16.1
 description: OpenTelemetry Collector Helm chart for Kubernetes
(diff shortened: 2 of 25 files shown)


dmitryax (approved):

charts/opentelemetry-collector/examples/daemonset-collector-logs/values.yaml
```diff
@@ -7,3 +7,9 @@ config:
   receivers:
     filelog:
       exclude: []
+  service:
+    pipelines:
+      logs:
+        receivers:
+          - otlp
+          - filelog
```
dmitryax, line 15: Once presets are done properly in https://github.com/open-telemetry/opentelemetry-helm-charts/issues/158, this whole config should not be needed anymore

dmitryax (approved):

PR #202, opened 2022-05-12
Add PrometheusRule

ben-natan:
This PR adds a an optional PrometheusRule to the opentelemetry-collector chart.

Signed-off-by: Adrien Bennatan <adrien.ben-natan@ensta-paris.fr>


TylerHelmuth:
Hello @ben-natan, thanks for the contribution.  Can you open an issue detailing the issue/feature request this PR solves?  Also, please run `make generate-examples`

dmitryax:
Why is this supposed to be part of the helm chart? Can't you create the prometheus rule separately?

ben-natan:
> Why is this supposed to be part of the helm chart? Can't you create the prometheus rule separately?

I could create the rule separately but it is not convenient.
This chart already provides `ServiceMonitor` and `PodMonitor` which are both Prometheus CRDs. Adding `PrometheusRule` is an extra step in the same direction.

TylerHelmuth:
@dmitryax what do you think?

dmitryax:
Sounds reasonable to me

TylerHelmuth:
@ben-natan please update `charts/opentelemetry-collector/README.md` stating that at this time Prometheus CRDs are supported but other CRDs are not.

Also, please rebase.

charts/opentelemetry-collector/templates/prometheusrule.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if .Values.prometheusRule.enabled }}
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: {{ include "opentelemetry-collector.fullname" . }}
+  labels:
+    {{- include "opentelemetry-collector.labels" . | nindent 4 }}
+    {{- range $key, $value := .Values.prometheusRule.extraLabels }}
+    {{- printf "%s: %s" $key (tpl $value $ | quote) | nindent 4 }}
```
dmitryax, line 9: is `tpl` needed? Will this work instead?
```suggestion
    {{- printf "%s: %s" $key (quote $value) | nindent 4 }}
```

TylerHelmuth: If the suggestion works, then [servicemonitor.yaml](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/templates/servicemonitor.yaml) and [podmonitor.yaml](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/templates/podmonitor.yaml) should be updated to match.

TylerHelmuth: Although I think changing servicemonitor.yaml and podmonitor.yaml would be a breaking change.  Isn't that tpl function letting user add things like `{{ .Values.fullnameOverride }}` in the `.extraLabels` section?

dmitryax: Oh ok, we can keep his for consistency them. probably can be addressed separately.

> Isn't that tpl function letting user add things like {{ .Values.fullnameOverride }} in the .extraLabels section?

I don't think so. I don't really see a reason to have `tpl` here

TylerHelmuth (approved):

TylerHelmuth:
@dmitryax please review

dmitryax (approved):

PR #212, opened 2022-05-20
Add annotations to DaemonSet

carlosjgp:
I've also changed the way annotations are "calculated", especially for `.Values.annotations` + `.Values.standaloneCollector.annotations` which I feel would help to migrate away from the standalone setup but I'm happy to add a more specific change for the DaemonSet only

## Update

`.Values.standaloneCollector.annotations` are meant to be ignored when `.Values.annotations` are set. Code updated accordingly


TylerHelmuth (commented):
Please run `make generate-examples`

charts/opentelemetry-collector/templates/_helpers.tpl
```diff
@@ -153,13 +153,12 @@ Return if ingress is stable.
...
-{{- if eq .Values.mode "deployment" }}
-{{- if .Values.annotations }}
+{{- if or .Values.annotations .Values.standaloneCollector.annotations -}}
 annotations:
+{{- if .Values.annotations }}
 {{- .Values.annotations | toYaml | nindent 2  }}
-{{- end }}
-{{- else if .Values.standaloneCollector.annotations }}
-annotations:
+{{- else }}
```
carlosjgp, line 160: As per conversation when `.annotations` is set `.standaloneCollector.annotations` is ignored

charts/opentelemetry-collector/templates/_helpers.tpl
```diff
@@ -153,13 +153,13 @@ Return if ingress is stable.
...
-{{- if .Values.annotations }}
+{{- if or .Values.annotations .Values.standaloneCollector.annotations -}}
 annotations:
+{{- if .Values.annotations }}
 {{- .Values.annotations | toYaml | nindent 2  }}
 {{- end }}
-{{- else if .Values.standaloneCollector.annotations }}
-annotations:
+{{- if .Values.standaloneCollector.annotations }}
 {{- .Values.standaloneCollector.annotations | toYaml | nindent 2  }}
```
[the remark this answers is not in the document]
carlosjgp: regarding `podLabels` it could be consolidated into one function making it easier to maintain
```
{{- define "opentelemetry-collector.daemonset-podLabels" -}}
{{- if eq .Values.mode "daemonset" }}
  {{- if .Values.podLabels }}
  {{- .Values.podLabels | toYaml }}
  {{- end }}
{{- else if .Values.agentCollector.podLabels }}
{{- .Values.agentCollector.podLabels | toYaml }}
{{- end }}
{{- end }}

{{- define "opentelemetry-collector.deployment-podLabels" -}}
{{- if eq .Values.mode "deployment" }}
  {{- if .Values.podLabels }}
  {{- .Values.podLabels | toYaml }}
  {{- end }}
{{- else if .Values.standaloneCollector.podLabels }}
{{- .Values.standaloneCollector.podLabels | toYaml }}
{{- end }}
{{- end }}
```
==>
```
{{- define "opentelemetry-collector.podLabels" -}}
{{- if or .Values.podLabels .Values.agentCollector.podLabels -}}
{{- .Values.podLabels | toYaml }}
{{- else }}
{{- .Values.agentCollector.podLabels | toYaml }}
{{- end }}
{{- end }}
```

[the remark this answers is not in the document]
TylerHelmuth: helm doesn't know how to short circuit right?  In the consolidated template, if `.Values.agentCollector.podLabels` is truthy but `.Values.podLabels` is falsy won't it pass the if condition and then use `.Values.podLabels` empty value?  Also it looks like this breaks backwards compatibility for `.Values.standaloneCollector.podLabels`

charts/opentelemetry-collector/templates/_helpers.tpl
TylerHelmuth: This version will still allow `.Values.standaloneCollector.annotations` values to be used when `mode` is set but `.Values.annotations` is not.  It's an unlikely scenario but technically possible.

I know this pain of supporting backwards compatibility with agentCollector and standaloneCollector lol if you're willing to wait a couple days, I will be removing agentCollector and standaloneCollector support as soon as collector 0.52.0 releases this week, which will make this this feature add simpler.

charts/opentelemetry-collector/templates/_helpers.tpl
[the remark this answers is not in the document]
carlosjgp: Sorry, I missed one line to check which was set... but after rereading all the comments I came up with the [change](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/212/commits/97fd6f0ea62dca35daf037276df46952a5b91fd8)

Which consolidates labels and annotations following:
- only use `podLabels`, `podAnnotaions` and `annotations` if `mode` is set with any value (not an empty string)
- only use `standaloneCollector.podLabels`, `standaloneCollector.podAnnotations` and `standaloneCollector.annotations` if `standaloneCollector.enabled = true`

charts/opentelemetry-collector/templates/_helpers.tpl
```diff
@@ -111,54 +111,27 @@ Return if ingress is stable.
...
-{{- define "opentelemetry-collector.deployment-podAnnotations" -}}
-{{- if eq .Values.mode "deployment" }}
-  {{- if .Values.podAnnotations }}
-  {{- .Values.podAnnotations | toYaml }}
-  {{- end }}
-{{- else if .Values.standaloneCollector.podAnnotations }}
+{{- define "opentelemetry-collector.podAnnotations" -}}
+{{- if and .Values.podAnnotations .Values.mode  }}
+{{- .Values.podAnnotations | toYaml }}
+{{- else if and .Values.standaloneCollector.enabled .Values.standaloneCollector.podAnnotations }}
```
TylerHelmuth, line 117: This change breaks backwards compatibility with .Values.agentCollector.podAnnotations

charts/opentelemetry-collector/templates/_helpers.tpl
```diff
@@ -111,54 +111,27 @@ Return if ingress is stable.
...
 {{- .Values.standaloneCollector.podAnnotations | toYaml }}
 {{- end }}
 {{- end }}
 
-{{- define "opentelemetry-collector.daemonset-podLabels" -}}
-{{- if eq .Values.mode "daemonset" }}
-  {{- if .Values.podLabels }}
+{{- define "opentelemetry-collector.podLabels" -}}
+{{- if and .Values.podLabels .Values.mode }}
```
TylerHelmuth, line 123: This change breaks backwards compatibility with .Values.agentCollector.podAnnotations

TylerHelmuth: The consolidation is still off.  With the removal of

```
{{- define "opentelemetry-collector.deployment-podAnnotations" -}}
{{- if eq .Values.mode "deployment" }}
  {{- if .Values.podAnnotations }}
  {{- .Values.podAnnotations | toYaml }}
  {{- end }}
{{- else if .Values.standaloneCollector.podAnnotations }}
{{- .Values.standaloneCollector.podAnnotations | toYaml }}
{{- end }}
{{- end }}
```

I don't see where users of .Values.standaloneCollector.podAnnotations would have their podAnnotations set.

I will be submitting a PR today to remove agentCollector and standaloneCollector.  Once that is out of the way adding pod annotations for daemonsets shouldn't need to worry about consolidation anymore.

PR #216, opened 2022-05-26
Remove agentCollector and standaloneCollector

### LINKED ISSUE #199: Remove Deprecated agentCollector and standaloneCollector configuration
The `agentCollector` and `standaloneCollector` configuration sections in `values.yaml` were deprecated as part of the 0.14.0 release of the charts (#159). These sections should be removed from `values.yaml` and all references to them in the templates should be cleaned up.  `NOTES.txt` should be updated to fail any install that attempts to use `agentCollector` or `standaloneCollector`.

This work should not be completed until the collector is at version 0.52.0 (current version is 0.50.0) to give time for existing users of the charts to switch to using `mode`.

Open question: should `mode` be defaulted to `daemonset`?  Once `agentCollector` is removed, the charts will not be able to be installed with the default `values.yaml` unless we set a default value for `mode`.

TylerHelmuth:
This PR removes the deprecated `agentCollector` and `standaloneCollector` sections from values.yaml.  Instead, `mode` must be used.  

An upgrade/install of the chart without `mode: daemonset` or `mode: deployment` will fail.  Since the current values.yaml uses `.Values.agentCollector.enabled: true`, users of the default values are using `agentCollector` settings. In order to ensure that users who use the default values.yaml are not hit with any new surprise values (the default `requests` are different for `mode`), `mode` has been left blank by default.  As a result, the default values.yaml cannot be used to install the chart; all users **must** provide a values.yaml that sets `mode` for the time being.

Users who have switched to `mode` as of 0.14.0 should not be impacted by this change.

Fixes #199


TylerHelmuth:
@dmitryax please take a look when you're free.

charts/opentelemetry-collector/examples/daemonset-and-deployment/rendered/service.yaml
```diff
@@ -42,4 +42,4 @@ spec:
   selector:
     app.kubernetes.io/name: opentelemetry-collector
     app.kubernetes.io/instance: example
-    component: standalone-collector
+    component: deployment-collector
```
dmitryax, line 45: I would not change the labels at this point since it breaks the upgrade and forces user to reinstall the chart. deployment-collector doesn't sound much better then standalone-collector. I would recommend handling this separately

TylerHelmuth: Ok I can set it back

TylerHelmuth: @dmitryax done.

dmitryax (approved):

PR #219, opened 2022-05-30
Add nodeport to ports configuration

kostya9:
Add nodePort entry to ports configuration.

This allows us to specify the exact port to use by the Service (type=NodePort).

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.19.0
+version: 0.18.1
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.18.0
+version: 0.19.0
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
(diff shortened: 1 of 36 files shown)


TylerHelmuth:
I am out for Memorial Day but will review tomorrow.

charts/opentelemetry-collector/values.schema.json
```diff
@@ -215,6 +215,9 @@
             "hostPort": {
               "type": "integer"
             },
+            "nodePort": {
+                "type": "integer"
+              },
```
TylerHelmuth, line 220:
```suggestion
            },
```

TylerHelmuth (commented):
Can you add in the values.yaml in the `port` section a commented out `nodePort:` for each port?  Also a comment that nodePort is only used with deployment mode.

charts/opentelemetry-collector/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.18.0
+version: 0.19.0
```
TylerHelmuth, line 3:
```suggestion
version: 0.18.1
```

kostya9:
Thanks for the review, addressed your comments in the last commit

charts/opentelemetry-collector/values.yaml
```diff
@@ -155,36 +155,43 @@ ports:
     containerPort: 4317
     servicePort: 4317
     hostPort: 4317
+    # nodePort could only be used with the deployment mode
```
TylerHelmuth, line 158:
```suggestion
    # nodePort can only be used with deployment mode
```

TylerHelmuth (approved):

charts/opentelemetry-collector/values.yaml
kostya9: fixed

TylerHelmuth:
@dmitryax please take a look when you're free.

TylerHelmuth:
@kostya9 can you take a look at the Dmitrii's feedback and also merge latest from main?  I'll need to bump the chart version and regenerate the examples probably.

Issue #225, opened 2022-06-09
Update open-telemetry-collector chart image to 0.53.0


hex1848:
The opentelemetry-collector and opentelemetry-collector-config repos got updated to 0.53.0.   The opentelemetry-collector chart should be updated to reflect this change.

PR #226, opened 2022-06-09
Update default opentelemetry-collector version to 0.53.0

hex1848:
This PR updates the default collector image version to 0.53.0.  Taking a look at the opentelemetry-collector and opentelemetry-collector-config changes for 0.51.0 and 0.52.0, It does not appear that there are any breaking changes.

Fixes:  https://github.com/open-telemetry/opentelemetry-helm-charts/issues/225


dmitryax (approved):
Thanks for taking care of it

PR #235, opened 2022-06-20
Allow operator pod to start when admissionWebhooks are disabled

Allex1:


Allex1:
@dmitryax @TylerHelmuth @tigrannajaryan I signed the CLA and the commit. Can you start the workflows pls ?

TylerHelmuth:
@Allex1 please add an issue or description to this PR describing what this PR changes and why it is needed.  Also, please take latest from main and resolve any conflicts.

TylerHelmuth:
@Allex1 it looks like your commits are not associated with your github user id.

Allex1:
#https://github.com/open-telemetry/opentelemetry-helm-charts/issues/238

TylerHelmuth (approved):
@dmitryax

charts/opentelemetry-operator/values.yaml
```diff
@@ -49,7 +49,7 @@ kubeRBACProxy:
       memory: 64Mi
 
 ## Admission webhooks make sure only requests with correctly formatted rules will get into the Operator.
-##
+## Disabling webhooks requires the ENABLE_WEBHOOKS env var to be set to false in order for the operator pod to start.
```
dmitryax, line 52: Should this env var just be set in the helm template instead of asking users for setting it manually?

Allex1: @dmitryax The env var can be set trough the helm values [here](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/values.yaml#L33)

Allex1: mm, not really sure. It doesn't feel natural to set env vars for users, but I can do it ...

dmitryax: Do you see a use case when `ENABLE_WEBHOOKS` shouldn't be set to false while `Values.admissionWebhooks.enabled=false`?

charts/opentelemetry-operator/templates/deployment.yaml
```diff
@@ -58,10 +58,12 @@ spec:
             initialDelaySeconds: 5
             periodSeconds: 10
           resources: {{ toYaml .Values.manager.resources | nindent 12 }}
+          {{- if and .Values.admissionWebhooks.enabled .Values.admissionWebhooks.certManager.enabled }}
```
dmitryax, line 61: Why not `{{- if .Values.admissionWebhooks.enabled }}`? Does setting `admissionWebhooks=false` and `admissionWebhooks.certManager=true` has any meaning/effect?

dmitryax: nvm. I misunderstood this

charts/opentelemetry-operator/values.yaml
Allex1: @dmitryax set the env var when both values set to false

charts/opentelemetry-operator/templates/deployment.yaml
```diff
@@ -32,6 +32,11 @@ spec:
             {{- end }}
           command:
             - /manager
+          {{- if not (and .Values.admissionWebhooks.enabled .Values.admissionWebhooks.certManager.enabled) }}
```
dmitryax, line 35: Shouldn't it be `{{- if .not Values.admissionWebhooks.enabled }}`? Is it valid/useful configuration when `ENABLE_WEBHOOKS` isn't set to `false` but `.Values.admissionWebhooks.enabled=true and .Values.admissionWebhooks.certManager.enabled=false`?

Allex1: So certmanager is the main dependency for to secrets volume to be created/mounted and the pod to start with webhooks enabled. So if we want to disable the webhooks we need all 3 flags disabled. I mean we could have certmanager enabled but I don't see the point as it has only this function. 
Anyway, give me some more time as the current pr can have a duplicate `env` key which I need to fix.

Allex1: But to answer your question yes there is a valid use case for what you said: webhooks enabled and the tls secret provided manually without certmanager.

Allex1: @dmitryax I think this is the best version so far. Problem is that it introduces a breaking change for current users that have env vars defined.
Let me know your thoughts.

charts/opentelemetry-operator/templates/deployment.yaml
```diff
@@ -32,9 +32,13 @@ spec:
             {{- end }}
           command:
             - /manager
-          {{- if .Values.manager.env }}
+          {{- if .Values.manager.env.enabled }}
```
dmitryax, line 35: Why is `env.enabled` added? I believe it should work as it used to be. If not, we may need `env: []` to be set as default value in `values.yaml`

charts/opentelemetry-operator/templates/deployment.yaml
dmitryax: I don't think it's going to be a significant breaking change if we don't introduce `env.enabled`.

dmitryax: Otherwise, the PR is in na good shape

charts/opentelemetry-operator/templates/deployment.yaml
Allex1: Before there was a corner case where you would have ended up with duplicate `env` key.
If we just put `env: {}` and don't specify any kv we end up with an empty `var` key in the deployment.

dmitryax: You can use this line instead?

```suggestion
          {{- if or (not .Values.admissionWebhooks.enabled) .Values.manager.env }}
```

Allex1: I'm pretty sure this will set the `ENABLE_WEBHOOKS: false` for any variable kv defined (only second condition matching)

dmitryax: Yes, if you remove another to check of `.Values.admissionWebhooks.enabled` inside this clause, but that's not what I'm proposing

Allex1: ok, so basically checking `.Values.admissionWebhooks.enabled` twice.

charts/opentelemetry-operator/values.yaml
```diff
@@ -29,9 +29,11 @@ manager:
       cpu: 100m
       memory: 64Mi
   ## Adds additional environment variables
-  ## e.g ENV_VAR: env_value
+  ## e.g ENV_VAR: ENV_VALUE
   env:
-
+    enabled: false
+    vars:
+      ENV_VAR: ENV_VALUE
```
dmitryax, line 36: Please revert changes to this file

Allex1:
@dmitryax I think this is your recommendation. Have another look.

dmitryax (approved):

Issue #239, opened 2022-06-21
Make ClusterRoles deployment configurable


Allex1:
In some multi tenant Kubernetes environments the user cannot load the ClusterRoles and ClusterRoleBindings during normal deployment. In this case we sideload them  via a different process before the actual helm deployment. For this to be possible we should have a feature flag (enabled by default) that allow the use to choose whether to load these resources while deploying the helm release.

PR #240, opened 2022-06-21
Make ClusterRoles deployment configurable

Allex1:
Addresses https://github.com/open-telemetry/opentelemetry-helm-charts/issues/239


charts/opentelemetry-operator/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.8.2
+version: 0.8.4
```
TylerHelmuth, line 3:
```suggestion
version: 0.8.3
```

charts/opentelemetry-operator/templates/clusterrole.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.rbac.create }}
```
TylerHelmuth, line 1: There are multiple ClusterRoles defined in this file.  Is there value in being able to enabled/disable each cluster role individually?  Or does it not ever make sense with the operator to have 1 cluster role but not the others.

charts/opentelemetry-operator/templates/clusterrolebinding.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.rbac.create }}
```
TylerHelmuth, line 1: Same questions here

charts/opentelemetry-operator/values.yaml
```diff
@@ -62,6 +62,11 @@ admissionWebhooks:
       # kind:
       # name:
 
+## Create the provided ClusterRoles and ClusterRoleBindings
+##
+rbac:
+  create: true
```
TylerHelmuth, line 68:
```suggestion
  enabled: true
```
The charts usually use `enabled` as the field to turn things on and off

charts/opentelemetry-operator/templates/clusterrole.yaml
Allex1: I would have the flag to switch between all or nothing. The issue in a multi tenant k8s cluster is that you cannot deploy ClusterRole at all.

charts/opentelemetry-operator/templates/clusterrolebinding.yaml
Allex1: I would have the flag to switch between all or nothing. The issue in a multi tenant k8s cluster is that you cannot deploy ClusterRole at all.

TylerHelmuth (approved):
@dmitryax

charts/opentelemetry-operator/values.yaml
```diff
@@ -62,6 +62,11 @@ admissionWebhooks:
       # kind:
       # name:
 
+## Create the provided ClusterRoles and ClusterRoleBindings
+##
+rbac:
+  enabled: true
```
dmitryax, line 68: I think we should use the same interface as we have in the collector chart:

```
clusterRole:
  create: true
```

dmitryax: Unless you you think that that interface is not the best (we will need to change collector chart in that case then)

Allex1: @dmitryax changed it to your suggestion

dmitryax (approved):

Allex1:
@dmitryax @TylerHelmuth fixed conflicts,  pls restart CI

dmitryax (approved):

PR #241, opened 2022-06-23
Add helper for podDisruptionBudget API version

n-oden:
The `policy/v1beta1` api is deprecated as of k8s 1.22 and will be fully
removed in 1.25:

https://kubernetes.io/docs/reference/using-api/deprecation-guide/#poddisruptionbudget-v125

The only behavior difference between the beta and v1 APIs is how empty
`spec.selector` blocks are treated, which is not applicable here.

Signed-off-by: Nathan J Mehl <n@oden.io>


dmitryax (approved):

dmitryax:
@n-oden thanks for the contribution!

Issue #238, opened 2022-06-20
controller-manager fails to start with admissionWebhooks disabled


Allex1:
Context:
We need the operator to start without webhooks enabled as we are running in a multi tenant Kubernetes platform where webhooks are currently disabled for users. We still want to use the operator for creating opentelemetry-collector deployments which do not require webhooks.

When trying to start the controller-manager with `admissionWebhooks.enabled=false` or `admissionWebhooks.certManager.enabled=false` the pod fails with:
```MountVolume.SetUp failed for volume "cert" : secret "opentelemetry-operator-controller-manager-service-cert" not found```
which makes these options unusable.

PR #245, opened 2022-06-27
Allow operator pod to start when admissionWebhooks are disabled ADDENDUM

Allex1:
It seems I managed to broke master. While testing today I get :
```Error: INSTALLATION FAILED: Deployment in version "v1" cannot be handled as a Deployment: json: cannot unmarshal bool into Go struct field EnvVar.spec.template.spec.containers.env.value of type string```

This pr fixes it. 

Initial issue is https://github.com/open-telemetry/opentelemetry-helm-charts/issues/238

### DIFF
--- a/charts/opentelemetry-operator/Chart.yaml
+++ b/charts/opentelemetry-operator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.8.5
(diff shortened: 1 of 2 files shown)


TylerHelmuth (approved):

Allex1:
> @Allex1 please bump the chart patch version

@TylerHelmuth done

Issue #252, opened 2022-06-29
Add support for ServiceMonitor


Allex1:
Add support for ServiceMonitor in order for Prometheus to be able to scrape the manager metrics endpoint.

PR #253, opened 2022-06-29
Add support for ServiceMonitor

Allex1:
https://github.com/open-telemetry/opentelemetry-helm-charts/issues/252

### DIFF
--- a/charts/opentelemetry-operator/templates/_helpers.tpl
+++ b/charts/opentelemetry-operator/templates/_helpers.tpl
@@ -29,4 +29,5 @@ Selector labels
 */}}
 {{- define "opentelemetry-operator.selectorLabels" -}}
 app.kubernetes.io/name: {{ include "opentelemetry-operator.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
 {{- end }}
--- a/charts/opentelemetry-operator/templates/_helpers.tpl
+++ b/charts/opentelemetry-operator/templates/_helpers.tpl
@@ -13,9 +13,20 @@ Create chart name and version as used by the chart label.
 {{- end }}
 
 {{/*
+Common labels
+*/}}
+{{- define "opentelemetry-operator.labels" -}}
+helm.sh/chart: {{ include "opentelemetry-operator.chart" . }}
+{{ include "opentelemetry-operator.selectorLabels" . }}
+{{- if .Chart.AppVersion }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+{{- end }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end }}
+
+{{/*
 Selector labels
 */}}
 {{- define "opentelemetry-operator.selectorLabels" -}}
 app.kubernetes.io/name: {{ include "opentelemetry-operator.name" . }}
-app.kubernetes.io/instance: {{ .Release.Name }}
 {{- end }}
--- a/charts/opentelemetry-operator/templates/deployment.yaml
+++ b/charts/opentelemetry-operator/templates/deployment.yaml
@@ -57,9 +57,6 @@ spec:
           image: "{{ .Values.manager.image.repository }}:{{ .Values.manager.image.tag }}"
           name: manager
           ports:
-            - containerPort: {{ .Values.manager.ports.metricsPort }}
-              name: metrics
-              protocol: TCP
             {{- if .Values.admissionWebhooks.enabled }}
             - containerPort: {{ .Values.manager.ports.webhookPort }}
               name: webhook-server
--- a/charts/opentelemetry-operator/templates/deployment.yaml
+++ b/charts/opentelemetry-operator/templates/deployment.yaml
@@ -23,7 +23,7 @@ spec:
       hostNetwork: {{ .Values.hostNetwork }}
       containers:
         - args:
-            - --metrics-addr=127.0.0.1:{{ .Values.manager.ports.metricsPort }}
+            - --metrics-addr=0.0.0.0:{{ .Values.manager.ports.metricsPort }}
             - --enable-leader-election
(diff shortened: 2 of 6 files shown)


charts/opentelemetry-operator/templates/service.yaml
```diff
@@ -12,6 +12,10 @@ spec:
       port: 8443
       protocol: TCP
       targetPort: https
+    - name: metrics
```
TylerHelmuth, line 15: Since service monitor is optional, this should also be optional.

charts/opentelemetry-operator/templates/deployment.yaml
```diff
@@ -23,7 +23,7 @@ spec:
       hostNetwork: {{ .Values.hostNetwork }}
       containers:
         - args:
-            - --metrics-addr=127.0.0.1:{{ .Values.manager.ports.metricsPort }}
+            - --metrics-addr=0.0.0.0:{{ .Values.manager.ports.metricsPort }}
```
TylerHelmuth, line 26: Is feels like it could be a breaking change.  If the service monitor needs `0.0.0.0` instead of `127.0.0.1` then it needs to be configurable based on whether or not the service monitor is enabled.

charts/opentelemetry-operator/templates/deployment.yaml
```diff
@@ -48,6 +48,9 @@ spec:
           image: "{{ .Values.manager.image.repository }}:{{ .Values.manager.image.tag }}"
           name: manager
           ports:
+            - containerPort: {{ .Values.manager.ports.metricsPort }}
```
TylerHelmuth, line 51: I don't see `.Values.manager.ports.metricsPort` in the default values.  What happens if the user doesn't manually add it in their values.yaml?  How do they know to add it?

TylerHelmuth: NVM, I wasn't looking at main on accident.

charts/opentelemetry-operator/values.yaml
```diff
@@ -32,6 +32,14 @@ manager:
   ## e.g ENV_VAR: env_value
   env:
 
+  ##Enable ServiceMonitor for Prometheus metrics scrape
+  ##
+  serviceMonitor:
+    enabled: true
```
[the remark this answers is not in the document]
Allex1: @TylerHelmuth should we disable the metrics endpoint that gets instantiated at start up then ?

charts/opentelemetry-operator/templates/deployment.yaml
Allex1: 0.0.0.0 includes 127.0.0.1 so I don't think it will be a breaking change. (Tested with a sidecar container hitting 127.0.0.1 on this branch)

Allex1:
```Error: unable to build kubernetes objects from release manifest: unable to recognize "": no matches for kind "ServiceMonitor" in version "monitoring.coreos.com/v1"```

The serviceMonitor needs the CRD present in the test cluster in order to work.

File: charts/opentelemetry-operator/values.yaml (hunk 4 above)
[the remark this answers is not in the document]
Comment (TylerHelmuth) [reply charts/opentelemetry-operator/values.yaml:38]: It is possible to use the metrics endpoint that gets instantiated at startup without the service monitor right?

File: charts/opentelemetry-operator/templates/deployment.yaml (hunk 2 above)
Comment (TylerHelmuth) [reply charts/opentelemetry-operator/templates/deployment.yaml:26]: Thanks for doing that test. @dmitryax do you want to do this as part of this PR or should it be separate?

File: charts/opentelemetry-operator/templates/service.yaml (hunk 1 above)
Comment (Allex1) [reply charts/opentelemetry-operator/templates/service.yaml:15]: @TylerHelmuth As long as the container is exposing that port and endpoint by default I think it's good practice to propagate it at the `Service` level. This is not needed just for the ServiceMonitor but it's a way to access metrics from the pod/s behind the service by any other application/client.

**Comment by TylerHelmuth** [comment]
> 

The opentelemetry-collector charts has the same issue so we don't have any CI scenarios that enable the service monitor.

**Comment by Allex1** [comment]
CI passes if ServiceMonitor is disabled by default :)

File: charts/opentelemetry-operator/templates/deployment.yaml (hunk 2 above)
Comment (dmitryax) [reply charts/opentelemetry-operator/templates/deployment.yaml:26]: Looks like this PR has many unrelated changes. Let's break them down.

This particular change can be considered insecure. I would say we should keep 127.0.0.1 by default and has an option to override it?

Comment (dmitryax) [reply charts/opentelemetry-operator/templates/deployment.yaml:26]: I don't really know from the top of my head. The operator chart was introduced by someone who is not involved in the project anymore. We are actively looking for code owners for the operator helm chart. Maybe @open-telemetry/operator-approvers can answer your question

File: charts/opentelemetry-operator/templates/service.yaml (hunk 1 above)
Comment (TylerHelmuth) [reply charts/opentelemetry-operator/templates/service.yaml:15]: I agree.  Can you move both this change and the new containerPort into a new PR?  Those 2 changes seem to fix an issue that the metrics port is not exposed and it should be.

File: charts/opentelemetry-operator/templates/_helpers.tpl [hunk 5]
```diff
@@ -12,10 +12,21 @@ Create chart name and version as used by the chart label.
...
+{{- end }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end }}
+
 {{/*
 Selector labels
 */}}
 {{- define "opentelemetry-operator.selectorLabels" -}}
 app.kubernetes.io/name: {{ include "opentelemetry-operator.name" . }}
-app.kubernetes.io/instance: {{ .Release.Name }}
```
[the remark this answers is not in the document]
Allex1: @dmitryax you are right, I forgot to add it back after testing. Fixed

TylerHelmuth (approved):

dmitryax (approved):

PR #266, opened 2022-07-07
Add prometheus-operator alerts support for manager

Allex1:


TylerHelmuth (commented):
Lets reuse the pattern that opentelemetry-collector chart uses for prometheus rules.  Users should be able to enable prometheus rules (disabled by default) and should be able to customize the rules.  We can supply a default set of rules as well.

Allex1:
@TylerHelmuth @dmitryax one more look pls.

TylerHelmuth (approved):

dmitryax (approved):

Issue #267, opened 2022-07-07
Add support for basic alerts for the operator/collector pod


Allex1:
Right now afaik we don't  have any alerts/dashboards for when the operator pod has issues/goes down.
We should start analysing the oob metrics and create alerts/grafana dashboards for making the operator helm production ready.

Allex1:
https://github.com/open-telemetry/opentelemetry-helm-charts/pull/266

PR #268, opened 2022-07-07
Include default alerts for collector

Allex1:
https://github.com/open-telemetry/opentelemetry-helm-charts/issues/267

### DIFF
--- a/charts/opentelemetry-collector/templates/prometheusrule.yaml
+++ b/charts/opentelemetry-collector/templates/prometheusrule.yaml
@@ -48,14 +48,6 @@ spec:
       annotations:
         description: '{{`The {{ $labels.processor }} is dropping metrics at a rate of {{ humanize $value }} per second `}}'
         runbook_url: 'https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/monitoring.md#data-loss'
-    - alert: ExporterDroppedSpans #metric not present in collector 0.54
-      expr: rate(otelcol_exporter_send_failed_spans[5m]) > 0
-      for: 2m
-      labels:
-        severity: critical
-      annotations:
-        description: '{{`The {{ $labels.exporter }} is dropping spans at a rate of {{ humanize $value }} per second `}}'
-        runbook_url: 'https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/monitoring.md#receive-failures'
     - alert: ExporterDroppedMetrics
       expr: rate(otelcol_exporter_send_failed_metric_points[5m]) > 0
       for: 2m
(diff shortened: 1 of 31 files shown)


TylerHelmuth (commented):
@Allex1 the opentelemetry-collector chart already has support for prometheus rules.  If this PR is adding default rules I think we can accomplish that by adding them to `.Values.prometheusRule.groups` in the values.yaml

Allex1:
@TylerHelmuth not sure dumping the whole yaml in values is a good idea, but I think there might be a workaround for that. I'll explore options in the next days.
In the meanwhile any idea why the linting fails ? Works fine on localhost :)

TylerHelmuth:
>  In the meanwhile any idea why the linting fails ? Works fine on localhost :)

You need to run `make generate-examples`.

Allex1:
@TylerHelmuth Added the default alerts under a feature flag with the option for the user to add his own alerts without exploding the values file.
Once an approach is agreed I'll replicate it in the operator chart as well.

charts/opentelemetry-collector/templates/prometheusrule.yaml
```diff
@@ -10,5 +10,78 @@ metadata:
...
+        runbook_url: 'https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/monitoring.md#data-loss'
+    - alert: ProcessorDroppedMetrics
+      expr: rate(otelcol_processor_dropped_metric_points[5m]) > 0
+      for: 2m
+      labels:
+        severity: critical
+      annotations:
+        description: '{{`The {{ $labels.processor }} is dropping metrics at a rate of {{ humanize $value }} per second `}}'
+        runbook_url: 'https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/monitoring.md#data-loss'
+    - alert: ExporterDroppedSpans #metric not present in collector 0.54
```
TylerHelmuth, line 51: If this isn't present in 0.54.0 by design then I don't think we should include it as a default rule since the chart's default collector version is 0.54.0

charts/opentelemetry-collector/templates/prometheusrule.yaml
```diff
@@ -10,5 +10,78 @@ metadata:
...
+    - alert: ExporterDroppedMetrics
+      expr: rate(otelcol_exporter_send_failed_metric_points[5m]) > 0
+      for: 2m
+      labels:
+        severity: critical
+      annotations:
+        description: '{{`The {{ $labels.exporter }} is dropping metrics at a rate of {{ humanize $value }} per second `}}'
+        runbook_url: 'https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/monitoring.md#receive-failures'
+    - alert: ExporterQueueSize
+      expr: otelcol_exporter_queue_size > 5000 # TODO: Define/Reuse queue_size (default = 5000)
```
TylerHelmuth, line 68: Can you make an issue for this TODO and remove the TODO

TylerHelmuth (commented):
@Allex1 I like this idea and it feels in the same vain as the ongoing presets work (#158).  @dmitryax please review.

charts/opentelemetry-collector/templates/prometheusrule.yaml
Allex1: Problem is that it's present in the docs ... I need to do a bit of digging

TylerHelmuth: The monitoring.md doc hasn't been updated in 5 months so I don't really trust it lol

Allex1: deleted alert

TylerHelmuth (approved):

dmitryax (approved):

PR #271, opened 2022-07-10
Updates Operator version to 0.54.0

angelokurtis:
This PR bumps the default Operator version to 0.54.0


TylerHelmuth:
@angelokurtis Please bump the minor version of the chart. Also, most of the time when the operator version changes the CRDs need updated as well. Can you confirm that those don't need changed in this instance?

Allex1 (approved):

angelokurtis:
Hi @TylerHelmuth, thank you for the support!
Chart version updated to 0.9.0. I also checked the Operator, and there were no changes in the CRDs from 0.53.0 to 0.54.0.

TylerHelmuth (approved):

dmitryax (approved):

PR #273, opened 2022-07-11
set proper resources for collector

### LINKED ISSUE #272: Set better default resources
The opentelemetry-collector [only defines limits by default](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/values.yaml#L194-L197),  if you don’t specify requests, Kubernetes will automatically set requests to equal to Pod’s limits. This means that with the default Helm  values, opentelemtry will take up 2Gb and 1cpu, which is insane.

I think there should be default requests set so that K8s does not schedule these pods with such high resources.

MarkAndersonTrocme:
Closes #272


Allex1:
@MarkAndersonTrocme pls bump the chart version. 
Also I would tend to give a bit more memory for limits.

MarkAndersonTrocme:
@Allex1 Let me know if those resources make more sense now? If not let me know what you recommend

Allex1 (approved):
LGTM

TylerHelmuth (commented):
Also, please run "make generate-examples"

charts/opentelemetry-collector/values.yaml
```diff
@@ -192,9 +192,12 @@ containerLogs:
...
 
 resources:
+  requests:
+    cpu: 100m
+    memory: 128Mi
   limits:
-    cpu: 1
-    memory: 2Gi
+    cpu: 200m
+    memory: 256Mi
```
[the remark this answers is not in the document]
MarkAndersonTrocme: @TylerHelmuth do you want to set that as the request or the limit?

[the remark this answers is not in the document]
MarkAndersonTrocme: ok done!

charts/opentelemetry-collector/values.yaml
```diff
@@ -192,9 +192,12 @@ containerLogs:
   enabled: false
 
 resources:
+  requests:
+    cpu: 100m
+    memory: 128Mi
```
TylerHelmuth, line 197:
```suggestion
```
I think we still want it to requests its limits by default.  As the potential monitor of the system it would be bad if the collector got scheduled with insufficient resources and then couldn't scale up.

TylerHelmuth (commented):
Please re-run `make generate-examples`

charts/opentelemetry-collector/values.yaml
MarkAndersonTrocme: makes sense, done!

TylerHelmuth:
@dmitryax @MarkAndersonTrocme @Allex1  I support this change but I think its a "breaking change".  Some customers may be relying on the higher default values and if they upgraded to this version their requests/limits would get halved.  Any ideas on how we can handle that gracefully?

Allex1:
We increase the minor version and maybe update the main Readme with a warning?
While the chart suggests sensible defaults users are responsible for setting their own resource allocation values.
Also users should expect some imminent breaking changes as the chart is not yet stable.
Once it's stable we could increase the major version when making breaking changes and document the upgrade path on our main Readme.
wdyt?

TylerHelmuth:
Ya without a major version I can't think of a good way to handle this without introducing a new field, which I don't really want to do.  We do have an UPGRADING.md doc that feels like a good place to callout this change.  @MarkAndersonTrocme can you add a new entry in UPGRADING.md that calls out this change and shows readers how to set the values back to the old values?

charts/opentelemetry-collector/UPGRADING.md
```diff
@@ -189,3 +194,16 @@ config:
       metrics: null
       logs: null
 ```
+
+### Set resources:
```
TylerHelmuth, line 198: Instead of its own section you can include this directly under the new `## 0.22.1 to 0.23.0` section

TylerHelmuth (approved):

TylerHelmuth:
@MarkAndersonTrocme please handle merge conflicts.

charts/opentelemetry-collector/UPGRADING.md
```diff
@@ -1,4 +1,18 @@
 # Upgrade guidelines
+## 0.22.1 to 0.23.0
```
TylerHelmuth, line 2: Looks like the charts version needs bumped to 0.24.0 instead, as another PR got merged before this.  Then this doc will need updated and examples will need to be regenerated.

MarkAndersonTrocme: yup just noticed, bumped and reran the make generate

TylerHelmuth (approved):
@dmitryax would appreciate your opinion on this PR since it has impact to existing users despite being a small change.

charts/opentelemetry-collector/UPGRADING.md
```diff
@@ -1,4 +1,18 @@
 # Upgrade guidelines
+## 0.23.0 to 0.24.0
```
dmitryax, line 2:
```suggestion

## 0.23.0 to 0.24.0
```

charts/opentelemetry-collector/values.yaml
```diff
@@ -197,8 +197,8 @@ containerLogs:
 
 resources:
   limits:
-    cpu: 1
-    memory: 2Gi
+    cpu: 256m
+    memory: 512Mi
```
dmitryax, line 201: This is a good default for daemonset but pretty low for a typical deployment mode installation. Let's call this out in a comment?

Allex1: Not sure what the comment should state. If we agree this is a sensible default where all use cases would work I think user can easily override this to his needs

Allex1:
@TylerHelmuth @dmitryax I've reopened this as I think it's useful to set lower sensible defaults. PTAL

charts/opentelemetry-collector/values.yaml
TylerHelmuth: It is sensible for the daemonset, but low for deployment.  If we want default values that work for both we'll need to up it some.

Allex1: @TylerHelmuth what would be proper values for deployment and can you expand on why this would not be enough? imo these values should allow any user to start a  collector and make a poc, not rely on them for prod.

TylerHelmuth: It will totally start, but you're correct, they wouldn't work for prod.  I think thats what @dmitryax wants to call out in a comment.  Keeping these values is ok, but we should say in the values.yaml that the values should typically be increased in order to handle typical production load.

Allex1: Added comment.

Issue #196, opened 2022-05-05
Allow for a setting such as `useExistingConfigMap`


snuggie12:
I see a few other issues (#90 and #158) related to the default config but this is a slightly different approach.

We use other tools to handle creating the config for otel-collector and would like a setting where the helm chart does not create a configmap but uses whatever I input as the configmap to load in.

example can be found here: https://github.com/prometheus-community/helm-charts/blob/ee700113d070664053bb041f97ae5356e6072fe6/charts/kube-prometheus-stack/values.yaml#L427

TylerHelmuth:
@snuggie12 thanks for the request.  Do you want to work on this?

dmitryax:
This must be resolved by presets. 

1. disable default presets
2. provide custom config

cc @TylerHelmuth

TylerHelmuth:
@dmitryax that would work, but this issue reads to me like they don't want the helm chart generating a configmap at all.  Sounds like that want an option for the collector to not generate a configmap and let them provide their own values for

```yaml
volumes:
  - name: opentelemetry-collector-configmap
    configMap:
      name: example-opentelemetry-collector-agent
      items:
        - key: relay
          path: relay.yaml
```

and

```yaml
containers:
  - name: opentelemetry-collector
    command:
      - /otelcol-contrib
      - --config=/conf/relay.yaml
```

dmitryax:
Oh, ok. That sounds reasonable. We can support this option

TylerHelmuth:
@JaredTan95 do you want to work on this?

JaredTan95:
> @JaredTan95 do you want to work on this?

Okay, glad to support this.

PR #274, opened 2022-07-12
Allow use existing config map

JaredTan95:
close https://github.com/open-telemetry/opentelemetry-helm-charts/issues/196


charts/opentelemetry-collector/templates/configmap.yaml
```diff
@@ -1,4 +1,5 @@
 {{- if eq .Values.mode "deployment" -}}
+  {{- if not .Values.useExistingConfigMap -}}
```
TylerHelmuth, line 2: Can you combine this if statement with the one above?

charts/opentelemetry-collector/values.yaml
```diff
@@ -8,6 +8,10 @@ fullnameOverride: ""
 # Valid values are "daemonset" and "deployment".
 mode: ""
 
+## If true then the user will be responsible to provide a configmap with otel-col configuration
+## So when true the config part will be ignored and the one in the volumes will be used
```
TylerHelmuth, line 12:
```suggestion
# When true, the chart will not use .Values.config and will not provide the collector a configuration. 
# The user is responsible for providing a configmap with the collector's configuration. 
# See the "deployment-use-existing-configmap" example for values.yaml details.
```

charts/opentelemetry-collector/values.yaml
```diff
@@ -8,6 +8,10 @@ fullnameOverride: ""
...
+useExistingConfigMap: false
```
TylerHelmuth, line 13: We could go as far as to disallow this value being true and .Values.config being anything but null.  We could add a rule to enforce this in NOTES.txt.

charts/opentelemetry-collector/templates/configmap.yaml
TylerHelmuth: Also, configmap-agent.yaml should have a similar check.

TylerHelmuth (commented):
Thanks for doing this.

charts/opentelemetry-collector/examples/deployment-use-existing-configMap/daemonset-values.yaml
```diff
@@ -0,0 +1,24 @@
+mode: daemonset
```
TylerHelmuth, line 1: You can remove this file, the deployment-values.yaml is sufficient for an example

charts/opentelemetry-collector/templates/NOTES.txt
```diff
@@ -13,3 +13,7 @@
 {{- if and (not (eq .Values.mode "daemonset")) (not (eq .Values.mode "deployment")) }}
 {{ fail "[ERROR] 'mode' must be set. See https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/UPGRADING.md for instructions." }}
 {{ end }}
+
+{{- if and .Values.useExistingConfigMap (eq (toString .Values.config) "<nil>") }}
+{{ fail "[ERROR] 'config' must be set." }}
```
TylerHelmuth, line 18:
```suggestion
{{- if and .Values.useExistingConfigMap (not (eq (toString .Values.config) "<nil>")) }}
{{ fail "[ERROR] 'config' must be null when using 'useExistingConfigMap'." }}
```

charts/opentelemetry-collector/values.yaml
```diff
@@ -8,6 +8,11 @@ fullnameOverride: ""
 # Valid values are "daemonset" and "deployment".
 mode: ""
 
+# When true, the chart will not use .Values.config and will not provide the collector a configuration.
+# The user is responsible for providing a configmap with the collector's configuration.
```
TylerHelmuth, line 12:
```suggestion
# The user is responsible for providing a configmap with the collector's configuration.
# When true, .Values.config must be set to null.
```

charts/opentelemetry-collector/examples/deployment-use-existing-configMap/deployment-values.yaml
```diff
@@ -0,0 +1,24 @@
+mode: deployment
+
+resources:
+  limits:
+    cpu: 100m
+    memory: 200M
+
+useExistingConfigMap: true
```
TylerHelmuth, line 8:
```suggestion
useExistingConfigMap: true

config: null
```

TylerHelmuth (commented):
@dmitryax can you take a look as well

charts/opentelemetry-collector/values.yaml
```diff
@@ -8,6 +8,11 @@ fullnameOverride: ""
...
+# See the "deployment-use-existing-configmap" example for values.yaml details.
+useExistingConfigMap: false
```
dmitryax, line 14: Should we follow the same approach that we have for `serviceAccount` and ` clusterRole` and use the following interface instead:

```yaml
configMap:
  # Specifies whether a configMap should be created
  create: false
  # The name of the configMap to use.
  # If not set and create is true, a name is generated using the chart name.
  name: ""
```

dmitryax: I don't get why `Command Args, extraVolumes, and extraVolumeMounts` are related to the `configMap` setting

TylerHelmuth: In this situation if we don't generate the configmap ourselves and the user wants to provide their own they have to set some entries in those 3 fields, like in the new example: https://github.com/open-telemetry/opentelemetry-helm-charts/blob/098d0712f3006ce122004b6c67f471a90f5b788f/charts/opentelemetry-collector/examples/deployment-use-existing-configMap/deployment-values.yaml.

charts/opentelemetry-collector/examples/deployment-use-existing-configMap/deployment-values.yaml
```diff
@@ -0,0 +1,26 @@
...
+  - name: custom-otelcol-configmap
+    configMap:
+      name: custom-otel-collector-config
+      items:
+        - key: config
+          path: config.yaml
+      defaultMode: 420
+extraVolumeMounts:
+  - name: custom-otelcol-configmap
+    mountPath: /conf/config.yaml
```
dmitryax, line 26: We tend to introduce examples for only significant functional use cases. Let's do not introduce one for this use case. If we  accumulate a lot of examples, any tiny change to the chart creates a huge diff :)

cc @TylerHelmuth

charts/opentelemetry-collector/values.yaml
dmitryax: What kind of configuration interface do you envision?

charts/opentelemetry-collector/examples/deployment-use-existing-configMap/deployment-values.yaml
TylerHelmuth: @dmitryax If we stick with the proposals solution of relying on `command.extraArgs`, `extraVolumes`, and `extraVolumeMounts` I would support this example.  Thats a lot of extra things to manage when turning on useExistingConfigMap. 

If we switch to your configMap interface idea and include in that interface options to set the content that was previous being set in `command.extraArgs`, `extraVolumes`, and `extraVolumeMounts` like in this example then I think we could forgo the example.

JaredTan95: @dmitryax How about moving this example's content into an existing example dir(e.g examples/daemonset-and-deployment) instead of creating a separate `deployment-use-existing-configMap` folder?

charts/opentelemetry-collector/values.yaml
TylerHelmuth: @dmitryax what do you think?

dmitryax: Yeah, this sounds good to me. Just to avoid any imlicit magic

charts/opentelemetry-collector/examples/deployment-use-existing-configMap/deployment-values.yaml
dmitryax: Ok, let's keep this example since it requires many fields to be changed

charts/opentelemetry-collector/values.yaml
TylerHelmuth: @JaredTan95 Can you update the PR to use this config interface?

JaredTan95: done

charts/opentelemetry-collector/templates/NOTES.txt
```diff
@@ -13,3 +13,7 @@
 {{- if and (not (eq .Values.mode "daemonset")) (not (eq .Values.mode "deployment")) }}
 {{ fail "[ERROR] 'mode' must be set. See https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/UPGRADING.md for instructions." }}
 {{ end }}
+
+{{- if and (not .Values.configMap.create) (or (eq (toString .Values.extraVolumes) "<nil>") (eq (toString .Values.extraVolumeMounts) "<nil>") ) }}
```
TylerHelmuth, line 17: I don't think we want to strictly enforce `.Values.extraVolumes` or `.Values.extraVolumeMounts` be set when `.Values.configMap.create` is false.  There are situations where users may build their own collector image with the config included and then not need a configmap at all.  I do think we should enforce that `.Values.config` be set to `null` when `.Values.configMap.create` is true.

JaredTan95: when `.Values.configMap.create` is true,  Whatever the value of the ` `.Values.config``, it does not work, I think we can give an warning msg.

TylerHelmuth (approved):
Thanks @JaredTan95!

Issue #275, opened 2022-07-12
[opentelemetry-operator] - Add support for imagePullSecrets


Allex1:
Add support for imagePullSecrets in order to reference one or more secrets to be used when pulling images from authenticated repositories.

PR #276, opened 2022-07-12
[opentelemetry-operator] - Add support for imagePullSecrets

Allex1:
Addresses https://github.com/open-telemetry/opentelemetry-helm-charts/issues/275

### DIFF
--- a/charts/opentelemetry-operator/templates/deployment.yaml
+++ b/charts/opentelemetry-operator/templates/deployment.yaml
@@ -98,8 +98,9 @@ spec:
           {{- with .Values.kubeRBACProxy.resources }}
           resources: {{ toYaml . | nindent 12 }}
           {{- end }}
-      {{- if .Values.imagePullSecrets }}
-      imagePullSecrets: {{ toYaml .Values.imagePullSecrets | nindent 8 }}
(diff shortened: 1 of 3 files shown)


charts/opentelemetry-operator/templates/deployment.yaml
```diff
@@ -98,6 +98,9 @@ spec:
           {{- with .Values.kubeRBACProxy.resources }}
           resources: {{ toYaml . | nindent 12 }}
           {{- end }}
+      {{- if .Values.imagePullSecrets }}
+      imagePullSecrets: {{ toYaml .Values.imagePullSecrets | nindent 8 }}
```
TylerHelmuth, line 102: This looks a little different than the collector chart's implementation.  Is that intentional?

TylerHelmuth: Shouldn't it be nested?

Allex1: Haven't looked at the collector implementation before. Changed it to match and tested.

TylerHelmuth (approved):

dmitryax (approved):

PR #280, opened 2022-07-20
add opentelemetry-demo chart

wph95:
https://github.com/open-telemetry/opentelemetry-demo-webstore/issues/153

### DIFF
--- a/charts/opentelemetry-demo/templates/component.yaml
+++ b/charts/opentelemetry-demo/templates/component.yaml
@@ -1,0 +1,28 @@
+
+{{ $servicePortMap := default dict }}
+{{- range $name, $config := .Values.components }}
+    {{ if $config.ports }}
+        {{- $servicePortMap := set $servicePortMap  $name (get (index $config.ports 0 ) "value") }}
+    {{- end}}
+    {{ if $config.servicePort }}
+        {{- $servicePortMap := set $servicePortMap  $name $config.servicePort }}
+    {{- end}}
+
+
+{{- end }}
+
+{{- range $name, $config := .Values.components }}
+    {{- $config := set . "name" $name }}
+    {{- $config := set . "Release" $.Release }}
+    {{- $config := set . "Chart" $.Chart }}
+    {{- $config := set . "imageConfig" $.Values.image }}
+    {{- $config := set . "serviceAccount" $.Values.serviceAccount }}
+    {{- $config := set . "observability" $.Values.observability }}
+    {{- $config := set . "servicePortMap" $servicePortMap }}
+
+    {{- if $config.enabled -}}
+        {{- include "otel.demo.component" $config -}}
+        {{- include "otel.demo.service" $config -}}
+    {{ end }}
+
+{{- end }}
--- a/charts/opentelemetry-demo/templates/services.yaml
+++ b/charts/opentelemetry-demo/templates/services.yaml
@@ -1,0 +1,28 @@
+{{- define "otel.demo.service" }}
+{{- if or .ports .servicePort}}
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "otel-demo.name" . }}-{{ .name }}
+  labels:
+    {{- include "otel-demo.labels" . | nindent 4 }}
+spec:
+  ports:
+    {{- if .ports }}
+    {{- range $port := .ports }}
+    - port: {{ $port.value }}
+      name: {{ $port.name}}
+      targetPort: {{ $port.value }} 
+    {{- end }}
+    {{- end }}
+
+    {{- if .servicePort }}
+    - port: {{.servicePort}}
+      name: service
+      targetPort: {{ .servicePort }} 
+    {{- end }}
+  selector:
+    {{- include "otel-demo.selectorLabels" . | nindent 4 }}
+{{- end}}
+{{- end}}
--- a/charts/opentelemetry-demo/values.yaml
+++ b/charts/opentelemetry-demo/values.yaml
@@ -1,0 +1,159 @@
+observability:
+  otelcol: 
+    enabled: true 
+    config: 
+      receivers:
+        otlp:
+          protocols:
+            grpc:
+            http:
+      exporters:
+        jaeger:
+          endpoint: "${JAEGER_ADDR}"
+          tls:
+            insecure: true
+        logging:
+
+      processors:
+        batch:
+
+      service:
+        pipelines:
+          traces:
+            receivers: [otlp]
+            processors: [batch]
+            exporters: [logging, jaeger]
+
+  jaeger:
+    enabled: true
+
+image:
+  repository: otel/demo
+  pullSecrets: []
+
+serviceAccount: ""
+
+components:
+  redis:
+    enabled: true
+    image: redis:alpine
+    ports:
+    - name: redis
+      value: 6379
+
+  ad-service:
+    enabled: true
+    servicePort: 8080
+
+  cart-service:
+    enabled: true
+    servicePort: 8080
+    env:
+    - name: ASPNETCORE_URLS
+      value: http://*:8080
+    depends:
+    - redis
+    
+
+  checkout-service:
+    enabled: true
+    servicePort: 8080
+    depends:
+    - cart-service
+    - currency-service
+    - email-service
+    - payment-service
+    - product-catalog-service
+    - shipping-service
+
+  currency-service:
+    enabled: true
+    servicePort: 8080
+    env:
+    - name: PORT
+      value: '8080'
+
+  email-service:
+    enabled: true
+    servicePort: 8080
+    env:
+    - name: APP_ENV
+      value: production
+    - name: PORT
+      value: '8080'
+
+  featureflag-service:
+    enabled: true
+    env:
+    - name: DATABASE_URL
+      value: ecto://ffs:ffs@ffs-postgres:5432/ffs
+    - name: GRPC_PORT
+      value: '50053'
+    - name: PORT
+      value: '50052'
+    ports:
+    - name: grpc
+      value: 50052
+    - name: http
+      value: 50053
+
+  ffs-postgres:
+    enabled: true
+    env:
+    - name: POSTGRES_DB
+      value: ffs
+    - name: POSTGRES_PASSWORD
+      value: ffs
+    - name: POSTGRES_USER
+      value: ffs
+    image: cimg/postgres:14.2
+    ports:
+    - name: postgres
+      value: 5432 
+
+  frontend:
+    enabled: true
+    servicePort: 8080
+    depends:
+    - ad-service
+    - cart-service
+    - checkout-service
+    - currency-service
+    - product-catalog-service
+    - recommendation-service
+    - shipping-service
+    env:
+    - name: FRONTEND_ADDR
+      value: :8080
+  
+  loadgenerator:
+    enabled: true
+    env:
+    - name: FRONTEND_ADDR
+      value: frontend:8080
+    - name: USERS
+      value: '10'
+  
+  payment-service:
+    enabled: true
+    servicePort: 8080
+
+  product-catalog-service:
+    enabled: true
+    servicePort: 8080
+
+  recommendation-service:
+    enabled: true
+    servicePort: 8080
+    depends:
+    - product-catalog-service
+    env:
+    - name: OTEL_PYTHON_LOG_CORRELATION
+      value: 'true'
+  
+  shipping-service:
+    enabled: true
+    servicePort: 8080
+    env:
+    - name: PORT 
+      value: '8080'
--- a/charts/opentelemetry-demo/.helmignore
+++ b/charts/opentelemetry-demo/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*.orig
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+.vscode/
--- a/charts/opentelemetry-demo/Chart.yaml
+++ b/charts/opentelemetry-demo/Chart.yaml
@@ -1,0 +1,12 @@
+apiVersion: v2
+type: application
+name: opentelemetry-demo
+version: 0.1.0
+description: opentelemetry demo webstore helm chart
+home: https://opentelemetry.io/
+sources:
+  - https://github.com/open-telemetry/opentelemetry-demo-webstore
+maintainers:
+  - name: justwph
+icon: https://opentelemetry.io/img/logos/opentelemetry-logo-nav.png
+appVersion: "0.2.0-alpha"
--- a/charts/opentelemetry-demo/templates/NOTES.txt
+++ b/charts/opentelemetry-demo/templates/NOTES.txt
@@ -1,0 +1,2 @@
+1. Get the OTEL-DEMO-WEBSTORE Front UI by running these commands:
+     kubectl port-forward svc/{{ include "otel-demo.name" . }}-frontend 8080:8080
--- a/charts/opentelemetry-demo/templates/_helpers.tpl
+++ b/charts/opentelemetry-demo/templates/_helpers.tpl
@@ -1,0 +1,78 @@
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "otel-demo.name" -}}
+{{- default .Release.Name | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Create chart name and version as used by the chart label.
+*/}}
+{{- define "otel-demo.chart" -}}
+{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Common labels
+*/}}
+{{- define "otel-demo.labels" -}}
+helm.sh/chart: {{ include "otel-demo.chart" . }}
+{{ include "otel-demo.selectorLabels" . }}
+{{- if .Chart.AppVersion }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+{{- end }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end }}
+
+{{/*
+Selector labels
+*/}}
+{{- define "otel-demo.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "otel-demo.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- if .name }}
+app.kubernetes.io/component: {{ .name}}
+{{- end}}
+{{- end }}
+
+
+{{/*
+Get Deployment Env 
+*/}}
+{{- define "otel-demo.env" -}}
+{{- if .env }}
+{{- toYaml .env }}
+{{- end }}
+
+{{- if .observability.otelcol.enabled }} 
+- name: OTEL_EXPORTER_OTLP_TRACES_ENDPOINT
+  value: http://{{ include "otel-demo.name" . }}-otelcol:4317
+- name: OTEL_RESOURCE_ATTRIBUTES
+  value: service.name={{ .name }}
+{{- end }}
+
+{{- if .servicePort}}
+- name: {{ printf "%s_PORT " .name | upper | replace "-" "_" }}
+  value: {{.servicePort | quote}} 
+{{- end }}
+
+{{- if .depends }}
+{{ $prefix := include "otel-demo.name" $ }}
+
+{{- range $depend := .depends }}
+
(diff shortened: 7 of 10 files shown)


dmitryax:
Hi @wph95, thanks for the contribution. Few suggestions/questions:
- I believe the chart should be called the same name as the repo
- I would recommend splitting deployments.yaml template by components
- Looks like prometheus is missing. Is this intentional?

wph95:
> * I believe the chart should be called the same name as the repo
done

> * Looks like prometheus is missing. Is this intentional?
Yes, the current version does not include metrics (grafana and prometheus)
I'm a little concerned that PR is too big to review, so metric is not included yet.

charts/opentelemetry-demo-webstore/templates/deployment.yaml
```diff
@@ -0,0 +1,35 @@
...
+      {{- if .imageConfig.pullSecrets }}
+      imagePullSecrets:
+        {{- toYaml .imageConfig.pullSecrets | nindent 8}}
+      {{- end }}
+      {{- with .serviceAccountName }}
+      serviceAccountName: {{ .serviceAccountName}} 
+      {{- end }}
+      containers:
+        - name: {{ .name }}
+          image: {{ .image | default (printf "%s:v%s-%s" .imageConfig.repository .Chart.AppVersion (.name | replace "-" "")) }}
```
JaredTan95, line 27: Can we support customize the image tag?

dmitryax:
> Yes, the current version does not include metrics (grafana and prometheus)
> I'm a little concerned that PR is too big to review, so metric is not included yet. plan add it in next PR :)

Sounds good 👍

charts/opentelemetry-demo-webstore/Chart.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v2
+type: application
+name: opentelemetry-demo-webstore
+version: 0.1.0
+description: opentelemetry demo webstore helm chart
+home: https://opentelemetry.io/
+sources:
+  - https://github.com/open-telemetry/opentelemetry-demo-webstore
+maintainers:
+  - name: wph95
```
TylerHelmuth, line 10: I took a look at the other charts' maintainers list and realized they are pretty out of date, assuming this list is supposed to be the maintainer of the chart and not the thing being installed.

Can you add myself and @dmitryax to this list?

charts/opentelemetry-demo-webstore/templates/deployment.yaml
```diff
@@ -0,0 +1,35 @@
+{{- define "otel.demo.deployment" }}
```
TylerHelmuth, line 1: Since this is being used in another template and isn't generating a file itself it should be named with an "_" like "_pod.tpl" and "_helpers.tpl"

charts/opentelemetry-demo-webstore/templates/otelcol.yaml
```diff
@@ -0,0 +1,113 @@
...
+          - name: JAEGER_ADDR
+            value: {{ include "otel-demo.name" . }}-jaeger:14250
+        {{- end }}
+        ports:
+          - containerPort: 4317
+            protocol: TCP
+          - containerPort: 4318
+            protocol: TCP
+          - containerPort: 9464 
+            protocol: TCP
```
dineshg13, line 75: What is this for ?

dineshg13 (changes requested):
otel-col is failing for me with error ```Error: failed to get config: cannot unmarshal the configuration: error reading top level configuration sections: 1 error(s) decoding:

* '' has invalid keys: config
2022/07/20 17:43:16 collector server run finished with error: failed to get config: cannot unmarshal the configuration: error reading top level configuration sections: 1 error(s) decoding:

* '' has invalid keys: config
```
This might becoz of bad indentation in  `exporter` section of `configmap`. I think it would be good idea to `yamllint` these files.

File: charts/opentelemetry-demo-webstore/Chart.yaml [hunk 5]
```diff
@@ -0,0 +1,12 @@
...
+icon: https://opentelemetry.io/img/logos/opentelemetry-logo-nav.png
+appVersion: "0.2.0-alpha"
```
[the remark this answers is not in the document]
wph95: For now, 0.2.0-alpha is necessary.

name: v0.2.00
tag_name: v0.2.0-alpha
docker image: https://hub.docker.com/layers/demo/otel/demo/v0.2.0-alpha-emailservice/images/sha256-1ff699b1715b67e1070afb116aac88793362599410b14f8585d8cca9ec33ec7e?context=explore

For that pr, I personally feel like using 0.2.0-alpha to make sure it works properly.
Also it has been discussed in the otel-demo slack channel. Standardizing the release process and tag name

charts/opentelemetry-demo-webstore/templates/otelcol.yaml
wph95: ref https://github.com/open-telemetry/opentelemetry-demo-webstore/blob/main/docker-compose.yml#L31

wph95: I will remove it, thank you review :)

charts/opentelemetry-demo-webstore/Chart.yaml
wph95: cool! added

charts/opentelemetry-demo-webstore/Chart.yaml
```diff
@@ -0,0 +1,14 @@
...
+type: application
+name: opentelemetry-demo-webstore
+version: 0.1.0
+description: opentelemetry demo webstore helm chart
+home: https://opentelemetry.io/
+sources:
+  - https://github.com/open-telemetry/opentelemetry-demo-webstore
+maintainers:
+  - name: dmitryax
+  - name: tigrannajaryan
```
tigrannajaryan, line 11:
```suggestion
```

tigrannajaryan: I am not a maintainer.

wph95: sorry for it, i set wrong name

wph95:
> otel-col is failing for me with error

@dineshg13  fixed it :)

saurabhdes:
What's the best way to test this PR. I tried below steps after looking at README but ran into below error : 
`helm repo add open-telemetry https://open-telemetry.github.io/opentelemetry-helm-charts`

`helm install my-otel-demo open-telemetry/opentelemetry-demo-webstore`
_Error: INSTALLATION FAILED: failed to download "open-telemetry/opentelemetry-demo-webstore"_

The error seems to be happening because the chart doesn't yet exist in the repo. Any suggestions on how to get workaround this issue so i can test this PR?

`helm search repo open-telemetry`

NAME                                  	CHART VERSION	APP VERSION	DESCRIPTION
open-telemetry/opentelemetry-collector	0.22.2       	0.55.0     	OpenTelemetry Collector Helm chart for Kubernetes
open-telemetry/opentelemetry-operator 	0.9.4        	0.54.0     	OpenTelemetry Operator Helm chart for Kubernetes

saurabhdes:
> > What's the best way to test this PR.
> 
> If you do `gh pr checkout 280` you'll be able to checkout this PR. Then you'd be able to do normal helm commands by pointing to ./charts/opentelemetry-demo-webstore. For example, you could do `helm template testing ./charts/opentelemetry-demo-webstore --output-dir ./testing` to see the files that would be rendered during a default install of the chart.

thanks @TylerHelmuth ! I was able to deploy the services successfully in my AKS cluster.

dineshg13:
@wph95 what the reason for not including `prometheus` & `grafana` ? we do have some metrics & I believe we are working to adding more

saurabhdes:
I noticed that in `Jaeger`  "cart-service" is missing in this K8s helm setup and present in docker-compose setup - attaching Jaeger pics for both. I do see t
Screen Shot 2022-07-20 at 10 54 22 PM

Screen Shot 2022-07-21 at 8 52 30 AM
he cart-service pod running in both cases  with no errors, so likely some instrumentation setup in K8s helm case but not sure.

wph95:
> I noticed that in `Jaeger` "cart-service" is missing in this K8s helm setup and present in docker-compose setup - attaching Jaeger pics for both. I do see the cart-service pod running in both cases with no errors, so likely some instrumentation setup in K8s helm case but not sure.

Thank you for your carefulness, it has been fixed.
(I was using the wrong env, and although this env `OTEL_EXPORTER_OTLP_ENDPOINT` , it is defined in the spec, some sdk's, such as dotnet for cart-service, do not support this env.

https://github.com/open-telemetry/opentelemetry-helm-charts/pull/280/commits/176aae6c45610237399d033f46311aa44552b08b

saurabhdes:
> > I noticed that in `Jaeger` "cart-service" is missing in this K8s helm setup and present in docker-compose setup - attaching Jaeger pics for both. I do see the cart-service pod running in both cases with no errors, so likely some instrumentation setup in K8s helm case but not sure.
> 
> Thank you for your carefulness, it has been fixed. (I was using the wrong env, and although this env `OTEL_EXPORTER_OTLP_ENDPOINT` , it is defined in the spec, some sdk's, such as dotnet for cart-service, do not support this env.
> 
> [176aae6](https://github.com/open-telemetry/opentelemetry-helm-charts/commit/176aae6c45610237399d033f46311aa44552b08b)
Screen Shot 2022-07-21 at 3 29 24 PM

thanks @wph95 for the quick fix ! I now see cart-service in Jaeger after doing helm deploy.

dmitryax:
I believe we should remove the chart back now give https://github.com/open-telemetry/opentelemetry-demo/issues/222 :)

TylerHelmuth:
> I believe we should rename the chart back now given https://github.com/open-telemetry/opentelemetry-demo/issues/222 :) 

Nice, I like it way more without "webstore"

charts/opentelemetry-demo-webstore/values.yaml
```diff
@@ -0,0 +1,137 @@
...
+    enabled: true
+    servicePort: 8080
+
+  cart-service:
+    enabled: true
+    servicePort: 8080
+    env:
+    - name: ASPNETCORE_URLS
+      value: http://*:8080
+    depends_on:
```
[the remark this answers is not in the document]
wph95: https://github.com/open-telemetry/opentelemetry-helm-charts/pull/280/files#diff-e02a306853f89e21f54ac77efe8a5a03f1f4df85f9d316a58f15a2a578d51382R22-R28

The inspiration here is the docker compose file from otel-demo.
For some reasons, the component needs to manually declare the endpoint of the dependent service via environment variables.

1. For example, the front service depends on shipping. 
If the shipping port is modified, you need to modify the environment variables of multiple services.
With depends_on, you only need to modify one port, and the other environment variables are automatically generated with the correct names.

2. Our k8s service includes the release name, so the pre-set environment variables should not include the release name.

wph95:
> @wph95 what the reason for not including `prometheus` & `grafana` ? we do have some metrics & I believe we are working to adding more

https://github.com/open-telemetry/opentelemetry-helm-charts/pull/280#issuecomment-1190375707
https://github.com/wph95/opentelemetry-helm-charts/pull/1 (wait this pr merge

charts/opentelemetry-demo-webstore/values.yaml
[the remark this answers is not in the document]
dmitryax: I understand the meaning of `depends_on`, I just don't understand why it has to be in values.yaml - why a helm chart user would need to change the `depend_on` list?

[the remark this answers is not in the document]
wph95: In general, users do not need to modify this depends on.
The default value in value.yaml is sufficient.
This way of generating all yaml in bulk is a bit special compared to the compoent one yaml file approach. But I think it's worth it because we can generate yaml for a dozen components with a small amount of code and declarations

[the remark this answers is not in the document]
TylerHelmuth: If users shouldn't be changing the items in `depends_on`, then we should probably move it to a helper template.  I think we'd be able to get the same functionality that way but it wouldn't need to be in values.yaml.

charts/opentelemetry-demo-webstore/templates/_deployment.yaml
```diff
@@ -0,0 +1,35 @@
+{{- define "otel.demo.deployment" }}
```
TylerHelmuth, line 1: Since this is a template lets keep the pattern of naming it .tpl

charts/opentelemetry-demo-webstore/values.yaml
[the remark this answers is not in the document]
wph95: move depends logic  to _pod.tpl, thx suggest

TylerHelmuth:
@Allex1 would appreciate your review on this PR.

JaredTan95:
@wph95 `test` meanings examples like https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-collector/examples? 
Will it i added in this PR or another separated?

wph95:
> separated

I think it is better to submit separately

> @wph95 `test` meanings examples like https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-collector/examples? Will it i added in this PR or another separated?

I think it is better to PR separately

JaredTan95:
> > @wph95 `test` meanings examples like https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-collector/examples? Will it i added in this PR or another separated?
> 
> I think it is better to PR separately

+1

JaredTan95 (approved):
LGTM

charts/opentelemetry-demo-webstore/values.yaml
```diff
@@ -0,0 +1,116 @@
...
+
+components:
+  redis:
+    enabled: true
+    image: redis:alpine
+    ports:
+    - name: redis
+      value: 6379
+
+  ad-service:
```
dmitryax, line 21: Can we make component names consistent? `-` is not the best symbol for values.yaml parameters. camelCase is usually recommended in helm ecosystem. But I think it's better to keep component names consistent with https://github.com/open-telemetry/opentelemetry-demo/tree/main/src and https://github.com/open-telemetry/opentelemetry-demo/blob/main/docker-compose.yml. So just removing dashes would be probably good.

Allex1 (approved):
@wph95 Some docs or comments in values would be appropriate for new users. We can probably do this in a future docs pr. Great contribution! Thanks!

TylerHelmuth (commented):
@wph95 can you add a values.json.schema like https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/values.schema.json

charts/opentelemetry-demo-webstore/values.yaml
wph95: Because I want to automatically set environment variables like PRODUCT_CATALOG_SERVICE_ADDR by name.
If it's all lowercase, it won't be able to distinguish between words.
**I will change it to camel.** 
I'll move to the demo repo to discuss this later. (In the future, it is likely that value.yaml will be used to automatically generate docker-compose, thus avoiding similar problems)

wph95: fixed it

wph95:
> @wph95 can you add a values.json.schema like https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/values.schema.json

Good suggestion, just added

charts/opentelemetry-demo-webstore/templates/NOTES.txt
```diff
@@ -0,0 +1,22 @@
+=======================================================================================
+
+
+ ██████╗ ████████╗███████╗██╗         ██████╗ ███████╗███╗   ███╗ ██████╗ 
+██╔═══██╗╚══██╔══╝██╔════╝██║         ██╔══██╗██╔════╝████╗ ████║██╔═══██╗
+██║   ██║   ██║   █████╗  ██║         ██║  ██║█████╗  ██╔████╔██║██║   ██║
+██║   ██║   ██║   ██╔══╝  ██║         ██║  ██║██╔══╝  ██║╚██╔╝██║██║   ██║
+╚██████╔╝   ██║   ███████╗███████╗    ██████╔╝███████╗██║ ╚═╝ ██║╚██████╔╝
+ ╚═════╝    ╚═╝   ╚══════╝╚══════╝    ╚═════╝ ╚══════╝╚═╝     ╚═╝ ╚═════╝
```
dmitryax, line 9: Nice :)

charts/opentelemetry-demo-webstore/templates/NOTES.txt
```diff
@@ -0,0 +1,22 @@
...
+
+ ██████╗ ████████╗███████╗██╗         ██████╗ ███████╗███╗   ███╗ ██████╗ 
+██╔═══██╗╚══██╔══╝██╔════╝██║         ██╔══██╗██╔════╝████╗ ████║██╔═══██╗
+██║   ██║   ██║   █████╗  ██║         ██║  ██║█████╗  ██╔████╔██║██║   ██║
+██║   ██║   ██║   ██╔══╝  ██║         ██║  ██║██╔══╝  ██║╚██╔╝██║██║   ██║
+╚██████╔╝   ██║   ███████╗███████╗    ██████╔╝███████╗██║ ╚═╝ ██║╚██████╔╝
+ ╚═════╝    ╚═╝   ╚══════╝╚══════╝    ╚═════╝ ╚══════╝╚═╝     ╚═╝ ╚═════╝ 
+                                                                          
+
+- Front UI: http://localhost:8080
```
dmitryax, line 12: ?
```suggestion
- Frontend UI: http://localhost:8080
```

dmitryax (commented):
the chart is still located at /charts/opentelemetry-demo-webstore. Can you please update the directory name?

charts/opentelemetry-demo-webstore/templates/NOTES.txt
TylerHelmuth: It's beautiful!

charts/opentelemetry-demo-webstore/templates/NOTES.txt
wph95: yep make sence

dmitryax (approved):
LGTM. Thanks @wph95!

charts/opentelemetry-demo/templates/component.yaml
```diff
@@ -0,0 +1,18 @@
...
+    {{- $config := set . "observability" $.Values.observability }}
+    {{- $config := set . "serviceMapping" $serviceMapping }}
+    {{- $config := set . "depends" $depends }}
+
+    {{- if $config.enabled -}}
+        {{- include "otel.demo.deployment" $config -}}
+        {{- include "otel.demo.service" $config -}}
+    {{ end }}
+
+{{- end }}
```
TylerHelmuth, line 18:
```suggestion
{{- end }}

```

charts/opentelemetry-demo/templates/services.yaml
```diff
@@ -0,0 +1,28 @@
+{{- define "otel.demo.service" }}
```
TylerHelmuth, line 1: Lets merge the `otel.demo.service`, `otel.demo.deployment`, and `otel-demo.otelcol.config` templates into 1 template file, similar to `charts/opentelemetry-collector/tempaltes/_config.tpl`.  Maybe something like `_objects.tpl`. I could also see an argument for putting `otel-demo.otelcol.config` in a different tpl file since it isn't a k8s object.

The intended goal being all templates are defined in `_*.tpl` files.

charts/opentelemetry-demo/values.schema.json
```diff
@@ -0,0 +1,188 @@
...
+                    "type": "boolean"
+                }
+            },
+            "required": [
+                "enabled"
+            ],
+            "title": "ObservabilityProp"
+        }
+    }
+}
```
TylerHelmuth, line 188:
```suggestion
}

```

charts/opentelemetry-demo/values.yaml
```diff
@@ -0,0 +1,116 @@
...
+    env:
+    - name: ASPNETCORE_URLS
+      value: http://*:8080
+
+  checkoutService:
+    enabled: true
+    servicePort: 8080
+
+  currencyService:
+    enabled: true
```
TylerHelmuth, line 37: If this component is disabled, what is the outcome for the checkout-service and frontend-service, both which have a dependency on the currency-service in `_deployment.tpl`?

TylerHelmuth (commented):
Couple more questions as it had been a while since I look through things in depth.

charts/opentelemetry-demo/values.yaml
wph95: For example, the frontend service has many dependencies (including currencyService), and if currencyService is disabled, the frontend service will still run normally. But when the order is placed, it will fail to place the order. In jaeger, you can see that frontend tries to access the currency but returns 500.

charts/opentelemetry-demo/templates/services.yaml
wph95: Thank you for the guidance, learned!

TylerHelmuth (approved):
Thanks @wph95 and everyone else involved! Excited to see this chart help our end users.

PR #281, opened 2022-07-20
Update maintainers list in charts

TylerHelmuth:
Opening this PR partially as a question and also as a fix, depending on the answer to the question.

Looking through the helm docs it looks to me like the maintainers list in the chart.yaml is ambiguous. It could refer only to the chart maintainers ([as with this prometheus example](https://github.com/prometheus-community/helm-charts/blob/main/PROCESSES.md#adding-chart-maintainers)) or it could refer to the chart maintainers _and_ the maintainers of the image being deployed.  

Either way, our list needs updated.  But this option should we use?  @dmitryax @Allex1 what do you think?

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.22.1
+version: 0.22.2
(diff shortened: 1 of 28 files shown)


dmitryax:
Sounds good to me. Let's stick to the list of maintainers of the repo. Later we will add code owners per chart and update the list in Chart.yaml accordingly

Issue #136, opened 2022-03-04
Introduce values schema for the operator chart


dmitryax:
Introduce [values.schema.json](https://helm.sh/docs/topics/charts/#schema-files) for values.yaml input validation in opentelemetry-operator helm chart. All the objects in values.yaml with predefined set of properties should not allow unknown properties.

TylerHelmuth:
@Allex1 are you available to work on this?

Allex1:
@TylerHelmuth I can give it a go

PR #289, opened 2022-07-26
Introduce values schema for the operator chart

Allex1:
Addresses https://github.com/open-telemetry/opentelemetry-helm-charts/issues/136

### DIFF
--- a/charts/opentelemetry-operator/values.schema.json
+++ b/charts/opentelemetry-operator/values.schema.json
@@ -1,0 +1,891 @@
+{
+    "$schema": "http://json-schema.org/schema#",
+    "type": "object",
+    "default": {},
+    "title": "Values",
+    "required": [
+        "replicaCount",
+        "nameOverride",
+        "imagePullSecrets",
+        "manager",
+        "kubeRBACProxy",
+        "admissionWebhooks",
+        "role",
+        "clusterRole",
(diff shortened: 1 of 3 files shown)


charts/opentelemetry-operator/values.schema.json
```diff
@@ -0,0 +1,891 @@
...
+        "kubeRBACProxy",
+        "admissionWebhooks",
+        "role",
+        "clusterRole",
+        "affinity",
+        "tolerations",
+        "nodeSelector",
+        "hostNetwork"
+    ],
+    "properties": {
```
TylerHelmuth, line 20: I think we should use  `"additionalProperties": false` at the top level, and then within objects when appropriate.  This is the field that helps use with deprecations or breaking changes in the future.  

With the addition of `additionalProperties` flag at the top level, we should also include 

```yaml
    "enabled": {
      "description": "Usually used when using Opentelemetry-collector as a subchart.",
      "type": "boolean"
    },
    "global": {
      "type": "object"
    },
```
in the properties.  These are fields that we've added in the collector chart to help with a helm defect and subcharts.

Allex1: @TylerHelmuth added

TylerHelmuth (approved):

PR #294, opened 2022-07-28
Presets for only logsCollection and hostMetrics

### LINKED ISSUE #158: Replace default config with config presets

TylerHelmuth:
Implements presets for logsCollection and hostMetrics.  config continues to contain default config.  Use `append` to add `filelog` and `hostmetrics` receivers into user config.

Closes #158


charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -42,7 +42,8 @@ Build config file for daemonset OpenTelemetry Collector
 {{- $data := dict "Values" $values | mustMergeOverwrite (deepCopy .) }}
 {{- $config := include "opentelemetry-collector.baseConfig" $data | fromYaml }}
 {{- $config := include "opentelemetry-collector.ballastConfig" $data | fromYaml | mustMergeOverwrite $config }}
-{{- $config := mustMergeOverwrite (include "opentelemetry-collector.daemonset.containerLogsConfig" $data | fromYaml) $config }}
+{{- $config := mustMergeOverwrite (include "opentelemetry-collector.daemonset.logsCollectionConfig" $data | fromYaml) $config }}
+{{- $config := mustMergeOverwrite (include "opentelemetry-collector.daemonset.hostMetricsConfig" $data | fromYaml) $config }}
```
dmitryax, line 46:
```
{{- if .Values.presets.hostMetrics.enabled }}
```
should probably go here. 

What if instead of `mustMergeOverwrite` we have `include "opentelemetry-collector.daemonset.applyHostMetricsConfig" $config ` that would do list merges where needed, e.g. `.service.pipelines.metrics.receivers | append "hostmetrics" | uniq`?

TylerHelmuth: I think we still have to do a `mustMergeOverwrite` for the actual component configuration, but I like the `append` idea for dealing with the pipelines.  I'll look into that tomorrow.

dmitryax: > I think we still have to do a mustMergeOverwrite for the actual component configuration

Yes, but inside `opentelemetry-collector.applyHostMetricsConfig` for particular parts of the config

TylerHelmuth: @dmitryax there is still cleanup work today but I think I got the "append" solution working

TylerHelmuth:
@dmitryax I have run `make generate-examples` and pushed the results. so we can see how the changes affect things.  Once we are good with this approach i'll bump the chart and regenerate the examples.

charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -53,6 +63,9 @@ Build config file for deployment OpenTelemetry Collector
 {{- $values := deepCopy .Values }}
 {{- $data := dict "Values" $values | mustMergeOverwrite (deepCopy .) }}
 {{- $config := include "opentelemetry-collector.baseConfig" $data | fromYaml }}
+{{- if .Values.presets.hostMetrics.enabled }}
```
TylerHelmuth, line 66: Should we allow hostmetrics to be used in deployment mode or keep it to daemonset mode like filelog is?

dmitryax: I don't think we should care about such misconfigurations. Probably just mention that `hostMetrics` supposed to be used with daemonset deployment only

TylerHelmuth: I would agree.  Kind feels like we should remove the restriction from logsCollection as well.

charts/opentelemetry-collector/UPGRADING.md
```diff
@@ -1,5 +1,32 @@
...
+If you are using `containerLogs.enabled` and also enabling collection of the collector logs you can use `includeCollectorLogs`
+
+```yaml
+presets:
+  logsCollection:
+    enabled: true
+    includeCollectorLogs: true
+```
+
+You no longer need to update `config.service.pipelines.logs` to include the filelog receiver yourself as the preset will automatically update the defau logs pipeline to include the filelog receiver.
```
dmitryax, line 24:
```suggestion
You no longer need to update `config.service.pipelines.logs` to include the filelog receiver yourself as the preset will automatically update the defaut logs pipeline to include the filelog receiver.
```

charts/opentelemetry-collector/examples/daemonset-hostmetrics/rendered/daemonset.yaml
```diff
@@ -110,6 +110,6 @@ spec:
               - key: relay
                 path: relay.yaml
         - name: hostfs
-          hostPath:
+          hostPath:
```
dmitryax, line 113:
```suggestion
          hostPath:
```

charts/opentelemetry-collector/templates/NOTES.txt
```diff
@@ -17,3 +17,11 @@
 {{- if not .Values.configMap.create }}
 [WARNING] "configMap" wil not be created and "config" will not take effect.
 {{ end }}
+
+{{- if (eq (toString .Values.secretMounts) "true") }}
+{{ fail "[ERROR] 'containerLogs' should not be used.  Use 'presets.logsCollection' instead. See https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/UPGRADING.md for instructions on how to migrate." }}
```
dmitryax, line 22:
```suggestion
{{ fail "[ERROR] 'containerLogs' should not be used.  Use 'presets.logsCollection' instead. See https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/UPGRADING.md#0230-to-0240 for instructions on how to migrate." }}
```

charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -111,13 +124,45 @@ Get otel memory_limiter ballast_size_mib value based on 40% of resources.memory.
...
+    collection_interval: 10s
+    scrapers:
+      cpu:
+      disk:
+      filesystem:
+      memory:
+      network:
+      load:
+      paging:
+      processes:
```
dmitryax, line 146: What if we remove `scrapers`? Does it have a default list so it can be easily overridden?

charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -111,13 +124,45 @@ Get otel memory_limiter ballast_size_mib value based on 40% of resources.memory.
 {{- div (mul (include "opentelemetry-collector.convertMemToMib" .) 40) 100 }}
 {{- end -}}
 
-{{- define "opentelemetry-collector.daemonset.containerLogsConfig" -}}
-{{- if .Values.containerLogs.enabled }}
+{{- define "opentelemetry-collector.applyHostMetricsConfig" -}}
+{{- $config := mustMergeOverwrite (include "opentelemetry-collector.hostMetricsConfig" .Values | fromYaml) .config }}
+{{- $newList := dict "service" (dict "pipelines" (dict "metrics" (dict "receivers" (append .config.service.pipelines.metrics.receivers "hostmetrics" | uniq))))}}
+{{- $config := mustMergeOverwrite $config $newList  }}
```
dmitryax, line 130: I believe we can use `set` http://masterminds.github.io/sprig/dicts.html with updated list of receivers instead of `mustMergeOverwrite`

charts/opentelemetry-collector/UPGRADING.md
```diff
@@ -1,5 +1,32 @@
...
+    enabled: true
+```
+
+If you are using `containerLogs.enabled` and also enabling collection of the collector logs you can use `includeCollectorLogs`
+
+```yaml
+presets:
+  logsCollection:
+    enabled: true
+    includeCollectorLogs: true
```
dmitryax, line 21: What if we add `filelogConfig` here?

```
presets:
  logsCollection:
    enabled: true
    includeCollectorLogs: true
    filelogConfig: 
       ...
```

Same for hostmetrics preset. Then it becomes fully overridable. `exclude` can be adjusted with `set` from http://masterminds.github.io/sprig/dicts.html if includeCollectorLogs=false.

TylerHelmuth: I thought about that too.  Feels natural for hostmetrics, but the default filelog config is massive and it felt strange to add it to the values.yaml.  The benefits probably outweigh the messy values.yaml tho.  I'll try it out this week.

charts/opentelemetry-collector/templates/_config.tpl
TylerHelmuth: Ya I don't think you can since we can't take advantage of helm's 'null' concept since it isn't in the values.yaml.  Adding it there like you suggested below would let the list be updated.

charts/opentelemetry-collector/templates/_config.tpl
TylerHelmuth: Unfortunately I don't think `set` knows how to handle nested objects.  retrying to set a nested object like `"service.pipelines.logs.receivers"` resulted in a new root-level field called `"service.pipelines.logs.receivers"`

TylerHelmuth: Done.

charts/opentelemetry-collector/UPGRADING.md
TylerHelmuth: I think this is the only sticking point:

`exclude: [ /var/log/pods/{{ .Release.Namespace }}_{{ include "opentelemetry-collector.fullname" . }}*_*/{{ .Chart.Name }}/*.log ]
`

The filelog config today is built with some knowledge about where the collector is going to run.  If we switch the default for `includeCollectorLogs` to be `true` we can default the config to be `exclude: []` in values.yaml and then then we can use the same `append`/`set` logic to update the exclude list to the value above if `includeCollectorLogs` is switched to `false`.

dmitryax: Even if `includeCollectorLogs: false` and `exclude: []` by default, we can do the same `_ := set exclude ... append | uniq` trick in the templating

dmitryax: > Feels natural for hostmetrics, but the default filelog config is massive and it felt strange to add it to the values.yaml

You're right. I just checked the filelog config. It's too big, and overriding operators would be not convenient at all. My only concern is making `scrapers` configurable with enabled hostmetrics preset. There should be another way to do that

TylerHelmuth: Maybe it's as simple as adding the hostmetrics config to the interface but not filelog? Or we reduce the number of default scrapers and make it less likely a user would need to remove something from the default config.

TylerHelmuth: But if we do add hostmetrics config to the interface then users would be able to add/update/remove anything as if it was a default entry in .Values.config.

dmitryax: > Or we reduce the number of default scrapers and make it less likely a user would need to remove something from the default config.

Let's probably do this. It's always possible to disable particular metrics if needed

dmitryax: Better to start with smaller configuration interface for presets and enhance it if needed in future

TylerHelmuth: @dmitryax done.  Reduced to cpu, load, memory, disk, and network.

TylerHelmuth:
@dmitryax I believe I've addressed all unresolved comments.  If all looks good I'll regenerate the examples for final review.

charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -1,3 +1,8 @@
+
+{{- define "opentelemetry-collector.var_dump" -}}
```
dmitryax, line 2: Debug?

charts/opentelemetry-collector/values.yaml
```diff
@@ -8,6 +8,29 @@ fullnameOverride: ""
...
+# .Values.config and use the other fields supplied in the
+# values.yaml to configure k8s as necessary.
+presets:
+  # Configures the collector to collect logs.
+  # Adds the filelog receiver to the logs pipeline
+  # and adds the necessary volumes and volume mounts.
+  logsCollection:
+    enabled: false
+    includeCollectorLogs: false
+    filelogConfig:
```
dmitryax, line 27: This line is not needed right?

charts/opentelemetry-collector/values.schema.json
```diff
@@ -27,6 +27,29 @@
         ""
       ]
     },
+    "presets": {
+      "type": "object",
+      "properties": {
+        "logsCollection": {
+          "type": "object",
+          "properties": {
+            "enabled": {
```
dmitryax, line 36: Let's add `"additionalProperties": false` in all the objects. Also schema for `includeCollectorLogs` is missed

charts/opentelemetry-collector/templates/_config.tpl
TylerHelmuth: Ya I'll remove

charts/opentelemetry-collector/values.yaml
TylerHelmuth: Correct, thought I removed it.

charts/opentelemetry-collector/values.schema.json
TylerHelmuth: Good catch

charts/opentelemetry-collector/templates/NOTES.txt
```diff
@@ -17,3 +17,7 @@
 {{- if not .Values.configMap.create }}
 [WARNING] "configMap" wil not be created and "config" will not take effect.
 {{ end }}
+
+{{- if not (eq (toString .Values.containerLogs) "<nil>") }}
+{{ fail "[ERROR] 'containerLogs' should not be used.  Use 'presets.logsCollection' instead. See https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/UPGRADING.md#0231-to-0240 for instructions on how to migrate." }}
```
dmitryax, line 22: Will this ever be triggered? My feeling is that the schema validation will fail first. If that's the case, I would go with a backward compatible approach.

dmitryax: Oh you kept the deprecated `containerLogs` option in the schema. Makes sense now.

TylerHelmuth:
@dmitryax @Allex1 ready for final review

charts/opentelemetry-collector/templates/NOTES.txt
dmitryax: Still why not make it backward compatible for some time? Shouldn't be that complicated, right?

charts/opentelemetry-collector/values.yaml
```diff
@@ -8,6 +8,28 @@ fullnameOverride: ""
...
+presets:
+  # Configures the collector to collect logs.
+  # Adds the filelog receiver to the logs pipeline
+  # and adds the necessary volumes and volume mounts.
+  logsCollection:
+    enabled: false
+    includeCollectorLogs: false
+  # Configures the collector to collect host metrics.
+  # Adds the hostmetrics receiver to the metrics pipeline
+  # and adds the necessary volumes and volume mounts.
```
dmitryax, line 29: Please mention that it should be used with deamonset mode only

charts/opentelemetry-collector/templates/NOTES.txt
TylerHelmuth: I think we could.  I had originally not allowed it bc of the override stuff, but I think now that we can update the list and are using `| uniq` it should be ok.  I'll update that.

TylerHelmuth: Done.  `containerLogs.enabled` functions like `presets.logsCollection.enabled`.  If customers are using a config similar to the previous logs-collection example, their experience should not change.

charts/opentelemetry-collector/templates/_config.tpl
TylerHelmuth: > Should we allow hostmetrics to be used in deployment mode or keep it to daemonset mode like filelog is?

charts/opentelemetry-collector/values.yaml
TylerHelmuth: Done.  Based on [this conversation](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/294#discussion_r933594388), I went with `Best used with ...`.

dmitryax (approved):
LGTM. @Allex1 PTAL

Allex1 (approved):
🐾 it

PR #301, opened 2022-08-02
enhance otel demo chart‘s schema

Frapschen:
if otel demo chart become a subchart, we need add a schema to suport `enabled` and `global` value.


TylerHelmuth:
@Frapschen please bump the chart version to `0.1.2`

TylerHelmuth (approved):

PR #302, opened 2022-08-03
enhance otel demo env schema

Frapschen:
In some scenarios, we want to set some value from k8s metadata to our `OTEL_RESOURCE_ATTRIBUTES`,  like namespace, we need the configuration below:
```
- name: OTEL_K8S_NAMESPACE
  valueFrom:
    fieldRef:
      apiVersion: v1
      fieldPath: metadata.namespace
  value: k8s.namespace.name=$(OTEL_K8S_NAMESPACE)
```

I propose to add the schema to catch the feature described above.

### DIFF
--- a/charts/opentelemetry-demo/templates/_pod.tpl
+++ b/charts/opentelemetry-demo/templates/_pod.tpl
@@ -48,20 +48,18 @@ Get Pod Env
 
 {{- if .default.enabled  }}
 {{- if .env }}
-{{- $filteredDefaultEnv := list -}}
+{{- $defaultEnvMap := dict }}
 {{- range $defaultEnvItem := $.default.env }}
-{{- $matched :=false }}
+{{- $defaultEnvMap = set $defaultEnvMap $defaultEnvItem.name $defaultEnvItem }}
+{{- end }}
+{{- $envMap := dict }}
 {{- range $envItem := $.env }}
-{{- if eq $defaultEnvItem.name $envItem.name  }}
-{{- $matched = true }}
+{{- $envMap := set $envMap $envItem.name $envItem }}
 {{- end }}
-{{- end }}
-{{- if eq $matched false }}
-{{- $filteredDefaultEnv = append $filteredDefaultEnv $defaultEnvItem -}}
-{{- end }}
-{{- end }}
-{{ toYaml $filteredDefaultEnv }}
-{{ toYaml .env }}
+{{- $mergedEnvMap := merge $envMap $defaultEnvMap }}
+{{- $otelResourceAttributesList := get $mergedEnvMap "OTEL_RESOURCE_ATTRIBUTES" | list }}
+{{ unset $mergedEnvMap "OTEL_RESOURCE_ATTRIBUTES" | values | toYaml }}
+{{ $otelResourceAttributesList | toYaml }}
 {{- else }}
 {{ toYaml .default.env }}
 {{- end }}
--- a/charts/opentelemetry-demo/templates/_pod.tpl
+++ b/charts/opentelemetry-demo/templates/_pod.tpl
@@ -72,12 +72,6 @@ Get Pod Env
 {{- if .observability.otelcol.enabled }}
 - name: OTEL_EXPORTER_OTLP_ENDPOINT
   value: http://{{ include "otel-demo.name" . }}-otelcol:4317
-- name: OTEL_RESOURCE_ATTRIBUTES
-{{- if .default.enabled  }}
-  value: service.name={{ .name | kebabcase }},k8s.namespace.name=$(OTEL_K8S_NAMESPACE),k8s.node.name=$(OTEL_RESOURCE_ATTRIBUTES_NODE_NAME),k8s.pod.name=$(OTEL_RESOURCE_ATTRIBUTES_POD_NAME)
-{{- else }}
-  value: service.name={{ .name | kebabcase }}
-{{- end }}
 {{- end }}
 
 {{- if .servicePort}}
--- a/charts/opentelemetry-demo/templates/_pod.tpl
+++ b/charts/opentelemetry-demo/templates/_pod.tpl
@@ -41,25 +41,29 @@ Get Services Port Mapping
 {{- end }}
 
 {{/*
-Get Pod Env 
+Get Pod Env
 */}}
 {{- define "otel-demo.pod.env" -}}
 {{- $prefix := include "otel-demo.name" $ }}
 
-{{- if .env }}
-{{- toYaml .env }}
+{{- if .default  }}
+{{- toYaml .default.env -}}
 {{- end }}
 
-{{- if .observability.otelcol.enabled }} 
+{{- if .env }}
+{{ toYaml .env }}
+{{- end }}
+
+{{- if .observability.otelcol.enabled }}
 - name: OTEL_EXPORTER_OTLP_ENDPOINT
   value: http://{{ include "otel-demo.name" . }}-otelcol:4317
 - name: OTEL_RESOURCE_ATTRIBUTES
-  value: service.name={{ .name | kebabcase }}
+  value: service.name={{ .name | kebabcase }},k8s.namespace.name=$(OTEL_K8S_NAMESPACE),k8s.node.name=$(OTEL_RESOURCE_ATTRIBUTES_NODE_NAME),k8s.pod.name=$(OTEL_RESOURCE_ATTRIBUTES_POD_NAME)
 {{- end }}
 
 {{- if .servicePort}}
 - name: {{ printf "%s_PORT" .name | snakecase | upper }}
-  value: {{.servicePort | quote}} 
+  value: {{.servicePort | quote}}
 {{- end }}
 
 # {{ $.depends }}
@@ -74,7 +78,7 @@ Get Pod Env
 {{- end }}
 
 {{/*
-Get Pod ports 
+Get Pod ports
 */}}
 {{- define "otel-demo.pod.ports" -}}
 {{- if .ports }}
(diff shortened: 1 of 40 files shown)


TylerHelmuth:
@Frapschen Please bump the chart version.

TylerHelmuth:
@Frapschen Can you open an issue that lays out in more details the issue being solved by this PR? It's _possible_ the solution will end up being a new section in the values.yaml instead.

wph95:
Are you ready to review this pr? I see that only values.schema.json/chart.yaml has been modified

For the need to add specific envs, I think issues need to be discussed. In the absence of a context, I can't tell if this addition is feasible.

But if it's a generic solution for custom environment variables. I think it is necessary (but low priority)
For example in value.yaml
```
default:
   env:
    - name: OTEL_K8S_NAMESPACE
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: metadata.namespace
      value: k8s.namespace.name=$(OTEL_K8S_NAMESPACE)


```

charts/opentelemetry-demo/values.yaml
```diff
@@ -4,6 +4,24 @@ observability:
   jaeger:
     enabled: true
 
+default:
```
TylerHelmuth, line 7: I think we should supply the option to allow downard api to be used in env vars and also a way to set more values in `OTEL_RESOURCE_ATTRIBUTES`, but unless the demo was changed to require these I don't think they should be default.

Frapschen: got it.

charts/opentelemetry-demo/templates/_pod.tpl
```diff
@@ -41,25 +41,29 @@ Get Services Port Mapping
...
+{{ toYaml .env }}
 {{- end }}
 
-{{- if .observability.otelcol.enabled }} 
+{{- if .observability.otelcol.enabled }}
 - name: OTEL_EXPORTER_OTLP_ENDPOINT
   value: http://{{ include "otel-demo.name" . }}-otelcol:4317
 - name: OTEL_RESOURCE_ATTRIBUTES
-  value: service.name={{ .name | kebabcase }}
+  value: service.name={{ .name | kebabcase }},k8s.namespace.name=$(OTEL_K8S_NAMESPACE),k8s.node.name=$(OTEL_RESOURCE_ATTRIBUTES_NODE_NAME),k8s.pod.name=$(OTEL_RESOURCE_ATTRIBUTES_POD_NAME)
```
TylerHelmuth: I think normally in a k8s deployment using otel these values would be added via the collector.  What is the advantage of using `OTEL_RESOURCE_ATTRIBUTES` instead?

Frapschen: sorry, the newest commit have recover this line. 

In my scenario, the collector and actual service is deploying in a different namespace, so the collector doesn't know which namespace the service is deployed or what the pod name which the service's container is running. Pod name is a very important uniform identity for us, in the service monitor scenario, we want to know every service instance's metrics.

TylerHelmuth: I guess since the k8sattributes processor requires elevated permissions and downard api doesn't this solution makes sense.

TylerHelmuth: Actually, these 3 values are pretty standard for k8s deployments using otel.  Maybe it does make sense to include them by default.

TylerHelmuth: @dmitryax @Allex1 @wph95 what do you think?

dmitryax: We can start with this approach. It makes sense to me. 

I'd say ideally we should use a fully featured collector installation with `k8sattributes` processor, but we don't even have that in the collector chart. @TylerHelmuth I think that is a good candidate for another preset, WDYT? 

Maybe later we can use the collector with enabled `k8sattributes` processor as a subchart here.

TylerHelmuth: @dmitryax now that we've got presets in place I think we could better support k8sattributes and its RBAC needs in the opentelemetry-collector chart.  

I'm not entirely sure we'll want to introduce the same RBAC requirements into the demo chart tho; my gut says the less the demo chart has to install the better.  This PR's approach means we won't need any cluster roles or cluster role bindings.  I'm open to the idea tho.

TylerHelmuth: @Frapschen please add back these 3 fields as defaults, but configurable.

Allex1 (approved):

charts/opentelemetry-demo/values.yaml
```diff
@@ -4,6 +4,25 @@ observability:
...
```
TylerHelmuth, line 7: @Frapschen can you add a comment above `default` that explains that items in this object will apply to all components?

charts/opentelemetry-demo/templates/_pod.tpl
```diff
@@ -41,25 +41,33 @@ Get Services Port Mapping
...
 {{- define "otel-demo.pod.env" -}}
 {{- $prefix := include "otel-demo.name" $ }}
 
+{{- if .default.enabled  }}
+{{- toYaml .default.env }}
+{{- end }}
+
 {{- if .env }}
-{{- toYaml .env }}
+{{ toYaml .env }}
```
TylerHelmuth, line 54: a component's individual environment variables should probably be merged with the defaults.  That way there are no duplicates and a individual component could change a value without having to change it for all other components.

charts/opentelemetry-demo/templates/_pod.tpl
```diff
@@ -41,25 +41,33 @@ Get Services Port Mapping
...
 {{- end }}
 
-{{- if .observability.otelcol.enabled }} 
+{{- if .observability.otelcol.enabled }}
 - name: OTEL_EXPORTER_OTLP_ENDPOINT
   value: http://{{ include "otel-demo.name" . }}-otelcol:4317
 - name: OTEL_RESOURCE_ATTRIBUTES
+{{- if .default.enabled  }}
+  value: service.name={{ .name | kebabcase }},k8s.namespace.name=$(OTEL_K8S_NAMESPACE),k8s.node.name=$(OTEL_RESOURCE_ATTRIBUTES_NODE_NAME),k8s.pod.name=$(OTEL_RESOURCE_ATTRIBUTES_POD_NAME)
```
TylerHelmuth, line 62: I think this line needs to be configurable.  These should be enabled by default but should be able to be removed.  probably another section in the `default` object in the values.yaml.

Frapschen: you means this?
```
default:
  enabled: true
  env:
    - name: OTEL_K8S_NAMESPACE
      enabled: true
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: metadata.namespace
    - name: OTEL_RESOURCE_ATTRIBUTES_POD_NAME
      enabled: false
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: metadata.name
```

and finally `OTEL_RESOURCE_ATTRIBUTES` would like:
```
  value: service.name=cart-service,k8s.namespace.name=$(OTEL_K8S_NAMESPACE)
```

How about use a prefix, like OTEL_RESOURCE, to mark which envs shoud add into the `OTEL_RESOURCE_ATTRIBUTES`?

charts/opentelemetry-demo/templates/_pod.tpl
Frapschen: finished,please take a look

charts/opentelemetry-demo/templates/_pod.tpl
TylerHelmuth: Actually, "OTEL_RESOURCE_ATTRIBUTES" is an environment variable right? Can we put it directly in default?

Frapschen: ok, put it directly in default section is esay and Flexible way.

charts/opentelemetry-demo/values.yaml
```diff
@@ -4,6 +4,33 @@ observability:
   jaeger:
     enabled: true
 
+# if the enabled is true, the env items in this object will apply to all components
+# and will add into the OTEL_RESOURCE_ATTRIBUTES env
```
TylerHelmuth, line 8:
```suggestion
# if the enabled is true, the items in this object will apply to all components
```

charts/opentelemetry-demo/templates/_pod.tpl
TylerHelmuth: I like it, but since `OTEL_RESOURCE_ATTRIBUTES` relies on other env vars, it definitely has to be after the others in the rendered yaml file.  I'm pretty sure `toYaml` won't preserve the order of a dictionary, so there isn't a guarantee that it says at the end of the list like in the values.yaml.  Can you test that?  Sorry, I didn't think about it till this morning.

TylerHelmuth: You. might be able to do something fancy with the [`omit` function](https://helm.sh/docs/chart_template_guide/function_list/#omit) and the [`get` function](https://helm.sh/docs/chart_template_guide/function_list/#get) to omit `OTEL_RESOURCE_ATTRIBUTES` from the initial `toYaml` of the dictionary and then do a specific `get` for `OTEL_RESOURCE_ATTRIBUTES` after.

TylerHelmuth:
@Frapschen I will be gone next week, but @dmitryax @Allex1 and @wph95 can help move this forward.

charts/opentelemetry-demo/templates/_pod.tpl
Frapschen: .default.env and .env  are not a map but slice containing env maps. I will try to use build-in funciton to implement the merge

charts/opentelemetry-demo/templates/_pod.tpl
Frapschen: I will change the logic to guarantee it.

Frapschen:
@wph95 can you have time to review?

wph95 (approved):

PR #311, opened 2022-08-07
[opentelemetry-collector] Add extra ingress configurations

clarkjohnd:
## Overview

This PR adds a new values parameter `extraIngresses` to the OpenTelemetry Collector Helm chart. Use case for this is where multiple receivers are deployed in a single OpenTelemetry Collector configuration that need different ingress annotations/configuration. For example:

- Receiver A (OTLP receiver etc.) uses NGINX ingress controller but needs to whitelist IPs to client CIDR ranges (with annotations) to only allow metrics sent from external client environments
- Receiver B (AWS CloudWatch receiver etc.) uses NGINX ingress controller but needs to whitelist AWS CIDR ranges (with annotations) to only allow metrics sent from AWS platform through CloudWatch Metric Streams

This can currently be acomplished by manually deploying another ingress resource or by deploying another Helm installation, but this change allows it to be done in a single installation instead.

## Changes

`extraIngresses` merges with `ingress` configuration in the Ingress resource template and builds from a `range` function, so a few minor changes in code but no changes to end result if no `extraIngresses` defined. Reason for doing it this way is to avoid breaking changes by changing `ingress` to a map list etc. `extraIngresses` default values is `[]`, so no impact on existing installations with chart version updates.

The schema for `extraIngresses` elements is the same as `ingress` with the following exceptions:
- No `enabled` key - this is inherited from `ingress` configuration
- A `name` key (required) exists to differentiate ingress resources - this is used as a suffix for the ingress resource to differentiate from the root ingress (no suffix) and other additional ingress resources.

A new CI values file has been added (and tested), and chart version minor updated to v0.26.0. Open to any feedback or suggested changes!


Allex1 (approved):

charts/opentelemetry-collector/values.yaml
```diff
@@ -285,6 +285,23 @@ ingress:
 #      hosts:
 #        - collector.example.com
 
+# Additional ingresses - only created if ingress.enabled is true
+# Useful for when differently annotated ingress services are required
+extraIngresses: []
```
dmitryax, line 290: I'm not sure this rare use case warrants adding another section in values.yaml. The configuration interface is already pretty complicated, and will be hard to read and maintain if we keep adding configurations for any possible use case. If this particular problem can be solved with another k8s object created manually, I wouldn't like to merge this. Looking for other opinions cc @TylerHelmuth

Allex1: @dmitryax I've mixed feelings as well on this, and I may be biased by my own env but I see 2 valid use cases for this pr:
1. Many companies use at least 2 networks (corporate and cloud/multi cloud), so just 1 collector might need at least 2 ingress types
2. sub-charting the collector chart multiple times for different types of collectors for multiple use cases.

dmitryax: > 2. sub-charting the collector chart multiple times for different types of collectors for multiple use cases.

Not sure I understand how this PR helps in such a use case.

Allex1: right, depending on the setup that might work with the default ingress config

clarkjohnd: @dmitryax @Allex1 @TylerHelmuth I understand the issue with adding more sections to `values.yaml`, but certainly from my perspective this could enhance functionality of a lot of different setups. Manually deploying ingresses isn't exactly complicated but replicating the default chart ingress multiple times (labels, names, links to services etc.) manually or through IaC which instead could be fully managed in the single chart without any breaking changes to existing installations seems like a feature some might take advantage of if they see it exists, for the cost of a new default empty list section in `values.yaml`.

I am of course completely biased as this feature would help me considerably, but I do see the OpenTelemetry Collector as somewhat niche compared to other charts as a single deployment can have a huge unpredictable combination of different receivers which can benefit from customisable endpoints, so for me at least this feature completes that package and for me at least removes any external Kubernetes resources required for a production setup of OTEL. Either way, I appreciate the feedback! Cheers.

dmitryax, line 290: Can we move it under `ingress->additionalIngresses`?

charts/opentelemetry-collector/values.yaml
```diff
@@ -285,6 +285,23 @@ ingress:
...
+#    hosts:
+#      - host: collector.example.com
+#        paths:
+#          - path: /
+#            pathType: Prefix
+#            port: 4318
+#    tls:
+#      - secretName: collector-tls
+#        hosts:
+#          - collector.example.com
```
dmitryax, line 303: Does every other ingress needs a separate `tls` option or `ingress.tls` can be used instead?

clarkjohnd: I would probably keep `tls` as different hosts could be configured in additional ingresses which require different TLS secrets? Could add an option in `additionalIngresses` to inherit TLS from the root ingress but maybe that's going too far.

dmitryax: It depends on how common the use case when TLS is shared across all the ingresses. If it's not common at all let's not inherit it. Otherwise we may even omit the `tls` option in `additionalIngresses` until it's needed and use `ingress.tls` for now

clarkjohnd: One of my specific use case just now uses seperate secrets per ingress (an AWS CloudWatch receiver with a standard public CA TLS certificate secret, and an OTLP receiver with a private CA TLS certificate secret used for client certificate authentication), so I'd prefer to just keep `tls` defined per additional ingress for now, and if there was feedback that a common TLS config was often being used across multiple ingresses by users, could add an `inheritTLS` option in `additionalIngresses` or something in the future?

charts/opentelemetry-collector/values.yaml
clarkjohnd: Moved to `ingress.additionalIngresses`.

charts/opentelemetry-collector/values.schema.json
```diff
@@ -474,6 +474,21 @@
         },
         "annotations": {
           "type": "object"
+        },
+        "additionalIngresses": {
+          "type": "array",
+          "items": {
+            "type": "object",
+            "additionalProperties": true,
```
dmitryax, line 482: Let's define the full schema here and disallow `additionalProperties`

charts/opentelemetry-collector/values.yaml
dmitryax: Ok let's add this configuration

charts/opentelemetry-collector/values.yaml
dmitryax: Ok, sounds good

charts/opentelemetry-collector/values.schema.json
clarkjohnd: I've defined the full schema for both root ingress and additional ingresses as both are using the same schema (except `enabled` and `name` being in each respectively). I've also set required properties based on [Kubernetes reference](https://kubernetes.io/docs/reference/kubernetes-api/service-resources/ingress-v1/) (basically the path properties) and set enum for path type.

dmitryax (approved):

PR #317, opened 2022-08-11
Remove conditional metrics address

jaronoff97:
Following up from [here](https://cloud-native.slack.com/archives/C03HVLM8LAH/p1660146696356489) we want to make it always possible to scrape metrics of the otel operator

### DIFF
--- a/charts/opentelemetry-operator/Chart.yaml
+++ b/charts/opentelemetry-operator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.11.3
+version: 0.11.4
 description: OpenTelemetry Operator Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-operator/templates/deployment.yaml
+++ b/charts/opentelemetry-operator/templates/deployment.yaml
@@ -23,11 +23,7 @@ spec:
       hostNetwork: {{ .Values.hostNetwork }}
       containers:
         - args:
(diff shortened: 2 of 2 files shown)


dmitryax:
@jaronoff97 please bump the patch version of the chart

dmitryax (approved):

Allex1 (approved):

ChristianCiach:
I am surprised that this has been merged without any security considerations.

I've read the linked Slack discussion, but no one there seemed to understand why the metrics-endpoint was only exposed on 127.0.0.1 to begin with. 

The whole point of exposing the metrics on localhost only was to *guarantee* that all clients have to use the kube-rbac-proxy (exposed on 0.0.0.0:8443) for authentication. Exposing the metrics on `127.0.0.1:8080` exposes to metrics to **everyone without any authentication**, rendering the kube-rbac-proxy completely useless.

Yes, I've seen that the kube-rbac-proxy [is subject to be removed](https://github.com/open-telemetry/opentelemetry-operator/issues/3369), but the discussion there makes it clear that it probably won't be removed *without any replacement*. 

Or am I wrong about all of this and the metrics of the operator are so little confidential that it's okay to always expose them to everybody?

As it stands, the kube-rbac-proxy serves no purpose whatsoever, since the metrics-port is always accessible by the whole cluster without authentication (unless there are some NetworkPolicies deployed). Since the metrics-endpoint is currently hardcoded to `0.0.0.0`, we should just remove the kube-rbac-proxy, because it currently suggests some kind of security where there really is none.

PR #323, opened 2022-08-17
[opentelemetry-collector] Refactoring labels

nlamirault:
- Refactoring the `app.kubernetes.io/component` label into Helm helper
- Support for additional labels into all resources (like `mycompany.io/stack: xxx`)


TylerHelmuth:
@nlamirault please break changes for different charts into separate PRs.

charts/opentelemetry-collector/templates/daemonset.yaml
```diff
@@ -9,15 +9,13 @@ spec:
   selector:
     matchLabels:
       {{- include "opentelemetry-collector.selectorLabels" . | nindent 6 }}
-      component: agent-collector
```
dmitryax, line 12: This change breaks the upgrade process. Users will have to reinstall the chart.

This has to be well documented. Also please provide more justification for this change with references.

And it would be nice if you can find an example of graceful handling of this change in other helm charts.

charts/opentelemetry-collector/templates/_helpers.tpl
```diff
@@ -48,6 +52,11 @@ Selector labels
 {{- define "opentelemetry-collector.selectorLabels" -}}
 app.kubernetes.io/name: {{ include "opentelemetry-collector.name" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
+{{- if eq .Values.mode "daemonset" }}
+app.kubernetes.io/component: agent-collector
```
TylerHelmuth, line 56: @dmitryax if this is changed from `app.kubernetes.io/component` to `component` it wouldn't break upgrades right?  But @nlamirault is this related to the ability to add additional labels?  Feels like we could keep hard coding `component` in the templates and add the ability to provide additional labels

TylerHelmuth: Also a side affect of adding `component` here would be that it is included in the templates' metadata labels.   Is that intended?

dmitryax: Any change to selector labels breaks the k8s deployment upgrade. deployment has to be recreated.

dmitryax: > @dmitryax if this is changed from app.kubernetes.io/component to component it wouldn't break upgrades right?

Should be right

nlamirault: i change `component` to `app.kubernetes.io/component` to be consistent with [kubernetes recommanded labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/).
Actually, i could also add a `component` label into in the templates' metadata labels to not perform a breaking change

TylerHelmuth: @nlamirault is the change necessary to meet the problem you are trying to solve?  Isn't the addition of `additionalLabels` meeting your needs?

TylerHelmuth: @nlamirault see my question above

nlamirault: We execute some policies on manifests, like check that Kubernetes recommanded labels exists.
It fails due to missing `app.kubernetes.io/component` label.
So i change the `component`  label to the Kubernetes recommanded label `app.kubernetes.io/component` 

Additional labels is another feature we would like.
To add to all resources a label like `mycompany/stack: observability`

TylerHelmuth: > Additional labels is another feature we would like.

Can you breakout this feature into its own PR?  

For `app.kubernetes.io/component` I think adding it in addition to `component` is an acceptable solution and works towards maybe dropping `component` in the future at some point.  @dmitryax do you agree?

dmitryax: Any change in pod labels that are used in `matchLabels` section requires removing and creating the whole deployment. This is a significant breaking change for existing users, they would have to run `helm delete` and `helm install` to upgrade to the new version. I'm not saying I'm against it, but this is not how it should be handled. To make this through, this PR needs at least the following:
- Clear justification for the change with an issue submitted that includes references
- Documented upgrade plan for existing users. Ideally somehow automated, I'm sure we are not the only ones who want to handle this gracefully, so just look around for inspiration in other charts

PR #321, opened 2022-08-15
[opentelemetry-operator] Add targets to update and check CRDs

TylerHelmuth:
Adds new make targets to update and check the operator CRDs.  

Closes #298

PR #324, opened 2022-08-17
[opentelemetry-operator] Remove operator chart release folder

TylerHelmuth:
This PR removes the release folder from the operator chart.  This folder's intention was to make updating the operator chart easier, but it has fallen into disrepair and no longer upgrades the chart correct.  Instead of troubleshooting the code, I suggest we remove it, relying instead on simpler makefile targets, manual changes, and CI tests.

Once this is removed I will create a second PR with new upgrade steps for each chart.


dmitryax (approved):

Allex1 (approved):

Allex1:
Related to https://github.com/open-telemetry/opentelemetry-helm-charts/pull/321

PR #333, opened 2022-08-19
[opentelemetry-demo] bump version to 0.3.4-alpha

TylerHelmuth:
Bumps the opentelemetry-demo chart to use 0.3.4-alpha demo images


xoscar (approved):

TylerHelmuth:
Closing until Demo images are available on docker

wph95 (approved):

wph95:
By the way, do we need to add a github action to automatically update the version.(when the otel-demo launches a new release)

TylerHelmuth:
> By the way, do we need to add a github action to automatically update the version.(when the otel-demo launches a new release)

Maybe, but it is possible when a demo version releases there will need to be changes to the chart itself (like right now, I can't get this PR's CI to pass and I am trying to figure out why), so there is always the possibility for manual intervention to be needed.  This happens mainly with the operator chart and not the collector chart.

For the demo chart, I am concerned that because there are so many changes happening right now that the chart will be fragile (demonstrated by this PR) so for now I'm ok requiring purposeful version upgrades.

TylerHelmuth:
Depends on #388

TylerHelmuth:
@wph95 please re-review. Added a new env var that product catalog depends on.

wph95 (approved):

dmitryax (approved):

PR #337, opened 2022-08-22
[opentelemetry-demo] Fix how DATABASE_URL gets set for Feature flag service

### LINKED ISSUE #335: [opentelemetry-demo] Issue with DATABASE_URL in Feature flag service
Database URL set in feature flag service is connecting to `ffs-postgres`, without taking into consideration the installation name.
https://github.com/open-telemetry/opentelemetry-helm-charts/blob/7421d9abecae37c5b9e04e31489a4d08188fb466/charts/opentelemetry-demo/values.yaml#L92-L93 
Say we run
```helm install mydemo ./opentelemetry-demo```

the postgres service would be `mydemo-ffs-postgres` 

we see error ```16:07:22.958 [error] Postgrex.Protocol (#PID<0.1854.0>) failed to connect: ** (DBConnection.ConnectionError) tcp connect (ffs-postgres:5432): non-existing domain - :nxdomain```

### PR DESCRIPTION
**Pull request by TylerHelmuth**
This PR updates the Feature flag service env vars to automatically set DATABASE_URL based on the ffs-postgres service name, which includes the release name.  If DATABASE_URL is set directly in the feature flag service's env vars it takes precedence.

Fixes #335

### DIFF
--- a/charts/opentelemetry-demo/templates/_pod.tpl
+++ b/charts/opentelemetry-demo/templates/_pod.tpl
@@ -54,7 +54,7 @@ Get Pod Env
 {{- end}}
 {{- end }}
 {{- if not $hasDatabaseUrl }}
-{{- $_ := set . "env" (append .env (dict "name" "DATABASE_URL" "value" (printf "ecto://ffs:ffs@%s-ffs-postgres:5432/ffs" $prefix)) | uniq) }}
+{{- $_ := set . "env" (append .env (dict "name" "DATABASE_URL" "value" (printf "ecto://ffs:ffs@%s-ffs-postgres:5432/ffs" $prefix))) }}
 {{- end}}
 {{- end }}
 
--- a/charts/opentelemetry-demo/templates/_pod.tpl
+++ b/charts/opentelemetry-demo/templates/_pod.tpl
@@ -46,6 +46,18 @@ Get Pod Env
 {{- define "otel-demo.pod.env" -}}
 {{- $prefix := include "otel-demo.name" $ }}
 
+{{- if eq .name "featureflag-service" }}
+{{- $hasDatabaseUrl := false }}
+{{- range .env }}
+{{- if eq .name "DATABASE_URL" }}
+{{- $hasDatabaseUrl = true }}
+{{- end}}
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by dineshg13 — APPROVED** [review]
LGTM . Thanks for fixing this .

**Review by wph95 — APPROVED** [review]
LGTM

File: charts/opentelemetry-demo/templates/_pod.tpl [hunk 1]
```diff
@@ -46,6 +46,18 @@ Get Pod Env
...
 
+{{- if eq .name "featureflag-service" }}
+{{- $hasDatabaseUrl := false }}
+{{- range .env }}
+{{- if eq .name "DATABASE_URL" }}
+{{- $hasDatabaseUrl = true }}
+{{- end}}
+{{- end }}
+{{- if not $hasDatabaseUrl }}
+{{- $_ := set . "env" (append .env (dict "name" "DATABASE_URL" "value" (printf "ecto://ffs:ffs@%s-ffs-postgres:5432/ffs" $prefix)) | uniq) }}
```
dmitryax, line 57: I don't think `uniq` will do the trick that you expect. If user provides a custom `DATABASE_URL`, only `name` key/value pair is the same, but the `value` pair will likely be different, so the items won't be identical for `uniq` to be applied. If you provide two `env` items with the same `name`, Kubernetes doesn't provide any guarantee which one is applied and behaves pretty wonky based on my experience.

TylerHelmuth: Ya we probably don't even need it because of the logic above. The logic above ensures that we only ever modify the list if DATABASE_URL was never present so I think we can safely remove the call to uniq. What do you think?

TylerHelmuth: Also, I think we should refactor something to make setting env var easier. Adding env vars for specific services that rely on other services (and therefore the Release name) doesn't feel graceful.

TylerHelmuth: @dmitryax done.

dmitryax (approved):

PR #338, opened 2022-08-22
bump helm/ct/kind

TylerHelmuth:
Bumps the dependencies in the `setup` action.


dmitryax:
Sad that the dependabot doesn't watch this directory

dmitryax (approved):

TylerHelmuth:
> Sad that the dependabot doesn't watch this directory

Ya I just realized that haha

Issue #343, opened 2022-08-25
[demo ] Feature flag service is failing to start


dineshg13:
When we run the latest demo chart application via command 

```
helm install mydemo ./opentelemetry-demo
```
we see Feature flag service is failing to start . 
```
k logs mydemo-featureflag-service-5d59b6cfb7-tl5mc                                                                                                                                                         ─╯
ERROR! Config provider Config.Reader failed with:
** (ArgumentError) errors were found at the given arguments:

  * 1st argument: not a binary

    :erlang.binary_to_integer(nil)
    (stdlib 4.0) erl_eval.erl:744: :erl_eval.do_apply/7
    (stdlib 4.0) erl_eval.erl:492: :erl_eval.expr/6
    (elixir 1.13.3) src/elixir.erl:296: :elixir.recur_eval/3
    (elixir 1.13.3) src/elixir.erl:274: :elixir.eval_forms/3
    (elixir 1.13.3) lib/code.ex:404: Code.validated_eval_string/3
    (elixir 1.13.3) lib/config.ex:260: Config.__eval__!/3
    (elixir 1.13.3) lib/config/reader.ex:92: Config.Reader.read!/2

{"init terminating in do_boot",{badarg,[{erlang,binary_to_integer,[nil],[{error_info,#{module=>erl_erts_errors}}]},{erl_eval,do_apply,7,[{file,"erl_eval.erl"},{line,744}]},{erl_eval,expr,6,[{file,"erl_eval.erl"},{line,492}]},{elixir,recur_eval,3,[{file,"src/elixir.erl"},{line,296}]},{elixir,eval_forms,3,[{file,"src/elixir.erl"},{line,274}]},{'Elixir.Code',validated_eval_string,3,[{file,"lib/code.ex"},{line,404}]},{'Elixir.Config','__eval__!',3,[{file,"lib/config.ex"},{line,260}]},{'Elixir.Config.Reader','read!',2,[{file,"lib/config/reader.ex"},{line,92}]}]}}
[...]

Issue #340, opened 2022-08-24
[opentelemetry-demo] Refactor component env vars to better handle service names


TylerHelmuth:
Also, I think we should refactor something to make setting env var easier. Adding env vars for specific services that rely on other services (and therefore the Release name) doesn't feel graceful.

I can open an issue for this later.

_Originally posted by @TylerHelmuth in https://github.com/open-telemetry/opentelemetry-helm-charts/pull/337#discussion_r953223476_

dineshg13:
Currently demo application isn't working.  See #358 . When i manually fixed that in my environment , am seeing that checkout service is unable to connect to email service 
```
{"message":"failed to send order confirmation to \"tobias@example.com\": failed POST to email service: Post \"demo-email-service:8080/send_order_confirmation\": unsupported protocol scheme \"demo-email-service\"","severity":"warning","timestamp":"2022-09-09T01:22:19.778637815Z"}
```

TylerHelmuth:
@2start would you like to take point on this issue?

2start:
#363 describes the issue in more detail.

PR #344, opened 2022-08-27
Rename environment variables for demo/featureflag-service

joshleecreates:
This fixes the issue with the featureflagservice pod not starting in https://github.com/open-telemetry/opentelemetry-helm-charts/issues/343

There is probably a better way to do this in the .tpl files, and the product-catalog-service appears to have the wrong port for the feature flag grpc.

### DIFF
--- a/charts/opentelemetry-demo/values.yaml
+++ b/charts/opentelemetry-demo/values.yaml
@@ -93,10 +93,12 @@ components:
     # If a different url is desired, set DATABASE_URL here.
     # - name: DATABASE_URL
     #   value:
-    - name: GRPC_PORT
-      value: '50053'
-    - name: PORT
+    - name: FEATURE_FLAG_GRPC_SERVICE_PORT
       value: '50052'
+    - name: FEATURE_FLAG_SERVICE_PORT
+      value: '50053'
+    - name: SECRET_KEY_BASE
+      value: 'Vive++6s/oRCd7xI3zSvyUGH1AJQS0BE3xjNrEOJEVmzOg5qPBlTYE+5kgz2cfBE'
     ports:
     - name: grpc
       value: 50052


TylerHelmuth:
@joshleecreates Can you fix the product catalog as well if it is broken?

TylerHelmuth:
Also please bump the patch version in the chart.yaml

joshleecreates:
@TylerHelmuth I'm not sure if using port 8080 was the right move here but it seemed reasonable.

charts/opentelemetry-demo/values.yaml
```diff
@@ -93,13 +93,15 @@ components:
...
     #   value:
-    - name: GRPC_PORT
+    - name: FEATURE_FLAG_GRPC_SERVICE_PORT
+      value: '8080'
+    - name: FEATURE_FLAG_SERVICE_PORT
       value: '50053'
-    - name: PORT
-      value: '50052'
+    - name: SECRET_KEY_BASE
+      value: 'Vive++6s/oRCd7xI3zSvyUGH1AJQS0BE3xjNrEOJEVmzOg5qPBlTYE+5kgz2cfBE'
```
TylerHelmuth, line 101: Is this required?

charts/opentelemetry-demo/values.yaml
```diff
@@ -93,13 +93,15 @@ components:
     # If a different url is desired, set DATABASE_URL here.
     # - name: DATABASE_URL
     #   value:
-    - name: GRPC_PORT
+    - name: FEATURE_FLAG_GRPC_SERVICE_PORT
```
TylerHelmuth, line 96: Based on the demo's env var file the GRPC port should be 50031 and the http port should be 8081.

https://github.com/open-telemetry/opentelemetry-demo/blob/c5831e8e632259260babdaf7c15423eb2a56d619/.env#L46-L49

charts/opentelemetry-demo/values.yaml
joshleecreates: It can probably be removed as a requirement from the underlying app.

TylerHelmuth: Secrets aren't a good thing to include in a chart by default typically.  If a secret needs set it is better to use helm's `--set` flag when running `helm install` to set a secret. Can we remove it from this PR?

charts/opentelemetry-demo/values.yaml
```diff
@@ -93,15 +93,15 @@ components:
...
+      value: '8080'
+    - name: FEATURE_FLAG_SERVICE_PORT
       value: '50053'
```
TylerHelmuth, line 99: Do these values need to match the ports exposed below?

charts/opentelemetry-demo/values.yaml
TylerHelmuth: [It looks like there is a value checked into source control](https://github.com/open-telemetry/opentelemetry-demo/blob/3ccd221fafefdb3c8b3fdcc302ad03ed5dca1f6f/src/featureflagservice/Dockerfile#L88) in the demo so maybe this env var is ok, but I'd rather see the values match.  Can you reach out in the Demo Slack channel and confirm this value should be committed and that its safe to include in the helm charts.

charts/opentelemetry-demo/values.yaml
joshleecreates: Doh...

joshleecreates: Should be all set.

charts/opentelemetry-demo/values.yaml
joshleecreates: Will do.

TylerHelmuth: @joshleecreates Looks like the value isn't needed in the helm charts env var bc it is built into the docker container the chart relies on. I think we can safely remove this.

joshleecreates: Ah makes sense, I was testing directly without Docker. This is not present on the current version of the PR, should be gtg

TylerHelmuth (approved):
Thanks for taking care of this

Allex1 (approved):

charts/opentelemetry-demo/templates/_pod.tpl
```diff
@@ -91,7 +91,7 @@ Get Pod Env
 
 {{- if eq .name "product-catalog-service" }}
 - name: FEATURE_FLAG_GRPC_SERVICE_ADDR
-  value: {{ (printf "%s-featureflag-service:%0.f" $prefix .servicePort ) }}
+  value: {{ (printf "%s-featureflag-service:50031" $prefix ) }}
```
TylerHelmuth, line 94: Logging this hard coded value as a temporary fix.  Will address as part of https://github.com/open-telemetry/opentelemetry-helm-charts/issues/340

PR #345, opened 2022-08-29
[opentelemetry-demo] Add demo example

TylerHelmuth:


TylerHelmuth:
the component file generation is not consistent.  Every time I `make generate-examples` things are rendered in a different order.  Need to think about how to handle this.

PR #347, opened 2022-08-29
[opentelemetry-demo] Refactor default env vars

TylerHelmuth:
This PR refactors the usage of default env vars to be simpler and more deterministic, which is important for https://github.com/open-telemetry/opentelemetry-helm-charts/pull/345.  

Taking a page from the presets work done on the collector chart, I think it is ok if we don't try to merge everything perfect all the time.  Instead of merging the default env vars with each component env var list, this PR changes the chart to allow each component to enable or disable the use of env vars.  

When enabled, the component will use all the values supplied to `.default.env`. If a user want to modify the default env vars for a specific component, they should disable the use of the default env vars for that component and use the component's env var section to set what is needed (like how we did presets).

This is the first of several PRs working towards https://github.com/open-telemetry/opentelemetry-helm-charts/issues/340


Allex1 (approved):

TylerHelmuth:
@dmitryax please review

charts/opentelemetry-demo/values.yaml
```diff
@@ -39,19 +38,28 @@ serviceAccount: ""
 components:
   redis:
     enabled: true
+    default:
+      env:
+        enabled: true
```
dmitryax, line 43: Am I understanding it right that, if I want to have other set of envs for a particular component, I need to:
```yaml
components:
  redis:
    default:
      env:
        enabled: false
      values:
        - name: CUSTOM_ENV
          value: custom_val 
```

I don't think this is an intuitive interface for users... 

Maybe something like?

```yaml
components:
  redis:
    useDefault:
      env: false
    env:
      - name: CUSTOM_ENV
        value: custom_val 
```

TylerHelmuth: @dmitryax the latter is what is proposed in this PR. I stuck with our standard "env.enabled" syntax but I am ok combining "enabled" and "env" into one.

dmitryax: > @dmitryax the latter is what is proposed in this PR. 

Not sure I understand what interface this PR provides then. Can you please post a snippet?

What I understood (the first section) just doesn't seem right. For example why `components.default.values` is used to override default `env`?

TylerHelmuth: I'll provide more clarification when I'm back from PTO on the 13th.

TylerHelmuth: @dmitryax done.

charts/opentelemetry-demo/templates/_pod.tpl
```diff
@@ -68,24 +68,11 @@ Note: Consider that dependent variables need to be declared before the reference
 {{- end}}
 {{- end }}
 
-{{- if .default.enabled  }}
+{{- if .useDefault.env  }}
+{{ toYaml .defaultValues.env }}
```
TylerHelmuth, line 72: `defaultValues` is set in `component.yaml` as a way to pass the default values into the pod template.  Due to our values schema, users cannot set it.

dmitryax:
@TylerHelmuth please rebase if this is still relevant

TylerHelmuth:
@dmitryax still relevant. I am going to do more work after this to address individual service env vars

dmitryax (approved):

Issue #351, opened 2022-09-02
[opentelemetry-operator]  use self-signed cert not works well.


JaredTan95:
As described from https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/README.md#tls-certificate-requirement 3rd way.

We deploy as following sets:
```bash
helm install \
  my-opentelemetry-operator open-telemetry/opentelemetry-operator \
--set admissionWebhooks.certManager.enabled=false
```

and we also create tls by `kubectl create secret tls...`

but the `opentelemetry-operator` pod occurs error :
`TLS handshake error from 10.233.96.0:39014: remote error: tls: bad certificate `

It can be fix by bundle `ca cert` for `MutatingWebhookConfiguration ` and `MutatingWebhookConfiguration ` like :
```yaml
{{- if  and (.Values.admissionWebhooks.enabled) (not .Values.admissionWebhooks.certManager.enabled) }} 👈
apiVersion: admissionregistration.k8s.io/v1
kind: MutatingWebhookConfiguration
metadata:
  labels:
    app.kubernetes.io/name: opentelemetry-operator
  name: {{ template "opentelemetry-operator.name" . }}-mutating-webhook-configuration
webhooks:
  - admissionReviewVersions:
      - v1
    clientConfig:
      caBundle:  "xxxxxxxx" 👈
      service:
        name: {{ template "opentelemetry-operator.name" . }}-webhook-service
        namespace: {{ .Release.Namespace }}
        path: /mutate-opentelemetry-io-v1alpha1-instrumentation
```

PR #355, opened 2022-09-07
[operator] fix self-signed cert not works well.

JaredTan95:
fix https://github.com/open-telemetry/opentelemetry-helm-charts/issues/351

Considering that the current self-signed does not take effect, and considering the ease of use of users, the self-signed certificate is built directly into the chart.


JaredTan95:
https://github.com/open-telemetry/opentelemetry-helm-charts/runs/8220627772?check_suite_focus=true#step:4:36

I did't change `collector` cahrt,  should I bump up collector chart version?

Allex1 (approved):
@JaredTan95 this is great work. Thanks

JaredTan95:
@TylerHelmuth @dmitryax hi, do you have time to help review this, I am eagerly waiting for this PR to be completed as soon as possible :-)

TylerHelmuth:
@JaredTan95 I will review this PR tomorrow when I am back from PTO.

charts/opentelemetry-operator/README.md
```diff
@@ -18,34 +18,7 @@ certificate that the API server is configured to trust. There are three ways for
...
-        namespace: opentelemetry-operator-system
-      type: kubernetes.io/tls
-      data:
-        tls.crt: |
-            # your signed cert
-        tls.key: |
-            # your private key
-      EOF
-      ```
+  - The last way is to use self-signed certificate by configuring the `admissionWebhooks.certManager.enabled` to `false`.
```
TylerHelmuth, line 21: @Allex1 @JaredTan95 it feels kinda strange for the chart to create a new object when a existing setting is set to false. Is there a way to implement this change so that the secret can be created based on a new entry in the values.yaml?

charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook.yaml
```diff
@@ -0,0 +1,182 @@
+{{- if and (.Values.admissionWebhooks.enabled) (not .Values.admissionWebhooks.certManager.enabled) }}
+{{- $altNames := list ( printf "%s.%s" "opentelemetry-operator-webhook-service" .Release.Namespace ) ( printf "%s.%s.svc" "opentelemetry-operator-webhook-service" .Release.Namespace ) -}}
+{{- $ca := genCA "opentelemetry-operator-operator-ca" 365 -}}
+{{- $cert := genSignedCert (include "opentelemetry-operator.fullname" .) nil $altNames 365 $ca -}}
+apiVersion: v1
+kind: Secret
```
TylerHelmuth, line 6: Is there a way to break this out into its own thing so the objects below don't have to be duplicated?

TylerHelmuth (commented):
@JaredTan95 were you able to test this change and confirm it works as expected?

charts/opentelemetry-operator/README.md
Allex1: @TylerHelmuth this can probably be achieved but to me it makes sense the way it is defined in the template : admissionWebhooks -> enabled and certManager -> disabled.
Maybe we can make it more clear in the docs.

Allex1:
> @JaredTan95 were you able to test this change and confirm it works as expected?

I actually tested in a kind cluster and it worked as advertised. 👍

charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook.yaml
JaredTan95: make sense, I will simplify this

JaredTan95: I tried like this:
```yaml
{{- if .Values.admissionWebhooks.enabled }}
{{- if not .Values.admissionWebhooks.certManager.enabled }}
{{- $altNames := list ( printf "%s.%s" "opentelemetry-operator-webhook-service" .Release.Namespace ) ( printf "%s.%s.svc" "opentelemetry-operator-webhook-service" .Release.Namespace ) -}}
{{- $ca := genCA "opentelemetry-operator-operator-ca" 365 -}}
{{- $cert := genSignedCert (include "opentelemetry-operator.fullname" .) nil $altNames 365 $ca -}}
apiVersion: v1
kind: Secret
type: kubernetes.io/tls
metadata:
  annotations:
    "helm.sh/hook": "pre-install,pre-upgrade"
    "helm.sh/hook-delete-policy": "before-hook-creation"
  labels:
    {{- include "opentelemetry-operator.labels" . | nindent 4 }}
  name: opentelemetry-operator-controller-manager-service-cert
data:
  tls.crt: {{ $cert.Cert | b64enc }}
  tls.key: {{ $cert.Key | b64enc }}
---
{{- end }}
apiVersion: admissionregistration.k8s.io/v1
kind: MutatingWebhookConfiguration
metadata:
{{- if .Values.admissionWebhooks.certManager.enabled }}
  annotations:
    cert-manager.io/inject-ca-from: {{ printf "%s/%s-serving-cert" .Release.Namespace (include "opentelemetry-operator.name" .) }}
{{- end }}
  labels:
  name: {{ template "opentelemetry-operator.name" . }}-mutating-webhook-configuration
webhooks:
  - admissionReviewVersions:
      - v1
    clientConfig:
{{- if not .Values.admissionWebhooks.certManager.enabled }}
      caBundle: {{ $ca.Cert | b64enc }}       👈 # parse error at (opentelemetry-operator/templates/admission-webhooks/operator-webhook.yaml:37): undefined variable "$ca"
{{- end }}
```

But It failed to parse variable "$ca":
 https://github.com/open-telemetry/opentelemetry-helm-charts/actions/runs/3060941363/jobs/4940081825#step:4:27

And I followed `tailingsidecar` project: 
- https://github.com/SumoLogic/tailing-sidecar/blob/main/helm/tailing-sidecar-operator/templates/_operator-webhook-with-cert-manager.tpl
- https://github.com/SumoLogic/tailing-sidecar/blob/main/helm/tailing-sidecar-operator/templates/_operator-webhook.tpl

they used seprated file to impl this 🤔

charts/opentelemetry-operator/README.md
Allex1: @TylerHelmuth a secret was manually created before as well with the exact same name `opentelemetry-operator-controller-manager-service-cert` , so an upgrade would just update that secret. We could explain in values.yaml via comments what the `admissionWebhooks.certManager.enabled: false` does in order to avoid confusion

As a last resort we could add another flag that the template would validate like `admissionWebhooks.selfSignedCert.create: True` to be mutually exclusive to `admissionWebhooks.certManager.enabled: False` and update the docs if you really think it's useful .

TylerHelmuth: If `admissionWebhooks.certManager.enabled: False` always implies `admissionWebhooks.selfSignedCert.create: True` then `admissionWebhooks.selfSignedCert.create` probably isn't useful, but I do think we should make it clear via the docs and values.yaml that `admissionWebhooks.certManager.enabled: False` means the secret will be created for you.

charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook.yaml
TylerHelmuth: Thanks for sharing that example.  I see now why it is done this way,

charts/opentelemetry-operator/README.md
TylerHelmuth: @JaredTan95 can you update the documentation? Once that's good we'll be good to go.

JaredTan95: okay

charts/opentelemetry-operator/README.md
```diff
@@ -18,34 +18,7 @@ certificate that the API server is configured to trust. There are three ways for
...
-        namespace: opentelemetry-operator-system
-      type: kubernetes.io/tls
-      data:
-        tls.crt: |
-            # your signed cert
-        tls.key: |
-            # your private key
-      EOF
-      ```
+  - The last way is to use self-signed certificate by configuring the `admissionWebhooks.certManager.enabled` to `false`. helm will create self-signd cert and secret for you.
```
TylerHelmuth, line 21:
```suggestion
  - The last way is to use self-signed certificate by configuring the `admissionWebhooks.certManager.enabled` to `false`. Helm will create self-signd cert and secret for you.
```

charts/opentelemetry-operator/values.yaml
```diff
@@ -94,6 +93,7 @@ admissionWebhooks:
   ## Provide the issuer kind and name to do the cert auth job.
   ## By default, OpenTelemetry Operator will use self-signer issuer.
   certManager:
+    # flase means helm will create self-signd cert and secret for you.
```
TylerHelmuth, line 96:
```suggestion
    # false means helm will create self-signd cert and secret for you.
```

TylerHelmuth (commented):
Couple nits

JaredTan95:
> Couple nits

😄 thx

TylerHelmuth (approved):

Issue #358, opened 2022-09-09
[opentelemetry-demo] Load generator is unable to connect to frontend service


dineshg13:
When we install demo application on k8s cluster, loadgenerator is unable to connect to frontend. We see error trace in the jaeger as well . 

![image](https://user-images.githubusercontent.com/2471669/189366284-6cef5cef-5110-4eb8-ba49-5dfc23cb1c93.png)

TylerHelmuth:
@dineshg13 do you plan to work on this?

PR #361, opened 2022-09-13
Fix resolution of dependent environment variable.

2start:
The PR fixes the connection between the loadgenerator and the frontend as described in #358. There are 2 issues at the same time:

1. The below snippet generates environment variables for the pod containing the address of another service that it depends on. 

```
# {{ $.depends }}
# {{ .name }}
{{- if hasKey $.depends .name }}
{{- range $depend := get $.depends .name }}
- name: {{ printf "%s_ADDR" $depend | snakecase | upper }}
  value: {{ printf "%s-%s:%0.f" $prefix ($depend | kebabcase) (get $.serviceMapping $depend )}}
{{- end }}
{{- end }}
```

E.g. the loadgenerator depends on the frontend and therefore `FRONTEND_ADDR: my-release-frontend:8080` is created. To reference `FRONTEND_ADDR` it must be created before the dependent environment variable. The order matters. (Source: [Docs](https://kubernetes.io/docs/tasks/inject-data-application/define-interdependent-environment-variables/#define-an-environment-dependent-variable-for-a-container))

I moved the above block to the top of the env variable definitions because the variables will get referenced but never reference other variables themselves.

2. A dependent variable is declared using parentheses instead of curly brackets. I.e. we need to reference `FRONTEND_ADDR` as follows: `http://$(FRONTEND_ADDR)`

After testing the changes in this PR, the error rate in the APM tool dropped from 100% to 0% and the whole demo seems to be working fine. The requests seem to propagate properly through all the services.
image

### DIFF
--- a/charts/opentelemetry-demo/Chart.yaml
+++ b/charts/opentelemetry-demo/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: opentelemetry-demo
-version: 0.2.4
+version: 0.2.5
 description: opentelemetry demo helm chart
 home: https://opentelemetry.io/
 sources:
--- a/charts/opentelemetry-demo/templates/_pod.tpl
+++ b/charts/opentelemetry-demo/templates/_pod.tpl
@@ -42,9 +42,19 @@ Get Services Port Mapping
 
 {{/*
 Get Pod Env
+Note: Consider that dependent variables need to be declared before the referenced env varibale.
 */}}
 {{- define "otel-demo.pod.env" -}}
 {{- $prefix := include "otel-demo.name" $ }}
+
+# {{ $.depends }}
+# {{ .name }}
+{{- if hasKey $.depends .name }}
+{{- range $depend := get $.depends .name }}
+- name: {{ printf "%s_ADDR" $depend | snakecase | upper }}
+  value: {{ printf "%s-%s:%0.f" $prefix ($depend | kebabcase) (get $.serviceMapping $depend )}}
+{{- end }}
+{{- end }}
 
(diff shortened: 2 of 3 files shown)


TylerHelmuth:
@2start Please bump the chart's patch version.

wph95 (approved):
just need fix 
- CLA check
- Helm chart version check

2start:
Should be both fixed! Can you try to rerun the workflows? I already signed the CLA yesterday after the bot commented yesterday. Now I signed it again. 

The tool is Datadog's APM tool btw.

dineshg13 (approved):
thanks @2start for fixing same. I wasn't aware that in k8s order matters

TylerHelmuth:
@2start thanks for the contribution.  To pass the easycla check you need to make sure that the latest commit in your branch is associated with your UserId: https://help.github.com/en/github/committing-changes-to-your-project/why-are-my-commits-linked-to-the-wrong-user

TylerHelmuth (approved):
@2start thanks for fixing this bug.  We are working on refactoring how this chart handles env vars (#340) as they seem to keep causing defects, but I like fixing this issue now with this simple change to ensure the chart is usable.

2start:
Thanks for the tip. I had to change my email to the one that I used signing the CLA. However, just changing the latest commit was not sufficient. Now that every commit belongs to the correct email it works. Happy to contribute. 

Regarding the new env setup. I agree. There is also another similar bug where the checkout service can't contact the email service which is a little harder to fix with the current setup.

TylerHelmuth:
@2start can you open an issue for that bug if you know the details?

PR #366, opened 2022-09-15
[opentelemetry-collector] Remove deprecated containerLogs option

dmitryax:
Remove deprecated `containerLogs` config option from the default values.yaml. Otherwise it always throws a deprecation warning even if user uses `presets.logsCollection` instead


TylerHelmuth (approved):
Needs a version bump

PR #367, opened 2022-09-15
Simplify dependency propagation

2start:
1. I removed the below comments because they are just confusing in the final template.

```
# {{ $.depends }}
# {{ .name }}
```

2. I also changed that each service only receives its own dependencies because it confused me in the first place that the complete dependency map was passed down.

### DIFF
--- a/charts/opentelemetry-demo/Chart.yaml
+++ b/charts/opentelemetry-demo/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: opentelemetry-demo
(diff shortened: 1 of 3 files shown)


TylerHelmuth (commented):
@2start were you able to test this change and confirm it works as expected?

TylerHelmuth:
works towards #340

2start:
Yes, it runs in my cluster and seems to work properly!

TylerHelmuth (approved):

Issue #374, opened 2022-09-19
Kubernetes metadata using `logsCollection` preset should be added to resource attributes


povilasv:
I believe Kubernetes metadata using `logsCollection` preset should be added to resource attributes not log attributes. As it is metadata about resource, not the individual log. Also 

stanza seems to support it via `resource` field https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/stanza/docs/types/entry.md#structure

Additionally, The k8s metadata processor only works with Resource attributes https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/k8sattributesprocessor/processor.go#L98-L103, which can add more kubernetes metadata

PR #375, opened 2022-09-19
fix: logsCollection use resouce attributes

povilasv:
Ref https://github.com/open-telemetry/opentelemetry-helm-charts/issues/374

I believe Kubernetes metadata using logsCollection preset should be added to resource attributes not log attributes. As it is metadata about resource, not the individual log.

Additionally, The k8s metadata processor only works with Resource attributes https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/k8sattributesprocessor/processor.go#L98-L103, which can add more kubernetes metadata

stanza supports it via resource field https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/stanza/docs/types/entry.md#structure

Tested this locally

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.31.3
+version: 0.31.4
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-collector/templates/NOTES.txt
+++ b/charts/opentelemetry-collector/templates/NOTES.txt
@@ -19,5 +19,9 @@
 {{ end }}
 
 {{- if not (eq (toString .Values.containerLogs) "<nil>") }}
-[WARNING] 'containerLogs' is deprecated.  Use 'presets.logsCollection' instead. See https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/UPGRADING.md#0231-to-0240 for instructions on how to migrate."
+[WARNING] 'containerLogs' is deprecated.  Use 'presets.logsCollection' instead. See https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/UPGRADING.md#0231-to-0240 for instructions on how to migrate.
+{{ end }}
+
+{{- if not (eq (toString .Values.presets.logsCollection) "<nil>") }}
+[WARNING] 'logsCollection' "k8s.container.name", "k8s.namespace.name", "k8s.pod.name", "k8s.container.restart_count", "k8s.pod.uid" log-level attributes are promoted to resource-level attributes. Next release will remove the log-level attributes. If you need log-level attributes, please configure log collection manually.
 {{ end }}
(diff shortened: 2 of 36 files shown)


charts/opentelemetry-collector/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.31.1
+version: 0.32.1
```
TylerHelmuth, line 3:
```suggestion
version: 0.32.0
```

djaglowski (approved):

djaglowski:
I agree with your assessment here - the `k8s.*` attributes should be resource attributes.

TylerHelmuth (commented):
Thanks for the contribution.  Please run `make generate-examples`.

TylerHelmuth:
@povilasv switching to `resource` is better, but a breaking change for existing users.  It seems relatively harmless, but could have big implications for users who are relying on those attributes as they'll disappear.

We might need to do this change in stages by adding an opt-in flag in the preset's configuration.  When opting in the attributes will be `resource`.  By default they'll stay as attributes on the log.  Then we can deprecate the log attributes and switch to resource attributes only in a future release.

That's a lot of overhead, though.  @dmitryax @Allex1 what do you think?

@povilasv in the meantime you should still be able to configure a filelog receiver and all the appropriate mounts manually if the logsCollection preset isn't meeting your needs.

TylerHelmuth:
> I was thinking on how do we fix it long term, and what do you think if we use copy transformation to set both resource attributes and log attributes?

Ya I like the simplicity of this.

TylerHelmuth (commented):
In NOTES.txt can you add a message that displays when logsCollection preset is enabled that lets users know that these fields on log attributes will be removed in a future release?

charts/opentelemetry-collector/templates/NOTES.txt
```diff
@@ -19,5 +19,9 @@
 {{ end }}
 
 {{- if not (eq (toString .Values.containerLogs) "<nil>") }}
-[WARNING] 'containerLogs' is deprecated.  Use 'presets.logsCollection' instead. See https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/UPGRADING.md#0231-to-0240 for instructions on how to migrate."
+[WARNING] 'containerLogs' is deprecated.  Use 'presets.logsCollection' instead. See https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/UPGRADING.md#0231-to-0240 for instructions on how to migrate.
+{{ end }}
+
+{{- if not (eq (toString .Values.presets.logsCollection) "<nil>") }}
+[WARNING] 'logsCollection' "k8s.container.name", "k8s.namespace.name", "k8s.pod.name", "k8s.container.restart_count", "k8s.pod.uid" log-level attributes are promoted to resource-level attributes. Next release will remove the log-level attributes. If you need log-level attributes, please configure log collection manually.
```
TylerHelmuth, line 26: We don't have a standard release cadence for the charts, so `Next Release` doesn't mean much.  In this situations we normally go by the Collector Release cycle.  In this case I think by 0.62.0 makes sense.

```suggestion
[WARNING] 'logsCollection' "k8s.container.name", "k8s.namespace.name", "k8s.pod.name", "k8s.container.restart_count", "k8s.pod.uid" log-level attributes are promoted to resource-level attributes. Once Collector version 0.62.0 is released we will remove the log-level attributes. If you need log-level attributes, please configure log collection manually.
```

TylerHelmuth (approved):

PR #378, opened 2022-09-20
fix: Use port number instead of name in targetPort

labaneilers:
When creating ports in a Service for a Deployment, the chart currently uses the  port name (instead of port number) for the targetPort property value. This works  in Kubernetes with Services and Ingresses, because it knows how to map port  names to numbers.

However, third-party controllers, including in my case, the AWS Load Balancer  Controller: https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/, may not properly map port names to numbers when translating service ports into  load balancer listener configurations. This means that this chart cannot create  a functional Ingress using an AWS Application Load Balancer.

This fix is to simply use the port number for the service port. This shouldn't  affect any existing Services created by the chart, because the port number  works in Kubernetes equally as well as the name. It also shouldn't affect any  externally created Ingresses that refer to the port, since the port name  remains available and unchanged.

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.31.4
+version: 0.32.0
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-collector/examples/daemonset-and-deployment/rendered/configmap-agent.yaml
+++ b/charts/opentelemetry-collector/examples/daemonset-and-deployment/rendered/configmap-agent.yaml
@@ -5,7 +5,7 @@ kind: ConfigMap
 metadata:
   name: example-opentelemetry-collector-agent
   labels:
-    helm.sh/chart: opentelemetry-collector-0.31.4
+    helm.sh/chart: opentelemetry-collector-0.32.0
     app.kubernetes.io/name: opentelemetry-collector
     app.kubernetes.io/instance: example
     app.kubernetes.io/version: "0.60.0"
--- a/charts/opentelemetry-collector/examples/daemonset-and-deployment/rendered/configmap.yaml
+++ b/charts/opentelemetry-collector/examples/daemonset-and-deployment/rendered/configmap.yaml
@@ -5,7 +5,7 @@ kind: ConfigMap
 metadata:
   name: example-opentelemetry-collector
   labels:
-    helm.sh/chart: opentelemetry-collector-0.31.4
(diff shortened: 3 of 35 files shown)


TylerHelmuth (commented):
@labaneilers thanks for the contribution.  Please bump the chart's minor version number and run `make generate-examples`.  Have you tested this change, specifically as part of an upgrade flow?

simpli-laban:
> @labaneilers thanks for the contribution. Please bump the chart's minor version number and run `make generate-examples`. Have you tested this change, specifically as part of an upgrade flow?

@TylerHelmuth I just want to double check you mean to bump the minor version, not patch, correct? I'm not 100% certain what would be the difference in terms of a helm chart, but my understanding was that this is more of a bug fix, which wouldn't affect the helm chart's "API", or the behavior for anyone who isn't being affected by the bug. I thought that would mean you'd want a "patch" bump. Can you please confirm?

simpli-laban:
@TylerHelmuth As for testing the change as part of an upgrade flow: I've re-installed my collector Deployment via `helm upgrade`, pointing at my local copy of the charts repo, with my fix. The Ingress, which was previously creating a broken load balancer (with a listener pointing to port 1), immediately created a healthy load balancer. Is that what you mean?

Thanks for your attention and patience!

TylerHelmuth:
@simpli-laban @labaneilers If the change affects users we'll do a Minor bump (since we don't have a Major version yet). But if the change doesn't affect anyone we could do a patch bump. 

When you tested did you use the AWS Load Balancer Controller or a standard k8s cluster? I want to make sure this change does affect users using standard k8s controllers.

labaneilers:
> @simpli-laban @labaneilers If the change affects users we'll do a Minor bump (since we don't have a Major version yet). But if the change doesn't affect anyone we could do a patch bump.
> 
> When you tested did you use the AWS Load Balancer Controller or a standard k8s cluster? I want to make sure this change does affect users using standard k8s controllers.

@TylerHelmuth I'm not 100% sure this is what you're asking, but: The cluster I'm testing with has both aws-load-balancer-controller and nginx controllers installed. Ingresses created with the nginx controller worked both before and after my fix. Ingresses using aws-load-balancer-controller work after my fix.

labaneilers:
lol, sorry I switched to my work account by accident- I was in the wrong chrome profile :)

This is meant to be a contribution via my personal github account.

TylerHelmuth (approved):

TylerHelmuth:
@labaneilers please handle merge conflicts (its probably the chart version)

labaneilers:
> @labaneilers please handle merge conflicts (its probably the chart version)

Done. It was just the chart version and generated examples.

Allex1 (approved):

Issue #223, opened 2022-06-07
Disable metrics and logging pipelines is not working as expected in Helm chat 0.18.0


AnujAroshA:
As mentioned in the [basic top level configuration section in the document](https://artifacthub.io/packages/helm/opentelemetry-helm/opentelemetry-collector#basic-top-level-configuration), we have set _null_ values as below.

```
      pipelines:
        traces:
          receivers:
            - otlp
          processors: {}
          exporters:
            - jaeger
            - logging
        metrics: null
        logs: null
```
But when I see the ConfigMap it shows all the pipelines.

```
  pipelines:
    logs:
      exporters:
      - logging
      processors:
      - memory_limiter
      - batch
      receivers:
      - otlp
    metrics:
      exporters:
      - logging
      processors:
      - memory_limiter
      - batch
      receivers:
      - otlp
      - prometheus
    traces:
      exporters:
      - jaeger
      - logging
      processors: {}
      receivers:
      - otlp
```

What we need is to get rid of _logs_ and _metrics_ and keep only _traces_.

Following are details of my testing environment, even though some of those are not directly impact with the issue.

- macOS Monterey (Version 12.4)
- Docker Desktop (4.8.2)
- Kubectl (1.24 both Client and Server)
- Helm v3.8.0

TylerHelmuth:
Are you using "mode" or "agentCollector"/"standaloneCollector"?

TylerHelmuth:
I was able to run this command and successfully disable the logs and metrics pipeline

`helm template testing open-telemetry/opentelemetry-collector --values ./charts/opentelemetry-collector/examples/deployment-otlp-traces/values.yaml`

```yaml
# Source: opentelemetry-collector/templates/configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: testing-opentelemetry-collector
  labels:
    helm.sh/chart: opentelemetry-collector-0.16.2
    app.kubernetes.io/name: opentelemetry-collector
    app.kubernetes.io/instance: testing
    app.kubernetes.io/version: "0.50.0"
    app.kubernetes.io/managed-by: Helm
data:
  relay: |
    exporters:
      logging: {}
    extensions:
      health_check: {}
      memory_ballast: {}
    processors:
      batch: {}
      memory_limiter:
        check_interval: 5s
        limit_mib: 1638
        spike_limit_mib: 512
    receivers:
      otlp:
        protocols:
          grpc:
[...]

PR #379, opened 2022-09-22
Support additional config stanzas for collector

austinlparker:
This PR adds functionality to add arbitrary collector configuration to the deployment, similar to how the config-extras config functions in the demo project today. If values.observability.additionalconfig.enabled is `true`, the values will be added and merged with the existing config via conventional collector config merging rules.

### DIFF
--- a/charts/opentelemetry-demo/templates/_objects.tpl
+++ b/charts/opentelemetry-demo/templates/_objects.tpl
@@ -96,3 +96,9 @@ service:
         - jaeger
         {{- end}}
 {{- end }}
+
+{{- define "otel-demo.otelcol.extras.config" -}}
+{{- if .Values.observability.additionalConfig.enabled }}
+{{ toYaml .Values.observability.additionalConfig.values }}
+{{- end }}
+{{- end }}
--- a/charts/opentelemetry-demo/templates/otelcol.yaml
+++ b/charts/opentelemetry-demo/templates/otelcol.yaml
@@ -1,5 +1,6 @@
 {{- $ := set . "name" "otelcol" }}
 {{- $otelconfig := include "otel-demo.otelcol.config" .}}
+{{- $otelconfigextras := include "otel-demo.otelcol.extras.config" .}}
 
 {{- if .Values.observability.otelcol.enabled -}}
 ---
@@ -13,6 +14,17 @@ data:
   otelcol-config.yml: |
 
     {{- $otelconfig | nindent 4 }}
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "otel-demo.name" . }}-{{ .name }}-extras-config
+  labels:
+    {{- include "otel-demo.labels" . | nindent 4 }}
+data:
+  otelcol-extras-config.yml: |
+
+    {{- $otelconfigextras | nindent 4 }}
 ---
 apiVersion: apps/v1
 kind: Deployment
@@ -34,7 +46,7 @@ spec:
       containers:
       - image: otel/opentelemetry-collector:0.55.0
         name: otel-col
-        args: ["--config=/etc/otelcol-config.yml"]
+        args: ["--config=/etc/otelcol-config.yml", "--config=/etc/otelcol-extras-config.yml"]
         {{- if .Values.observability.jaeger.enabled }}
         env:
           - name: JAEGER_ADDR
@@ -51,10 +63,16 @@ spec:
           - mountPath: /etc/otelcol-config.yml
             name: otel-config 
             subPath: otelcol-config.yml
+          - mountPath: /etc/otelcol-extras-config.yml
+            name: otel-extras-config
+            subPath: otelcol-extras-config.yml
       volumes:
         - configMap:
             name: {{ include "otel-demo.name" . }}-{{.name}}-config
-          name: otel-config 
+          name: otel-config
+        - configMap:
+            name: {{ include "otel-demo.name" . }}-{{.name}}-extras-config
+          name: otel-extras-config
 ---
 apiVersion: v1
 kind: Service
--- a/charts/opentelemetry-demo/values.schema.json
+++ b/charts/opentelemetry-demo/values.schema.json
@@ -282,11 +282,15 @@
                 },
                 "jaeger": {
                     "$ref": "#/definitions/ObservabilityProp"
+                },
+                "additionalConfig": {
+                  "$ref": "#/definitions/AdditionalConfigProp"
                 }
             },
             "required": [
                 "jaeger",
-                "otelcol"
+                "otelcol",
+                "additionalConfig"
             ],
             "title": "Observability"
         },
@@ -302,6 +306,19 @@
                 "enabled"
             ],
             "title": "ObservabilityProp"
+        },
+        "AdditionalConfigProp": {
+          "type": "object",
+          "additionalProperties": true,
+          "properties": {
+            "enabled": {
+              "type": "boolean"
+            }
+          },
+          "required": [
+            "enabled"
+          ],
+          "title": "AdditionalConfigProp"
         }
     }
 }
--- a/charts/opentelemetry-demo/values.yaml
+++ b/charts/opentelemetry-demo/values.yaml
@@ -3,6 +3,27 @@ observability:
     enabled: true
   jaeger:
     enabled: true
+  additionalConfig:
+    enabled: false
+    values:
+      exporters:
+        otlp/2: 
+            endpoint: "sample.endpoint"
+            headers: 
+              sample-header: "foo"
+      service:
+        pipelines:
+          traces/2:
+            receivers: [otlp]
+            processors: [batch]
+            exporters: [otlp/2]
+          metrics/2:
+            receivers: [otlp]
+            processors: [batch]
+            exporters: [otlp/2]
+
+
+
 
 # if the enabled is true, the items in this object will apply to all components
 default:


TylerHelmuth (commented):
Being able to modify the collector's config seems super useful, but I was imagining refactoring this chart's collector config to be closer to the [collector chart's pattern](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/b7150b1fb19f8e53d8cb5b235a4be3548d3affef/charts/opentelemetry-collector/values.yaml#L39).  Defining the entire config in the values.yaml allows users to modify whatever they want.  

I am making an assumption, though, that it is acceptable to modify whatever they want.  Supplying an `additionalConfig` section seems to imply it is not.  Based on the setup of the demo, should users be able to configure any part of the collector?

austinlparker:
Un-implemented in this PR, but somewhat implied, is the idea that a user may wish to replace the `otelcol` image with the `otelcontribcol` image. In that case, arbitrary modification of the config file feels more useful in order to be adaptable to the entire range of configuration available to contrib plugins.

austinlparker:
Ultimately, I think my goal here is for _distributions_ of the demo to be performed via a modification of the `values.yaml` rather than modifications of the raw chart; This is a bit of a band-aid solution for that.

TylerHelmuth:
>  In that case, arbitrary modification of the config file feels more useful in order to be adaptable to the entire range of configuration available to contrib plugins.

I'd agree, especially if swapping out collector images is on the table (which would also be in the values.yaml.  It's more work than what you've already done, but I'd rather do the collector config refactor now versus adding config options that would be removed in the future.

austinlparker:
> > In that case, arbitrary modification of the config file feels more useful in order to be adaptable to the entire range of configuration available to contrib plugins.
> 
> I'd agree, especially if swapping out collector images is on the table (which would also be in the values.yaml). It's more work than what you've already done, but I'd rather do the collector config refactor now versus adding config options that would be removed in the future.

Is there a way to do this more easily? Can we make the collector part of the demo chart use the existing collector chart? It seems like there's a decent amount of complexity in the collector chart and it strikes me that copying it over would be foolish -- if we could refactor the demo chart to use the collector chart itself, then that'd make more sense?

TylerHelmuth:
> if we could refactor the demo chart to use the collector chart itself, then that'd make more sense?

I think this is possible but I've not done it myself before.  Would need to research more about helm chart dependencies. Assuming it is possible, it could be a good solution, but I am worried about the complexity for user.  I'd rather this chart's implementation be more complex but the values.yaml remain simple; I wouldn't want users to have to manage 2 values.yamls.  

If we only care about the collector's yaml configuration, I think introducing that capability to this chart makes sense.  If the demo needs all the flexibility/functionality that the collector chart provides, then a dependency sounds like a great idea.

TylerHelmuth:
@dmitryax when you're back from vacation would love to hear your thoughts as well.

wph95:
My personal suggestion is that we divide the users into two types
1. otel new/normal users
2. otel master

For 1, he only cares about a parameter in enable/disable values.yaml
For 2, we open up the complete configuration. (For example, you can change the collector config in values.yaml)

Take the pr for example, the requirement is currently 2. We can do this by enhancing how simple and intuitive it is to modify the collect config in values.yaml.
If the feature is generic and we want the majority of users to use it, we can then encapsulate and simplify the feature by `observability.stanzas.enabled` to enable

TylerHelmuth:
Also, we've had issues in the past trying to remove default values when using a subchart (#223) so I'd definitely lean towards adding basic collector configuration directly to the demo chart.  Maybe something like:


```yaml
observability:
  otelcol:
    enabled: true
    image:
      repository: otel/opentelemetry-collector-contrib
      tag: 0.60.0
    config:
      receivers:
        otlp:
          protocols:
            grpc:
            http:
      exporters:
        logging:
      processors:
        batch:
      service:
        pipelines:
          traces:
            receivers:
              - otlp
            processors:
              - batch
            exporters:
              - logging
```

Would need to figure out how to fit jaeger in there based on its enabled/disabled status.

TylerHelmuth:
I propose https://github.com/open-telemetry/opentelemetry-helm-charts/pull/386 to solve this issue

dmitryax:
Can we just apply collector chart as sub chart? In that case user will get all ways to configure collector as if it's installed independently. It should be way more DRY :)

TylerHelmuth:
@dmitryax we absolutely could, but I have a couple concerns. 

1. Subcharts don't seem to be able to handle using null to remove default values (see #223). This would severely limit the configurability. 

2. There are a lot of options available in the collector chart, most that won't get used for the demo. Using a subchart means users would need to understand the collector's values.yaml as well, most of which they won't need. 

Of the two concerns the first is the top of my mind. I can create another PR tomorrow that uses the subchart method to reconfirm that fear and so that we can compare the 2 options.

dmitryax:
> Subcharts don't seem to be able to handle using null to remove default values (see https://github.com/open-telemetry/opentelemetry-helm-charts/issues/223). This would severely limit the configurability.

Do we really depend on this functionality with the presets?

> There are a lot of options available in the collector chart, most that won't get used for the demo. Using a subchart means users would need to understand the collector's values.yaml as well, most of which they won't need.

I don't think this is a big problem. If we provide reasonable default values, they don't need to understand it unless they need to customize anything.

I can imagine a few asks that can come soon, e.g. deploying collector as deamonset to collect logs from demo apps. This is pretty obvious one. What will we do in that case? Bring more copy/paste from the collector chart?

TylerHelmuth:
> Do we really depend on this functionality with the presets?

I believe the initial ask for this feature is for vendors to be able to supply their own config, so being able to turn off some of the collector charts default will be important. 

> I can imagine a few asks that can come soon, 

This is the question I posed at the beginning of this PR. The more  functionality required of the collector the more a subchart becomes necessary.

TylerHelmuth:
@dmitryax here is a draft implementation of subchart: https://github.com/open-telemetry/opentelemetry-helm-charts/pull/390

When I run `helm template testing ./charts/opentelemetry-demo` in main, the generate collector file looks like:

```yaml
---
# Source: opentelemetry-demo/templates/otelcol.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: testing-otelcol-config
  labels:
    helm.sh/chart: opentelemetry-demo-0.3.0
    app.kubernetes.io/name: testing
    app.kubernetes.io/instance: testing
    app.kubernetes.io/component: otelcol
    app.kubernetes.io/version: "0.4.0-alpha"
    app.kubernetes.io/managed-by: Helm
data:
  otelcol-config.yml: |
    receivers:
      otlp:
        protocols:
          grpc:
          http:
    exporters:
      jaeger:
        endpoint: "${JAEGER_ADDR}"
        tls:
          insecure: true
      logging:
    
    processors:
      batch:
    
    service:
      pipelines:
        traces:
          receivers:
            - otlp
          processors:
            - batch
          exporters:
            - logging
            - jaeger
---
# Source: opentelemetry-demo/templates/otelcol.yaml
apiVersion: v1
kind: Service
metadata:
  name: testing-otelcol
  labels:
    helm.sh/chart: opentelemetry-demo-0.3.0
    app.kubernetes.io/name: testing
    app.kubernetes.io/instance: testing
    app.kubernetes.io/component: otelcol
    app.kubernetes.io/version: "0.4.0-alpha"
    app.kubernetes.io/managed-by: Helm
spec:
  ports:
    - name: otlp
      port: 4317
      protocol: TCP
      targetPort: 4317
    - name: otlphttp
      port: 4318
      protocol: TCP
      targetPort: 4318
    - name: metric 
      port: 8888
      protocol: TCP
      targetPort: 8888
  selector:
    app.kubernetes.io/name: testing
    app.kubernetes.io/instance: testing
    app.kubernetes.io/component: otelcol
---
# Source: opentelemetry-demo/templates/otelcol.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: testing-otelcol
  labels:
    helm.sh/chart: opentelemetry-demo-0.3.0
    app.kubernetes.io/name: testing
    app.kubernetes.io/instance: testing
    app.kubernetes.io/component: otelcol
    app.kubernetes.io/version: "0.4.0-alpha"
    app.kubernetes.io/managed-by: Helm
spec:
  replicas: 1
  selector:
    matchLabels:
      app.kubernetes.io/name: testing
      app.kubernetes.io/instance: testing
      app.kubernetes.io/component: otelcol
  template:
    metadata:
      labels:
        checksum/otel_config: 4730a8fedeb3189a4835e3ea5c65ff15d2051838261f9da7deca79a7c3048b6
        app.kubernetes.io/name: testing
        app.kubernetes.io/instance: testing
        app.kubernetes.io/component: otelcol
    spec:
      containers:
      - image: otel/opentelemetry-collector:0.55.0
        name: otel-col
        args: ["--config=/etc/otelcol-config.yml"]
        env:
          - name: JAEGER_ADDR
            value: testing-jaeger:14250
        ports:
          - containerPort: 4317
            protocol: TCP
          - containerPort: 4318
            protocol: TCP
          - containerPort: 8888
            protocol: TCP
        volumeMounts:
          - mountPath: /etc/otelcol-config.yml
            name: otel-config 
            subPath: otelcol-config.yml
      volumes:
        - configMap:
            name: testing-otelcol-config
          name: otel-config
```

When I run the same command with the collector as a subchart the result is

```yaml
---
# Source: opentelemetry-demo/charts/opentelemetry-collector/templates/configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: testing-otelcol
  labels:
    helm.sh/chart: opentelemetry-collector-0.33.0
    app.kubernetes.io/name: otelcol
    app.kubernetes.io/instance: testing
    app.kubernetes.io/version: "0.60.0"
    app.kubernetes.io/managed-by: Helm
data:
  relay: |
    exporters:
      jaeger:
        endpoint: ${JAEGER_ADDR}
        tls:
          insecure: true
      logging: {}
    extensions:
      health_check: {}
      memory_ballast: {}
    processors:
      batch: {}
      memory_limiter:
        check_interval: 5s
        limit_mib: 409
        spike_limit_mib: 128
    receivers:
      jaeger:
        protocols:
          grpc:
            endpoint: 0.0.0.0:14250
          thrift_compact:
            endpoint: 0.0.0.0:6831
          thrift_http:
            endpoint: 0.0.0.0:14268
      otlp:
        protocols:
          grpc:
            endpoint: 0.0.0.0:4317
          http:
            endpoint: 0.0.0.0:4318
      prometheus:
        config:
          scrape_configs:
          - job_name: opentelemetry-collector
            scrape_interval: 10s
            static_configs:
            - targets:
              - ${MY_POD_IP}:8888
      zipkin:
        endpoint: 0.0.0.0:9411
    service:
      extensions:
      - health_check
      - memory_ballast
      pipelines:
        logs:
          exporters:
          - logging
          processors:
          - memory_limiter
          - batch
          receivers:
          - otlp
        metrics:
          exporters:
          - logging
          processors:
          - memory_limiter
          - batch
          receivers:
          - otlp
          - prometheus
        traces:
          exporters:
          - logging
          - jaeger
          processors:
          - memory_limiter
          - batch
          receivers:
          - otlp
      telemetry:
        metrics:
          address: 0.0.0.0:8888
---
# Source: opentelemetry-demo/charts/opentelemetry-collector/templates/deployment.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: testing-otelcol
  labels:
    helm.sh/chart: opentelemetry-collector-0.33.0
    app.kubernetes.io/name: otelcol
    app.kubernetes.io/instance: testing
    app.kubernetes.io/version: "0.60.0"
    app.kubernetes.io/managed-by: Helm
  
spec:
  replicas: 1
  selector:
    matchLabels:
      app.kubernetes.io/name: otelcol
      app.kubernetes.io/instance: testing
  strategy:
    type: RollingUpdate
  template:
    metadata:
      annotations:
        checksum/config: c286a03802d0242cd7a2cdae6429f4690c1905c0820468dd76350305e987325a
        
      labels:
        app.kubernetes.io/name: otelcol
        app.kubernetes.io/instance: testing
        
    spec:
      
      serviceAccountName: testing-otelcol
      securityContext:
        {}
      containers:
          command:
            - /otelcol
            - --config=/conf/relay.yaml
          securityContext:
            {}
          image: "otel/opentelemetry-collector:0.60.0"
          imagePullPolicy: IfNotPresent
          ports:
            - name: metrics
              containerPort: 8888
              protocol: TCP
            - name: otlp
              containerPort: 4317
              protocol: TCP
            - name: otlp-http
              containerPort: 4318
              protocol: TCP
          env:
            - name: MY_POD_IP
              valueFrom:
                fieldRef:
                  apiVersion: v1
                  fieldPath: status.podIP
          livenessProbe:
            httpGet:
              path: /
              port: 13133
          readinessProbe:
            httpGet:
              path: /
              port: 13133
          resources:
            limits:
              cpu: 256m
              memory: 512Mi
          volumeMounts:
            - mountPath: /conf
              name: opentelemetry-collector-configmap
      volumes:
        - name: opentelemetry-collector-configmap
          configMap:
            name: testing-otelcol
            items:
              - key: relay
                path: relay.yaml
---
# Source: opentelemetry-demo/charts/opentelemetry-collector/templates/service.yaml
apiVersion: v1
kind: Service
metadata:
  name: testing-otelcol
  labels:
    helm.sh/chart: opentelemetry-collector-0.33.0
    app.kubernetes.io/name: otelcol
    app.kubernetes.io/instance: testing
    app.kubernetes.io/version: "0.60.0"
    app.kubernetes.io/managed-by: Helm
spec:
  type: ClusterIP
  ports: 
    
    - name: metrics
      port: 8888
      targetPort: 8888
      protocol: TCP
    - name: otlp
      port: 4317
      targetPort: 4317
      protocol: TCP
    - name: otlp-http
      port: 4318
      targetPort: 4318
      protocol: TCP
  selector:
    app.kubernetes.io/name: otelcol
    app.kubernetes.io/instance: testing
---
# Source: opentelemetry-demo/charts/opentelemetry-collector/templates/serviceaccount.yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: testing-otelcol
  labels:
    helm.sh/chart: opentelemetry-collector-0.33.0
    app.kubernetes.io/name: otelcol
    app.kubernetes.io/instance: testing
    app.kubernetes.io/version: "0.60.0"
    app.kubernetes.io/managed-by: Helm

```

Which are very close, except for the configmap.  The configmap when using a subchart results in bot logs and metrics pipelines and extra components.  This is due to the fact that the null's included in the subchart branch's yaml are not honored (https://github.com/helm/helm/issues/9136).

The demo's values.yaml came out simple enough, but I do really dislike how any demo users are forced to use the collector chart's default configuration.  I haven't been able to find any workarounds like that we could use.  I think we'd have to make changes to the collector chart, going back to the concept have having zero default config and making every component a presets, which we have tried to avoid.

dmitryax:
> The demo's values.yaml came out simple enough, but I do really dislike how any demo users are forced to use the collector chart's default configuration.

Why do you see it as a problem? We should have a reasonable default collector's config that can be used with apps like the demo microservices

TylerHelmuth:
> Why do you see it as a problem?

I don't like the idea of including extraneous config in the demo, which in my mind should be as slim and precise as possible.  I don't want this extra stuff to trip up any users, especially since it wouldn't be in the docker-compose version of the demo.

But maybe I'm being over cautious and it isn't a big deal.

dmitryax:
> I don't like the idea of including extraneous config in the demo, which in my mind should be as slim and precise as possible. I don't want this extra stuff to trip up any users, especially since it wouldn't be in the docker-compose version of the demo.

Would you rather copy paste stuff from the collector chart for any other ask like this and keep them in sync afterwards? :)

dmitryax:
What is docker-compose? :D

TylerHelmuth:
> Would you rather copy paste stuff from the collector chart for any other ask like this

Guess its naive to think configuration will be the only thing the demo chart needs 😬 

I'll work more on the subchart solution tomorrow

austinlparker:
> What is docker-compose? :D

A miserable little pile of `SECRETS`

austinlparker:
Could you use the collector's config merging functionality? So all of the collector-specific stuff lives in a new file... I guess the difficulty would be in making the collector chart aware of config merging.

dmitryax:
> Could you use the collector's config merging functionality? So all of the collector-specific stuff lives in a new file... I guess the difficulty would be in making the collector chart aware of config merging.

config in the collector chart is set via values.yaml so you can actually merge anything in from your custom values.yaml in. Its essentially the same result as applying another config, but merging is done by helm, not collector itself

TylerHelmuth:
Closing since we have https://github.com/open-telemetry/opentelemetry-helm-charts/pull/390

Issue #360, opened 2022-09-13
Error message while deploying opentelemetry-operator with flux


adomenech73:
I'm trying to deploy `opentelemetry-operator` using flux and I always come to the point of receiving some error messages:

```
Helm install failed: rendered manifests contain a resource that already exists. 
Unable to continue with install: ClusterRole "opentelemetry-operator-manager-role" in namespace "" exists and cannot be imported into the current release: 
invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-namespace" must equal "observability": current value is "opentelemetry-operator-system"
``` 

I'm probably missing something but I tried to to add the annotation the message points and still getting the same message, then I disabled the clusterRole creation with `clusterRole.create: false` and then I got another error message:

```
Helm install failed: rendered manifests contain a resource that already exists. Unable to continue with install:
MutatingWebhookConfiguration "opentelemetry-operator-mutating-webhook-configuration" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: 
key "meta.helm.sh/release-namespace" must equal "observability": current value is "opentelemetry-operator-system"
```
I finally also disabled the `admissionWebhooks` and seems that finally I got the HelmRelease reconciled but my guess is that there is something wrong with this prune approach as this elements seem to be necessary and should be created and reinstalled anytime.  My impression is that they are created first time and then raise errors whenever the helm-controller tries to apply them again.

My helmRelease definitions now looks like this:

```yaml
apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
metadata:
  name: opentelemetry-operator
  namespace: observability
  annotations:
    "meta.helm.sh/release-namespace": "observability"
spec:
  releaseName: opentelemetry-operator
  chart:
    spec:
      chart: opentelemetry-operator
      sourceRef:
        kind: HelmRepository
        name: open-telemetry
        namespace: flux-system
      version: 0.11.9
  interval: 10m
  values:
    admissionWebhooks:
      enabled: false 
    clusterRole:
      create: false
    prometheusRule:
      enabled: true
      defaultRules:
        enabled: true
```
Any help would be much appreciated

Allex1:
@adomenech73 I'm not very familiar with how Flux handles helm releases but from your error messages it seems related to the fact that some of the resources outlined are missing the `namespace` metadata explicitly defined. I'll submit a pr to fix that and you can test it.

dmitryax:
[...]

PR #384, opened 2022-09-26
[opentelemetry-operator] - Explicitly define namespace metadata on all resources

Allex1:
I think addresses https://github.com/open-telemetry/opentelemetry-helm-charts/issues/360

### DIFF
--- a/charts/opentelemetry-operator/Chart.yaml
+++ b/charts/opentelemetry-operator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.13.0
+version: 0.13.1
 description: OpenTelemetry Operator Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook-with-cert-manager.yaml
+++ b/charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook-with-cert-manager.yaml
@@ -7,6 +7,7 @@ metadata:
   labels:
     app.kubernetes.io/name: opentelemetry-operator
   name: {{ template "opentelemetry-operator.name" . }}-mutating-webhook-configuration
+  namespace: {{ .Release.Namespace }}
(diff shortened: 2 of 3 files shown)


TylerHelmuth (approved):
LGTM

TylerHelmuth:
@adomenech73 can you test if this PR solves your issue?

charts/opentelemetry-operator/templates/clusterrole.yaml
```diff
@@ -6,6 +6,7 @@ metadata:
   labels:
     app.kubernetes.io/name: opentelemetry-operator
   name: {{ template "opentelemetry-operator.name" . }}-manager-role
+  namespace: {{ .Release.Namespace }}
```
dmitryax, line 9: `ClusterRoles` are not a namespaced resource so I believe this will fail

dmitryax (approved):

dmitryax (changes requested):
Didn't mean to approve actually

charts/opentelemetry-operator/templates/clusterrole.yaml
Allex1: @dmitryax great catch. Thanks. Also `ClusterRoleBinding` is not a namespaced resource.

dmitryax:
I don't think this will fix https://github.com/open-telemetry/opentelemetry-helm-charts/issues/360 . Let's discuss the issue first. helm adds namespace field automatically to all namespaced resources, so I don't think we need to do this

Allex1:
> I don't think this will fix #360 . Let's discuss the issue first. helm adds namespace field automatically to all namespaced resources, so I don't think we need to do this

Helm does this, I agree, but depends on how the CD tool is applying the manifests. If Flux does `helm template` and then applies the  resulted manifests via `kubectl apply -f` then it will fail because of not not explicitly having the namespace defined.

Allex1:
@dmitryax Actually both MutatingWebhookConfiguration and ValidatingWebhookConfiguration are not namespaced resources either, so closing this .

PR #386, opened 2022-09-27
[opentelemetry-demo] Add ability to configure the collector

TylerHelmuth:
This PR adds to the demo chart the ability to configure the collector and its image.  It uses similar techniques as the collector chart's preset to handle enabling and disabling of jaeger.


TylerHelmuth:
Looks like I formatted some stuff unexpectedly. I'll address it tomorrow.

dmitryax:
Let's wait until we con to conclusion in https://github.com/open-telemetry/opentelemetry-helm-charts/pull/379

I would like to avoid copy pasting stuff from the collectors chart

Issue #381, opened 2022-09-26
Add Kubernetes Events preset


povilasv:
Hey,

This https://github.com/open-telemetry/opentelemetry-helm-charts/issues/350#issue-1358927903 issue mentioned support for Kubernetes Event as a preset. So I'm creating a seperate issue. What are thoughts around this?

I'm thinking UX would be something like this:
```
presets:
  eventsCollection:
    enabled: false

```

and then configure k8s_event receiver and use the ClusterRole mentioned in https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/k8seventsreceiver#rbac

FYI I can work on this :)

TylerHelmuth:
@povilasv assigned.  Thanks!

nicolastakashi:
@povilasv and @TylerHelmuth now we have k8sobjectreceiver working as expected to collect Kubernetes events, shall we add this as a preset, to collect events?

PR #388, opened 2022-09-28
add Kubernetes events preset

povilasv:
First pass at implementing the preset, let me know what you think :)

Ref https://github.com/open-telemetry/opentelemetry-helm-charts/issues/381

Tested with:
```
 helm template --debug --set mode=deployment --set "presets.eventsCollection.enabled=true" .
 ```
 and actually installing the chart:
 
 ```
 helm install --set "presets.eventsCollection.enabled=true" --set "mode=deployment" --generate-name .
 ```

### DIFF
--- a/charts/opentelemetry-collector/README.md
+++ b/charts/opentelemetry-collector/README.md
@@ -150,7 +150,7 @@ To enable this feature, set the  `presets.eventsCollection.enabled` property to 
 Here is an example `values.yaml`:
 
 ```yaml
-mode: daemonset
+mode: deployment
 
 presets:
   eventsCollection:
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.35.0
+version: 0.35.1
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-collector/examples/daemonset-and-deployment/rendered/configmap-agent.yaml
+++ b/charts/opentelemetry-collector/examples/daemonset-and-deployment/rendered/configmap-agent.yaml
@@ -5,7 +5,7 @@ kind: ConfigMap
 metadata:
   name: example-opentelemetry-collector-agent
   labels:
-    helm.sh/chart: opentelemetry-collector-0.35.0
+    helm.sh/chart: opentelemetry-collector-0.35.1
     app.kubernetes.io/name: opentelemetry-collector
     app.kubernetes.io/instance: example
     app.kubernetes.io/version: "0.61.0"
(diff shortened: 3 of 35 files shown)

### CODE REVIEW

File: charts/opentelemetry-collector/README.md [hunk 1]
```diff
@@ -138,6 +138,25 @@ config:
           - otlphttp
 ```
 
+### Configuration for Kubernetes events
+
+The collector can be used to collect Kubernetes Events as logs.
+This feature is disabled by default. It has the following requirements:
+
+- It requires the [contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib) version
```
TylerHelmuth, line 146: Technically any distribution of the collector with the k8s_events receiver should work.

charts/opentelemetry-collector/README.md
```diff
@@ -138,6 +138,25 @@ config:
...
+of the collector image.
+
+To enable this feature, set the  `presets.eventsCollection.enabled` property to `true`.
+Here is an example `values.yaml`:
+
+```yaml
+mode: daemonset
```
TylerHelmuth, line 153:
```suggestion
mode: deployment
```

suggesting deployment based on the comments in values.yaml

charts/opentelemetry-collector/templates/clusterrole.yaml
```diff
@@ -9,5 +9,66 @@ metadata:
   annotations:
     {{- toYaml . | nindent 4 }}
   {{- end }}
-rules: {{ toYaml .Values.clusterRole.rules | nindent 2 -}}
+rules:
+  {{- if .Values.clusterRole.create -}}
+  {{ toYaml .Values.clusterRole.rules | nindent 2 -}}
+  {{- end }}
+  {{- if .Values.presets.eventsCollection.enabled}}
```
TylerHelmuth, line 16: It is possible that both `.Values.clusterRole.create` and `.Values.presets.eventsCollection.enabled` are `true`, which will result in both `.Values.clusterRole.rules` and these default k8s events rules being used.  that could result in some duplication.

I think `.Values.clusterRole.rules` should take precedence so I think we can use and `else if` here.

TylerHelmuth (commented):
Thanks for adding this preset

TylerHelmuth (dismissed):
@povilasv thanks for adding this preset.  @dmitryax @Allex1 please take a look.

dmitryax:
Given that k8s events receiver is in alpha and potentially can be replaced (see https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/14474#issuecomment-1268819633) I'd recommend holding off introducing this preset.

Enabling k8s attributes processor on tracing seems much more important to me. WDYT?

TylerHelmuth:
> Given that k8s events receiver is in alpha and potentially can be replaced (see [open-telemetry/opentelemetry-collector-contrib#14474 (comment)](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/14474#issuecomment-1268819633)) I'd recommend holding off introducing this preset.

Thanks for bringing this to my attention.  Based on the conversation I'd agree with holding off.


> Enabling k8s attributes processor on tracing seems much more important to me.

I too would love to see a preset for this.  Issue is being tracked via #389.

PR #390, opened 2022-09-28
[opentelemetry-demo] Collector as subchart

TylerHelmuth:
Adds opentelemetry-collector as a subchart to opentelemetry-demo.  This allows the demo to take advantage of all the collector chart's settings.


charts/opentelemetry-demo/values.yaml
```diff
@@ -165,18 +165,64 @@ components:
...
+    jaeger-compact:
+      enabled: false
+    jaeger-thrift:
+      enabled: false
+    jaeger-grpc:
+      enabled: false
+    zipkin:
+      enabled: false
+    metrics:
+      enabled: true
```
dmitryax, line 228: Why do we care about open ports if it's not a daemonset?

charts/opentelemetry-demo/values.yaml
```diff
@@ -165,18 +165,64 @@ components:
...
+      zipkin: null
+    exporters:
+      jaeger:
+        endpoint: "${JAEGER_ADDR}"
+        tls:
+          insecure: true
+    service:
+      pipelines:
+        logs: null
+        metrics: null
```
dmitryax, line 211: No harm to keep them right?

charts/opentelemetry-demo/values.yaml
```diff
@@ -165,18 +165,64 @@ components:
...
+        traces:
+          receivers:
+            - otlp
```
dmitryax, line 214: Same here, all traces receivers can stay I believe

charts/opentelemetry-demo/values.yaml
```diff
@@ -165,18 +165,64 @@ components:
...
+    receivers:
+      jaeger: null
+      otlp:
+        protocols:
+          grpc:
+            endpoint: 0.0.0.0:4317
+          http:
+            endpoint: 0.0.0.0:4318
+      prometheus: null
+      zipkin: null
```
dmitryax, line 202: This can stay as default as well

charts/opentelemetry-demo/values.yaml
```diff
@@ -70,82 +70,82 @@ components:
...
+        value: "8081"
     ports:
-    - name: grpc
-      value: 50053
-    - name: http
-      value: 8081
+      - name: grpc
+        value: 50053
+      - name: http
+        value: 8081
```
dmitryax, line 104: I like this formatting as well. Can you submit another PR for that?

charts/opentelemetry-demo/values.yaml
```diff
@@ -165,18 +165,34 @@ components:
...
+      jaeger:
+        endpoint: "${JAEGER_ADDR}"
+        tls:
+          insecure: true
+    service:
+      pipelines:
+        traces:
+          exporters:
+            - logging
+            - jaeger
```
dmitryax, line 198: This looks pretty good to me

charts/opentelemetry-demo/values.yaml
TylerHelmuth: Will do

charts/opentelemetry-demo/values.yaml
TylerHelmuth: Yes, this is simple enough for me as well.  Still don't love that the collector is doing more than the demo requires, but maybe the subchart is worth that complexity.

charts/opentelemetry-demo/values.yaml
TylerHelmuth: https://github.com/open-telemetry/opentelemetry-helm-charts/pull/392/files

TylerHelmuth:
I suspect the test is failing for a naming reason.  I will investigate tomorrow, I haven't actually tried to test this myself yet.

charts/opentelemetry-demo/templates/otelcol.yaml
```diff
@@ -1,81 +0,0 @@
...
-        {{- include "otel-demo.selectorLabels" . | nindent 8 }}
-    spec:
-      containers:
-      - image: otel/opentelemetry-collector:0.55.0
-        name: otel-col
-        args: ["--config=/etc/otelcol-config.yml"]
-        {{- if .Values.observability.jaeger.enabled }}
-        env:
-          - name: JAEGER_ADDR
-            value: {{ include "otel-demo.name" . }}-jaeger:14250
```
TylerHelmuth, line 41: @dmitryax we need to be able to tell the subchart what the endpoint will be fore the jaeger exporter.  The demo chart currently creates the the name of the jaeger deployment based on the release name, but we can't get access to the release name in the values.yaml when setting the subchart's env vars or collector config.

How do you want to handle this?  Hardcode the jaeger name seems like a simple option, but feels like an anti-pattern.  What do you think?

charts/opentelemetry-demo/templates/jaeger.yaml
```diff
@@ -44,7 +44,7 @@ spec:
 apiVersion: v1
 kind: Service
 metadata:
-  name: {{ include "otel-demo.name" . }}-{{ .name }}
+  name: opentelemetry-demo-jaeger
```
TylerHelmuth, line 47: example of hard coding

dmitryax: Yeah, we shouldn't do that. There must be a way. Let me think about it

TylerHelmuth: I've come up with a couple intrusive ideas, but I don't like the idea of change the collector chart specifically for this use case either.

TylerHelmuth: But maybe there is a more generic "oh you're using the collector as a subchart, then here is a solution to this problem" option.  We know there are other users using the collector as a subchart today.

TylerHelmuth: I think `tpl` could be a solution.

dmitryax: I think it's a good idea. We should clearly define which values options support templating, not just jaeger endpoint as one workaround. support of value templating can be useful for other use cases

dmitryax: Probably even all of them, if it's not going to add a lot of complexity

TylerHelmuth: @dmitryax I've pushed a version with tpl.  If we like it, I'll open a separate PR to get that into the collector chart.

TylerHelmuth: https://github.com/open-telemetry/opentelemetry-helm-charts/pull/395

TylerHelmuth:
@dmitryax @Allex1 ready for review.

dmitryax (approved):
LGTM

PR #391, opened 2022-09-28
[chore] remove core comment

TylerHelmuth:
Since opentelemetry-collector requires the healthcheck extension it can't be used with the Core image.  Removing comment that suggests it can.


dmitryax:
health check extension is part of the core image https://github.com/open-telemetry/opentelemetry-collector-releases/blob/main/distributions/otelcol/manifest.yaml#L39

PR #395, opened 2022-09-30
[opentelemetry-collector] Use tpl with configmap

TylerHelmuth:
Updates the configmap helpers to run the configuration through `tpl`.  `.Values.config` now supports templating.  Affects daemonset, deployment, and statefulset configmaps.

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.34.0
+version: 0.35.0
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-collector/UPGRADING.md
+++ b/charts/opentelemetry-collector/UPGRADING.md
@@ -4,7 +4,7 @@
 
 [config supports templating](TBD)
 
-The chart now supports templating in `.Values.config`.  If you are currently using any `{{ }}` syntax in `.Values.yaml` it will now be rendered.  To escape existing instances of `{{`, use `{{ "{{" }}`. To escape existing instances of `}}` `{{ "}}" }}`.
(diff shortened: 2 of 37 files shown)


dmitryax (approved):
Looks good to me

dmitryax:
Actually I think there are might be corner cases when this is undesirable. We should document how to escape it or make it optional/configurable feature

Allex1 (approved):

Allex1:
> Actually I think there are might be corner cases when this is undesirable. We should document how to escape it or make it optional/configurable feature

Can you pls provide some examples @dmitryax? We're using this internally in multiple charts.

dmitryax:
For example when user want to use something with `{{ }}` in a value, for example:

```yaml
config:
  processors:
    attributes:
      actions:
        - key: email
           action: update
           value: "{{ REDACTED_EMAIL }}" 
```

I think this will fail. 

I believe at least we should document a way to escape such values

TylerHelmuth:
@dmitryax @Allex1 we can handle that situation with this escape sequence.

```yaml
config:
  processors:
    attributes:
      actions:
        - key: email
          action: update
          value: '{{ "{{" }} REDACTED_EMAIL {{ "}}" }}'
```

The result it

```yaml
config:
  processors:
    attributes:
      actions:
        - key: email
          action: update
          value: '{{ REDACTED_EMAIL }}'
```

Allex1:
> @dmitryax @Allex1 we can handle that situation with this escape sequence.
> 
> ```yaml
> config:
>   processors:
>     attributes:
>       actions:
>         - key: email
>           action: update
>           value: '{{ "{{" }} REDACTED_EMAIL {{ "}}" }}'
> ```
> 
> The result it
> 
> ```yaml
> config:
>   processors:
>     attributes:
>       actions:
>         - key: email
>           action: update
>           value: '{{ REDACTED_EMAIL }}'
> ```
@TylerHelmuth this also works and looks better
``` `{{ {{ REDACTED_EMAIL }} `}} ```

dmitryax:
Yes, we just need to document this in values.yaml probably

dmitryax (approved):

TylerHelmuth:
@Allex1 Thanks, I'll update accordingly.

Issue #397, opened 2022-10-03
Add quote service to demo charts


joshleecreates:
The new PHP quote service is missing from the demo charts.

joshleecreates:
I can't assign myself but I'm working on this.

PR #398, opened 2022-10-03
Adds Quote Service to Demo Helm Chart

joshleecreates:
Fixes https://github.com/open-telemetry/opentelemetry-helm-charts/issues/397

The first commit could be made a separate PR, just some small fixes I noticed while working on this.

### DIFF
--- a/charts/opentelemetry-demo/templates/_pod.tpl
+++ b/charts/opentelemetry-demo/templates/_pod.tpl
@@ -23,6 +23,8 @@ loadgenerator:
   - frontend
 recommendation-service:
   - product-catalog-service
+shipping-service:
+  - quote-service
 {{- end}}
 
 
@@ -97,6 +99,10 @@ Exclude email service and treat differently because the addr. for the email serv
 {{- if .observability.otelcol.enabled }}
 - name: OTEL_EXPORTER_OTLP_ENDPOINT
   value: http://{{ include "otel-demo.name" . }}-otelcol:4317
+{{- if eq .name "shipping-service" }}
+- name: OTEL_EXPORTER_OTLP_TRACES_ENDPOINT
+  value: http://{{ include "otel-demo.name" . }}-otelcol:4317
+{{- end }}
 {{- end }}
 
 {{- if .servicePort}}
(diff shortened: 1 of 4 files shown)


charts/opentelemetry-demo/templates/_pod.tpl
```diff
@@ -106,7 +112,7 @@ Exclude email service and treat differently because the addr. for the email serv
 
 {{- if eq .name "product-catalog-service" }}
 - name: FEATURE_FLAG_GRPC_SERVICE_ADDR
-  value: {{ (printf "%s-featureflag-service:50031" $prefix ) }}
+  value: {{ (printf "%s-featureflag-service:50053" $prefix ) }}
```
TylerHelmuth, line 115: Is this related to the shipping service? If not, let's put it in its own PR.

TylerHelmuth (commented):
Thanks for taking this. Have you been able to test locally and confirm the shipping services is working as expected?

joshleecreates:
> Thanks for taking this. Have you been able to test locally and confirm the shipping services is working as expected?

This works locally when I rebuild the images with the latest from the app repo. I think the published images are out of date with recent changes to the shipping and quote services.

TylerHelmuth:
> This works locally when I rebuild the images with the latest from the app repo. I think the published images are out of date with recent changes to the shipping and quote services.

In that case we should probably hold off on this PR until the demo's official images are up to date. We don't want to release a chart that relies on demo code that hasn't been released yet. Let me know if I've misunderstood anything.

joshleecreates:
> > This works locally when I rebuild the images with the latest from the app repo. I think the published images are out of date with recent changes to the shipping and quote services.
> 
> In that case we should probably hold off on this PR until the demo's official images are up to date. We don't want to release a chart that relies on demo code that hasn't been released yet. Let me know if I've misunderstood anything.

That makes sense. I'll figure out what's going on tomorrow.

charts/opentelemetry-demo/templates/_pod.tpl
joshleecreates: I thought this was unrelated and removed it, but we actually need this so that the checkout service will complete it's flow to shipping and quote services.

PR #400, opened 2022-10-04
Update demo chart to v0.5.0-alpha

austinlparker:


charts/opentelemetry-demo/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: opentelemetry-demo
-version: 0.3.1
+version: 0.5.0
```
TylerHelmuth, line 4:
```suggestion
version: 0.4.0
```
Chart version doesn't need to align with the demo version used by default.

TylerHelmuth (commented):
Thanks for doing this so quickly. Were you able to test locally and confirm things work as expected (sans quote service)

dmitryax (approved):

joshleecreates:
This works as expected for me, thanks Austin!

PR #401, opened 2022-10-04
Add PHP quote service

joshleecreates:
This is tested and working as expected, with traces from shipping and quote services. It's a little bit kludgy to accommodate the slight variances in how each service wants the environment variables formatted w/ or w/out protocols, but it all works (even without the ffs port fix).

Fixes https://github.com/open-telemetry/opentelemetry-helm-charts/issues/397

### DIFF
--- a/charts/opentelemetry-demo/Chart.yaml
+++ b/charts/opentelemetry-demo/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: opentelemetry-demo
-version: 0.5.0
+version: 0.5.1
 description: opentelemetry demo helm chart
 home: https://opentelemetry.io/
 sources:
--- a/charts/opentelemetry-demo/values.yaml
+++ b/charts/opentelemetry-demo/values.yaml
@@ -181,6 +181,22 @@ components:
     podAnnotations: {}
     #  instrumentation.opentelemetry.io/inject-sdk: "true"
 
+  quoteService:
+    enabled: true
+    servicePort: 8080
+    env:
+      - name: OTEL_TRACES_SAMPLER
+        value: "parentbased_always_on"
+      - name: OTEL_TRACES_EXPORTER
+        value: "otlp"
+      - name: OTEL_EXPORTER_OTLP_TRACES_PROTOCOL
+        value: "grpc"
+      - name: OTEL_PHP_TRACES_PROCESSOR
+        value: "simple"
+    podAnnotations: {}
+    #  instrumentation.opentelemetry.io/inject-sdk: "true"
+
+
 opentelemetry-collector:
   nameOverride: otelcol
   mode: deployment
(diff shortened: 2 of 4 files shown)


charts/opentelemetry-demo/Chart.yaml
```diff
@@ -11,4 +11,4 @@ maintainers:
   - name: tylerhelmuth
   - name: wph95
 icon: https://opentelemetry.io/img/logos/opentelemetry-logo-nav.png
-appVersion: "0.4.0-alpha"
+appVersion: "0.5.0-alpha"
```
TylerHelmuth, line 14: Version bump should be done in its own PR, hopefully https://github.com/open-telemetry/opentelemetry-helm-charts/pull/400

TylerHelmuth (commented):
@joshleecreates thanks for handling this.  I agree, the env vars are a mess right now, I plan to take care of it soon once things slow down with the demo's feature freeze.  I'd prefer the image bump to be in its own PR but if things can't go independently I understand.  If we do need to bump the version in this PR then the chart's version should bump the minor version instead of the patch.  Also, please handle merge conflict.  @dmitryax please review.

charts/opentelemetry-demo/Chart.yaml
joshleecreates: Will wait for that PR and then bump the chart version.

TylerHelmuth: @joshleecreates if you tested on 0.5.0-alpha, please comment on that PR to confirm it works as expected.

joshleecreates: > I plan to take care of it soon once things slow down with the demo's feature freeze. 

Agreed, this makes the most sense.

> I'd prefer the image bump to be in its own PR but if things can't go independently I understand. If we do need to bump the version in this PR then the chart's version should bump the minor version instead of the patch. 

I leave it up to your discretion how you'd like to merge things. I've tested all of the variations and here is the status:

- with #400, (as with the current version of the chart), the shipping service fails because the quote service is not present and the shipping service has no address for it.
- with this PR and app release 0.4.0-alpha, the shipping service error is fixed and the communication between shipping/quote service seems to working, but the quote service fails to send traces to the collector.
- with this PR and app release 0.5.0-alpha, all services work and send traces.

> Also, please handle merge conflict.
✅

TylerHelmuth: @joshleecreates 0.5.0-alpha change merged in via #400.  Please rebase.

charts/opentelemetry-demo/values.yaml
```diff
@@ -195,4 +211,4 @@ opentelemetry-collector:
         traces:
           exporters:
             - logging
-            - jaeger
+            - jaeger
```
dmitryax, line 214: Please revert this line

dmitryax (approved):

TylerHelmuth (approved):

PR #402, opened 2022-10-04
Fix mistyped port for feature flag service

joshleecreates:
Fixing my own dumb mistake. This needs to match the port provided in values.yaml.


TylerHelmuth (approved):
@joshleecreates please bump the patch version of the chart.

Allex1 (approved):

PR #407, opened 2022-10-06
[opentelemetry-demo] Move env vars into values.yaml

### LINKED ISSUE #340: [opentelemetry-demo] Refactor component env vars to better handle service names

TylerHelmuth:
This PR moves all the env var logic into the values.yaml instead of the template helper.  

The original intent of the helper was to handle setting env vars that all services shared based on their dependencies.  In reality, there are so exceptions that it became very difficult to modify env vars and to add more exceptions.

I propose explicitly listing out each service's env vars in the values.yaml, utilizing `tpl` to handle situations where services rely on each other. This has the advantage of allowing users to set the env vars for each service exactly how they need. 

Fixes #340

### DIFF
--- a/charts/opentelemetry-demo/Chart.yaml
+++ b/charts/opentelemetry-demo/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: opentelemetry-demo
-version: 0.5.4
+version: 0.5.5
 description: opentelemetry demo helm chart
 home: https://opentelemetry.io/
 sources:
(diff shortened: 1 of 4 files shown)


dmitryax (approved):
This looks great!

PR #408, opened 2022-10-06
[opentelemetry-demo] Refactor image configuration

### LINKED ISSUE #406: [demo] - allow image names/tags to be configured
Since the OpenTelemetry community demo is meant to be forked/vendored, we want to ensure that a vendor can modify an image of one or more of the service components and leverage the official chart to deploy.

Ideally, we can have a default image name and tag prefix that can be configured for all service components, as well as the ability to override just an individual service component.

TylerHelmuth:
This PR refactors the chart's image configuration options to be more flexible and apparent.  Users may now:

1. change the jaeger image settings
2. change the default image settings for otel demo services.  this was doable before, but only for repository and pullSecrets.  Now includes tag and pullPolicy.
3. change individual service image settings.  Previously could only set the repository and tag.

Fixes #406

### DIFF
--- a/charts/opentelemetry-demo/templates/_objects.tpl
+++ b/charts/opentelemetry-demo/templates/_objects.tpl
@@ -28,7 +28,7 @@ spec:
       {{- end }}
       containers:
         - name: {{ .name }}
-          image: '{{ .imageOverride.repository | default .defaultValues.image.repository }}:{{ .imageOverride.tag | default (printf "v%s-%s" (default .Chart.AppVersion .defaultValues.image.tag) (.name | replace "-" "" | lower)) }}'
+          image: '{{ .imageOverride.repository | default .defaultValues.image.repository }}:{{ .imageOverride.tag | default (printf "v%s-%s" (default .Chart.AppVersion .defaultValues.image.tag) (replace "-" "" .name)) }}'
           imagePullPolicy: {{ .imageOverride.pullPolicy | default .defaultValues.image.pullPolicy }}
           {{- if or .ports .servicePort}}
           ports:
--- a/charts/opentelemetry-demo/templates/_objects.tpl
+++ b/charts/opentelemetry-demo/templates/_objects.tpl
@@ -28,7 +28,7 @@ spec:
       {{- end }}
       containers:
         - name: {{ .name }}
-          image: "{{ .imageOverride.repository | default .defaultValues.image.repository }}:{{ .imageOverride.tag | default .defaultValues.image.tag | default .Chart.AppVersion }}"
+          image: '{{ .imageOverride.repository | default .defaultValues.image.repository }}:{{ .imageOverride.tag | default (printf "v%s-%s" (default .Chart.AppVersion .defaultValues.image.tag) (.name | replace "-" "" | lower)) }}'
           imagePullPolicy: {{ .imageOverride.pullPolicy | default .defaultValues.image.pullPolicy }}
           {{- if or .ports .servicePort}}
           ports:
--- a/charts/opentelemetry-demo/templates/_objects.tpl
+++ b/charts/opentelemetry-demo/templates/_objects.tpl
@@ -19,16 +19,17 @@ spec:
         {{- toYaml .podAnnotations | nindent 8 }}
       {{- end }}
     spec:
-      {{- if .imageConfig.pullSecrets }}
+      {{- if or .defaultValues.image.pullSecrets .imageOverride.pullSecrets }}
       imagePullSecrets:
-        {{- toYaml .imageConfig.pullSecrets | nindent 8}}
+        {{- .imageOverride.pullSecrets | default .defaultValues.image.pullSecrets | toYaml | nindent 8}}
       {{- end }}
       {{- with .serviceAccountName }}
       serviceAccountName: {{ .serviceAccountName}}
       {{- end }}
       containers:
         - name: {{ .name }}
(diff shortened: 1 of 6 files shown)


charts/opentelemetry-demo/values.yaml
```diff
@@ -63,6 +77,7 @@ components:
     enabled: true
     useDefault:
       env: true
+    imageOverride: {}
```
dmitryax, line 80: We can call it `image` and use yaml anchors here
```suggestion
    image: *default-image
```

charts/opentelemetry-demo/values.yaml
```diff
@@ -28,10 +35,13 @@ default:
           fieldPath: metadata.name
     - name: OTEL_RESOURCE_ATTRIBUTES
       value: service.name=$(OTEL_SERVICE_NAME),k8s.namespace.name=$(OTEL_K8S_NAMESPACE),k8s.node.name=$(OTEL_K8S_NODE_NAME),k8s.pod.name=$(OTEL_K8S_POD_NAME)
-
-image:
-  repository: otel/demo
-  pullSecrets: []
+  image:
```
dmitryax, line 38: to put anchor
```suggestion
  image: &default-image
```

dmitryax:
not sure if user can override default image with the anchors tho... please let me know if you can try this out

TylerHelmuth:
> not sure if user can override default image with the anchors tho... please let me know if you can try this out

I was able to override values where the anchor was referenced, but not the anchor itself.  With this in `values.yaml`

```yaml
  image: &default-image
    repository: "otel/demo"
    # Overrides the image tag whose default is the chart appVersion.
    # The service's name will be applied to the end of this value.
    tag: ""
    pullPolicy: IfNotPresent
    pullSecrets: []

serviceAccount: ""

components:
  redis:
    enabled: true
    useDefault:
      env: true
    # Options to override the default image settings.
    image:
      <<: *default-image
      repository: "redis"
      tag: "alpine"
    ports:
      - name: redis
        value: 6379

  adService:
    enabled: true
    useDefault:
      env: true
    image:
      <<: *default-image
    servicePort: 8080
    env:
        value: 'http://{{ include "otel-demo.name" . }}-otelcol:4317'
      - name: AD_SERVICE_PORT
        value: "8080"
    podAnnotations: {}
    #  sidecar.opentelemetry.io/inject: "false"
    #  instrumentation.opentelemetry.io/inject-java: "true"
```

and this in my `testing.values.yaml`

```yaml
default:
  image:
    repository: otel/testing
    tag: "0.0.0"

components:
  adService:
    image:
      repository: testing/testing
      tag: "0.0.1"
```

I get this output

```yaml
.
.
.
    spec:
      containers:
        - name: ad-service
          image: 'testing/testing:0.0.1'
          imagePullPolicy: IfNotPresent
          ports:
.
.
.
    spec:
      containers:
        - name: checkout-service
          image: 'otel/demo:v0.0.0-checkoutservice'
          imagePullPolicy: IfNotPresent
```

`checkout-service` should've used `otel/testing` as the repository but still uses `otel/demo`.  If I set `repository` to `""` in the default values.yaml, like `tag` is, then it works. 

Printing out the values before they are used I can see that the default value is getting overridden, but the service's copy of it wasn't updated.

```json
{
  "defaultValues": {
    "image": {
      "pullPolicy": "IfNotPresent",
      "pullSecrets": [],
      "repository": "otel/testing",
      "tag": "0.0.0"
    }
  },
  "image": {
    "pullPolicy": "IfNotPresent",
    "pullSecrets": [],
    "repository": "otel/demo",
    "tag": ""
  }
}
```

charts/opentelemetry-demo/templates/_objects.tpl
```diff
@@ -19,16 +19,17 @@ spec:
...
-        {{- toYaml .imageConfig.pullSecrets | nindent 8}}
+        {{- .imageOverride.pullSecrets | default .defaultValues.image.pullSecrets | toYaml | nindent 8}}
       {{- end }}
       {{- with .serviceAccountName }}
       serviceAccountName: {{ .serviceAccountName}}
       {{- end }}
       containers:
         - name: {{ .name }}
-          image: {{ .image | default (printf "%s:v%s-%s" .imageConfig.repository .Chart.AppVersion (.name | replace "-" "" | lower)) }}
+          image: '{{ .imageOverride.repository | default .defaultValues.image.repository }}:{{ .imageOverride.tag | default (printf "v%s-%s" (default .Chart.AppVersion .defaultValues.image.tag) (.name | replace "-" "" | lower)) }}'
```
dmitryax, line 31: Why `replace "-" ""` is needed? Do we have any names with `-`?

TylerHelmuth: Ya, because we pipe the values.yaml name through kebabcase: https://github.com/open-telemetry/opentelemetry-helm-charts/blob/edb0c9316a43c687395cb6ae32c3ff69dac8aa32/charts/opentelemetry-demo/templates/component.yaml#L3.  This is great for the name, but the tag doesn't use it: https://hub.docker.com/r/otel/demo/tags

TylerHelmuth: We could get rid of the `| lower` tho

dmitryax (approved):

TylerHelmuth:
@Allex1 @joshleecreates please review.

joshleecreates (approved):
This works for me. Having access to imagePullPolicy is a welcome change.

Issue #410, opened 2022-10-10
Allow byo webhooks and cert


Allex1:
In some multi tenant kubernetes deployments users do not posses the RBAC to deploy webhooks but have a process to sideload a scrutinised version of the provided ones. Same for the certificate.

PR #411, opened 2022-10-10
[operator] Allow byo webhooks and cert

Allex1:
Addresses https://github.com/open-telemetry/opentelemetry-helm-charts/issues/410

### DIFF
--- a/charts/opentelemetry-operator/UPGRADING.md
+++ b/charts/opentelemetry-operator/UPGRADING.md
@@ -1,0 +1,9 @@
+# Upgrade guidelines
+
+## 0.13.0 to 0.14.0
+
+[Allow byo webhooks and cert](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/411)
+
+The ability to use admission webhooks has been moved from `admissionWebhooks.enabled` to `admissionWebhooks.create` as it now supports more use cases.
+
+In order to completely disable admission webhooks you need to explicitly set the environment variable `ENABLE_WEBHOOKS: "false"` in `.Values.manager.env` .
--- a/charts/opentelemetry-operator/templates/NOTES.txt
+++ b/charts/opentelemetry-operator/templates/NOTES.txt
@@ -2,7 +2,3 @@
   kubectl --namespace {{ .Release.Namespace }} get pods -l "release={{ $.Release.Name }}"
 
 Visit https://github.com/open-telemetry/opentelemetry-operator for instructions on how to create & configure OpenTelemetryCollector and Instrumentation custom resources by using the Operator.
-
-{{- if and .Values.admissionWebhooks.enabled .Values.customAdmissionWebhooks.enabled }}
-{{ fail "[ERROR] 'admissionWebhooks' flag is mutually exclusive with 'customAdmissionWebhooks' and should not be enabled at the same time." }}
-{{ end }}
--- a/charts/opentelemetry-operator/templates/NOTES.txt
+++ b/charts/opentelemetry-operator/templates/NOTES.txt
@@ -1,0 +1,8 @@
+{{ $.Chart.Name }} has been installed. Check its status by running:
+  kubectl --namespace {{ .Release.Namespace }} get pods -l "release={{ $.Release.Name }}"
+
+Visit https://github.com/open-telemetry/opentelemetry-operator for instructions on how to create & configure OpenTelemetryCollector and Instrumentation custom resources by using the Operator.
+
+{{- if and .Values.admissionWebhooks.enabled .Values.customAdmissionWebhooks.enabled }}
+{{ fail "[ERROR] 'admissionWebhooks' flag is mutually exclusive with 'customAdmissionWebhooks' and should not be enabled at the same time." }}
+{{ end }}
--- a/charts/opentelemetry-operator/values.schema.json
+++ b/charts/opentelemetry-operator/values.schema.json
@@ -11,6 +11,7 @@
         "manager",
         "kubeRBACProxy",
         "admissionWebhooks",
+        "customAdmissionWebhooks",
         "role",
         "clusterRole",
         "affinity",
@@ -772,6 +773,37 @@
                 }
             }]
         },
+        "customAdmissionWebhooks": {
+            "type": "object",
+            "default": {},
+            "title": "The customAdmissionWebhooks Schema",
+            "required": [
+                "enabled",
+                "secretName"
+            ],
+            "properties": {
+                "enabled": {
+                    "type": "boolean",
+                    "default": false,
+                    "title": "The enabled Schema",
+                    "examples": [
+                        false
+                    ]
+                },
+                "secretName": {
+                    "type": "string",
+                    "default": "",
+                    "title": "The secretName Schema",
+                    "examples": [
+                        ""
+                    ]
+                }
+            },
+            "examples": [{
+                "enabled": false,
+                "secretName": ""
+            }]
+        },
         "role": {
             "type": "object",
             "default": {},
@@ -1025,6 +1057,10 @@
                 "enabled": true,
                 "issuerRef": {}
             }
+        },
+        "customAdmissionWebhooks": {
+            "enabled": false,
+            "secretName": ""
         },
         "role": {
             "create": true
--- a/charts/opentelemetry-operator/values.schema.json
+++ b/charts/opentelemetry-operator/values.schema.json
@@ -11,7 +11,6 @@
         "manager",
         "kubeRBACProxy",
         "admissionWebhooks",
-        "customAdmissionWebhooks",
         "role",
         "clusterRole",
         "affinity",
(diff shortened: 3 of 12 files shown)


charts/opentelemetry-operator/values.yaml
```diff
@@ -99,6 +99,12 @@ admissionWebhooks:
       # kind:
       # name:
 
+## BYO webhooks and certificate while disabling chart ones.
+##
```
TylerHelmuth, line 103:
```suggestion
```

TylerHelmuth (commented):
Does values.json.schema need updated?

charts/opentelemetry-operator/values.yaml
Allex1: done

Allex1:
> Does values.json.schema need updated?

done

TylerHelmuth (approved):

Allex1:
> @Allex1 would it be valuable to add something to NOTES.txt that ensures admissionWebhooks and customAdmissionWebhooks are not both enabled at the same time?

@TylerHelmuth created NOTES.txt and added the info , as well as in values.yaml

charts/opentelemetry-operator/templates/NOTES.txt
```diff
@@ -0,0 +1,6 @@
+{{ $.Chart.Name }} has been installed. Check its status by running:
+  kubectl --namespace {{ .Release.Namespace }} get pods -l "release={{ $.Release.Name }}"
+
+Visit https://github.com/open-telemetry/opentelemetry-operator for instructions on how to create & configure OpenTelemetryCollector and Instrumentation custom resources by using the Operator.
+
+admissionWebhooks flag is mutually exclusive with customAdmissionWebhooks and should not be enabled at the same time.
```
TylerHelmuth, line 6: We can go a step further and check if they are both set and fail the install if they are.  The collector chart has an example of this technique.

Allex1:
@TylerHelmuth @dmitryax PTAL

TylerHelmuth (approved):

charts/opentelemetry-operator/values.yaml
```diff
@@ -85,7 +85,7 @@ kubeRBACProxy:
       memory: 64Mi
 
 ## Admission webhooks make sure only requests with correctly formatted rules will get into the Operator.
-##
+## They also enable the sidecar injection for OpenTelemetryCollector and Instrumentation CR's
 admissionWebhooks:
   failurePolicy: Fail
   enabled: true
```
dmitryax, line 91: I think using `create` bool flag would be the most common approach for such use cases instead of adding `customAdmissionWebhooks`. See https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/values.yaml#L137 as an example.

I believe we can deprecate `enabled` in favor of `create` and `secretName` options.

Allex1: @dmitryax I think this covers the 4 possible scenarios:
- using provided web-hooks + cert: needs `admissionWebhooks.create: true`
- using provided web-hooks but a different cert name: needs `admissionWebhooks.create: true` and `admissionWebhooks.secretName: "name"`
- not using web-hooks + cert at all: needs admissionWebhooks.create: false
- using custom sideloaded web-hooks + cert: needs admissionWebhooks.create: false(so no webhooks are created by the chart) + admissionWebhooks.secretName: "name"

WDYT

Allex1:
@dmitryax @TylerHelmuth 🙏

charts/opentelemetry-operator/values.yaml
dmitryax: Sounds good to me.

> not using web-hooks + cert at all: needs admissionWebhooks.create: false and setting env var to ENABLE_WEBHOOKS: "false"

Do we need setting env var in that case? Can't `admissionWebhooks.create: false` and `admissionWebhooks.secretName: ""` mean that?

Allex1: > Sounds good to me.
> 
> > not using web-hooks + cert at all: needs admissionWebhooks.create: false and setting env var to ENABLE_WEBHOOKS: "false"
> 
> Do we need setting env var in that case? Can't `admissionWebhooks.create: false` and `admissionWebhooks.secretName: ""` mean that?

the `ENABLE_WEBHOOKS` is true by default and if I remember correctly the manager pod will throw an error.

Allex1: @dmitryax , just tested, the manager pod does not start with that env var set to `false` when AW and cert are missing

dmitryax: I mean why do we need to ask user to set the ver? We can set to flase in the template based if `admissionWebhooks.create: false` and `admissionWebhooks.secretName: ""`, right?

Allex1: @dmitryax imo we can't set it based on only these 2 flags and encompass the 4 options outlined above. I believe that this setting will be used by a very small % of our users and caters to more advanced(or that work in more complex envs) users so I would rather have it set explicit rather than implicit.

dmitryax: SGTM . in that case we need to add `ENABLE_WEBHOOKS: "true"` into the default config to keep existing behavior, right?

dmitryax: Another thing. If you want to make a breaking change, please document the Upgrade process in a file similar to https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/UPGRADING.md

charts/opentelemetry-operator/values.yaml
```diff
@@ -34,6 +34,7 @@ manager:
   ## Adds additional environment variables
   ## e.g ENV_VAR: env_value
   env: {}
+    # ENABLE_WEBHOOKS: "false"
```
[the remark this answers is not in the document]
Allex1: @dmitryax default behaviour is set in code without needing to set the `ENABLE_WEBHOOKS` to true, but we can set it in the default values as well to be explicit.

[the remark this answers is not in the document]
dmitryax: > default behaviour is implicitly set in code without needing to set the ENABLE_WEBHOOKS to true

Do you mean that the absent `ENABLE_WEBHOOKS` env var is treated by the operator the same as `ENABLE_WEBHOOKS=true`?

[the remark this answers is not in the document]
Allex1: > > default behaviour is implicitly set in code without needing to set the ENABLE_WEBHOOKS to true
> 
> Do you mean that absent `ENABLE_WEBHOOKS` env var is treated by the operator the same as `ENABLE_WEBHOOKS=true`?
Yes, exactly . I've set it explicitly as I think it's clearer. PTAL.

charts/opentelemetry-operator/UPGRADING.md
```diff
@@ -0,0 +1,7 @@
+# Upgrade guidelines
+
+## 0.13.0 to 0.14.0
+
+[Allow byo webhooks and cert](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/411)
+
+In order to completely disable admission webhooks you need to explicitly set the environment variable `ENABLE_WEBHOOKS: "false"` in `.Values.manager.env` .
```
dmitryax, line 7: Please also mention that `admissionWebhooks.create` must be used now instead of `admissionWebhooks.enabled` as it's a breaking change

charts/opentelemetry-operator/values.schema.json
```diff
@@ -685,48 +689,45 @@
...
                     "default": {},
                     "title": "The objectSelector Schema",
-                    "examples": [{
-                        "matchLabels": {
-                            "foo": "bar"
-                        }
-                    }]
+                    "required": [],
+                    "properties": {},
+                    "examples": [{}]
```
dmitryax, line 730: Why example are removed from here and above? seems like an unrelated change

dmitryax (commented):
Couple last nits, otherwise LGTM

charts/opentelemetry-operator/values.schema.json
Allex1: regenerated the schema based on current values.

charts/opentelemetry-operator/UPGRADING.md
Allex1: done

charts/opentelemetry-operator/values.schema.json
dmitryax: I mean `"examples": [{}]` is pretty useless now :)

dmitryax (approved):

charts/opentelemetry-operator/values.schema.json
Allex1: oh, it's pretty late here :) Maybe @Sadzeih  manually added them in the json

dmitryax:
I merging this, but the new `"examples": [{}]` still don't look right. I'd be great to revert or remove them as a follow up PR

charts/opentelemetry-operator/values.schema.json
Allex1: @dmitryax we can probably address this in another pr , my schema is generated based on the current values. Can we merge ? Tomorrow is Monday :)

Allex1:
> I merging this, but the new `"examples": [{}]` still don't look right. I'd be great to revert or remove them as a follow up PR

@dmitryax the behaviour is the same for all empty dicts.
```
        "affinity": {
            "type": "object",
            "default": {},
            "title": "The affinity Schema",
            "required": [],
            "properties": {},
            "examples": [{}]

dmitryax:
> @dmitryax the behaviour is the same for all empty dicts.

I see. That's ok if we use a tool to generate it consistently. In that case we can use it in all the charts and force the same format with linter. The Q here is how to set enums with that tool which are pretty important in https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/values.schema.json?

PR #413, opened 2022-10-11
[operator] polish testframework image schema

JaredTan95:
Signed-off-by: Jared Tan <jian.tan@daocloud.io>

### DIFF
--- a/charts/opentelemetry-operator/values.schema.json
+++ b/charts/opentelemetry-operator/values.schema.json
@@ -17,7 +17,8 @@
         "tolerations",
         "nodeSelector",
         "hostNetwork",
-        "securityContext"
+        "securityContext",
+        "testFramework"
     ],
     "properties": {
         "replicaCount": {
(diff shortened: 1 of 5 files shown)


Allex1 (approved):
LGTM

charts/opentelemetry-operator/values.schema.json
```diff
@@ -858,6 +859,53 @@
                 "runAsUser": 65534,
                 "fsGroup": 65534
             }]
+        },
+        "testFramework": {
+            "type": "object",
+            "default": {},
+            "title": "The testFramework Schema",
+            "required": [
+                "image"
```
TylerHelmuth, line 868: "tag" is also required I believe based on the usage of the settings.  If we don't have a tag we'd end up with `<image-namge>:`

TylerHelmuth: Ya you're right I was in the wrong spot

TylerHelmuth (approved):

PR #422, opened 2022-10-13
[demo] Use docker library postgres image which has multiarch images

chancez:
The docker library postgres has multiarch images which works with ARM64 hosts. Currently the demo works poorly on M1 Macs because a number of images being used are amd64 only, this improves the situation by fixing the Postgres image.

I'm also assuming `cimg/postgres` doesn't have anything special for OTEL specifically. I checked their Dockerfile and it looked pretty vanilla, so I think this change should be minimal.

### DIFF
--- a/charts/opentelemetry-demo/Chart.yaml
+++ b/charts/opentelemetry-demo/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: opentelemetry-demo
-version: 0.6.1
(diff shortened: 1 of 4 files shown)


dmitryax:
@chancez please bumb patch version of the chart

dmitryax:
@chancez please also run `make generate-examples`

TylerHelmuth (approved):

PR #423, opened 2022-10-13
opentelemetry-demo: Support overriding OTEL_EXPORTER_OTLP_ENDPOINT and d OTEL_EXPORTER_OTLP_TRACES_ENDPOINT

chancez:
I've got my own OTEL collector running as a daemonset, and would like to be able to use it instead of the provided Otel collector. However the endpoint is hardcoded, so this change allows users to override the endpoint. Additionally, when running the collector as a daemonset, you need to send to the hostIP of the node in Kubernetes, so this exposes that as the `HOST_IP` env-var so pods can use this to send to the local Otel agent.

Example values.yaml:
```
observability:
  otelExporterOTLPGRPCEndpoint: "http://$(HOST_IP):4317"
  otelExporterOTLPHTTPEndpoint: "http://$(HOST_IP):4318"
  otelcol:
    enabled: false
  jaeger:
    enabled: false
```

### DIFF
--- a/charts/opentelemetry-demo/values.yaml
+++ b/charts/opentelemetry-demo/values.yaml
@@ -1,4 +1,7 @@
 observability:
+  # override the otel exporter endpoint to use
+  otelExporterOTLPGRPCEndpoint: ""
+  otelExporterOTLPHTTPEndpoint: ""
   # collector settings are configured in the opentelemetry-collector section.
   otelcol:
     enabled: true
@@ -13,6 +16,11 @@ observability:
 
 default:
   env:
+    # if sending to an otel-collector daemonset, this is required
+    - name: HOST_IP
+      valueFrom:
+        fieldRef:
+          fieldPath: status.hostIP
     - name: OTEL_SERVICE_NAME
       valueFrom:
         fieldRef:
@@ -35,6 +43,8 @@ default:
           fieldPath: metadata.name
     - name: OTEL_RESOURCE_ATTRIBUTES
       value: service.name=$(OTEL_SERVICE_NAME),k8s.namespace.name=$(OTEL_K8S_NAMESPACE),k8s.node.name=$(OTEL_K8S_NODE_NAME),k8s.pod.name=$(OTEL_K8S_POD_NAME)
+    - name: OTEL_EXPORTER_OTLP_ENDPOINT
+      value: '{{ include "otel-demo.otelExporterOTLPGRPCEndpoint" . }}'
   image:
     repository: otel/demo
     # Overrides the image tag whose default is the chart appVersion.
@@ -65,8 +75,6 @@ components:
     imageOverride: {}
     servicePort: 8080
     env:
-      - name: OTEL_EXPORTER_OTLP_ENDPOINT
-        value: 'http://{{ include "otel-demo.name" . }}-otelcol:4317'
       - name: AD_SERVICE_PORT
         value: "8080"
     podAnnotations: {}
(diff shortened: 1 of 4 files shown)


charts/opentelemetry-demo/values.yaml
```diff
@@ -1,4 +1,7 @@
 observability:
+  # override the otel exporter endpoint to use
+  otelExporterOTLPGRPCEndpoint: ""
+  otelExporterOTLPHTTPEndpoint: ""
```
dmitryax, line 4: Let's don't introduce these options. You can just override `default.env` instead

chancez: But then I have to re-define all the existing default values in `default.env` right?

chancez: Plus, I still have to override individual ones that require HTTP vs GRPC which is quite cumbersome.

dmitryax: Correct. If it's configurable (even with more lines of code), we don't want to introduce another yaml option. Otherwise the the yaml interface blows up and becomes hard to read and maintain

dmitryax: > Plus, I still have to override individual ones that require HTTP vs GRPC which is quite cumbersome, especially since the values are arrays, that means for shippingService.env and emailService.env, as well as default.env, I have to copy the whole set of existing values to my values.yaml

I'm ok with moving the env vars to default, so you will only need to override `default.env`, right?

TylerHelmuth: I think one option is to switch the env vars from a list to a dict so that they can merge properly. We don't do it anywhere in this repo, but I'd be interested to see examples of other charts using that technique.

chancez: The problem is the fact the the merge behavior is replace for arrays, so because env-vars are in an array, it means I have to copy existing defaults, otherwise the service literally won't even run. So you can't even override knowing the defaults. That means if you change the default values, users overriding the component-level envs, won't get those updates.

chancez: >  I'm ok with moving the env vars to default, so you will only need to override default.env, right?

No because the same env-var can't specify both HTTP and GRPC OTLP endpoints, and different applications use different values here. The email and shipping service use HTTP, while other apps use GRPC. Setting them once at the default-env won't let you do this.

dmitryax: > I think one option is to switch the env vars from a list to a dict so that they can merge properly. We don't do it anywhere in this repo, but I'd be interested to see examples of other charts using that technique.

Yes, we can you an approach similar to https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/values.yaml#L186 where keys are used for overriding only

chancez: That could work, though it would be a breaking change, and would still require more configuration on the user's side, and I feel like this would be simpler for what IMO, is likely quite a common use-case.

dmitryax: I believe it's ok to make such breaking change given that it's still days of the demo chart. Better than introducing redundant configuration options

joshleecreates: I think there is a lot of value in being able to easily configure the collector endpoints in one place for all services. I like @TylerHelmuth's suggestion to use a map for env vars to make it easier to override specific variables, FWIW.

chancez: Refactoring the chart to use a map for env-vars is a bigger task than I can take on right now, so I'll have to defer that.

TylerHelmuth:
@chancez with the merging of https://github.com/open-telemetry/opentelemetry-helm-charts/pull/484 can this PR be closed?

PR #424, opened 2022-10-14
[opentelemtry-collector] Fixed the annotations section for StatefulSet / Deployment

sherifkayad:
`annotations` on the metadata level of Deployments / StatefulSets were not working properly. That PR attempts to fix that.

### DIFF
--- a/charts/opentelemetry-collector/templates/_helpers.tpl
+++ b/charts/opentelemetry-collector/templates/_helpers.tpl
@@ -139,13 +139,6 @@ Return if ingress is stable.
 {{- end }}
 {{- end }}
 
-{{- define "opentelemetry-collector.annotations" -}}
-{{- if and (eq .Values.mode "deployment") .Values.annotations }}
-annotations:
-  {{- .Values.annotations | toYaml | nindent 2  }}
-{{- end }}
-{{- end }}
-
 {{/*
 Return the appropriate apiVersion for podDisruptionBudget.
 */}}
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.36.2
+version: 0.36.3
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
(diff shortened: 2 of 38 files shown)


sherifkayad:
@TylerHelmuth am not really sure why the linting of the demo chart fails .. tried bumping the version and upgrading the dependency didn't work .. then tried downgrading the dependency also didn't work .. any ideas?

sherifkayad:
@TylerHelmuth I will revert all changes done to the demo chart .. once you can review would be cool to have a look at the CI job and let me know what might be wrong there

TylerHelmuth:
@sherifkayad it looks like you've deleted the demo example.  Make sure your branch has the latest from main and then run "make generate-examples". Only the collector examples should change.

sherifkayad:
@TylerHelmuth I managed to fix the checks. Would be great if you can have another look

Allex1 (approved):

charts/opentelemetry-collector/templates/_helpers.tpl
```diff
@@ -139,13 +139,6 @@ Return if ingress is stable.
 {{- end }}
 {{- end }}
 
-{{- define "opentelemetry-collector.annotations" -}}
-{{- if and (eq .Values.mode "deployment") .Values.annotations }}
```
TylerHelmuth, line 143: Instead of removing this helper, can we at statefulset to the condition?

dmitryax: I actually agree @sherifkayad. I don't think why the helper is needed anymore. @TylerHelmuth don't we want to apply annotation to daemonset?

TylerHelmuth: I don't have a good reason other than at some point we decided to exclude them for daemonsets. I can't think of a good reason to do that so you're both right, removing this is a good idea.

TylerHelmuth (approved):

charts/opentelemetry-collector/templates/_helpers.tpl
sherifkayad: @dmitryax @TylerHelmuth thanks guys for the feedback. So I understand we can get this one merged already?

dmitryax (approved):

PR #427, opened 2022-10-15
Allow service/ingress exposure in any mode

### LINKED ISSUE #415: Allow user to enable ingress in agent/daemonset mode
Currently user is restricted to use ingress only when collector is deployed in `deployment` mode and user explicitly enable it.
We have use case where collector is running in agent mode, but at same time it could also ingest trace spans from outside of cluster.
Can this [artificial restriction](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/dcaeddac11897bc9e7c452efced12307665bee84/charts/opentelemetry-collector/templates/ingress.yaml#L1) be removed from chart?

rkosegi:
- Simplify logic in service creation
- Allow user to expose ingress in any mode

Closes #415

Signed-off-by: Richard Kosegi <richard.kosegi@gmail.com>

### DIFF
--- a/charts/opentelemetry-collector/templates/service.yaml
+++ b/charts/opentelemetry-collector/templates/service.yaml
@@ -1,4 +1,4 @@
-{{- if or (eq .Values.mode "deployment") (eq .Values.mode "statefulset") -}}
+{{- if or (eq .Values.mode "deployment") (eq .Values.mode "statefulset") (.Values.ingress.enabled) -}}
 apiVersion: v1
 kind: Service
 metadata:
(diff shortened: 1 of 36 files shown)


charts/opentelemetry-collector/templates/service.yaml
```diff
@@ -1,4 +1,3 @@
-{{- if or (eq .Values.mode "deployment") (eq .Values.mode "statefulset") -}}
```
TylerHelmuth, line 1: Instead of removing this restriction, let's add ingress.enabled to the or list. I don't think we want to start adding services for daemonset users who aren't using ingress options.

TylerHelmuth (commented):
Thanks for contributing! Please run "make generate-examples".

TylerHelmuth (approved):

Allex1 (approved):
LGTM

TylerHelmuth:
@rkosegi please rebase

Issue #425, opened 2022-10-14
Update the opentelemetry-collector helm chart to use v0.62.0 docker image


bogdandrutu:

TylerHelmuth:
Handled via https://github.com/open-telemetry/opentelemetry-helm-charts/pull/418

bogdandrutu:
Sorry about this @TylerHelmuth . We have to use v0.62.1, forgot to update the title.

PR #428, opened 2022-10-15
Bump collector version to 0.62.1

Allex1:
https://github.com/open-telemetry/opentelemetry-helm-charts/issues/425

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.36.0
+version: 0.36.1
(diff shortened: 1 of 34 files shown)


charts/opentelemetry-collector/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.36.0
+version: 0.37.0
```
dmitryax, line 3: I would say we should do a patch version bump in this case

TylerHelmuth (approved):

charts/opentelemetry-collector/Chart.yaml
Allex1: done

TylerHelmuth (approved):

Issue #426, opened 2022-10-14
[opentelemetry-demo] bump default image version to 0.6.0-beta


TylerHelmuth:

TylerHelmuth:
@joshleecreates @wph95

PR #432, opened 2022-10-16
[opentelemetry-demo] Update demo chart to 0.6.0-beta

joshleecreates:
Fixes https://github.com/open-telemetry/opentelemetry-helm-charts/issues/426

There are errors coming from recommendation service for me using the current version as well as this version. Need to investigate.

### DIFF
--- a/charts/opentelemetry-demo/Chart.yaml
+++ b/charts/opentelemetry-demo/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: opentelemetry-demo
-version: 0.6.2
+version: 0.7.0
 description: opentelemetry demo helm chart
 home: https://opentelemetry.io/
 sources:
--- a/charts/opentelemetry-demo/Chart.yaml
+++ b/charts/opentelemetry-demo/Chart.yaml
@@ -11,7 +11,7 @@ maintainers:
   - name: tylerhelmuth
   - name: wph95
 icon: https://opentelemetry.io/img/logos/opentelemetry-logo-nav.png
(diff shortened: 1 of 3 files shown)


joshleecreates:
Ok, so I have confirmed that the errors I'm seeing on the recommendation service is an issue with the current version and not new. I will make a separate issue for that and I don't think it should block this PR, but I defer to any reviewer who disagrees.

TylerHelmuth:
@joshleecreates no need to block the bump if it's an issue with the demo itself and not it's usage in k8s. The issue appears in docker compose as well?

joshleecreates:
> @joshleecreates no need to block the bump if it's an issue with the demo itself and not it's usage in k8s. The issue appears in docker compose as well?

It does not occur in docker compose. It does occur with the current main commit of the helm chart.

TylerHelmuth:
Thanks for creating the issue. I think we can move forward with the bump then and prioritize fixing the issue in a separate PR.

charts/opentelemetry-demo/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: opentelemetry-demo
-version: 0.6.2
+version: 0.6.3
```
TylerHelmuth, line 4:
```suggestion
version: 0.7.0
```
Normally we do appVersion bumps as a minor bump.

TylerHelmuth: You'll have to regenerate the examples.

TylerHelmuth (approved):
Thanks for doing this.

PR #434, opened 2022-10-17
Update README with links to charts

povilasv:
A minor update to link other charts with README

PR #437, opened 2022-10-18
[chore] Add dependency step to setup

TylerHelmuth:
Adds the dependency chart so that `helm dependency build` works as expected.  This hasn't been causing issues because any time `check-examples` was run we had already run `ct lint`, which added the dependency repo.  But for PRs that don't need linting, such as this PR and https://github.com/open-telemetry/opentelemetry-helm-charts/pull/434, we still need something to add the dependency repo.

### DIFF
--- a/.github/actions/setup/action.yaml
+++ b/.github/actions/setup/action.yaml
@@ -23,3 +23,7 @@ runs:
     - name: Create kind cluster
       uses: helm/kind-action@v1.3.0
       if: ${{ inputs.create-kind-cluster == 'true' }}
+
+    - name: Add Dependencies
+      shell: bash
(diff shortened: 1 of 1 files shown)


.github/actions/setup/action.yaml
```diff
@@ -23,3 +23,7 @@ runs:
     - name: Create kind cluster
       uses: helm/kind-action@v1.3.0
       if: ${{ inputs.create-kind-cluster == 'true' }}
+
+    - name: Add Dependencies
+      shell: bash
+      run: helm repo add open-telemetry https://open-telemetry.github.io/opentelemetry-helm-charts
```
TylerHelmuth, line 29: I thought about adding this to the top of `generate-examples` and `check-examples`, but I didn't want affect contributor's local repos.

dmitryax (approved):

PR #440, opened 2022-10-19
[demo] - update jaeger config

puckpuck:
Updates the Jaeger config to be inline with what is used by the otel demo's docker setup

- Set max traces to 10000, this can also be configured by the user with a values.yaml setting.
- Use OTLP instead of Jaeger protocol
- Removed all unused ports

Because the default OpenTelemetry collector config changed, including the exporter used, a chart minor version bump was in order.

### DIFF
--- a/charts/opentelemetry-demo/values.yaml
+++ b/charts/opentelemetry-demo/values.yaml
@@ -10,6 +10,7 @@ observability:
       tag: "latest"
       pullPolicy: IfNotPresent
       pullSecrets: []
+    args: ["--memory.max-traces", "10000"]
 
 default:
   env:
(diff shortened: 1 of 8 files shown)


charts/opentelemetry-demo/values.yaml
```diff
@@ -356,13 +357,13 @@ opentelemetry-collector:
   mode: deployment
   config:
     exporters:
-      jaeger:
-        endpoint: '{{ .Release.Name }}-jaeger:14250'
+      otlp:
+        endpoint: '{{ .Release.Name }}-jaeger:4317'
```
TylerHelmuth, line 361: Interesting, I wasn't aware jaeger new how to receive otlp data

TylerHelmuth (commented):
Run `make generate-examples` to fix the lint issues

charts/opentelemetry-demo/values.yaml
puckpuck: They are also deprecating their SDKs, pushing people towards OTLP 😄

TylerHelmuth (approved):

PR #444, opened 2022-10-19
[demo] - set memory limits

puckpuck:
This sets memory limits for all components including Jaeger and the Collector in line with what is set in the OpenTelemetry demo.  Ref: https://github.com/open-telemetry/opentelemetry-demo/pull/460

### DIFF
--- a/charts/opentelemetry-demo/values.yaml
+++ b/charts/opentelemetry-demo/values.yaml
@@ -10,9 +10,6 @@ observability:
       tag: "latest"
       pullPolicy: IfNotPresent
       pullSecrets: []
-    resources:
-      limits:
-        memory: 275Mi
 
 default:
   env:
@@ -60,9 +57,6 @@ components:
     ports:
       - name: redis
         value: 6379
-    resources:
-      limits:
-        memory: 20Mi
 
   adService:
     enabled: true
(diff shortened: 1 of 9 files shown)


puckpuck:
Is there a way to just have it take everything from the PR branch for the generated files in the merge conflict or am I better off just doing a new PR? Manually resolving 30+ conflicts on generated files seems excessive.

charts/opentelemetry-demo/values.yaml
```diff
@@ -348,12 +378,17 @@ components:
...
+    resources:
+      limits:
+        memory: 30Mi
 
 
 opentelemetry-collector:
   nameOverride: otelcol
   mode: deployment
+  resources:
+    limits:
```
TylerHelmuth, line 390: This will result in a collector with 

```yaml
resources:
  limits:
    cpu: 256m
    memory: 100Mi
```

Do you want to decrease the cpu as well?

TylerHelmuth (commented):
@puckpuck how did you decide what the limits should be?

charts/opentelemetry-demo/values.yaml
puckpuck: When I did the long-running tests I only looked at memory, though given we don't use any processors, we should likely lower CPU as well, I just don't have any data to support what it should be.

puckpuck:
> @puckpuck how did you decide what the limits should be?

This PR on the demo has details on what I did to determine memory usage for all services
https://github.com/open-telemetry/opentelemetry-demo/pull/460

TylerHelmuth (approved):

TylerHelmuth (approved):

PR #445, opened 2022-10-19
[demo] - allow configurable serviceType

puckpuck:
Allows the `serviceType` for each component to be configured (defaults to ClusterIP). This will allow users to deploy the demo and set some service components to be exposed based on what their infra allows (load balancers, etc) within their K8s cluster.

### DIFF
--- a/charts/opentelemetry-demo/templates/_objects.tpl
+++ b/charts/opentelemetry-demo/templates/_objects.tpl
@@ -36,6 +36,8 @@ spec:
           {{- end }}
           env:
             {{- include "otel-demo.pod.env" . | nindent 10 }}
+          resources:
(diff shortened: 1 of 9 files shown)


charts/opentelemetry-demo/templates/_objects.tpl
```diff
@@ -47,6 +47,7 @@ metadata:
   labels:
     {{- include "otel-demo.labels" . | nindent 4 }}
 spec:
+  type: {{ .serviceType | default .defaultValues.serviceType }}
```
TylerHelmuth, line 50: Did you mean to add an entry to `.Values.default` for `serviceType`?

TylerHelmuth (approved):

PR #452, opened 2022-10-20
Bump Demo Chart to v0.6.1-beta

austinlparker:


TylerHelmuth (commented):
I merged in another demo PR so we need another bump and example regeneration

austinlparker:
@TylerHelmuth ok, this should be g2g

Issue #421, opened 2022-10-13
Add k8s cluster metrics preset


dmitryax:
Similar to k8s attributes processor, we need a preset for k8s API metrics collection using k8scluster receiver. This preset must require deployment mode with replica = 1.

PR #454, opened 2022-10-20
[opentelemetry-collector] add cluster metrics preset

povilasv:
Ref: https://github.com/open-telemetry/opentelemetry-helm-charts/issues/421

Tested with:
```
helm install --set "presets.clusterMetrics.enabled=true" --set "mode=deployment" --generate-name .
```

Atm we dont enforce `replicaCount=1` and deployment mode, Should we force this somehow? (I'm not sure if helm supports this)

### DIFF
--- a/charts/opentelemetry-collector/templates/clusterrole.yaml
+++ b/charts/opentelemetry-collector/templates/clusterrole.yaml
@@ -1,4 +1,4 @@
-{{- if or (.Values.clusterRole.create) (.Values.presets.kubernetesAttributes.enabled) -}}
+{{- if or (.Values.clusterRole.create) (.Values.presets.kubernetesAttributes.enabled) (.Values.presets.clusterMetrics.enabled) -}}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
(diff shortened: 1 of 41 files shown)


charts/opentelemetry-collector/templates/clusterrole.yaml
```diff
@@ -18,4 +18,26 @@ rules:
     resources: ["pods", "namespaces"]
     verbs: ["get", "watch", "list"]
   {{- end }}
+  {{- if .Values.presets.clusterMetrics.enabled}}
+  - apiGroups: [""]
+    resources: ["pods", "namespaces"]
+    verbs: ["get", "watch", "list"]
+  {{- end }}
+  {{- if .Values.presets.clusterMetrics.enabled}}
```
TylerHelmuth, line 26: These if conditions look the same, can we use one?

TylerHelmuth (commented):
We could enforce in NOTES.txt that mode=deployment and replicaCount=1 if presets.clusterMetrics.enabled=true, but we've started being less prescriptive with the charts recently.  This might still be a good place to enforce some best practices tho.  @dmitryax @Allex1 what do you think?

TylerHelmuth:
@povilasv please rebase

dmitryax:
Кunning еру k8s cluster receiver with more than one replica is just wrong, so I think we should fail if it's set to something else. If users want otherwise for any weird reason, they can always configure the k8s cluster receiver without using the preset

TylerHelmuth:
@povilasv can you update NOTES.txt to enforce one replica and deployment mode when using this preset?

dmitryax:
> @povilasv can you update NOTES.txt to enforce one replica and deployment mode when using this preset?

Stateful set with one replica works as well

TylerHelmuth (approved):

PR #455, opened 2022-10-20
[opentelemetry-collector] cleanup log preset attributes

povilasv:
I think it's time to get rid of attributes, as we warned:

> Once Collector version 0.62.0 is released we will remove the log-level attributes. If you need log-level attributes, please configure log collection manually.

:)


TylerHelmuth (approved):
Thanks for taking care of this.

PR #456, opened 2022-10-21
[demo] add grafana and prometheus

puckpuck:
This adds Grafana and Prometheus to the demo app. The sample OpenTelemetry Collector dashboard is also part of this deployment. Both are done as subcharts.

Grafana is setup to use anonymous login (similar to the demo)

Because the Prometheus chart does not use `tpl` for ConfigMaps I couldn't use the same static scrape configs as specified in the demo. Instead, this uses a Kubernetes scraper for pods. The scrape will target pods with a specific annotation that is only present on the OpenTelemetry collector pod.  The Collector itself also includes an additional Prometheus receiver to scrape its own self-metrics and emit them as part of the standard Prometheus exporter that gets scraped by the Prometheus server.


charts/opentelemetry-demo/values.yaml
```diff
@@ -500,14 +516,1674 @@ opentelemetry-collector:
...
+      providers:
+        - name: 'default'
+          orgId: 1
+          folder: ''
+          type: file
+          disableDeletion: false
+          editable: true
+          options:
+            path: /var/lib/grafana/dashboards/default
+  dashboards:
```
TylerHelmuth, line 632: I'm not psyched about this massive dashboard in the values.yaml.  Is there a way to provide it via its own file/template?

charts/opentelemetry-demo/Chart.yaml
```diff
@@ -17,3 +17,11 @@ dependencies:
...
     repository: https://open-telemetry.github.io/opentelemetry-helm-charts
     condition: observability.otelcol.enabled
+  - name: prometheus
+    version: 15.16.1
+    repository: https://prometheus-community.github.io/helm-charts
+    condition: observability.prometheus.enabled
+  - name: grafana
+    version: 6.42.2
+    repository: https://grafana.github.io/helm-charts
+    condition: observability.prometheus.enabled
```
TylerHelmuth, line 27:
```suggestion
    condition: observability.grafana.enabled
```

TylerHelmuth (commented):
Were you able to test locally and confirm everything is hooked up with the collector as expected?

puckpuck:
> Were you able to test locally and confirm everything is hooked up with the collector as expected?

Yes. This is the OpenTelemetry collector dashboard
Screen Shot 2022-10-21 at 5 27 24 PM

charts/opentelemetry-demo/values.yaml
puckpuck: I moved it to a configmap ... much cleaner this way

TylerHelmuth (approved):
@dmitryax @Allex1 please review.  One thing that does concern me is how we'll keep the chart's dependencies up-to-date.  I haven't found anything like dependabot for helm chart dependencies

Allex1:
@TylerHelmuth for managing helm dependencies [Renovate](https://github.com/renovatebot/renovate) is a good option.

@puckpuck @dmitryax @TylerHelmuth I may be biased here but while I can see the value of adding Prometheus/Alertmanager/Grafana components to any otel setup, I think the user should deploy and manage them separately and not as part of this helm chart as dependencies.

Some thoughts:

- it will be difficult to keep all values in sync and make it convoluted for the average user.
- why not sub-chart in the otel collector/otel operator charts as well ?
- why not sub-chart [kube-prometheus-stack](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack) which would give us Prometheus-Operator CRD's which are k8s native objects and way more powerful than the vanilla Prometheus config ?
- More advanced users can already create a wrapper chart within their env which mixes and matches between otel/Prometheus/other sub-charts.

puckpuck:
@Allex1 the intent here is to keep this on par with deploying the demo today with [docker compose](https://github.com/open-telemetry/opentelemetry-demo/blob/main/docs/docker_deployment.md). 

This isn't meant for people to use in production, but rather have one-stop batteries included solution to seeing it all just work. I'm up for suggestions on other ways to include these components, but to make this work for anyone on a new cluster we will need them.

charts/opentelemetry-demo/values.yaml
```diff
@@ -500,14 +516,118 @@ opentelemetry-collector:
...
+              static_configs:
+                - targets: ['0.0.0.0:8888']
+
     exporters:
       otlp:
-        endpoint: '{{ .Release.Name }}-jaeger:4317'
+        endpoint: '{{ include "otel-demo.name" . }}-jaeger:4317'
         tls:
           insecure: true
+      logging:
```
dmitryax, line 536: logging exporter is already defined in the collector chart. No need to redefine it here

charts/opentelemetry-demo/values.yaml
```diff
@@ -500,14 +516,118 @@ opentelemetry-collector:
...
+      prometheus:
+        endpoint: '0.0.0.0:9464'
+
+    processors:
+      batch:
```
dmitryax, line 541: Same for batch

charts/opentelemetry-demo/values.yaml
```diff
@@ -500,14 +516,118 @@ opentelemetry-collector:
...
+        metrics:
+          level: detailed
+          address: 0.0.0.0:8888
       pipelines:
         traces:
-          exporters:
-            - logging
-            - otlp
+          receivers: [otlp]
+          processors: [spanmetrics, batch]
```
dmitryax, line 553: Please add `memory_limiter` processor

charts/opentelemetry-demo/values.yaml
```diff
@@ -500,14 +516,118 @@ opentelemetry-collector:
...
+    processors:
+      batch:
+      spanmetrics:
+        metrics_exporter: prometheus
+
     service:
+      telemetry:
+        metrics:
+          level: detailed
+          address: 0.0.0.0:8888
```
dmitryax, line 549: This is part of the default config. not need to redefine

dmitryax:
I think it makes sense to add Grafana and Prometheus to the demo chart to see a fully featured e2e open-source observability solution. We also need Elasticsearch for logs, right? 

My concern is similar to the one raised by @Allex1, I don't like how complicated the values.yaml becomes and how many options we supply to the sub-charts to get the desired behavior. I would like to ask you to reduce the custom sub-chart values as much as possible and rely more on defaults. E.g. if we really need spanmetrics processor, it would be nice to add it as a preset in the collector chart to keep the `opentelemetry-collector` part of values.yaml much simpler.

TylerHelmuth:
I can crank out a spanmetrics processor preset tomorrow since I'm not at Kubecon.

austinlparker:
> I think it makes sense to add Grafana and Prometheus to the demo chart to see a fully featured e2e open-source observability solution. We also need Elasticsearch for logs, right?

We haven't added logging to the demo yet, once it's there we'll integrate it into the helm chart.

> My concern is similar to the one raised by @Allex1, I don't like how complicated the values.yaml becomes and how many options we supply to the sub-charts to get the desired behavior. I would like to ask you to reduce the custom sub-chart values as much as possible and rely more on defaults. E.g. if we really need spanmetrics processor, it would be nice to add it as a preset in the collector chart to keep the `opentelemetry-collector` part of values.yaml much simpler.

Part of the goal of the demo that it should be very customizable; We don't want forks to proliferate. I also don't necessarily think that the demo should be setting guidance for the other helm components (e.g., we use spanmetrics processor because we don't do any sampling right now, this way we can get RED metrics for each endpoint using spans.)

I'm not opposed to pulling parts out into the other charts, but tbf I'd also like to reduce the surface area of this chart entirely and migrate things like the collector to an operator config rather than being deployed with the application itself.

TylerHelmuth:
@dmitryax I actually don't think the preset I would create would help the demo chart?  

I think a default spanmetricsprocessor preset would be configured to use the existing metrics pipeline and the logging exporter, whereas the demo needs to use the prometheus exporter.  We wouldn't want the collector preset to make any assumptions about where the data would get exported.

So we can still make a spanmetrics processor as a preset as it holds individual value, but it won't help simplify this PR.

charts/opentelemetry-demo/values.yaml
```diff
@@ -500,14 +516,118 @@ opentelemetry-collector:
...
         traces:
-          exporters:
-            - logging
-            - otlp
+          receivers: [otlp]
+          processors: [spanmetrics, batch]
+          exporters: [otlp, logging]
+        metrics:
+          receivers: [otlp, prometheus]
+          processors: [batch]
```
TylerHelmuth, line 557: Please add the memory_limiter processor here as well.

TylerHelmuth (commented):
@puckpuck can you review the 3 subchart entries in the values.yaml and confirm that all values defined are necessary?

dmitryax:
> Part of the goal of the demo that it should be very customizable; We don't want forks to proliferate. I also don't necessarily think that the demo should be setting guidance for the other helm components (e.g., we use spanmetrics processor because we don't do any sampling right now, this way we can get RED metrics for each endpoint using spans.)

I'm not saying we should the ability to customize the helm chart including sub-charts. I just want to avoid adding unnecessary predefined configuration in the default values.yaml for the sub-charts. PTAL at the PR comments. For example, if we partially override the collector's config as it's currently done in this PR, it'll be hard to maintain it going forward - a change to the collector's config interface or default collector chart configuration may break it. So I think it's better to rely on the default configuration of the collector chart instead of partially overriding it.

dmitryax:
> So we can still make a spanmetrics processor as a preset as it holds individual value, but it won't help simplify this PR.

@TylerHelmuth why not? Won't it be 2 changes in this PR: add spanmetrics preset + replacement of the exporter in the config?

dmitryax:
Also, I don't think we should strife for 1to1 parity with the docker-compose setup. The demo deployment in k8s likely needs other features like k8s attributes processor which is not needed in the docker-compose setup.

Allex1:
Imo helm charts should be as contained as possible. As an end user you have all the flexibility to add any other charts next to this one under a parent chart that you maintain within your project.
Personally I prefer kube-prometheus-stack and not prometheus/grafana charts and this allows me to just merge the 2 under my own wrapper chart.

charts/opentelemetry-demo/values.yaml
puckpuck: I removed the processor definitions for logs and metrics since they would have been the same as the defaults.

puckpuck:
Sorry about the delay. I was at Kubecon for the week.

I believe I have addressed all the concerns noted. I did my best to minimize the setting overrides for the sub-charts. I know this is a big add to this chart, but these are 2 key components required for that all-in experience we want to provide. For people that already have a Prometheus and Grafana setup, we have facilities in this chart to use their own existing services, and I will make sure that experience is well documented here and/or the demo repository in a future PR.

charts/opentelemetry-demo/Chart.yaml
```diff
@@ -11,9 +11,17 @@ maintainers:
   - name: tylerhelmuth
   - name: wph95
 icon: https://opentelemetry.io/img/logos/opentelemetry-logo-nav.png
-appVersion: "0.6.1-beta"
+appVersion: "1.0.0"
```
TylerHelmuth, line 14: Can we take the 1.0 images in this PR without the frontendproxy stuff?

puckpuck: Yes.  The frontend proxy is an additional layer to help for browser-initiated tracing span data, which is a separate thing.  We also use it to limit the exposed ports. All that is part of the next PR.

charts/opentelemetry-demo/values.yaml
```diff
@@ -500,14 +516,112 @@ opentelemetry-collector:
...
+          receivers: [otlp]
+          processors: [memory_limiter, spanmetrics, batch]
+          exporters: [otlp, logging]
+        metrics:
+          receivers: [otlp, prometheus]
+          exporters: [prometheus, logging]
+        metrics/spanmetrics:
+          receivers: [otlp/spanmetrics]
+          processors: [memory_limiter, batch]
+          exporters: [prometheus, logging]
```
dmitryax, line 555: Why do we need this pipeline? It doesn't seem to be used...

charts/opentelemetry-demo/values.yaml
```diff
@@ -500,14 +516,112 @@ opentelemetry-collector:
...
+        protocols:
+          grpc:
+            endpoint: "localhost:65535"
+      prometheus:
+        config:
+          scrape_configs:
+            - job_name: 'otel-collector'
+              scrape_interval: 5s
+              static_configs:
+                - targets: ['0.0.0.0:8888']
```
dmitryax, line 529: This is not needed, already included in the default config https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/values.yaml#L91-L95

charts/opentelemetry-demo/values.yaml
```diff
@@ -500,14 +516,112 @@ opentelemetry-collector:
...
       pipelines:
         traces:
-          exporters:
-            - logging
-            - otlp
+          receivers: [otlp]
+          processors: [memory_limiter, spanmetrics, batch]
+          exporters: [otlp, logging]
+        metrics:
+          receivers: [otlp, prometheus]
```
dmitryax, line 550: No need to redefine receivers

charts/opentelemetry-demo/values.yaml
```diff
@@ -500,14 +516,112 @@ opentelemetry-collector:
...
+      spanmetrics:
+        metrics_exporter: prometheus
+
     service:
       pipelines:
         traces:
-          exporters:
-            - logging
-            - otlp
+          receivers: [otlp]
```
dmitryax, line 546: No need to redefine receivers

charts/opentelemetry-demo/values.yaml
```diff
@@ -500,14 +516,112 @@ opentelemetry-collector:
               allowed_origins:
                 - "http://*"
                 - "https://*"
+      otlp/spanmetrics:
+        protocols:
+          grpc:
+            endpoint: "localhost:65535"
```
dmitryax, line 522: Shouldn't be needed if `metrics/spanmetrics` pipeline is removed

charts/opentelemetry-demo/values.yaml
puckpuck: This is how we have the collector configured to generate RED metrics from all spans, and send them into Prometheus. This is a copy from what we have for the collector running in the demo repo's config https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/otelcollector/otelcol-config.yml

charts/opentelemetry-demo/values.yaml
puckpuck: we need the metrics/spanmetrics pipeline

charts/opentelemetry-demo/values.yaml
dmitryax: > This is how we have the collector configured to generate RED metrics from all spans, and send them into Prometheus.

I see that spanmetrics receiver is configured to send metrics to prometheus exporter, can please you point me to a source that is sending otlp metric to localhost:65535?

puckpuck: This is part of how you configure the spanmetrics receiver/processor to work. It needs a dummy receiver to set up a pipeline, so we set it to an endpoint that should not be used for anything else.

https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/spanmetricsprocessor

dmitryax: Submitted a PR for the demo repo https://github.com/open-telemetry/opentelemetry-demo/pull/553

puckpuck: The spanmetrics processor is absolutely being used in the demo.  This is a screenshot of some of the metrics it generates in Grafana.
Screen Shot 2022-11-02 at 8 27 20 PM

dmitryax: I'm not saying that spanmetrics processor is not needed

puckpuck: The way the spanmetrics processor works is it requires a receiver and its own pipeline to generate metrics from traces and send them along.  We have it configured properly per the docs, and it does work in our demo as shown here.

dmitryax: Can you remove this pipeline and try again? I'm pretty sure it's not needed

puckpuck: I understand the mis-communication here now.  The spanmetrics processor was using the prometheus receiver, and not the otlp/spanmetrics one. That must of changed on the demo side some time ago and I missed it.

I did try without the otlp/spanmetrics receiver and pipeline, and this is working as expected.

TylerHelmuth: Ya the spanmetricsprocessor readme is a little misleading (which is what I think prompted this extra pipeline in the demo)

TylerHelmuth:
@dmitryax ready for re-review

Issue #458, opened 2022-10-21
[opentelemetry-operator] Usage of labels and selector labels


edwintye:
The OpenTelemtry Operator helm chart doesn't use `"opentelemetry-operator.labels"` and `"opentelemetry-operator.selectorLabels"` apart from: webhook, service, service monitor, and prometheus rule. See for example [deployment.yaml#L4](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/templates/deployment.yaml#L4) and [deployment.yaml#L12](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/templates/deployment.yaml#L12). 

This behaviour is different to the collector and causes an issue because we are not able to just grab `helm.sh/chart` from the labels and detect drift between the environments easily. 

I feel like this is a relatively easy fix and would be happy to raise a PR.

edwintye:
Ah, I was complaining about this at work to @LinuxAd about this and it seems like both of us decided that an upstream contribution will be faster 🤣

TylerHelmuth:
@edwintye @LinuxAd thanks for contributing. Can you choose one of the PRs to centralize on and close the duplicate?

LinuxAd:
Sure, I'll close mine now.

PR #463, opened 2022-10-21
[opentelmetry-operator] Use labels and selector labels from helpers

edwintye:
Addresses #458 throughout

### DIFF
--- a/charts/opentelemetry-operator/templates/deployment.yaml
+++ b/charts/opentelemetry-operator/templates/deployment.yaml
@@ -2,7 +2,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   labels:
-    app.kubernetes.io/name: opentelemetry-operator
+    {{- include "opentelemetry-operator.labels" . | nindent 4 }}
     control-plane: controller-manager
   name: {{ template "opentelemetry-operator.name" . }}-controller-manager
   namespace: {{ .Release.Namespace }}
@@ -10,7 +10,7 @@ spec:
   replicas: {{ .Values.replicaCount }}
   selector:
     matchLabels:
-      app.kubernetes.io/name: opentelemetry-operator
+      {{- include "opentelemetry-operator.selectorLabels" . | nindent 6 }}
       control-plane: controller-manager
   template:
     metadata:
@@ -18,7 +18,7 @@ spec:
         kubectl.kubernetes.io/default-container: manager
         {{- include "opentelemetry-operator.podAnnotations" . | nindent 8 }}
       labels:
-        app.kubernetes.io/name: opentelemetry-operator
+        {{- include "opentelemetry-operator.labels" . | nindent 8 }}
         control-plane: controller-manager
     spec:
       hostNetwork: {{ .Values.hostNetwork }}
--- a/charts/opentelemetry-operator/Chart.yaml
+++ b/charts/opentelemetry-operator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.18.2
+version: 0.18.3
 description: OpenTelemetry Operator Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-operator/README.md
+++ b/charts/opentelemetry-operator/README.md
@@ -35,14 +35,14 @@ _See [helm repo](https://helm.sh/docs/helm/helm_repo/) for command documentation
 
 ```console
 $ helm install \
-  my-opentelemetry-operator open-telemetry/opentelemetry-operator
+  opentelemetry-operator open-telemetry/opentelemetry-operator
 ```
 
 If you created a custom namespace, like in the TLS Certificate Requirement section above, you will need to specify the namespace with the `--namespace` helm option:
 
 ```console
 $ helm install --namespace opentelemetry-operator-system \
-  my-opentelemetry-operator open-telemetry/opentelemetry-operator
+  opentelemetry-operator open-telemetry/opentelemetry-operator
 ```
 
 _See [helm install](https://helm.sh/docs/helm/helm_install/) for command documentation._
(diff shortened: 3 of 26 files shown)


Allex1:
@edwintye pls run `make generate-examples` and commit examples.

edwintye:
@Allex1 thanks for letting me know the requirement, I have generated and committed the example

Allex1:
@edwintye pls rebase

Allex1 (approved):
lgtm

charts/opentelemetry-operator/templates/deployment.yaml
```diff
@@ -2,22 +2,22 @@ apiVersion: apps/v1
...
+    {{- include "opentelemetry-operator.labels" . | nindent 4 }}
     control-plane: controller-manager
   name: {{ template "opentelemetry-operator.name" . }}-controller-manager
   namespace: {{ .Release.Namespace }}
 spec:
   replicas: {{ .Values.replicaCount }}
   selector:
     matchLabels:
-      app.kubernetes.io/name: opentelemetry-operator
+      {{- include "opentelemetry-operator.selectorLabels" . | nindent 6 }}
```
TylerHelmuth, line 13: @edwintye Can you test a `helm upgrade` with these changes?  I want to make sure a user running `0.16.0` will be able to successfully run `helm upgrade` for this chart and not have the new labels added via this PR affect anything.

edwintye: Thanks for the suggestion, you are right in that `helm upgrade` would indeed run into issues due to the change in `selectorLabels`.  I have not considered that as we run `helm template` and `kubectl apply --force` internally.

The only way to make this a save change is to move the [L51 in helpers](https://github.com/edwintye/opentelemetry-helm-charts/blob/use-helper-labels/charts/opentelemetry-operator/templates/_helpers.tpl#L51)  `app.kubernetes.io/instance: {{ .Release.Name }}` from `selectorLabels` to `labels`.  This approach deviates a little bit from the other chart in this repo.  Alternatively, this is a "breaking change", or at a minimum require a change in their upgrade path.

Annoyingly the line `app.kubernetes.io/name: {{ include "opentelemetry-operator.name" . }}` is permitted in `helm upgrade` to my amusement as it is changes the name of all the component i.e. forces a recreate anyway.

TylerHelmuth: cc @dmitryax

TylerHelmuth: @dmitryax how do you want to handle this?

edwintye: I just made the change as per above so that the upgrade can be done while still adding the labels. More concretely, the `selectorLabel` is a single field such that we have (for example in the `service`)
```yaml
  selector:
    control-plane: controller-manager
```

which allows the upgrade.  I am happy to preserve the upgrade path if that makes all our life easier :)

TylerHelmuth: Thanks, I'll take a look tomorrow.

TylerHelmuth: I believe moving `app.kubernetes.io/instance: {{ .Release.Name }}` out of selector labels will harm anyone who has installed the chart multiple times under different releases.  IDK if thats actually a risk for this chart though.

edwintye: The current selector labels don't use `app.kubernetes.io/instance` at all, e.g. in [service](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/templates/service.yaml#L21) and in [deployment](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/templates/deployment.yaml#L12). This can be verified in our generated examples where the selector labels has not changed.

I think what you are thinking of is `app.kubernetes.io/name: {{ include "opentelemetry-operator.name" . }}` where the person upgrading may decide to suddenly introduce `nameOverride`.  This does not cause the upgrade to fail because `metadata.name` uses the definition `"opentelemetry-operator.name"` so an upgrade with `nameOverride` will force all the resources to be re-created. 

BUT, I have found a problem while doing my little tests against `main` (that is not part of this PR and a bug in general).  If I install using
```shell
helm install otel open-telemetry/opentelemetry-operator --version 0.17.2 --set nameOverride=whatever --set admissionWebhooks.certManager.enabled=false
```

then the chart fails because volume mount for secret adjusts against `nameOverride` as per [deployment.yaml#L119](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/templates/deployment.yaml#L119) but the secret name does not as per [operator-webhook.yaml#L14](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook.yaml#L14).

Allex1: @edwintye see https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-operator#tls-certificate-requirement and choose one of the ise cases.

edwintye: yes, and my claim is that if you use the third option `admissionWebhooks.certManager.enabled` with `nameOverride` at the same time the chart install fails because the name of the cert (secret to be exact) does not match up.

Allex1: Ah, my bad, you are right. I haven't really tried out that option. Would you like to contribute a fix for it in a different pr ?

TylerHelmuth: @edwintye sorry I lost track of this.  I agree that moving the `app.kubernetes.io/instance: {{ .Release.Name }}` out of `opentelemetry-operator.selectorLabels` will work since the helper isn't being used today.  The fact that the examples' selector labels don't change is proof of this I think.

TylerHelmuth (approved):
@edwintye thanks for cleaning this up, I wasn't even aware we had those label helpers and they weren't being used.

Allex1:
@edwintye Please also update the main readme[ install and uninstall commands ](https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-operator#install-chart) with the release name from `my-opentelemetry-operator` to `opentelemetry-operator` to match the existing labels.
Also please solve conflicts and rebase and the we can probably merge.
Thanks for your contribution! 🥳

Allex1:
ready to merge imo

edwintye:
Thanks both for your help.  Please let me know if there is any other issues. I am slightly busy with work so may have to defer until the weekend but I am definitely still keeping track of the PR :)

Issue #438, opened 2022-10-18
[DEMO] Support updating environment variables instead of overwriting env-vars


chancez:
Currently it's challenging add  "extra" env-vars to individual components because it overwrites their own env-vars, meaning you have to copy the existing env-var lists and put those in your own values.yaml. This is problematic because many of these env-vars are necessary for the component to even run, such as it's listen port, and it also means users doing this won't get new env-vars as the chart is updated upstream.

It was suggested in https://github.com/open-telemetry/opentelemetry-helm-charts/pull/423#discussion_r995157372 to that instead of defining env-vars in a list, we could use a map which would allow better merging behavior, instead of overwriting all env-vars, we can merge env-vars on the env-var name, making it possible to add new values or overwrite existing values per component without having to worry about copying the defaults.

This would then make it easy to override values like OTEL_EXPORTER_OTLP_ENDPOINT and OTEL_EXPORTER_OTLP_TRACES_ENDPOINT so that users can provide their own collector endpoint.

a-thaler:
While implementation we realized in that thread https://github.com/open-telemetry/opentelemetry-helm-charts/pull/465#discussion_r1003388710 that the order of the map keys are not preserved in Helm. With that the resulting order of the env-var list can be deterministic (sorted alphanumerical) but different to the provided order.
In consequence you have problems with env variables having dependencies to other variables like the OTEL_RESOURCE_ATTRIBUTES in the defaultEnv list.

There seem to be no clean solution and we either need to find a hacky solution by treating dependent envs special or we accept that variables cannot be overridden.

In order to find a decision, I like to provide my motivation of having the envs overwritable. In our project we want to provide the collector infrastructure as part of the cluster setup. Now I want to leverage the demo app for tutorials and demos by having some few instructions in place to deploy the app using the provided collector and backend.
For that instructions I want to provide a pre-configured values.yaml to disable the collector and the jaeger backend in the chart and configure the components to use the provided collector. The result looks like that:

<details>
    <summary>Expand me</summary>

      observability:
        otelcol:
          enabled: false
        jaeger:
          enabled: false
      
      components:
        redis:
          enabled: true
      
        adService:
          enabled: true
          env:
            - name: OTEL_EXPORTER_OTLP_ENDPOINT
              value: 'http://my-collector.my-system:4317'
            - name: AD_SERVICE_PORT
              value: "8080"
      
        cartService:
          enabled: true
          env:
[...]

PR #465, opened 2022-10-24
[demo] change env settings from array to map

a-thaler:
**Description**
Enables https://github.com/open-telemetry/opentelemetry-helm-charts/issues/438 by converting the env variables settings from an array to a map.

**Note** That is a breaking change as users who already set custom env variables need to adopt to it.

**Changes**
- Changed pod generation logic
- Adjusted default values.yaml
- Adjusted json validation schema, enforce a schema for the map attributes supporting oneOf "values" and "valuesFrom"
- Re-generated example components.yaml


charts/opentelemetry-demo/examples/default/rendered/component.yaml
```diff
@@ -382,32 +382,32 @@ spec:
...
               fieldRef:
                 apiVersion: v1
                 fieldPath: metadata.name
-          - name: OTEL_RESOURCE_ATTRIBUTES
+          - name: "OTEL_RESOURCE_ATTRIBUTES"
             value: service.name=$(OTEL_SERVICE_NAME),k8s.namespace.name=$(OTEL_K8S_NAMESPACE),k8s.node.name=$(OTEL_K8S_NODE_NAME),k8s.pod.name=$(OTEL_K8S_POD_NAME)
-          - name: OTEL_EXPORTER_OTLP_ENDPOINT
-            value: http://example-otelcol:4317
-          - name: AD_SERVICE_PORT
+          - name: "OTEL_SERVICE_NAME"
```
TylerHelmuth, line 402: This is the downside to switching the env vars from a list to a map: `range` doesn't preserve order 😞.  But we need order preserved because we need to ensure `OTEL_RESOURCE_ATTRIBUTES` is defined after the env vars it depends on. In order to be able to switch from a list to a map we need to be able to keep order I think.

TylerHelmuth (commented):
@a-thaler thanks for contributing.  Please bump the chart's minor version and re-render the examples.

charts/opentelemetry-demo/examples/default/rendered/component.yaml
a-thaler: I see, do you know if the order is kept stable inside the map and only the range is spoiling it? So are you confident on that there can be a solution, then I will try out to find something.

Otherwise, we could introduce a deterministic order by sorting and treat the service name special which of course is ugly

TylerHelmuth: I'm 99% sure that it a problem with Maps and not `range`.  I think there _might_ be a solution that uses `| sortAlpha`, but that will only sort alphabetically.  We could use it to our advantage, but it feels hacky.

a-thaler:
I bumped the minor version and re-rendered the examples

charts/opentelemetry-demo/examples/default/rendered/component.yaml
a-thaler: As far as I can see Helm dictionaries have no order, so sorting the list of keys will make the order stable and deterministic. Still I don't see how the OTEL_SERVICE_NAME can be moved to the last element without an explicit hack taking care of that.

So either we accept that there can be no dependency between env variables or we accept that you cannot override specific elements (and find a different way to set especially the collector URLs more simple)

Another idea: We already have 2 layers of env variables with the defaultEnvs, we could also say that dependencies can exist between the two layers only and we move the OTEL_SERVICE_NAME down to each component section.

TylerHelmuth: We could move `OTEL_RESOURCE_ATTRIBUTES` down, but it breaks the notion of being a default attribute.  If you want to update `OTEL_RESOURCE_ATTRIBUTES` for every component, you'd have to update each component's version vs only changing the instance in `default.env`.

Circling back the the larger issue prompting the PR, when you had to redefine variables, what did your values.yaml end up looking like?  It's annoying to have to redefine lists, but their consistency in these types of situations is useful.

a-thaler: I see your point and understand that this is much better from a consistency perspective.

However, I want to provide instructions on how to install the demo app using a different collector infrastructure. For that I want to provide an examplaric values.yaml defining the required overrides. That values.yaml contains close to 200 lines at the moment where I mainly want to reconfigure the collector URL and disable the bundled collector. That 200 lines I would need to check for every update if they are still complete or adjusted.

Going back to the initial discussion https://github.com/open-telemetry/opentelemetry-helm-charts/pull/423#discussion_r995157372, having the collector domain configurable and re-using that variable in the env definitions could be a better solution. However, it will not solve the override problem in a generic way...

TylerHelmuth: @a-thaler would you be willing to share your values.yaml in the #438?  I'd love to see the issue first hand so I can experience the pain.

a-thaler: I shared my scenario with the values.yaml at https://github.com/open-telemetry/opentelemetry-helm-charts/issues/438#issuecomment-1290237295

a-thaler:
Closing the PR in favor of https://github.com/open-telemetry/opentelemetry-helm-charts/pull/484

Issue #446, opened 2022-10-19
[operator] Opentelemetry Operator Collector contrib image


povilasv:
At the moment we default to regular opentelemetry-collector image in the Operator chart. Lines:

 https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/values.yaml#L20-L22
 
I was thinking maybe instead of this image we should use contrib one? As it has way more receivers / processors / exporters and gives better default user experience.

Also it would be the same as we do for opentelemetry-collector chart -> https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/values.yaml#L124-L125

And I believe we use contrib for the opentelemetry-demo too

Thoughts?

TylerHelmuth:
I agree on standardizing on contrib for the charts.  @Allex1 @dmitryax do you agree?

dmitryax:
Makes sense let's use the same image by default across all the charts

PR #467, opened 2022-10-26
[opentelemetry-operator] change collector image to contrib

povilasv:
Closes: https://github.com/open-telemetry/opentelemetry-helm-charts/issues/446

Not sure how do we approach this breaking change?

I added a notice, is it good enough? Or should we roll-out this by first adding a Notice and then after some time making the change?

Example of it rendered:
```                                                           
helm install opentelemetry-operator --set "admissionWebhooks.certManager.enabled=false" .

NAME: opentelemetry-operator
LAST DEPLOYED: Wed Oct 26 11:24:31 2022
NAMESPACE: sys-mon
STATUS: deployed
REVISION: 1
NOTES:
opentelemetry-operator has been installed. Check its status by running:
  kubectl --namespace sys-mon get pods -l "release=opentelemetry-operator"

Visit https://github.com/open-telemetry/opentelemetry-operator for instructions on how to create & configure OpenTelemetryCollector and Instrumentation custom resources by using the Operator.

[Important] The v0.17.0 helm chart version changes OpenTelemetry Collector image to the contrib version. If you want to use the core version, set `manager.collectorImage.repository` to `otel/opentelemetry-collector`.
```

### DIFF
--- a/charts/opentelemetry-operator/Chart.yaml
+++ b/charts/opentelemetry-operator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.16.0
+version: 0.17.0
 description: OpenTelemetry Operator Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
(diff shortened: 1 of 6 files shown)


TylerHelmuth (commented):
Since contrib's image contains everything in Core's image no upgrades will break, but we do want to inform users that they are using a different image.  Please also add a section to the UPGRADING.md doc as well.

Allex1 (approved):
@povilasv 🚢 it after addressing @TylerHelmuth comment.

TylerHelmuth (approved):

PR #470, opened 2022-10-26
[opentelemetry-collector] Add spanmetrics processor preset

TylerHelmuth:
Adds a spanmetrics processor preset.  Allows configuring the name of which exporter is used.  

As it relates to https://github.com/open-telemetry/opentelemetry-helm-charts/pull/456 IDK if it will simplify the demo's values.yaml since the preset will need configured and the collector's config will still need to create the prometheus exporter and add it to the metrics pipeline, but its always nice to use presets where they exist.


Allex1 (approved):

TylerHelmuth:
@dmitryax I agree.

PR #471, opened 2022-10-26
[opentelemetry-collector] Reorder preset processors

TylerHelmuth:
Switch the k8sattributes preset to use `prepend` instead of `append` to ensure that `batch` continues to be the final processor in the list.


TylerHelmuth:
@Allex1 unrelated to this PR, can you DM me in CNCF slack?  I can't figure out which account is yours lol

PR #472, opened 2022-10-27
[demo] add miss config for opentelemetry-collector

Frapschen:
I find a bug in demo chart's values, it miss receiver config.


TylerHelmuth:
@Frapschen closing this as I don't believe it is needed (see the comment on my review).  If I am misunderstanding please let me know.

Frapschen:
you are right, think for your explanation :)

PR #476, opened 2022-10-28
Modified default values file to disable logging and tracing blocks

datsabk:
Modified default values file to disable logging and tracing blocks which need to be explicitly disabled when trying to override values. 

These values sneak into the configuration without user knowledge when directly using the public chart.


TylerHelmuth (changes requested):
@datsabk sorry for any confusing, but we are not interested in disabling traces and logs by default.  The chart's default collector config very purposely includes traces, metrics, and logs, and common components to support those pipelines.

If you like to disable anything in the configuration you can use `null` in the values.yaml.  Checkout [this example](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/examples/deployment-otlp-traces/values.yaml) of how to disable pipelines.

Also note, disabling default helm values when using a chart as a subchart currently doesn't work: https://github.com/helm/helm/issues/9027

datsabk:
@TylerHelmuth - Thanks for clarifying. Was just taking off the burden of adding too many null values. There is easy possibility that the user might miss out on a null value and make a mistake. For instance, configuring exporter and receiver for traces but not making processors as null - That might end up doing additional processing outside of user's knowledge.

PR #478, opened 2022-10-28
[opentelemetry-collector] Allow to reach local collector

saez0pub:
When host port is not available (Ex: gke autopilot).
We need to reach a local collector for k8attributes processor
Enabling local service internal traffic pollicy could solve this.

https://kubernetes.io/docs/concepts/services-networking/service-traffic-policy/

### DIFF
--- a/charts/opentelemetry-collector/values.yaml
+++ b/charts/opentelemetry-collector/values.yaml
@@ -279,8 +279,10 @@ lifecycleHooks: {}
 #       - "5"
 
 service:
+  enabledIfDaemonset: false
   type: ClusterIP
   annotations: {}
+  internalTrafficPolicy: null
 
 ingress:
   enabled: false
--- a/charts/opentelemetry-collector/examples/daemonset-localtraffic/rendered/configmap-agent.yaml
+++ b/charts/opentelemetry-collector/examples/daemonset-localtraffic/rendered/configmap-agent.yaml
@@ -1,0 +1,93 @@
+---
+# Source: opentelemetry-collector/templates/configmap-agent.yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: example-opentelemetry-collector-agent
+  labels:
+    helm.sh/chart: opentelemetry-collector-0.37.2
+    app.kubernetes.io/name: opentelemetry-collector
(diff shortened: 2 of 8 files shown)


charts/opentelemetry-collector/values.yaml
```diff
@@ -279,8 +279,10 @@ lifecycleHooks: {}
 #       - "5"
 
 service:
+  enabledIfDaemonset: false
```
TylerHelmuth, line 282: Maybe instead of labeling this with `daemonset` we can use something more generic like `explicitlyEnable`?  That way if something else in the future wants to enable the service it will be able to.  You can also then remove the check for `mode = daemonset` in the template's `if` statement.

TylerHelmuth (commented):
@saez0pub thanks for the contribution.  The example was helpful while reviewing, but I don't think this scenario warrants a new example dedicated to it.  Can you remove it?  Also please bump the chart's patch version and run `make generate-examples`.

charts/opentelemetry-collector/values.yaml
dmitryax: Can't we use `internalTrafficPolicy==Local && mode=daemonset` as the condition for now? If we will need someone to have a regular service with a daemonset in future, we can always introduce this option that would take precedence.

dmitryax:
@saez0pub also don't forget to bump the chart version

dmitryax:
> We need to reach a local collector for k8attributes processor

@saez0pub do you need local collector just for k8attributes processor? The processor works in deployment mode as well.

charts/opentelemetry-collector/values.yaml
TylerHelmuth: Yes that'll work

saez0pub:
> > We need to reach a local collector for k8attributes processor
> 
> @saez0pub do you need local collector just for k8attributes processor? The processor works in deployment mode as well.

Interesting, I wasn't able to make it working as a deployment 🤔.
No need to manage internalTrafficPolicy if I can make it available as a deployment 👍.
I will search to make it working as a deployment again and then open an issue If I can't.

Closing it for now, thank you.

PR #484, opened 2022-11-04
[demo] support overriding of env variables

a-thaler:
**Description**
Enables https://github.com/open-telemetry/opentelemetry-helm-charts/issues/438 by adding new properties to override environment variables. It supports to overwrite and/or add variables to the default section or to every individual component without the need of repeating the whole env lists.

It indirectly also adds support for templated default envs, so using templates in the env variable of the default section.

It adds only new properties and with that the approach is fully compatible.

The merge logic for the environment variables is implemented in a dedicated helper template which gets applied twice; for generating the default envs and the component specific ones.

The PR does not support to override in the component an environment variable defined in the default section. That will require to print the env variables all together (default and component specific) and would not allow to have a re-usable template for the merge logic.
You would need a way to first merge the default envs, then append the result of that to the component env list, then merge that with the component overrides and then print it all together. That is possible and worked out but had repeating logic and the overall template looked very complex. So I decided for now to take a more clean approach but not having the capability supported.

**Changes**
- Introduced new "envOverrides" properties for the default section and for every component
- Extracted env generation logic into another template which is now doing the merge logic
- Re-generated example components.yaml
- Updated chart version

### DIFF
--- a/charts/opentelemetry-demo/values.yaml
+++ b/charts/opentelemetry-demo/values.yaml
@@ -21,6 +21,7 @@ observability:
     enabled: true
 
 default:
+  # list of environment variables applied to all components
   env:
     - name: OTEL_SERVICE_NAME
       valueFrom:
@@ -44,6 +45,11 @@ default:
           fieldPath: metadata.name
     - name: OTEL_RESOURCE_ATTRIBUTES
       value: service.name=$(OTEL_SERVICE_NAME),k8s.namespace.name=$(OTEL_K8S_NAMESPACE),k8s.node.name=$(OTEL_K8S_NODE_NAME),k8s.pod.name=$(OTEL_K8S_POD_NAME)
+  # list of environment variables which get applied to all components additional to the list provided with the "env" section.
+  # Use this section to override or add environment variables selectively 
+  envOverrides: []
+  #  - name: OTEL_K8S_NODE_NAME
+  #    value: "someConstantValue"
   image:
     repository: otel/demo
     # Overrides the image tag whose default is the chart appVersion.
@@ -96,6 +102,7 @@ components:
         value: 'http://{{ include "otel-demo.name" . }}-otelcol:4317'
       - name: AD_SERVICE_PORT
         value: "8080"
+    envOverrides: []
     podAnnotations: {}
     resources:
       limits:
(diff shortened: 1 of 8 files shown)


charts/opentelemetry-demo/values.yaml
```diff
@@ -44,6 +45,11 @@ default:
           fieldPath: metadata.name
     - name: OTEL_RESOURCE_ATTRIBUTES
       value: service.name=$(OTEL_SERVICE_NAME),k8s.namespace.name=$(OTEL_K8S_NAMESPACE),k8s.node.name=$(OTEL_K8S_NODE_NAME),k8s.pod.name=$(OTEL_K8S_POD_NAME)
+  # list of environment variables which get applied to all components additional to the list provided with the "env" section.
+  # Use this section to override or add environment variables selectively 
+  envOverrides: []
+  #  - name: OTEL_K8S_NODE_NAME
+  #    value: "someConstantValue"
```
dmitryax, line 52: The PR description says that this section is not supported or am I missing something?

a-thaler: It is supported to override envs in the default section as part of the default section. So overwrites will apply for all components.
What is not supported is to use the envOverride in the component section to override an env which is introduced in the default section.
So the two env/envOverride pairs are isolated from each other as they are getting rendered independently

TylerHelmuth: @a-thaler looking at `_pod.tpl`, this field allows users to override `.Values.default.env` correct?  If that is the case, lets update the comment to be something like `# Allows overriding and additions to .Values.default.env`.  Hopefully that will make it clear that this field is only applied to `.Values.default.env` and nothing else.

The fact that `.Values.default.env` is applied to all components is already being documented with your new comment.

a-thaler: Hi @TylerHelmuth , have adjusted the docu accordingly and bumped the versions once more

TylerHelmuth (approved):

TylerHelmuth:
@a-thaler whenever you want to generate the examples after bumping the chart version you can use `make generate-examples`.

dmitryax (approved):

dmitryax:
Thanks, @a-thaler!

PR #486, opened 2022-11-04
Added annotations for operator service account

### LINKED ISSUE #485: Allow setting annotations for the operator service account
Currently, you can't set any annotations on the operator's service account which means that you can't have it use something like GCP workload identity or autoscaling eviction annotations.

jaronoff97:
closes #485

### DIFF
--- a/charts/opentelemetry-operator/templates/serviceaccount.yaml
+++ b/charts/opentelemetry-operator/templates/serviceaccount.yaml
@@ -2,8 +2,13 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  labels:
-    app.kubernetes.io/name: opentelemetry-operator
   name: {{ template "opentelemetry-operator.serviceAccountName" . }}-controller-manager
   namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "opentelemetry-operator.labels" . | nindent 4 }}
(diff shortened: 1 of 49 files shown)


Allex1 (approved):

charts/opentelemetry-operator/templates/serviceaccount.yaml
```diff
@@ -2,8 +2,13 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  labels:
-    app.kubernetes.io/name: opentelemetry-operator
   name: {{ template "opentelemetry-operator.serviceAccountName" . }}-controller-manager
   namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "opentelemetry-operator.labels" . | nindent 4 }}
+    {{- include "opentelemetry-operator.selectorLabels" . | nindent 4 }}
```
dmitryax, line 9: This looks like something unrelated to this PR. I don't believe we have these labels consistently applied to other resources, so it should be handled separately.

jaronoff97: sure, i can remove this section. Figured while I was there i should give it what the expectation of the consumer would be.

dmitryax (approved):

PR #489, opened 2022-11-08
feat: add filesystem scraper to hostmetrics preset

povilasv:
IMO we should support filesystem receiver in hostmetrics preset. It's supported on all OS'es and I think it's very useful :)

Docs:

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/hostmetricsreceiver/internal/scraper/filesystemscraper/documentation.md


dmitryax:
I agree

dmitryax (approved):

Issue #172, opened 2022-04-15
use unexposed port to check healthy


caibirdme:
in _pod.tpl
```yaml
    livenessProbe:
      httpGet:
        path: /
        port: 13133
    readinessProbe:
      httpGet:
        path: /
        port: 13133
```
But port 13133 is not exposed

TylerHelmuth:
Hello @caibirdme, can you expand on what you mean?   LivenessProbe and readinessProbe ports are normally not exposed in the deployment/daemonset's `ports` section.  [You can see both probes working as expected in our tests.](https://github.com/open-telemetry/opentelemetry-helm-charts/runs/6026270800?check_suite_focus=true) Are you experiencing an issue getting K8s to interact with the probes?

navin-rai:
Liveness probe failed: Get "http://********:13133/": dial tcp ******:13133: connect: connection refused

Liveness probe failed: Get "http://********:13133/": dial tcp ******:13133: connect: connection refused

I am getting this issue while deploying the collector, what may be the reason ?

PR #490, opened 2022-11-09
Make health check port configurable

datsabk:
In scenarios where we are looking to deploy multiple helm charts in a single cluster - to allow different configurations, there is a conflict with healthCheck port causing one of the charts to fail. 

This conflict is observed only when running multiple charts - all with hostMode enabled. 

This PR allows easier configuration of this health check port while templatising the url for the extension configuration.

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.38.2
+version: 0.38.3
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-collector/examples/daemonset-and-deployment/rendered/configmap-agent.yaml
+++ b/charts/opentelemetry-collector/examples/daemonset-and-deployment/rendered/configmap-agent.yaml
@@ -5,7 +5,7 @@ kind: ConfigMap
 metadata:
   name: example-opentelemetry-collector-agent
   labels:
-    helm.sh/chart: opentelemetry-collector-0.38.2
+    helm.sh/chart: opentelemetry-collector-0.38.3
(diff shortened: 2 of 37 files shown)


dmitryax:
@datsabk doesn't `config.extensions.health_check.endpoint` work for your use case? we don't want to expose configuration options to high level for every possible use case

datsabk:
@dmitryax That does work for configuration of health check extension. But the port is hard coded into the pod.tpl file leading to this problem. Hard-coded port practically makes it impossible to configure a different endpoint. 

So with this PR, I make both of these items configurable

datsabk:
> > This conflict is observed only when running multiple charts - all with hostMode enabled.
> 
> Do you mean `hostNetwork`?

Yes

dmitryax:
Curios why do you need hostNetwork on a deployment?

datsabk:
To hijack the port and ensure the applications can talk to the agent on the same port on each node. This helps us in faster communication (reduced network hops) and also save cost on cross az traffic(when using service instead of local communication)

datsabk:
@dmitryax Let me know if you have the config parameters belongs to a different location.

dmitryax:
> To hijack the port and ensure the applications can talk to the agent on the same port on each node. 

So you run collector in hostNetwork as a deamonset not deployment, right? If so, how do you run into a conflict? Do you have another daemonset or deployment in hostNetwork?

dmitryax:
> @dmitryax Let me know if you have the config parameters belongs to a different location.

Yes, the right location for this should be `ports`. You need to add `healthCheck` there and use it in the pod template. No other configuration changes should be applied

TylerHelmuth:
Related to https://github.com/open-telemetry/opentelemetry-helm-charts/issues/172

dmitryax:
> Related to https://github.com/open-telemetry/opentelemetry-helm-charts/issues/172

Right, we can expose that port but it shouldn't be enabled on the service or as node port

PR #496, opened 2022-11-11
As kubelet gives us container fs usage we don't need to scrape this

oded-dd:
mount points

We might consider checking if the preset for kubelete is enabled, but with this PR I used the naive approach.

### DIFF
--- a/charts/opentelemetry-collector/templates/_config.tpl
+++ b/charts/opentelemetry-collector/templates/_config.tpl
@@ -156,6 +156,13 @@ receivers:
         memory:
         disk:
         filesystem:
+          exclude_mount_points:
+            mount_points:
+            - /snap/*
+            - /boot/*
+            - /var/lib/kubelet/pods/*
+            - /var/lib/kubelet/plugins/*
+            match_type: regexp
         network:
 {{- end }}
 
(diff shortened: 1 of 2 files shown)


charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -156,6 +156,13 @@ receivers:
         memory:
         disk:
         filesystem:
+          exclude_mount_points:
```
TylerHelmuth, line 159: I like your idea of making this optional, making something like

```yaml
hostMetrics:
    enabled: false
    excludeKubeletStatsFilesystems: false
```

or something like that.  A comment explaining when to use it would be good as well.

TylerHelmuth (commented):
@oded-dd thanks for the contribution!  I am not very familiar with the exact output of the kubletstats receiver, which metrics in https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/kubeletstatsreceiver/documentation.md are duplicated by hostmetrics watching these mount points?

Also, please bump the patch version of the chart and run `make generate-examples`

TylerHelmuth:
@povilasv I'm cool with either direction as long as its configurable and the user can turn things off/on based on if they are using just the hostmetrics preset, just the kubletstats preset, or both.

charts/opentelemetry-collector/templates/_config.tpl
oded-dd: @TylerHelmuth thanks. I'll add support for that structure.

PR #497, opened 2022-11-11
Following opentelemetry security best practices Users SHOULD bind rec…

oded-dd:
…eivers' servers to addresses that limit connections to authorized users

Setting chart defaults to the pod.ip should guarantee that.

https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/security-best-practices.md

Signed-off-by: Oded David <oded@coralogix.com>

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.39.1
+version: 0.39.2
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-collector/values.yaml
+++ b/charts/opentelemetry-collector/values.yaml
@@ -74,17 +74,17 @@ config:
     jaeger:
       protocols:
         grpc:
-          endpoint: 0.0.0.0:14250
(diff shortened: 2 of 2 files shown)


TylerHelmuth (commented):
@oded-dd the security recommendation for the collector endpoint is 

https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/security-best-practices.md?plain=1#L144-L145

which we are following.  I don't think using the pod ip by default is necessary for the chart to be in compliance.

TylerHelmuth:
Ya that is annoying.  Does this change break `kubectl port-forward`?

TylerHelmuth:
The demo relies on port forwarding so I really prefer not to stop using 0.0.0.0 since it is still meeting the security guidance.

oded-dd:
Hey. I am ok with not changing the code, but I do think the readme should be updated, as localhost does not follow any deployment type (e.g. hostNetwork = true vs hostNetwork = false). I think the readme should be changed/improved. @TylerHelmuth by your permission.

TylerHelmuth:
@oded-dd sure, want to submit a PR for that?

oded-dd:
Sure. Sorry for the delay. I'll take care of it in the next couple of days. Thank you.

TylerHelmuth:
Closing in favor of #534.  Thanks for raising this issue!

PR #502, opened 2022-11-14
Add a judgement while creating the admissionWebhooks' secret

dashanji:
If we have set up the `admissionWebhooks.secretName`, it looks like no need to install the secret.

### DIFF
--- a/charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook.yaml
+++ b/charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook.yaml
@@ -2,6 +2,7 @@
 {{- $altNames := list ( printf "%s.%s" "opentelemetry-operator-webhook-service" .Release.Namespace ) ( printf "%s.%s.svc" "opentelemetry-operator-webhook-service" .Release.Namespace ) -}}
 {{- $ca := genCA "opentelemetry-operator-operator-ca" 365 -}}
 {{- $cert := genSignedCert (include "opentelemetry-operator.fullname" .) nil $altNames 365 $ca -}}
+{{- if (not .Values.admissionWebhooks.secretName) }}
 apiVersion: v1
 kind: Secret
 type: kubernetes.io/tls
@@ -16,6 +17,7 @@ metadata:
 data:
   tls.crt: {{ $cert.Cert | b64enc }}
   tls.key: {{ $cert.Key | b64enc }}
+{{- end }}
 ---
 apiVersion: admissionregistration.k8s.io/v1
 kind: MutatingWebhookConfiguration


Allex1:
@dashanji[ that value](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/templates/deployment.yaml#L114) makes sure that the secret is mounted on the operator pod so your pr completely disables webhooks. Can you explain your use case?
You may need to use point 4 of https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-operator#tls-certificate-requirement where you can sideload your custom cert/secret  .

dashanji:
@Allex1 Make sense. I found a specific scenery. if users want to set up the webhook with the default secret but set up the `admissionWebhooks.secretName` at the same time, so I think it's a bug as the secret is wrong. For that, should we move the judgment to [the head](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/0e9ce252fe0551cec6d4eb76414990340af83e43/charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook.yaml#L1) to avoid it?

Allex1:
> @Allex1 Make sense to me. In fact, I found a specific scenery. if users want to set up the webhook with the default secret but set up the `admissionWebhooks.secretName` at the same time, I think it's a bug as the secret is wrong. For that, should we move the judgment to [the head](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/0e9ce252fe0551cec6d4eb76414990340af83e43/charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook.yaml#L1) to avoid it?

@dashanji I don't see why someone would want to use the default secret and also define another secret. That would be a corner case and also violate the [recommendations](https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-operator#tls-certificate-requirement) but I don't think it would brake anything if you add this condition to the [head](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/0e9ce252fe0551cec6d4eb76414990340af83e43/charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook.yaml#L1) like you mention, but also in the [webhooks with certmanager head](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/0e9ce252fe0551cec6d4eb76414990340af83e43/charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook-with-cert-manager.yaml#L1) .

@dmitryax @TylerHelmuth wdyt ?

dashanji:
Yes, actually it's a corner case.

Allex1:
> Yes, actually it's a corner case.

@dashanji so if you want to continue this pr pls rebase and bump the chart

PR #503, opened 2022-11-15
[opentelemetry-demo] Bump demo and collector version

TylerHelmuth:
Bumps demo images to 1.1.0 which fixes issues using the chart locally on Apple Silicon.

Bumps collector dependency to 0.39.1

### DIFF
--- a/charts/opentelemetry-demo/Chart.yaml
+++ b/charts/opentelemetry-demo/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: opentelemetry-demo
-version: 0.11.1
+version: 0.12.0
 description: opentelemetry demo helm chart
 home: https://opentelemetry.io/
 sources:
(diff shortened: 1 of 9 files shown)


charts/opentelemetry-demo/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: opentelemetry-demo
-version: 0.11.1
+version: 0.12.0
```
puckpuck, line 4: Does this warrant a minor version bump?

dmitryax: We usually bump minor chart version for minor version upgrades in images. Patch version bumps are usually made for fixes and minor changes in values.yaml interface. But we could reconsider that practice

TylerHelmuth: Id vote a minor bump for this one since it was a minor bump for the demo itself.

dmitryax (approved):

PR #505, opened 2022-11-15
[demo] Feat/scheduling rules for jaeger

### LINKED ISSUE #504: Add "schedulingRules" to jaeger
In the demo chart - I'd like to have the ability to specify node affinities for Jaeger's pod which from my understanding, is currently not support.

Is it reasonable to add the existing schedulingRules to the [values schema](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-demo/values.schema.json) and update [Jaeger template](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-demo/templates/jaeger.yaml) accordingly?

Thanks.

tglanz:
Closes #504.

Added the capability to specify scheduling rules for jaeger. 

Changes made:
- Added types to jaeger's schema
- Added default values
- Re-generated examples using make

Tested by diffing generated examples and verifying against a working cluster.

### DIFF
--- a/charts/opentelemetry-demo/examples/default/rendered/jaeger.yaml
+++ b/charts/opentelemetry-demo/examples/default/rendered/jaeger.yaml
@@ -5,7 +5,7 @@ kind: Service
 metadata:
   name: example-jaeger
   labels:
-    helm.sh/chart: opentelemetry-demo-0.12.0
+    helm.sh/chart: opentelemetry-demo-0.12.1
     app.kubernetes.io/name: example
     app.kubernetes.io/instance: example
     app.kubernetes.io/component: jaeger
@@ -34,7 +34,7 @@ kind: Deployment
 metadata:
   name: example-jaeger
   labels:
-    helm.sh/chart: opentelemetry-demo-0.12.0
+    helm.sh/chart: opentelemetry-demo-0.12.1
     app.kubernetes.io/name: example
     app.kubernetes.io/instance: example
     app.kubernetes.io/component: jaeger
--- a/charts/opentelemetry-demo/examples/default/rendered/jaeger.yaml
+++ b/charts/opentelemetry-demo/examples/default/rendered/jaeger.yaml
@@ -54,7 +54,7 @@ spec:
         app.kubernetes.io/name: example
         app.kubernetes.io/instance: example
         app.kubernetes.io/component: jaeger
-    spec:      
+    spec:
       containers:
         - name: jaeger
           image: "jaegertracing/all-in-one:latest"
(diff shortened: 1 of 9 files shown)


charts/opentelemetry-demo/examples/default/rendered/jaeger.yaml
```diff
@@ -54,7 +54,7 @@ spec:
         app.kubernetes.io/name: example
         app.kubernetes.io/instance: example
         app.kubernetes.io/component: jaeger
-    spec:
+    spec:
```
TylerHelmuth, line 57: Small nit, but can it be made so that these extra spaces aren't rendered?

charts/opentelemetry-demo/templates/_helpers.tpl
```diff
@@ -57,3 +57,20 @@ app.kubernetes.io/component: {{ .name}}
 {{- $mergedEnvs = concat $mergedEnvs $envOverrides }}
 {{- tpl (toYaml $mergedEnvs) . }}
 {{- end }}
+
+{{- define "otel-demo.schedulingRules" -}}
```
TylerHelmuth, line 61: Currently the `.Values.default` values only apply to components and not to jaeger.  My initial gut reaction is to continue to keep them separate (and we can rename the field to `componentDefaults` in a separate PR to denote this).  

The `.Values.observability.jaeger` section could still have `schedulingRules` so that they can be configured and used directly in the jaeger template.

@dmitryax @puckpuck @puckpuck what do you think?

TylerHelmuth (commented):
@tglanz thanks for the contribution.  Please sign the CLA and make sure the email you sign with is associated with all commits in the PR.

charts/opentelemetry-demo/templates/_helpers.tpl
puckpuck: I think we should keep components and Jaeger scheduling rules separate.

dmitryax: I agree it should be separate. 

But this brings another question: We use sub-charts for the collector, prometheus and grafana. Can we do the same for jaeger? In that case, this PR won't be needed.

tglanz: reverted the code and added jaeger specific logic to generate scheduling rules.

tglanz: @dmitryax i'll resolve this any case you want to keep/discard the PR

TylerHelmuth (approved):
@tglanz thanks!

PR #506, opened 2022-11-16
[demo] - change order of default and component env vars

puckpuck:
Before this change,  the overall default environment variables are added before the component-level environment variables for each pod. This change reverses that order.

This is trying to solve a situation, where I want to add an additional attribute as part of the `OTEL_RESOURCE_ATTRIBUTES`   variable defined at the overall level. My definition would include a new environment variable that I add to each component, but since those variables are added to the pod template after, they are not picked up using the `$(MY_VAR_NAME)` syntax.  With this change, the following values passed into a helm install will work as expected.

```
default:
  envOverrides:
      value: service.name=$(OTEL_SERVICE_NAME),k8s.namespace.name=$(OTEL_K8S_NAMESPACE),k8s.node.name=$(OTEL_K8S_NODE_NAME),k8s.pod.name=$(OTEL_K8S_POD_NAME),app.hive.team=$(HIVE_TEAM_NAME)


components:
  adService:
    envOverrides:
      - name: HIVE_TEAM_NAME
        value: pollen

  cartService:
    envOverrides:
      - name: HIVE_TEAM_NAME
        value: apiary
```

In the above example, each component defines a new var called `HIVE_TEAM_NAME` which gets referenced as part of the default overall `OTEL_RESOURCE_ATTRIBUTES` var added to each component automatically.

### DIFF
--- a/charts/opentelemetry-demo/templates/_pod.tpl
+++ b/charts/opentelemetry-demo/templates/_pod.tpl
@@ -1,14 +1,29 @@
 {{/*
 Get Pod Env
-Note: Consider that dependent variables need to be declared before the referenced env variable.
+Merges default environment variables (if used) with component environment variables.
+If using defaults, will pull out OTEL_RESOURCE_ATTRIBUTES from the list and add it to the end.
+The OTEL_RESOURCES_ATTRIBUTES environment variable uses Kubernetes environment variable expansion and should be last.
 */}}
 {{- define "otel-demo.pod.env" -}}
-{{- if .env }}
-{{ include "otel-demo.envOverriden" . }}
-{{- end }}
+{{- $resourceAttributesEnv := dict }}
+{{- $allEnvs := list }}
 {{- if .useDefault.env  }}
-{{ include "otel-demo.envOverriden" (dict "env" .defaultValues.env "envOverrides" .defaultValues.envOverrides "Template" $.Template) }}
+{{-   $defaultEnvs := include "otel-demo.envOverriden" (dict "env" .defaultValues.env "envOverrides" .defaultValues.envOverrides) | mustFromJson }}
+{{-   range $defaultEnvs }}
+{{-     if eq .name "OTEL_RESOURCE_ATTRIBUTES" }}
+{{-       $resourceAttributesEnv = . }}
+{{-     else }}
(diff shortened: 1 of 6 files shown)


charts/opentelemetry-demo/templates/_pod.tpl
```diff
@@ -3,12 +3,12 @@ Get Pod Env
 Note: Consider that dependent variables need to be declared before the referenced env variable.
 */}}
 {{- define "otel-demo.pod.env" -}}
-{{- if .useDefault.env  }}
-{{ include "otel-demo.envOverriden" (dict "env" .defaultValues.env "envOverrides" .defaultValues.envOverrides "Template" $.Template) }}
-{{- end }}
 {{- if .env }}
 {{ include "otel-demo.envOverriden" . }}
 {{- end }}
+{{- if .useDefault.env  }}
```
TylerHelmuth, line 9: I think no matter which order we chose someone is gonna be unhappy.  I think wanting to add content to `OTEL_RESOURCE_ATTRIBUTES` is a good reason to enact this switch.

Also, the number of default env vars is small, so if someone needed to reference `OTEL_SERVICE_NAME` in a component it wouldn't be too hard to remove it from default and add it to the component's env vars.

TylerHelmuth (approved):

charts/opentelemetry-demo/templates/_pod.tpl
puckpuck: I think the proper foolproof solution is because OTEL_RESOURCE_ATTRIBUTES depends on Kubernetes env var expansion, we should find a way to force it to always be last.

TylerHelmuth: We could definitely do that by pulling it out of the default list, applying the default list, applying the component env vars, and then applying the `OTEL_RESOURCE_ATTRIBUTES` env var.  Do you want to try that instead of switching the order?

puckpuck: rewrote the logic for this.

Does defaults env vars before component ones, but now it will pull OTEL_RESOURCE_ATTRIBUTES out of the default ones, and add it back in at the of the concatenated list.

TylerHelmuth (approved):

Issue #508, opened 2022-11-16
logsCollection storage extension


povilasv:
What do you think if we configure `storage` extension for the filelog receiver using logsCollection preset?

Then we would  set up the `file_storage` extension for it?

As I understand it stores the last file offsets to a file.

Something like:
```
extensions:
  file_storage:
    directory: /var/log/pods/
receivers:
  filelog:
    include: [ /var/log/pods/*/*/*.log ]
    ...
    storage: file_storages
```

Though there is one question - where do we store offsets db?

TylerHelmuth:
I think this is a good idea and safe since the file_storage extension is beta.  I would opt to but this feature behind a setting in the logCollection preset so that users have to opt in to this extra component.

PR #509, opened 2022-11-16
[opentelemetry-collector] add storage extension to logsCollection preset

povilasv:
ref https://github.com/open-telemetry/opentelemetry-helm-charts/issues/508

### DIFF
--- a/charts/opentelemetry-collector/values.yaml
+++ b/charts/opentelemetry-collector/values.yaml
@@ -25,6 +25,9 @@ presets:
   logsCollection:
     enabled: false
     includeCollectorLogs: false
+    # Enabling this writes checkpoints in /var/lib/otelcol/ host directory.
+    # Note this changes collector's user to root, so that it can write to host directory.
+    storeCheckpoints: false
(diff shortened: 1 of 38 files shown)


charts/opentelemetry-collector/values.yaml
```diff
@@ -25,6 +25,7 @@ presets:
   logsCollection:
     enabled: false
     includeCollectorLogs: false
+    storeOffsets: false
```
TylerHelmuth, line 28: Can you add a comment describing what this setting does and the security implications of using this setting?

TylerHelmuth, line 28: I noticed the filelogreceiver readme used "checkpoints" to describe what the `storage` option does.  Would that be a more user-friendly word instead of `Offsets`?

TylerHelmuth (commented):
Left a couple nits, but overall looks great.

TylerHelmuth (commented):
@povilasv please take a look at the merge conflicts.

Issue #491, opened 2022-11-09
[opentelemetry-collector] Cannot disable logs/metrics with kubernetesAttributes preset


sebbegg:
When enabling the `kubernetesAttributes` preset and attempting to disable the logs, traces or metrics pipeline fails:

```bash
❯ helm template --debug --set presets.kubernetesAttributes.enabled=true --set mode=deployment --set config.service.pipelines.metrics=null --version 0.38.2 open-telemetry/opentelemetry-collector
install.go:178: [debug] Original chart version: "0.38.2"
install.go:195: [debug] CHART PATH: /Users/eckweis/Library/Caches/helm/repository/opentelemetry-collector-0.38.2.tgz


Error: template: opentelemetry-collector/templates/deployment.yaml:31:28: executing "opentelemetry-collector/templates/deployment.yaml" at <include (print $.Template.BasePath "/configmap.yaml") .>: 
error calling include: template: opentelemetry-collector/templates/configmap.yaml:10:8: executing "opentelemetry-collector/templates/configmap.yaml" at <include "opentelemetry-collector.deploymentConfig" .>: 
error calling include: template: opentelemetry-collector/templates/_config.tpl:80:15: executing "opentelemetry-collector.deploymentConfig" at <include "opentelemetry-collector.applyKubernetesAttributesConfig" (dict "Values" $data "config" $config)>: 
error calling include: template: opentelemetry-collector/templates/_config.tpl:275:21: executing "opentelemetry-collector.applyKubernetesAttributesConfig" at <$config.service.pipelines.metrics>: wrong type for value; expected map[string]interface {}; got interface {}
helm.go:84: [debug] template: opentelemetry-collector/templates/deployment.yaml:31:28: executing "opentelemetry-collector/templates/deployment.yaml" at <include (print $.Template.BasePath "/configmap.yaml") .>: 
error calling include: template: opentelemetry-collector/templates/configmap.yaml:10:8: executing "opentelemetry-collector/templates/configmap.yaml" at <include "opentelemetry-collector.deploymentConfig" .>: 
error calling include: template: opentelemetry-collector/templates/_config.tpl:80:15: executing "opentelemetry-collector.deploymentConfig" at <include "opentelemetry-collector.applyKubernetesAttributesConfig" (dict "Values" $data "config" $config)>: 
[...]

PR #510, opened 2022-11-16
[opentelemetry-collector] check for empty pipelines

povilasv:
closes https://github.com/open-telemetry/opentelemetry-helm-charts/issues/491

Tested with:

helm template --debug --set presets.kubernetesAttributes.enabled=true --set mode=deployment --set config.service.pipelines.metrics=null .


dmitryax (approved):

dmitryax:
Thanks @povilasv !

PR #512, opened 2022-11-19
fix targetPort to target containerPort

Stelminator:
this might not be the only instance, I think I saw something similar in a file with "demo" in the name.

I want to create a service with multiple external ports targeting the same internal port. As is, I'll just create two duplicate helm installs with just the port different.

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.40.4
+version: 0.40.5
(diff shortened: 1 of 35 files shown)


TylerHelmuth (approved):
@Stelminator thanks for the contribution.  Please bump the chart's patch version and then run `make generate-examples`

Stelminator:
@TylerHelmuth rebased, bumped patch version, ran `make generate-examples`

> something similar in a file with "demo" in the name.
Confirmed this was in the `-demo` chart, and it's not clear to me if there's a problem there. It's in the other chart, so, a separate problem if there is one.

PR #513, opened 2022-11-19
[collector] - use percentage for memory limits

puckpuck:
This uses percentage limits for the `memory_limiter` processor and the `memory_ballast` extension instead of computing limits from defined pod resources.

I also noticed the `memory_ballast` extension was not getting configured for `deployment` mode collectors, so I added that in as well.

### DIFF
--- a/charts/opentelemetry-collector/templates/_config.tpl
+++ b/charts/opentelemetry-collector/templates/_config.tpl
@@ -28,8 +28,8 @@ Merge user supplied config into memory ballast config.
 */}}
 {{- define "opentelemetry-collector.ballastConfig" -}}
 {{- $memoryBallastConfig := get .Values.config.extensions "memory_ballast" }}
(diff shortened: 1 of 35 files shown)


charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -67,6 +67,7 @@ Build config file for deployment OpenTelemetry Collector
 {{- $values := deepCopy .Values }}
 {{- $data := dict "Values" $values | mustMergeOverwrite (deepCopy .) }}
 {{- $config := include "opentelemetry-collector.baseConfig" $data | fromYaml }}
+{{- $config = include "opentelemetry-collector.ballastConfig" $data | fromYaml | mustMergeOverwrite $config }}
```
TylerHelmuth, line 70: If we want include ballast in both deployment/statefulset and daemonset then I we can instead add it to `opentelemetry-collector.baseConfig`

PR #514, opened 2022-11-23
[opentelemetry-collector] add zpages extension

povilasv:
hey :)

I think we should by default add zpages extension, I used it today to debug a slow exporter:

Tested with:
```
helm install --set "mode=daemonset" --set "presets.logsCollection.enabled=true" --generate-name .

kubectl port-forward chart-1669198614-opentelemetry-collector-agent-4jnnb 55679
```

Example UI:
image

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.39.3
+version: 0.39.4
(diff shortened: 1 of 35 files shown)


TylerHelmuth (commented):
I like the idea of making it easier to use zpages with the chart, but since it is a debugging tool maybe its better as a preset that users opt into?

dmitryax:
> Regarding the preset option - I think this extension is pretty easy to configure even without preset, so maybe we just close this PR?

I agree. I don't think it should be a preset

Issue #515, opened 2022-11-23
Enable "env" configuration for jaeger deployment


tglanz:
I'd like to use an ElasticSearch cluster as a storage backend for Jaeger. For this I need to set some environment variables in Jaeger's containers. Some of the variables should be taken from the k8s cluster's secrets so it would be ideal to have an option configuring something like

```yaml
jaeger:
  env:
  - name: ELASTIC_USERNAME
     valueFrom:
       secretKeyRef:
         name: elastic-credentials
         key: username
  - name: ELASTIC_PASSWORD
     valueFrom:
       secretKeyRef:
         name: elastic-credentials
         key: password
```

tglanz:
Am wiling to work on it if that's something your'e interested in adding

tglanz:
@TylerHelmuth I'm not sure but it looks like the chart doesn't support in-memory mode? Which is this chart's default

puckpuck:
Moving to a Jaeger chart is a good idea here.

The downside is we may break backward compatibility for anyone that configured something specific to the Jaeger deployment.

dmitryax:
Yes, I believe we need to use jaeger as a sub-chart. We may be able to support the current jaeger config interface as deprecated and translate it into the sub-chart config, but it may be too much work

puckpuck:
The breaking change would also be the catalyst to do a proper readme for the chart as well finally.

puckpuck:
I went down the path of using Jaeger as a sub-chart.  

Unfortunately, the sub-chart doesn't implement a way to pass in extra args for the `all-in-one` mode. I still think doing Jaeger as a sub-chart is a good idea, but now I need to figure out how to PR a change into the Jaeger helm chart.

PR #516, opened 2022-11-23
[demo] feat: add "env" to obserbability.jaeger

tglanz:
Addresses #515

### DIFF
--- a/charts/opentelemetry-demo/Chart.yaml
+++ b/charts/opentelemetry-demo/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: opentelemetry-demo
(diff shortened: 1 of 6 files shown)


dmitryax:
Let's hold off on this PR until we have a decision in https://github.com/open-telemetry/opentelemetry-helm-charts/issues/515 to avoid introducing more problems if we decide to switch to the sub-chart

puckpuck:
#523 is moving to a sub-chart that will allow you to configure Jaeger to run in many different forms. Let's hold off on this pending that PR.

TylerHelmuth:
Closing in favor of https://github.com/open-telemetry/opentelemetry-helm-charts/pull/523.

PR #517, opened 2022-11-24
[opentelemetry-collector] Bump collector to 0.66.0

povilasv:
Ref: #518 

I did some testing with 0.66.0 and went thru https://github.com/open-telemetry/opentelemetry-collector/releases/tag/v0.66.0 notes everything seems okay, didn't see breaking changes

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.39.3
+version: 0.40.0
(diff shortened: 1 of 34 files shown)


charts/opentelemetry-collector/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.39.3
+version: 0.39.4
```
TylerHelmuth, line 3: @povilasv We normally bump the minor version when we do the collector image bump since the collector itself does a minor version bump.

TylerHelmuth (approved):

PR #67, opened 2021-08-13
Defined configuration for headless service

kurumoorthi89:
Defined configuration for headless service in open-telemetry helm charts. 

This allows us to directly interact with PODs instead of proxy.

Co-authored-by: Betsy Behun <betsybehun@gmail.com>

closes https://github.com/open-telemetry/opentelemetry-helm-charts/issues/61

PR #519, opened 2022-11-24
[ opentelemetry-collector ]fix schema validation headless service

anavarreteuk:
While trying to enable the service on a [headless](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services) mode, the values schema validation was blocking the possibility of passing additional properties. 

This PR allows us to pass the property "None" to the clusterIP key. 
Relates to https://github.com/open-telemetry/opentelemetry-helm-charts/pull/67

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.40.0
+version: 0.41.0
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.41.0
+version: 0.40.1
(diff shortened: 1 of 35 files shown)


charts/opentelemetry-collector/values.schema.json
```diff
@@ -493,7 +493,7 @@
     },
     "service": {
       "type": "object",
-      "additionalProperties": false,
+      "additionalProperties": true,
```
TylerHelmuth, line 496: Instead of setting this to true please add "None" to the type Enum list below.

TylerHelmuth (commented):
Please bump the patch version of the chart and then run "make generate-examples"

charts/opentelemetry-collector/values.schema.json
anavarreteuk: fixed, thanks.

charts/opentelemetry-collector/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.40.0
+version: 0.41.0
```
TylerHelmuth, line 3: Please change this to be 0.40.1 instead of 0.41.0. This change only needs the patch number to be changed. Also please run "make generate-examples".

TylerHelmuth (approved):

PR #521, opened 2022-11-25
[ opentelemetry-collector ] add additional property for clusterIP none

anavarreteuk:
Apologies but [this](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/519#discussion_r1031794877) is not working as expected; I have added another key into the schema validation, allowing the string "None" to be pass to clusterIP key.

### DIFF
--- a/charts/opentelemetry-collector/values.schema.json
+++ b/charts/opentelemetry-collector/values.schema.json
@@ -501,8 +501,13 @@
             "ClusterIP",
             "NodePort",
             "LoadBalancer",
-            "ExternalName",
-            "None"
+            "ExternalName"
+          ]
+        },
+        "clusterIP": {
+          "type": "string",
+          "enum": [
+           "None" 
           ]
         },
         "annotations": {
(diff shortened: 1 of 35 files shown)


charts/opentelemetry-collector/values.schema.json
```diff
@@ -501,8 +501,13 @@
...
             "NodePort",
             "LoadBalancer",
-            "ExternalName",
-            "None"
+            "ExternalName"
+          ]
+        },
+        "clusterIP": {
+          "type": "string",
+          "enum": [
```
TylerHelmuth, line 509: Instead of an Enum let's just enforce that clusterIP is a string.

TylerHelmuth (commented):
@anavarreteuk Thanks for following up on this fix, that's what I get for reviewing on my phone during thanksgiving.

The service template is already setup to use this field, so adding it to the schema will make it actually usable. Thanks!

TylerHelmuth (approved):

TylerHelmuth (approved):

PR #522, opened 2022-11-26
[demo] update sub-chart dependencies

puckpuck:
This updates Collector, Prometheus, and Grafana sub-chart dependencies.

The Prometheus chart made changes to how they implement the node exporter and push gateway. Those name updates are reflected in values.yaml.


TylerHelmuth (commented):
@puckpuck needs a version bump/example regeneration since that other PR was merged first.

TylerHelmuth (approved):

PR #523, opened 2022-11-26
[demo] use Jaeger sub-chart

puckpuck:
Fixes: #515 

Moves to using a Jaeger sub-chart instead of templated within this chart itself.

This does introduce a potential breaking change, where the Jaeger parameters are moving to a top-level `jaeger.` section instead. Even though this is a breaking change, the chart is early enough, and the current section isn't likely to of been modified by existing deployments. 

A readme is also added to the chart to document the upgrade process and all parameters in the chart itself. 

Closing this would render #516 no longer required.


puckpuck:
This one is ready for review now

julianocosta89 (approved):
Not an approver here, but it LGTM 🚀

puckpuck:
I think we can certainly entertain using helm-docs to generate Docs instead.  For the demo chart in particular, because we have a section that repeats for each service component, it would take time to create the proper go template for the output docs and would prefer to defer that effort to another PR.

TylerHelmuth:
That's true. I don't think it should block this PR, but something to think about.

TylerHelmuth:
@dmitryax please review.

dmitryax (approved):
LGTM. Thanks for adding this!

PR #531, opened 2022-11-28
Specify grpc endpoint in otlp receiver config

### LINKED ISSUE #530: OTel Demo App Helm Chart Install with specified values.yaml was not sending traces to Jaeger
**Steps taken:**

Ran the OpenTelemetry Demo App Helm Chart with `values.yaml` as follows:

```
helm install my-otel-demo open-telemetry/opentelemetry-demo -n otel-demo --create-namespace -f charts/opentelemetry-demo/values.yaml
```

Note that this was [values.yaml](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-demo/values.yaml) from the helm repo itself - I did not make any modifications to this file.

Set up k8s port-forwarding to access the Jaeger UI:

```
kubectl port-forward svc/my-otel-demo-frontendproxy 8080:8080
```

Opened up the Jaeger UI at http://localhost:8080/jaeger/ui

**Actual result:**
Did not see any services of the OTel Demo App listed in the Jaeger UI at http://localhost:8080/jaeger/ui.

**Expected result:**
I expected to see 12 services listed in the Jaeger UI.

I would have expected that:

```
helm install my-otel-demo open-telemetry/opentelemetry-demo -n otel-demo --create-namespace -f charts/opentelemetry-demo/values.yaml
```

and 

```
helm install my-otel-demo open-telemetry/opentelemetry-demo -n otel-demo --create-namespace
```

Would have yielded the same results, since they are both using the same `values.yaml` file.

avillela:
Fixes #530 

To fix this, I first rendered the k8s manifests for each scenario as follows:

```
helm template my-otel-demo open-telemetry/opentelemetry-demo -n otel-demo -f charts/opentelemetry-demo/values.yaml > otel-demo.yaml
```
and

```
helm template my-otel-demo open-telemetry/opentelemetry-demo -n otel-demo > otel-demo-2.yaml
```

I ran a diff of the 2 files, and noticed that `otel-demo.yaml` was missing the OTLP gRPC endpoint configuration:

```yaml
      otlp:
        protocols:
          grpc:
            endpoint: 0.0.0.0:4317
```

After adding this configuration to `values.yaml`, it appeared to fix the issue.

### DIFF
--- a/charts/opentelemetry-demo/values.yaml
+++ b/charts/opentelemetry-demo/values.yaml
@@ -578,8 +578,6 @@ opentelemetry-collector:
     receivers:
       otlp:
         protocols:
-          grpc:
-            endpoint: 0.0.0.0:4317          
           http:
             cors:
               allowed_origins:
--- a/charts/opentelemetry-demo/values.yaml
+++ b/charts/opentelemetry-demo/values.yaml
@@ -579,6 +579,7 @@ opentelemetry-collector:
       otlp:
         protocols:
           grpc:
+            endpoint: 0.0.0.0:4317          
           http:
             cors:
               allowed_origins:
--- a/charts/opentelemetry-demo/Chart.yaml
+++ b/charts/opentelemetry-demo/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: opentelemetry-demo
-version: 0.12.4
+version: 0.12.5
 description: opentelemetry demo helm chart
 home: https://opentelemetry.io/
 sources:
--- a/charts/opentelemetry-demo/examples/default/rendered/component.yaml
+++ b/charts/opentelemetry-demo/examples/default/rendered/component.yaml
@@ -5,7 +5,7 @@ kind: Service
 metadata:
   name: example-adservice
   labels:
-    helm.sh/chart: opentelemetry-demo-0.12.4
+    helm.sh/chart: opentelemetry-demo-0.12.5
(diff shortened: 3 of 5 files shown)


charts/opentelemetry-demo/values.yaml
```diff
@@ -579,6 +579,7 @@ opentelemetry-collector:
       otlp:
         protocols:
           grpc:
+            endpoint: 0.0.0.0:4317
```
TylerHelmuth, line 582: This is actually already the default value for the grpc object in the collector chart: https://github.com/open-telemetry/opentelemetry-helm-charts/blob/22bfae673b305d73ad99815d38bc16aff6bec953/charts/opentelemetry-collector/values.yaml#L85

I actually think we can get away with removing the `grpc` section and rely on the default collector chart config.  @avillela can you try that out?

avillela: Yeah, that did the trick. Removing it altogether caused it to get picked up from the collector chart. I think maybe because the original entry was:

```
otlp:
    protocols:
        grpc:
```

It probably overrode the inherited value. Config has been updated.

TylerHelmuth (commented):
@avillela please bump the patch version of the demo chart and then run `make generate-examples`.

avillela:
> @avillela please bump the patch version of the demo chart and then run `make generate-examples`.

@TylerHelmuth done!

TylerHelmuth (approved):

PR #532, opened 2022-11-29
Add support to pass extraContainers for sidecars

ekarlso:


charts/opentelemetry-collector/values.yaml
```diff
@@ -284,6 +284,8 @@ lifecycleHooks: {}
 #       - /test/sleep
 #       - "5"
 
+extraContainers: []
+
```
dmitryax, line 288: Can you please move it closer to `initContainers`?

dmitryax (commented):
Please bump the version and run `make generate-examples`

dmitryax:
Also values.schema.json needs to be updated

charts/opentelemetry-collector/values.yaml
ekarlso: Done.

ekarlso: Done!

TylerHelmuth (approved):

dmitryax (approved):

PR #533, opened 2022-11-30
[opentelemetry-collector] add hostmetrics filesystem config

povilasv:
I've played around filesystem filtering for hostmetrics, and I kinda like this "generic" solution, where users can include / exclude whatever they see fit

Superseeds https://github.com/open-telemetry/opentelemetry-helm-charts/pull/496

Any thoughts around this implementation?

cc @oded-dd

Tested with:
```
presets:
  hostMetrics:
    enabled: true
    filesystem:
     include_fs_types:
       fs_types: [ "ext4","xfs" ]
       match_type: strict
     exclude_mount_points:
       mount_points:
        - /snap/*
        - /boot/*
        - /var/lib/kubelet/pods/*
        - /var/lib/kubelet/plugins/*
       match_type: regexp
```

Gives:

```
    receivers:
      hostmetrics:
        collection_interval: 10s
        scrapers:
          cpu: null
          disk: null
          filesystem:
            exclude_mount_points:
              match_type: regexp
              mount_points:
              - /snap/*
              - /boot/*
              - /var/lib/kubelet/pods/*
              - /var/lib/kubelet/plugins/*
            include_fs_types:
              fs_types:
              - ext4
              - xfs
              match_type: strict
          load: null
          memory: null
          network: null
      jaeger:
        protocols:

```

### DIFF
--- a/charts/opentelemetry-collector/values.yaml
+++ b/charts/opentelemetry-collector/values.yaml
@@ -31,6 +31,17 @@ presets:
   # Best used with mode = daemonset.
   hostMetrics:
     enabled: false
+    # adding filesystem field allows you to include and exclude filesystems
+    # filesystem:
+    #   include_devices:
+    #     devices: [ <device name>, ... ]
+    #     match_type: <strict|regexp>
+    #   include_fs_types:
+    #     fs_types: [ <filesystem type>, ... ]
+    #     match_type: <strict|regexp>
+    #   include_mount_points:
+    #     mount_points: [ <mount point>, ... ]
+    #     match_type: <strict|regexp>
   # Configures the Kubernetes Processor to add Kubernetes metadata.
   # Adds the k8sattributes processor to all the pipelines
   # and adds the necessary rules to ClusteRole.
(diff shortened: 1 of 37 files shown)


charts/opentelemetry-collector/values.yaml
```diff
@@ -31,6 +31,17 @@ presets:
...
+    # filesystem:
+    #   include_devices:
+    #     devices: [ <device name>, ... ]
+    #     match_type: <strict|regexp>
+    #   include_fs_types:
+    #     fs_types: [ <filesystem type>, ... ]
+    #     match_type: <strict|regexp>
+    #   include_mount_points:
+    #     mount_points: [ <mount point>, ... ]
+    #     match_type: <strict|regexp>
```
TylerHelmuth, line 44: My initial reaction is that this capability is counter to the goal of a preset which is to allow users to enable features without having to have deep k8s/component understanding.  Requiring the user to specify the include/exclude points and know the path names feels like too much knowledge requirement for a preset.

I'd prefer a solution that is more like a switch: "if you enable this then you get X".  Since it is a preset we can make assumptions about what people need and limit the switch to only those needs.  If people require more they can manually configure the component. @povilasv @dmitryax what do you think?

TylerHelmuth: If we don't have a good set of assumptions to use to pre-define a list of includes/excludes it may mean that this isn't actually suitable for a preset and users should configure the component themselves. @dmitryax.

Issue #525, opened 2022-11-27
[demo] Add Ingress for frontend proxy


puckpuck:
An ingress for the frontend proxy would allow people to easily expose the demo beyond having to use `kubectl port-fortward` or a LoadBalancer service which both may be restrictive in many deployment scenarios.

I'm willing to do the effort, but would like to discuss what is the best way to define the parameters to add this. Should it be done as a new `ingress` section or as part of the `components` section (applicable to all components)?

PR #540, opened 2022-12-06
[demo] - add ingress 

puckpuck:
Fixes: #525 

Adds capability to specify Ingress(es) for each component. This follows the same pattern used for the collector chart.

The following config will create an Ingress using the default ingress controller.  As with the collector chart, annotations are also supported.

```
components:
  frontendProxy:
    ingress:
      enabled: true
      hosts:
        - host: otel-demo.example.com
          paths:
            - path: /
              pathType: Prefix
              port: 8080
```


puckpuck:
Linters are happy, this one is ready for review now

dmitryax (approved):
LGTM. @puckpuck please rebase

charts/opentelemetry-demo/templates/_helpers.tpl
```diff
@@ -57,3 +57,30 @@ app.kubernetes.io/component: {{ .name}}
 {{- $mergedEnvs = concat $mergedEnvs $envOverrides }}
 {{- mustToJson $mergedEnvs }}
 {{- end }}
+
+{{/*
+Return the appropriate apiVersion for ingress.
+*/}}
+{{- define "ingress.apiVersion" -}}
+  {{- if and (.Capabilities.APIVersions.Has "networking.k8s.io/v1") (semverCompare ">= 1.19-0" .Capabilities.KubeVersion.Version) -}}
```
TylerHelmuth, line 65: Is this to support old versions of k8s?  I'd rather the chart target only versions of k8s that the k8s team are still supporting.

charts/opentelemetry-demo/templates/_objects.tpl
```diff
@@ -82,3 +83,72 @@ spec:
     {{- include "otel-demo.selectorLabels" . | nindent 4 }}
 {{- end}}
 {{- end}}
+
+{{- define "otel-demo.ingress" }}
```
TylerHelmuth, line 87: I think there is enough logic here to warrant a new CI scenario that tests the chart with ingress enabled.  The collector chart does the same.

charts/opentelemetry-demo/templates/_helpers.tpl
puckpuck: yeah, the code was copied from the Collector chart.  I'm happy to simplify this so we only support recent versions of K8s.  What would be the min version we are targeting for support? 1.23?

TylerHelmuth: Dang collector chart, the maintainers of that thing should really up their game lol

I believe 1.23 is the minimum supported version ya.

charts/opentelemetry-demo/templates/_objects.tpl
puckpuck: Added 2 new CI scenarios, and 1 example

TylerHelmuth (approved):

PR #545, opened 2022-12-07
[demo] add securityContext and mandatory named users

styblope:
Part of [#615](https://github.com/open-telemetry/opentelemetry-demo/pull/615).

Added UIDs to the following non-root containers:
* postgres
* frontendproxy
* redis
* quoteservice
* kafka


TylerHelmuth (commented):
@styblope Thanks! please bump the demo's patch version and then run `make generate-examples`

puckpuck:
This will require the demo 1.2 images to be released before all of this will work.

puckpuck:
@styblope now that the Demo 1.2.x images are released can you make sure this change incorporates proper securityContext for all components?

puckpuck:
In an effort to document these parameters we need this to include:

1. In `values.yaml` at the top of the `components:` section is a commented example of a component. Please add this to that example with a short comment
2. document the new parameters in the `Component Parameters` section of the chart's `README.md` file

TylerHelmuth (approved):

puckpuck (approved):

PR #551, opened 2022-12-09
[opentelemetry-collector] add service.loadbalancerIP

Stelminator:
I know it's deprecated because it doesn't support dualstack, but, that doesn't mean it doesn't still work, and afaik, there isn't yet an alternative.

https://github.com/kubernetes/kubernetes/pull/107235

I'm targeting GKE, and have confirmed that this works as expected in that environment for IPV4 reserved external regional addresses.


Stelminator:
@puckpuck 

> Can you add a commented entry for this in values.yaml with how to use it?

is this about what you had in mind? https://github.com/open-telemetry/opentelemetry-helm-charts/pull/551/commits/035fce99c14e9cbc58477176699a1a4325837886

```
  # type: LoadBalancer
  # loadBalancerIP: 1.2.3.4
```

puckpuck (approved):

TylerHelmuth (approved):

PR #552, opened 2022-12-09
[demo] - release 1.2.1

puckpuck:
The demo 1.2 release had some issues, that needed to get fixed.  This picks up those changes (v1.2.1)

Also cleans up a few small things that were causing pain (exasperated by the issue in the demo 1.2 release).

- Prometheus will now only scrape in its own namespace
- Quote service properly routes OTLP data
- Jaeger routing from frontendproxy is fixed
- Adjusts memory limits for several components

This PR depends on the Demo 1.2.1 release images being created.


TylerHelmuth (approved):
👍

PR #555, opened 2022-12-10
Fix Grafana datasource for Jaeger.

philipz:
Fixed the datasource URL for the Jaeger within Grafana configmap, missed /jaeger/ui.

### DIFF
--- a/charts/opentelemetry-demo/values.yaml
+++ b/charts/opentelemetry-demo/values.yaml
@@ -737,7 +737,7 @@ grafana:
         - name: Jaeger
           uid: webstore-traces
           type: jaeger
-          url: 'http://{{ include "otel-demo.name" . }}-jaeger:16686'
+          url: 'http://{{ include "otel-demo.name" . }}-jaeger:16686/jaeger/ui'
           editable: true
           isDefault: false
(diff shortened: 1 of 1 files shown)


charts/opentelemetry-demo/values.yaml
```diff
@@ -737,7 +737,7 @@ grafana:
         - name: Jaeger
           uid: webstore-traces
           type: jaeger
-          url: 'http://{{ include "otel-demo.name" . }}-jaeger:16686'
+          url: 'http://{{ include "otel-demo.name" . }}-jaeger:16686/jaeger/ui'
```
puckpuck, line 740: This should be 
```
          url: 'http://{{ include "otel-demo.name" . }}-jaeger-query:16686'
```

When we moved to use the Jaeger sub-chart, the query service name changed. We had to do a similar change for other spots, and missed this one.

philipz: @puckpuck You are right, the svc name should be "jaeger-query". But, if the url missing "/jaeger/ui" path, it will not work out of the box.

philipz:
The svc name should be "jaeger-query".

PR #556, opened 2022-12-10
Fix Grafana datasource for Jaeger.

philipz:
Fixed the datasource URL for the Jaeger within Grafana configmap.

### DIFF
--- a/charts/opentelemetry-demo/values.yaml
+++ b/charts/opentelemetry-demo/values.yaml
@@ -737,7 +737,7 @@ grafana:
         - name: Jaeger
           uid: webstore-traces
           type: jaeger
-          url: 'http://{{ include "otel-demo.name" . }}-jaeger:16686'
+          url: 'http://{{ include "otel-demo.name" . }}-jaeger:16686/jaeger/ui'
           editable: true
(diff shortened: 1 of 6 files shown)


TylerHelmuth:
@philipz Please bump the patch version of the chart and then run "make generate-examples".

charts/opentelemetry-demo/values.yaml
```diff
@@ -737,7 +737,7 @@ grafana:
         - name: Jaeger
           uid: webstore-traces
           type: jaeger
-          url: 'http://{{ include "otel-demo.name" . }}-jaeger:16686'
+          url: 'http://{{ include "otel-demo.name" . }}-jaeger-query:16686/jaeger/ui'
```
TylerHelmuth, line 740: @puckpuck can you confirm this is needed?

puckpuck: @TylerHelmuth I just tested, and your suggestion above is what works: https://otel-demo.honeydemo.io/grafana/datasources/edit/webstore-traces

TylerHelmuth (approved):

PR #558, opened 2022-12-12
[operator] fix failurePolicy conf inconsistency

JaredTan95:


Allex1:
@JaredTan95 Thanks for your contribution. 
I think the inconsistencies stem from the way the policies are defined in the [operator repo](https://github.com/open-telemetry/opentelemetry-operator/blob/main/bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml#L450). I think we should align between the 2 repos.

JaredTan95:
> I think the inconsistencies stem from the way the policies are defined in the [operator repo](https://github.com/open-telemetry/opentelemetry-operator/blob/main/bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml#L450). I think we should align between the 2 repos.

ok, make sense.

PR #561, opened 2022-12-12
[opentelemetry-demo] Add ability to configure & command components.

ancostas:
This PR allows a user to set the following in `values.yaml` (as an example):
```
    redis:
      command:
        - /usr/local/bin/redis-server
        - /etc/config/redis.conf
      configuration:
        redis.conf: |
          <redis configuration goes here>
```

And generate the appropriate ConfigMap, volume/mounts, and command to make use of it.

I can separate the PR into one for the ability to add a command, and a separate one for the configuration setup, if so desired. The design is simple enough that I didn't think it worth it right now, but the configuration part of this may warrant some constructive criticism 😅 .


charts/opentelemetry-demo/values.yaml
```diff
@@ -583,7 +583,7 @@ components:
     enabled: true
     useDefault:
       env: true
-
+    serviceType: ClusterIP
```
ancostas, line 586: This was a minor cleanup, I think it was missing accidentally.

puckpuck: we have another PR to remove this from being required on all components (it will default to ClusterIP).

ancostas: ok, I'll prune it then.

charts/opentelemetry-demo/values.yaml
```diff
@@ -598,7 +597,8 @@ components:
       nodeSelector: {}
       affinity: {}
       tolerations: []
-
+    command: []
```
puckpuck, line 600: When #557 lands, this would make sense to be included in the sample documented component at the top of the components section instead of inline within just the redis component.

puckpuck:
Can you add a reference to the new parameters in the `Component Parameters` section of the chart's readme: https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-demo/README.md#component-parameters

charts/opentelemetry-demo/values.yaml
puckpuck: #557 is merged now. After a rebase, please move these 2 parameters into the commented example at the top of the components section, adding comments about how to use them.

ancostas:
@puckpuck sounds good! I'll do so.

charts/opentelemetry-demo/templates/_objects.tpl
```diff
@@ -83,6 +97,21 @@ spec:
     {{- include "otel-demo.selectorLabels" . | nindent 4 }}
 {{- end}}
 {{- end}}
+{{- define "otel.demo.configmap" }}
```
puckpuck, line 100: use otel-demo instead of otel.demo for the name here to stay consistent with other names.

```suggestion
{{- define "otel-demo.configmap" }}
```

charts/opentelemetry-demo/templates/component.yaml
```diff
@@ -12,6 +12,7 @@
         {{- include "otel-demo.deployment" $config -}}
         {{- include "otel-demo.service" $config -}}
         {{- include "otel-demo.ingress" $config -}}
+        {{- include "otel.demo.configmap" $config -}}
```
puckpuck, line 15: use otel-demo instead of otel.demo for the name here to stay consistent with other names.

```suggestion
        {{- include "otel-demo.configmap" $config -}}
```

puckpuck (changes requested):
Can you fix the 2 `otel-demo` names to be consistent with other defines, then I can approve.

puckpuck (approved):
I updated the defined names to be `otel-demo` instead of `otel.demo`.  The rest looks great.  Approved!

puckpuck:
We can do it as a separate PR, but we should add a ci test and rendered example showcasing these new parameters.

TylerHelmuth (approved):

ancostas:
@puckpuck Sounds good, thank you! I have a card on my board to add some tests at some point in a separate PR since I think also the tolerations/affinity fix I did was since we only test default config right now.

PR #565, opened 2022-12-14
[opentelemetry-collector] Remove old ingress logic

TylerHelmuth:
Removes old ingress logic that was checking for k8s versions older than 1.19.


puckpuck (approved):

puckpuck:
Should we add a min K8s version to the chart's readme?

TylerHelmuth:
> Should we add a min K8s version to the chart's readme?

Ya for each chart.  The operator chart may need to target 1.19+ since the operator targets that.

PR #570, opened 2022-12-20
[opentelemetry-collector] Add ability to template initcontainers.

ancostas:
I have a use case where I need to interpolate global values into my initcontainer, so I've added the ability to template it. I think this is backwards compatible.


Allex1 (approved):
lgtm

puckpuck (approved):

puckpuck:
@ancostas you'll need to update and generate-examples, then we can merge this.

ancostas:
@puckpuck Done!

TylerHelmuth (commented):
@ancostas to keep the number of examples to a minimum, can you merge your new lifecycle-hooks-with-globals example with the existing lifecycle-hookes example?

ancostas:
@TylerHelmuth sure thing! I've also ignored IntelliJ IDE config in case any future contributors use that.

TylerHelmuth (approved):

Issue #571, opened 2022-12-23
Improve otel demo's metric monitoring.


wph95:
# Motivation

The current demo chart will install prometheus and collect metric for some components.
However, it does not collect k8s metric.

I've done some local test and not found that a simple modification of values.yaml collect k8s metric.

# PLAN

So I want to solve the problem with the following goal.
1. the user only needs `prometheus.kube-state-metrics: enable` to let prometheus collect k8s metric (currently we have custom scrape_configs, so just enable configuration can't collect data)

2. the user can collect container information in a simple way (via cadvisor)

Saki042:
@TylerHelmuth , Can I more info about the issue and is it open to work?

TylerHelmuth:
We tend to let the chart follow features driven by the demo sig, closing as Not Planned for now

PR #572, opened 2022-12-23
improve prometheus scrape

wph95:
part of #571 

When I just modify value.yaml to enable `prometheus-node-exporter` and `kube-state-metrics`. prometheus doesn't have the data I expected.
After a brief research, I found out that the prometheus helm chart is scrape through service annotation.
https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus/values.yaml#L764-L803

So submit this PR
1. prometheus defaults to scrape via service annotation
2. modify the otel-col configuration so that it is configured with the correct service annotation
3. demo component supports serviceAnnotations

### DIFF
--- a/charts/opentelemetry-demo/values.yaml
+++ b/charts/opentelemetry-demo/values.yaml
@@ -56,6 +56,8 @@ components:
   #   serviceType: ClusterIP
   ## Service Port to use to expose this component. Default is nil
   #   servicePort: 8080
+  ## Service Annotations to add to this component
+  #   serviceAnnotations: {}
   ## Additional service ports to use to expose this component
   #   ports:
   #     - name: extraServicePort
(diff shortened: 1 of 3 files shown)


charts/opentelemetry-demo/values.yaml
```diff
@@ -558,10 +563,6 @@ opentelemetry-collector:
       containerPort: 9464
       servicePort: 9464
       protocol: TCP
-  podAnnotations:
-    prometheus.io/scrape: "true"
-    prometheus.io/port: "9464"
-    opentelemetry_community_demo: "true"
```
TylerHelmuth, line 564: Lets leave this annotation

TylerHelmuth (commented):
@wph95 please rebase, bump the chart's patch version, and then run `make generate-examples`

PR #575, opened 2022-12-28
[opentelemetry-demo] Fix schedulingRules handling for demo components

tonyo:
The problem we're fixing here is the following: after change https://github.com/open-telemetry/opentelemetry-helm-charts/pull/557, the `{{- if .schedulingRules }}` check yields "false" by default, which means that configuring scheduling rules via `default` map in chart values (i.e., having non-empty `default.schedulingRules`) is no longer working (all [logic here](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-demo/templates/_objects.tpl#L30-L41) is simply ignored).

This PR removes the `{{- if .schedulingRules }}` check, but also introduces a temporary variable `$schedulingRules` that is always meant to be non-null.

### DIFF
--- a/charts/opentelemetry-demo/Chart.yaml
+++ b/charts/opentelemetry-demo/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: opentelemetry-demo
-version: 0.15.2
+version: 0.15.3
 description: opentelemetry demo helm chart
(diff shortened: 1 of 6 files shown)


puckpuck (approved):
Thanks for this

dmitryax (approved):

Issue #576, opened 2022-12-29
[collector] expose 4317 port by grpc protocol enhancement.


JaredTan95:
In the Istio turn on OTEL and send data to the OTEL-Col experiment, we found that the TCP protocol of OTEL-Col will cause Istio could not to push tracing data normally.

more details can be following: https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/#explicit-protocol-selection

so, I would like to polish the service expose logic in opentelemetry-collecotr chart.

PR #577, opened 2022-12-29
[opentelemetry-collector] add grpc-otlp prot

JaredTan95:
Signed-off-by: Jared Tan <jian.tan@daocloud.io>

close https://github.com/open-telemetry/opentelemetry-helm-charts/issues/576

### DIFF
--- a/charts/opentelemetry-collector/examples/daemonset-and-deployment/rendered/service.yaml
+++ b/charts/opentelemetry-collector/examples/daemonset-and-deployment/rendered/service.yaml
@@ -5,7 +5,7 @@ kind: Service
 metadata:
   name: example-opentelemetry-collector
   labels:
-    helm.sh/chart: opentelemetry-collector-0.43.2
+    helm.sh/chart: opentelemetry-collector-0.43.3
     app.kubernetes.io/name: opentelemetry-collector
     app.kubernetes.io/instance: example
     app.kubernetes.io/version: "0.67.0"
@@ -15,6 +15,10 @@ spec:
   type: ClusterIP
   ports: 
     
+    - name: grpc-otlp
+      port: 4317
+      targetPort: 4317
+      protocol: TCP
     - name: jaeger-compact
       port: 6831
       targetPort: 6831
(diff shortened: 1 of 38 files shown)


charts/opentelemetry-collector/values.yaml
```diff
@@ -199,7 +199,7 @@ extraVolumeMounts: []
 
 # Configuration for ports
 ports:
-  otlp:
+  grpc-otlp:
```
Allex1, line 202: This should probably be called `otlp-grpc` in order to be consistent with `otlp-http`

JaredTan95: no, it must be `grpc-xxx` as prefix:https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/#explicit-protocol-selection

JaredTan95: if we do not use `grpc-otlp`, we can use `appProtocol`

puckpuck: I would opt to use `appProtocol` here instead.

charts/opentelemetry-collector/examples/daemonset-and-deployment/rendered/service.yaml
```diff
@@ -27,10 +31,6 @@ spec:
       port: 14268
       targetPort: 14268
       protocol: TCP
-    - name: otlp
-      port: 4317
-      targetPort: 4317
-      protocol: TCP
```
dmitryax, line 33: Isn't it just a renaming since we already have the otlp port here?

dmitryax: Nvm. I see it's not the only change

dmitryax (approved):

charts/opentelemetry-collector/values.yaml
dmitryax: This is actually a breaking change for the users. So I would keep the existing name. otherwise, the renaming should be done in a backward compatible way

dmitryax (changes requested):
Just to discard my previous approval

JaredTan95:
okay, I will update by `appProtocol `

puckpuck:
This is already merged, but should we a specify default `appProtocol` in values.yaml as part of a separate PR?

dmitryax:
I though the rename was reverted and merged accidentally. @puckpuck can you please take a look on the revert of the renaming? https://github.com/open-telemetry/opentelemetry-helm-charts/pull/578

PR #582, opened 2023-01-09
[BUMP] - operator 0.67/collector 0.68

Allex1:
Upgrade main components


Allex1:
fails bc of https://github.com/open-telemetry/opentelemetry-operator/blob/main/tests/e2e/route/00-assert.yaml#L7

Issue #585, opened 2023-01-09
[opentelemetry-collector] Cannot set clusterRole


kedare:
Hello,

When trying to install the following helm chart

```sh
helm install otlp open-telemetry/opentelemetry-collector --values staging.values.yaml
```

With the following minimal values:

```yaml
mode: "daemonset"

presets:
  kubernetesAttributes:
    enabled: true

clusterRole:
  create: true
  name: otlp-cluster-role

  clusterRoleBinding:
    name: otlp-cluster-binding
```

(to test, I have more values, commented)

I get the following error message:

```
Error: INSTALLATION FAILED: YAML parse error on opentelemetry-collector/templates/clusterrole.yaml: error converting YAML to JSON: yaml: line 12: did not find expected key
```

Also when I don't specify the `clusterRole`, I get one created with the name `default`, same for the `clusterRoleBinding`, I don't think this is expected ?

versions:

```
helm
version.BuildInfo{Version:"v3.7.0", GitCommit:"eeac83883cb4014fe60267ec6373570374ce770b", GitTreeState:"clean", GoVersion:"go1.16.8"}

kubectl
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3", GitCommit:"1e11e4a2108024935ecfcb2912226cedeafd99df", GitTreeState:"clean", BuildDate:"2020-10-14T12:50:19Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.4", GitCommit:"d360454c9bcd1634cf4cc52d1867af5491dc9c5f", GitTreeState:"clean", BuildDate:"2020-11-13T00:17:00Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"linux/amd64"}

chart:
open-telemetry/opentelemetry-collector  0.43.5          0.67.0          OpenTelemetry Collector Helm chart for Kubernetes
```

TylerHelmuth:
@kedare the issue is happening because you specified `clusterRole.create: true` but didn't provide any rules.  The template is make an assumption that if you specified to create a cluster role then you also supplied rules

https://github.com/open-telemetry/opentelemetry-helm-charts/blob/972673377767d3df1de322d4797f862068432d19/charts/opentelemetry-collector/templates/clusterrole.yaml#L12-L15

[...]

PR #586, opened 2023-01-09
[opentelemetry-collector] Update clusterrole and clusterrolebinding naming

TylerHelmuth:
Fixes https://github.com/open-telemetry/opentelemetry-helm-charts/issues/585

- Updates cluster role and cluster role binding naming logic to always default to fullname, even when a preset is creating the objects.  
- Updates clusterrole to only attempt rendering rules if there are rules preset.  
- Allow rendering rules even if `clusterRole.create` is not `true`.  This allows adding extra rules to cluster roles created by presets.
- Several cluster role and cluster role binding doc updates.

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.44.0
+version: 0.45.0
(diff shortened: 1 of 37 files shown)


Allex1 (approved):

kedare:
Maybe also changes the comment on top of `clusterRole.create` as the current one is not true ? (it may be created even if it's `false`)

TylerHelmuth:
@kedare updated the comment

Issue #380, opened 2022-09-26
Add option to set the nodePort mapping.


ydan287:
When choosing NodePort kind, the port exposed is randomized from a range. in some runtimes (Kind for example) you need to explicitly do port forwarding:
```yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
  # port forward 30080 on the host to 30080 on this node
  extraPortMappings:
  - containerPort: 30080
    hostPort: 30080
- role: worker
```

This is impossible if the nodePort cannot be predetermined.

TylerHelmuth:
@ydan287 are you referring to the opentelemetry-collector chart's `ports` section?

```yaml
# Configuration for ports
ports:
  otlp:
    enabled: true
    containerPort: 4317
    servicePort: 4317
    hostPort: 4317
    protocol: TCP
  otlp-http:
    enabled: true
    containerPort: 4318
    servicePort: 4318
    hostPort: 4318
    protocol: TCP
  jaeger-compact:
    enabled: true
    containerPort: 6831
    servicePort: 6831
    hostPort: 6831
    protocol: UDP
  jaeger-thrift:
    enabled: true
    containerPort: 14268
    servicePort: 14268
    hostPort: 14268
    protocol: TCP
  jaeger-grpc:
    enabled: true
    containerPort: 14250
    servicePort: 14250
    hostPort: 14250
    protocol: TCP
  zipkin:
    enabled: true
    containerPort: 9411
    servicePort: 9411
    hostPort: 9411
    protocol: TCP
  metrics:
    # The metrics port is disabled by default. However you need to enable the port
    # in order to use the ServiceMonitor (serviceMonitor.enabled) or PodMonitor (podMonitor.enabled).
    enabled: false
    containerPort: 8888
    servicePort: 8888
    protocol: TCP
```

You would like to specify nodePort as well?

ydan287:
Yes, exactly. That should allow running the chart with Kind locally.

PR #598, opened 2023-01-12
[collector] Add option to set the nodePort mapping

mllab-nl:
Fixes https://github.com/open-telemetry/opentelemetry-helm-charts/issues/380


TylerHelmuth (commented):
@mllab-nl thanks for your contribution.  Modifying the schema.json will be necessary for resolving the linked issue, but you'll also need to update the values.yaml and templates.

Once everything is implemented, please bump the patch version of the chart and then run `make generate-examples CHARTS=opentelemetry-collector`

mllab-nl:
@TylerHelmuth nodePort is only make sense when you use NodePort service type and only when you specifically want to. 
What changes would you expect to values.yaml and templates? 

I will bump up the patch version

TylerHelmuth:
@mllab-nl I believe for the new `nodePort` field to be used you'll need to update https://github.com/open-telemetry/opentelemetry-helm-charts/blob/808ed313d4283f2bfe123eea12cb0a67691e21a9/charts/opentelemetry-collector/templates/_config.tpl#L295 to set the `nodePort`.  That template is used by service.yaml to set the service's ports.

Lets also add a comment in the values.yaml's port section that specifies that `nodePort` is an allowed field.

mllab-nl:
Thx. For guidance. Done. 
Bumped the version again. 

What is your approach to testing and examples - what is their primary role examples or tests  ?   
Did a test in charts\opentelemetry-collector\examples\deployment-otlp-traces\
Current config of the example renders the ClusrterIp service with Node port which is expected, but service is not functional - ClusterIP service with a NodePort.
IMO adding an additional example for NodePort only is a bit too much, since all those are always regenerated and checked in. 

Should the test stay or should I revert it ?

charts/opentelemetry-collector/examples/deployment-otlp-traces/values.yaml
```diff
@@ -1,6 +1,9 @@
 mode: deployment
 
 ports:
+  otlp:
+    enabled: true
+    nodePort: 30317
```
TylerHelmuth, line 6: @mllab-nl we use the rendered examples to help users with common situations or complicated configuration.  I don't think `nodePort` is going to be common nor is it a complicated field to use when needed so I don't think we need to update any examples.

For your local testing you should create a values.yaml with all the fields set that will test `nodePort` (you can change service type with `service.type` in the values.yaml)

Allex1:
@mllab-nl pls use a meaningful title/commit name

charts/opentelemetry-collector/examples/deployment-otlp-traces/values.yaml
mllab-nl: Removed. 
Local check confirm that with following config I get proper service with correct ports deployed:

> 
> 
> mode: deployment
> 
> ports:
>   otlp:
>     nodePort: 30117
>   otlp-http:    
>     nodePort: 30118
>   jaeger-compact:
>     enabled: false
>   jaeger-thrift:
>     enabled: false
>   jaeger-grpc:
>     enabled: false
>   zipkin:
>     enabled: false
> 
> service:
>   type: NodePort

mllab-nl:
@TylerHelmuth Updated the fork, resolved conflicts, tested  - Ok
Thank you !

Issue #69, opened 2021-08-23
Allow option to create opentelemetry-collector CR along with otel-operator helm installation


VineethReddy02:
I am currently evaluating opentelemetry-operator helm chart. I am stuck on an issue. I wanted to add the opentelemetry-collector custom resource along with otel-operator helm chart installation. 

The otel-col CR creation fails during installation with the below error:

```
Internal error occurred: conversion webhook for cert-manager.io/v1alpha2, Kind=Certificate failed: Post "https://cert-manager-webhook.cert-manager.svc:443/convert?timeout=30s": dial tcp 10.102.92.197:443: connect: connection refused
```

Because the CR creation triggers the mutating webhook request to the operator and on creating otel-col CR during installation the otel-operator is still in the container creating phase, so the CR creation is failing. Is there any solution to fix this issue? I want to create otelcol CR along with the otel operator installation. 

As a workaround, I am installing otel-col CR alongside otel-operator installation by adding helm `pre-install` webhook to create otel-col CR prior to the creation of webhooks and I am myself adding the otel-col default mode as deployment and labels that are configured using mutating webhook as below

```
apiVersion: opentelemetry.io/v1alpha1
kind: OpenTelemetryCollector
metadata:
  name: opentelemetry-collector
  labels:
    "app.kubernetes.io/managed-by": "opentelemetry-operator"
  annotations:
    "helm.sh/hook": "pre-install"
spec:
  mode: deployment
 ``` 

Are there any alternative solutions to do this? Is my workaround an appropriate fix?
cc @Saber-W @jpkrohling

Saber-W:
> Are there any alternative solutions to do this? Is my workaround an appropriate fix?

I think using pre-install mechnism is an appropriate workaround if you want to "create otelcol CR along with the otel operator installation". If you can accept Collector to be installed slight after the Operator's readiness, I think `kubectl wait` would also be a good fix.

VineethReddy02:
But how can I use `kubectl wait` inside helm chart installation? 

I'm adding a otel-col CR inside `/templates` directory to create a otel-col with provided config alongside the otel-operator installation.

jpkrohling:
If you install the CR before the hook is installed, your CR will not be validated and we might not normalize it (defaulter). Not sure how to do it with Helm, but you do need to create the CR after the operator is ready.

VineethReddy02:
[...]

PR #599, opened 2023-01-12
[operator] Updated webhooks to respect failurePolicy field

TylerHelmuth:
While experimenting with installing CRs using the operator chart I noticed that the webhooks weren't all following the values.yaml's `failurePolicy` field.  This PR updates both versions of the webhooks to respect the `failurePolicy` field

Related to https://github.com/open-telemetry/opentelemetry-helm-charts/issues/69


Allex1 (approved):

TylerHelmuth:
Per the [operator's webhook manifest](https://github.com/open-telemetry/opentelemetry-operator/blob/main/config/webhook/manifests.yaml) I left the following wehbooks as Ignored:
- mpod.kb.io
- vinstrumentationdelete.kb.io
- vopentelemetrycollectordelete.kb.io

Setting these 3 to failure can cause issues with install and uninstall. All other webhooks use Fail by default but can be set to Ignore if the user wished.

Allex1:
@TylerHelmuth this is not really related to https://github.com/open-telemetry/opentelemetry-helm-charts/issues/69 in this form right ?

TylerHelmuth:
@Allex1 its related in the sense that if we want to have the ability for the operator chart to create OpenTelemetryCollector and Instrumentation resources during install we need to be able to set the failurePolicy to `Ignore`. Realistically if we do that then the opt-in will do the setting for you, but I found the inconsistent use of `.Values.admissionWebhooks.failurePolicy` while researching that topic.

PR #602, opened 2023-01-13
[demo] Bump demo dependencies

TylerHelmuth:


charts/opentelemetry-demo/Chart.lock
```diff
@@ -1,15 +1,15 @@
 dependencies:
 - name: opentelemetry-collector
   repository: https://open-telemetry.github.io/opentelemetry-helm-charts
-  version: 0.40.7
+  version: 0.44.1
```
puckpuck, line 4: Should we set this to the latest just released?

```suggestion
  version: 0.45.0
```

puckpuck (approved):

charts/opentelemetry-demo/Chart.lock
TylerHelmuth: @puckpuck ya I'll get this PR updated.

TylerHelmuth:
@puckpuck since I'm on an M1 the featureflag service doesn't work so I can't get a full test run.  Are you able to checkout this branch and test?

puckpuck (approved):

PR #610, opened 2023-01-18
Fix minstrumentation.kb.io webhook with certmanager

Allex1:


charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook-with-cert-manager.yaml
```diff
@@ -18,6 +18,14 @@ webhooks:
         path: /mutate-opentelemetry-io-v1alpha1-instrumentation
     failurePolicy: Fail
     name: minstrumentation.kb.io
+    {{- if .Values.admissionWebhooks.namespaceSelector }}
```
TylerHelmuth, line 21: Does the other webhook file need the same changes?

Allex1: No, That's fine.

PR #611, opened 2023-01-20
[operator] Update operator-webhook.yaml

EmpireErwinKooi:
Problem I need to solve:
Change the kubernetes opentelemetry webhook name so that it is called before other webhooks, as kubernetes calls the webhooks in alphabetical order.

Issue:
- When using nameOverride, to set the operator name to, say "aa-opentelemetry-operator";
- When setting certManager.enabled to false (in my org, we don't use a cert-manager);
- When applying applying a `kind: Instrumentation` ```kubectl apply -f instr.yaml ```
I get this error:
```
Error from server (InternalError): error when creating "instr.yaml": Internal error occurred: failed calling webhook "minstrumentation.kb.io": failed to call webhook: Post "https://aa-opentelemetry-operator-webhook-service.tracing.svc:443/mutate-opentelemetry-io-v1alpha1-instrumentation?timeout=10s": x509: certificate is valid for opentelemetry-operator-webhook-service.tracing, opentelemetry-operator-webhook-service.tracing.svc, not aa-opentelemetry-operator-webhook-service.tracing.svc
```

Solution:
This pull request uses the actual `opentelemetry-operator.name` to generate the self-signed certificate.

### DIFF
--- a/charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook.yaml
+++ b/charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook.yaml
@@ -1,5 +1,5 @@
 {{- if and (.Values.admissionWebhooks.create) (not .Values.admissionWebhooks.certManager.enabled) }}
-{{- $altNames := list ( printf "%s.%s" "opentelemetry-operator-webhook-service" .Release.Namespace ) ( printf "%s.%s.svc" "opentelemetry-operator-webhook-service" .Release.Namespace ) -}}
+{{- $altNames := list ( printf "%s-webhook-service.%s" (include "opentelemetry-operator.name" .) .Release.Namespace ) ( printf "%s-webhook-service.%s.svc" (include "opentelemetry-operator.name" .) .Release.Namespace ) -}}
 {{- $ca := genCA "opentelemetry-operator-operator-ca" 365 -}}
 {{- $cert := genSignedCert (include "opentelemetry-operator.fullname" .) nil $altNames 365 $ca -}}
 apiVersion: v1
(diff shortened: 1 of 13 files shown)


Allex1 (approved):

charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if and (.Values.admissionWebhooks.create) (not .Values.admissionWebhooks.certManager.enabled) }}
-{{- $altNames := list ( printf "%s.%s" "opentelemetry-operator-webhook-service" .Release.Namespace ) ( printf "%s.%s.svc" "opentelemetry-operator-webhook-service" .Release.Namespace ) -}}
+{{- $altNames := list ( printf "%s-webhook-service.%s" (include "opentelemetry-operator.name" .) .Release.Namespace ) ( printf "%s-webhook-service.%s.svc" (include "opentelemetry-operator.name" .) .Release.Namespace ) -}}
```
TylerHelmuth, line 2: @Allex1 should we use the `fullname` template here instead

Allex1: sure, that seems more appropriate

EmpireErwinKooi: @TylerHelmuth 
I think that the alt name should match whatever is set as the webhook service name, eg. see
https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook.yaml#L34
```
name: {{ template "opentelemetry-operator.name" . }}-webhook-service
```
Here, opentelemetry-operator.name is used, not fullname

TylerHelmuth (approved):

Issue #612, opened 2023-01-20
[operator] Use fullname for naming objects instead of name


TylerHelmuth:
> I think thats probably not best either.  I think in almost all resource naming cases we should be using `fullname`, but it looks like the chart is using `name`.  Since switching to `fullname` is a part of bigger issue, we can handle it in its own PR. I'm ok with using `name` for this PR.

_Originally posted by @TylerHelmuth in https://github.com/open-telemetry/opentelemetry-helm-charts/pull/611#discussion_r1082889862_

sfenman:
May I open a PR for this?

Allex1:
sure. Thanks

PR #617, opened 2023-01-29
[opentelemetry-operator] Use fullname instead of name for objects

sfenman:
Closes https://github.com/open-telemetry/opentelemetry-helm-charts/issues/612

### DIFF
--- a/.github/workflows/operator-test.yaml
+++ b/.github/workflows/operator-test.yaml
@@ -19,7 +19,7 @@ jobs:
         uses: ./.github/actions/setup
         with:
           create-kind-cluster: "true"
-      
+
       - name: Check CRDs
         run: make check-operator-crds
 
--- a/.github/workflows/operator-test.yaml
+++ b/.github/workflows/operator-test.yaml
@@ -63,7 +63,7 @@ jobs:
       - name: Install opentelemetry-operator chart
         run: |
           helm install --namespace=opentelemetry-operator-system --create-namespace my-opentelemetry-operator ./charts/opentelemetry-operator
-          kubectl wait --timeout=5m --for=condition=available deployment opentelemetry-operator-controller-manager -n opentelemetry-operator-system
+          kubectl wait --timeout=5m --for=condition=available deployment my-opentelemetry-operator -n opentelemetry-operator-system
 
       - name: Run e2e tests
         run:  kubectl kuttl test ./opentelemetry-operator/tests/e2e --config ./charts/opentelemetry-operator/kuttl-test.yaml
--- a/charts/opentelemetry-operator/UPGRADING.md
+++ b/charts/opentelemetry-operator/UPGRADING.md
@@ -1,5 +1,10 @@
 # Upgrade guidelines
 
+## 0.21 to 0.22.0
+Kubernetes resource names will now use `{{opentelemetry-operator.fullname}}` as the default value which will change the name of many resources.
+Some CI/CD tools might create duplicate resources when upgrading from an older version because of this change.
+`fullnameOverride` can be used to keep `deployment` resource consistent with the same name during an upgrade.
+
 ## 0.16.0 to 0.17.0
 
 The v0.17.0 helm chart version changes OpenTelemetry Collector image to the contrib version. If you want to use the core version, set `manager.collectorImage.repository` to `otel/opentelemetry-collector`.
--- a/charts/opentelemetry-operator/templates/certmanager.yaml
+++ b/charts/opentelemetry-operator/templates/certmanager.yaml
@@ -5,23 +5,23 @@ metadata:
   labels:
     {{- include "opentelemetry-operator.labels" . | nindent 4 }}
     app.kubernetes.io/component: webhook
-  name: {{ template "opentelemetry-operator.name" . }}-serving-cert
+  name: {{ template "opentelemetry-operator.fullname" . }}-serving-cert
   namespace: {{ .Release.Namespace }}
 spec:
   dnsNames:
-    - {{ template "opentelemetry-operator.name" . }}-webhook-service.{{ .Release.Namespace }}.svc
-    - {{ template "opentelemetry-operator.name" . }}-webhook-service.{{ .Release.Namespace }}.svc.cluster.local
+    - {{ template "opentelemetry-operator.fullname" . }}-webhook-service.{{ .Release.Namespace }}.svc
+    - {{ template "opentelemetry-operator.fullname" . }}-webhook-service.{{ .Release.Namespace }}.svc.cluster.local
   issuerRef:
     {{- if .Values.admissionWebhooks.certManager.issuerRef }}
     {{- toYaml .Values.admissionWebhooks.certManager.issuerRef | nindent 4 }}
     {{- else }}
     kind: Issuer
-    name: {{ template "opentelemetry-operator.name" . }}-selfsigned-issuer
+    name: {{ template "opentelemetry-operator.fullname" . }}-selfsigned-issuer
     {{- end }}
-  secretName: {{ default (printf "%s-controller-manager-service-cert" (include "opentelemetry-operator.name" .)) .Values.admissionWebhooks.secretName }}
+  secretName: {{ default (printf "%s-controller-manager-service-cert" (include "opentelemetry-operator.fullname" .)) .Values.admissionWebhooks.secretName }}
   subject:
     organizationalUnits:
-      - {{ template "opentelemetry-operator.name" . }}
+      - {{ template "opentelemetry-operator.fullname" . }}
 {{- if not .Values.admissionWebhooks.certManager.issuerRef }}
 ---
 apiVersion: cert-manager.io/v1
@@ -30,7 +30,7 @@ metadata:
   labels:
     {{- include "opentelemetry-operator.labels" . | nindent 4 }}
     app.kubernetes.io/component: webhook
-  name: {{ template "opentelemetry-operator.name" . }}-selfsigned-issuer
+  name: {{ template "opentelemetry-operator.fullname" . }}-selfsigned-issuer
   namespace: {{ .Release.Namespace }}
 spec:
   selfSigned: {}
--- a/charts/opentelemetry-operator/templates/certmanager.yaml
+++ b/charts/opentelemetry-operator/templates/certmanager.yaml
@@ -9,8 +9,8 @@ metadata:
   namespace: {{ .Release.Namespace }}
 spec:
   dnsNames:
-    - {{ template "opentelemetry-operator.fullname" . }}-webhook-service.{{ .Release.Namespace }}.svc
-    - {{ template "opentelemetry-operator.fullname" . }}-webhook-service.{{ .Release.Namespace }}.svc.cluster.local
+    - {{ template "opentelemetry-operator.fullname" . }}-webhook.{{ .Release.Namespace }}.svc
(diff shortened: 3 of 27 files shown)


Allex1:
@sfenman Failing with : 
`Error: INSTALLATION FAILED: Service "opentelemetry-operator-dsh5sv7p6b-no-cert-manager-webhook-service" is invalid: metadata.name: Invalid value: "opentelemetry-operator-dsh5sv7p6b-no-cert-manager-webhook-service": must be no more than 63 characters`
Can you make sure to trim to max 63 chars?

sfenman:
@Allex1 
Sure, just let me know how we should handle this.
For example the service name using a `fullname` without override is something like this:
`myrelease-opentelemetry-operator-controller-manager-metrics-service` which is 67 characters.

One way could be to further trim the default `fullname` value.

Maybe we could also use shorter suffixes (aka the `controller-manager-metrics-service`) or something else?

Allex1:
> Maybe we could also use shorter suffixes (aka the `controller-manager-metrics-service`) or something else?
That would be a breaking change, but we might do it anyway , just need to come to an agreement.

Allex1:
> > Maybe we could also use shorter suffixes (aka the `controller-manager-metrics-service`) or something else?
> 
> That would be a breaking change, but we might do it anyway , just need to come to an agreement.

I think a good compromise would be  to replace `controller-manager` with `manager` which would make it consistent with the key in values.yaml.
@TylerHelmuth @dmitryax thoughts?

TylerHelmuth:
I agree that `-controller-manager-metrics-service` seems excessively long.  `controller-manager` is reflective of the operator though right?  Would it be better to drop `-metrics-service`?  We don't include `-service` in the collector or demo charts' service template.

Allex1:
> I agree that `-controller-manager-metrics-service` seems excessively long. `controller-manager` is reflective of the operator though right? Would it be better to drop `-metrics-service`? We don't include `-service` in the collector or demo charts' service template.

@TylerHelmuth that name comes from the [operator examples](https://github.com/open-telemetry/opentelemetry-operator/blob/55c37bc5a5048fe6cb54766746ab0d3eda92ddc3/config/rbac/auth_proxy_service.yaml#LL7C37-L7C37)
Same for `controller-manager` which for some reason is called manager in [values.yaml](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/values.yaml#L16)

TylerHelmuth:
> that name comes from the [operator examples](https://github.com/open-telemetry/opentelemetry-operator/blob/55c37bc5a5048fe6cb54766746ab0d3eda92ddc3/config/rbac/auth_proxy_service.yaml#LL7C37-L7C37)

Then I feel even less inclined to keep it.  Using the prometheus operator chart as an example, they use `{{fullname}}-operator` https://github.com/prometheus-community/helm-charts/blob/c3b697d31b8631fb8847f7f313155deeb5962f98/charts/kube-prometheus-stack/templates/prometheus-operator/service.yaml#L5.  I'd be inclined to do the same.

Allex1:
>  I'd be inclined to do the same.

Then both `service` and `controller-manager` should be removed and replaced with `operator`

@sfenman if you are willing to do this work make sure you bump the chart to 0.22.0 and explain our reasoning on the [upgrading](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/UPGRADING.md) page so users understand it's a breaking change.
It's debatable but some ci/cd tools will duplicate resources bc of this change.

sfenman:
@Allex1 I will start with it shortly
Just a quick thought though, since the default `opentelemetry-operator.fullname = opentelemetry-operator` , I think it would be better to use only the {{fullname}} instead of {{fulname}}-operator in the cases you've mentioned.

TylerHelmuth:
@sfenman great catch.  `{{fullname}}` only is good.

TylerHelmuth (commented):
@sfenman you'll need to update https://github.com/open-telemetry/opentelemetry-helm-charts/blob/36d6b236a92745b82858fd2e703cd5669ad23cfa/.github/workflows/operator-test.yaml#L68 to match the new naming style.

I hope this doesn't break our use of the operator's e2e tests, but we'll cross that bridge when we get there.

.github/workflows/operator-test.yaml
```diff
@@ -65,7 +65,7 @@ jobs:
           until kubectl get ns opentelemetry-operator-system 2>&1 | grep "namespaces \"opentelemetry-operator-system\" not found"; do sleep 1; done
           kubectl create namespace opentelemetry-operator-system
           helm install --namespace=opentelemetry-operator-system my-opentelemetry-operator ./charts/opentelemetry-operator
-          kubectl wait --timeout=5m --for=condition=available deployment opentelemetry-operator-controller-manager -n opentelemetry-operator-system
+          kubectl wait --timeout=5m --for=condition=available deployment opentelemetry-operator -n opentelemetry-operator-system
```
TylerHelmuth, line 68:
```suggestion
          kubectl wait --timeout=5m --for=condition=available deployment my-opentelemetry-operator -n opentelemetry-operator-system
```

Since we use fullname now and the chart is installed as `my-opentelemetry-operator` I believe this is the name of the deployment the CI is looking for.

sfenman: Ah yes I've missed it, thank you.

But chart-testing also failing with:

```
Run until kubectl get ns opentelemetry-operator-system 2>&1 | grep "namespaces 

\"opentelemetry-operator-system\" not found"; do sleep 1; done

Error from server (NotFound): namespaces "opentelemetry-operator-system" not found
namespace/opentelemetry-operator-system created
```

Am I missing something?

TylerHelmuth: You can ignore that namespace error.  The CI should be taking advantage of helm's --create-namespace feature I think.  I'll handle that in a separate PR.

charts/opentelemetry-operator/templates/certmanager.yaml
```diff
@@ -5,23 +5,23 @@ metadata:
...
     app.kubernetes.io/component: webhook
-  name: {{ template "opentelemetry-operator.name" . }}-serving-cert
+  name: {{ template "opentelemetry-operator.fullname" . }}-serving-cert
   namespace: {{ .Release.Namespace }}
 spec:
   dnsNames:
-    - {{ template "opentelemetry-operator.name" . }}-webhook-service.{{ .Release.Namespace }}.svc
-    - {{ template "opentelemetry-operator.name" . }}-webhook-service.{{ .Release.Namespace }}.svc.cluster.local
+    - {{ template "opentelemetry-operator.fullname" . }}-webhook-service.{{ .Release.Namespace }}.svc
+    - {{ template "opentelemetry-operator.fullname" . }}-webhook-service.{{ .Release.Namespace }}.svc.cluster.local
```
TylerHelmuth, line 13:
```suggestion
    - {{ template "opentelemetry-operator.fullname" . }}-webhook.{{ .Release.Namespace }}.svc
    - {{ template "opentelemetry-operator.fullname" . }}-webhook.{{ .Release.Namespace }}.svc.cluster.local
```

We changed the names of the webhooks so this has to change to match.

charts/opentelemetry-operator/templates/clusterrolebinding.yaml
```diff
@@ -5,14 +5,14 @@ metadata:
   labels:
     {{- include "opentelemetry-operator.labels" . | nindent 4 }}
     app.kubernetes.io/component: controller-manager
-  name: {{ template "opentelemetry-operator.name" . }}-manager-rolebinding
+  name: {{ template "opentelemetry-operator.fullname" . }}-manager
```
Allex1, line 8: @sfenman this means that both roles and rolebindings will have the same resource name. Maybe we should add `rb` suffix as a differentiator? Same for other resources with duplicate names ?

sfenman: I removed it because they are already differentiated by their `Kind` attribute. For sure I can add an `rb` suffix if you think it is confusing though.

sfenman:
@Allex1 Please let me know if there are other things I need to include in the `UPGRADING.md` guide

charts/opentelemetry-operator/templates/clusterrolebinding.yaml
Allex1: @TylerHelmuth @dmitryax thoughts?

TylerHelmuth: I think `Kind` is enough.  The collector chart doesn't do any extra naming for clusterrole vs clusterrolebinding.

TylerHelmuth (approved):
@sfenman thanks for doing this and following through with all our reviews!

Allex1 (approved):
🐾 it after comment

charts/opentelemetry-operator/UPGRADING.md
```diff
@@ -1,5 +1,9 @@
 # Upgrade guidelines
 
+## 0.21 to 0.22.0
+Kubernetes resource names will now use `{{opentelemetry-operator.fullName}}` as the default value.
+Some systems might try to create duplicated resources because of this change.
+
```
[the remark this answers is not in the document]
sfenman: I will amend it in the previous commit

charts/opentelemetry-operator/UPGRADING.md
```diff
@@ -1,5 +1,9 @@
 # Upgrade guidelines
 
+## 0.21 to 0.22.0
+Kubernetes resource names will now use `{{opentelemetry-operator.fullName}}` as the default value which will change the name of many resources.
+Some CI/CD tools might create duplicate resources when upgrading from an older version because of this change.
```
dmitryax, line 5: Can you please also provide guidance on how to keep the existing names for an upgrade? Should be possible with `nameOverride`, right?

dmitryax (commented):
@shalevr thanks for the contribution. Please rebase

charts/opentelemetry-operator/UPGRADING.md
TylerHelmuth: @dmitryax because so many of the objects had hard-coded postfixes I don't think we can use `nameOverride`.  The mutating webhooks are an example.  Same with clusterrole and clusterrolebinding.

dmitryax: It should be fine to rename `clusterrole` and `clusterrolebinding` during the upgrade. I'm mostly worried about deployments. Just wanted to make sure we provide a non-disruptive upgrade experience

TylerHelmuth: In that case yes, we can give recommendations on how to keep the original deployment name with nameOverride.  @sfenman can you make that change to UPGRADING.md?

dmitryax: And please test that recommendation. The goal is to have a non-blocking upgrade experience. AFAIR, you cannot change a deployment name during the upgrade

sfenman: Yes, I will test it. AFAIK deployment will be replaced during the `helm upgrade` if it has a new name. Also, I think users should use `fullnameOverride` instead of `nameOverride` to get the same deployment name.

Allex1:
@sfenman the [tests](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/.github/workflows/operator-test.yaml#L66) need to be updated in order for the CI to pass. 
Currently getting : `Error from server (NotFound): deployments.apps "opentelemetry-operator-controller-manager" not found`

Issue #596, opened 2023-01-12
paymentservice OOMKilled, CrashLoopBackOff


mviitane:
Is there anyone else having a problem with the payment service not properly starting up?

```
$ kubectl get pods my-otel-demo-paymentservice-758cd56c47-7lxxh
my-otel-demo-paymentservice-758cd56c47-7lxxh          0/1     OOMKilled   4 (56s ago)    108s

After a while:
$ kubectl get pods
NAME                                                  READY   STATUS             RESTARTS         AGE
my-otel-demo-prometheus-server-57789ccfff-8t6wv       0/1     Pending            0                45m
my-otel-demo-emailservice-57d4cf7f8c-r75xl            1/1     Running            0                45m
my-otel-demo-checkoutservice-bf487f555-nv66x          1/1     Running            0                45m
my-otel-demo-jaeger-b54c97c5b-mqr4p                   1/1     Running            0                45m
my-otel-demo-quoteservice-6886646668-s55mv            1/1     Running            0                45m
my-otel-demo-kafka-59bb6fc9c6-djbgr                   1/1     Running            0                45m
my-otel-demo-accountingservice-c89755b9f-lfc2j        1/1     Running            0                45m
my-otel-demo-otelcol-6587f49484-lfrgv                 1/1     Running            0                45m
my-otel-demo-loadgenerator-7d7d8c5cf7-rz6p9           1/1     Running            0                45m
my-otel-demo-ffspostgres-59c75468f6-t9s9w             1/1     Running            0                45m
my-otel-demo-adservice-7fbdf74578-xfc97               1/1     Running            0                45m
my-otel-demo-recommendationservice-7d669c79bb-tn6kk   1/1     Running            0                45m
my-otel-demo-currencyservice-5668f894c5-vwx9p         1/1     Running            0                45m
my-otel-demo-cartservice-5f497c4fc8-qgptv             1/1     Running            0                45m
my-otel-demo-frontend-7f88b74488-xh9vm                1/1     Running            0                45m
my-otel-demo-redis-5b4b84644c-cgskx                   1/1     Running            0                45m
my-otel-demo-productcatalogservice-6cb65cc977-6rlb2   1/1     Running            0                45m
my-otel-demo-shippingservice-654b7c9694-vmfl4         1/1     Running            0                45m
my-otel-demo-frauddetectionservice-6477d4fbff-vpgzc   1/1     Running            0                45m
[...]

PR #618, opened 2023-01-30
Increase memory allocation to paymentservice to prevent OOMKills

irvingpop:
I was hitting the same issues as #596 - and based on my observations of the process under light load the RSS is in the range of 78-84 MB, but even at cold startup it's above the 70MB limit.

I propose increasing to 100Mi so there's a bit of safety margin here. 

Signed-off-by: Irving Popovetsky <irving@honeycomb.io>

### DIFF
--- a/charts/opentelemetry-demo/values.yaml
+++ b/charts/opentelemetry-demo/values.yaml
@@ -386,7 +386,7 @@ components:
         value: "8080"
     resources:
       limits:
-        memory: 70Mi
+        memory: 100Mi
 
   productCatalogService:
     enabled: true
--- a/charts/opentelemetry-demo/values.yaml
+++ b/charts/opentelemetry-demo/values.yaml
@@ -386,7 +386,7 @@ components:
         value: "8080"
     resources:
       limits:
-        memory: 100Mi
+        memory: 120Mi
 
   productCatalogService:
     enabled: true
--- a/charts/opentelemetry-demo/Chart.yaml
+++ b/charts/opentelemetry-demo/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: opentelemetry-demo
-version: 0.16.1
+version: 0.16.2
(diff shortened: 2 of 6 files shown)


TylerHelmuth:
@puckpuck I know you've been doing extensive stress testing of the demo chart, do you already have a PR ready that handles https://github.com/open-telemetry/opentelemetry-helm-charts/issues/596 as well as any other fixes you've discovered with your testing?

puckpuck:
I do have a PR ready with improvements to memory limits for a few services, including Payment.  it will be part of the 1.3 release.  I can break it out as a separate PR, but I think we will cut 1.3 tomorrow.

TylerHelmuth:
@puckpuck unless there is something performance-related to 1.3's image I'd vote for breaking it out.  Our 1.2 app version of the chart could use the performance tweaks now.

TylerHelmuth:
If paymentservice is the main culprit, we can move forward with this PR as the solution.

puckpuck:
It should be 120Mi ... that's what my PR is setting it to.

charts/opentelemetry-demo/values.yaml
```diff
@@ -386,7 +386,7 @@ components:
         value: "8080"
     resources:
       limits:
-        memory: 70Mi
+        memory: 100Mi
```
TylerHelmuth, line 389:
```suggestion
        memory: 120Mi
```

TylerHelmuth (commented):
@irvingpop please bump the chart's patch version and then run `make generate-examples CHARTS=opentelemetry-demo`.  To generate the examples you'll need the chart's dependencies added to your local help repos.

irvingpop:
thanks for the tip @TylerHelmuth I hopefully did all that correctly.

irvingpop:
whoops now I see what you mean about bumping the chart version.

TylerHelmuth (approved):

PR #619, opened 2023-01-31
[demo] Support for OpenTelemetry Demo 1.3.0

puckpuck:
Supports the release of [OpenTelemetry Demo 1.3.0](https://github.com/open-telemetry/opentelemetry-demo/releases/tag/1.3.0)

Memory updates for: frontendproxy, Kafka, and Jaeger

The demo release contains several enhancements including support for metrics exemplars to traces (Prometheus, Grafana), a new failure scenario (adservice), and significant updates to the PHP SDK (quotservice)


charts/opentelemetry-demo/values.yaml
```diff
@@ -375,7 +379,7 @@ components:
       - name: PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION
         value: python
       - name: OTEL_EXPORTER_OTLP_ENDPOINT
-        value: 'http://{{ include "otel-demo.name" . }}-otelcol:4317'
+        value: 'http://{{ include "otel-demo.name" . }}-otelcol:4318/v1/traces'
```
TylerHelmuth, line 382: According to the spec `OTEL_EXPORTER_OTLP_ENDPOINT` should not use the full endpoint as it will be interpreted as a base and have `/v1/metrics` and `/v1/traces` appended automatically.  What changed in the demo to require the full url?  Should we use `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`  instead?

charts/opentelemetry-demo/values.yaml
```diff
@@ -519,9 +519,17 @@ components:
     env:
       - name: KAFKA_ADVERTISED_LISTENERS
         value: 'PLAINTEXT://{{ include "otel-demo.name" . }}-kafka:9092'
+      - name: OTEL_EXPORTER_OTLP_TRACES_ENDPOINT
+        value: 'http://{{ include "otel-demo.name" . }}-otelcol:4317'
```
TylerHelmuth, line 523: does this need `/v1/traces`?

charts/opentelemetry-demo/values.yaml
```diff
@@ -519,9 +519,17 @@ components:
...
+      - name: OTEL_EXPORTER_OTLP_METRICS_ENDPOINT
+        value: 'http://{{ include "otel-demo.name" . }}-otelcol:4317'
```
TylerHelmuth, line 525: does this need `/v1/metrics`?

charts/opentelemetry-demo/values.yaml
```diff
@@ -615,10 +626,19 @@ jaeger:
...
   allInOne:
     enabled: true
-    args: ["--memory.max-traces", "10000", "--query.base-path", "/jaeger/ui"]
+    args:
+      - "--memory.max-traces"
+      - "10000"
+      - "--query.base-path"
+      - "/jaeger/ui"
+      - "--prometheus.server-url"
+      - 'http://{{ include "otel-demo.name" . }}-prometheus-server:9090'
```
TylerHelmuth, line 635: For my own knowledge, why does jaeger need the prometheus server?

charts/opentelemetry-demo/templates/_objects.tpl
```diff
@@ -41,7 +41,7 @@ spec:
       {{- end }}
       containers:
         - name: {{ .name }}
-          image: '{{ ((.imageOverride).repository) | default .defaultValues.image.repository }}:{{ ((.imageOverride).tag) | default (printf "v%s-%s" (default .Chart.AppVersion .defaultValues.image.tag) (replace "-" "" .name)) }}'
+          image: '{{ ((.imageOverride).repository) | default .defaultValues.image.repository }}:{{ ((.imageOverride).tag) | default (printf "%s-%s" (default .Chart.AppVersion .defaultValues.image.tag) (replace "-" "" .name)) }}'
```
TylerHelmuth, line 44: Will this break the use of the chart with older image versions?

charts/opentelemetry-demo/values.yaml
puckpuck: Jaeger has a way to display span metrics in the tool.  To do so, it needs to know where the metrics are.

charts/opentelemetry-demo/values.yaml
puckpuck: good catch.  Moving this to `OTLP_EXPORTER_TRACES_ENDPOINT` instead.

charts/opentelemetry-demo/values.yaml
puckpuck: It works without the path suffix. Though I now see the inconsistency across the demo components of using the standard vs telemetry-specific endpoints with and without paths.

charts/opentelemetry-demo/values.yaml
puckpuck: It works without the path suffix. Though I now see the inconsistency across the demo components of using the standard vs telemetry-specific endpoints with and without paths.

charts/opentelemetry-demo/values.yaml
puckpuck: This is the feature specifically that it enables
Screenshot 2023-01-31 at 12 09 37 PM

charts/opentelemetry-demo/values.yaml
TylerHelmuth: That is suspicious.  Do we need both env vars are can we use `OTEL_EXPORTER_OTLP_ENDPOINT`?

puckpuck: reading the original PR that made this change on the demo side, it seems like the TRACES_ENDPOINT isn't needed and shouldn't of been part of this at all.  I'll make some changes and test just using the non-telemetry specific endpoint and update this PR.

TylerHelmuth: Sweet, ping me when you're ready for another review.

puckpuck:
@TylerHelmuth ready for another review

TylerHelmuth (approved):

PR #621, opened 2023-02-01
[demo] - enable complex service definitions

### LINKED ISSUE #609: Helm scheme doesn't support service annotations for frontendProxy
When configuring the frontendProxy component, It is beneficial to configure the frontendProxy component to support additional annotations for managed services like AWS Load Balancer Controller, such as Network Load Balancer (NLB) with instance or IP targets through a Kubernetes service of type LoadBalancer with relevant annotations. However the scheme does not support this today, and static IPs can not be assigned. 

i.e.
```
service.beta.kubernetes.io/aws-load-balancer-type
service.beta.kubernetes.io/aws-load-balancer-eip-allocations
service.beta.kubernetes.io/aws-load-balancer-subnet
```

puckpuck:
Fixes #609 

Moves `serviceType` and `servicePort` parameters to be members of a `service` parameter as children called `type` and `port`. It also adds support for `annotations` and `nodePort` under the same parent parameter.

This will be a breaking change for people who overrode either of the component service parameters before.


charts/opentelemetry-demo/examples/default/rendered/component.yaml
```diff
@@ -1812,10 +1764,6 @@ spec:
         - name: shippingservice
           image: 'ghcr.io/open-telemetry/demo:1.3.0-shippingservice'
           imagePullPolicy: IfNotPresent
-          ports:
-          
-          - containerPort: 8080
```
TylerHelmuth, line 1817: Is the removal of these ports expected?

charts/opentelemetry-demo/README.md
```diff
@@ -27,6 +27,13 @@ helm install my-otel-demo open-telemetry/opentelemetry-demo
 
 ## Upgrading Chart
 
+### To 0.18
```
TylerHelmuth, line 30: How do you feel about moving this to an `UPGRADING.md` file?  This is what the other charts use.

charts/opentelemetry-demo/examples/default/rendered/component.yaml
puckpuck: no... I need to look into why this happened

charts/opentelemetry-demo/README.md
puckpuck: done

charts/opentelemetry-demo/examples/default/rendered/component.yaml
puckpuck: this is fixed now

charts/opentelemetry-demo/examples/public-hosted-ingress/rendered/component.yaml
```diff
@@ -1857,29 +1857,3 @@ spec:
           resources:
             limits:
               memory: 20Mi
----
-# Source: opentelemetry-demo/templates/component.yaml
-apiVersion: "networking.k8s.io/v1"
```
TylerHelmuth, line 1862: @puckpuck I believe the new service object refactor broke the `otel-demo.ingress` template.

puckpuck:
I did a search on the entire code base for `serviceType` and `servicePort` to make sure we aren't referencing the old values anymore and that they are adopted for the new structure.

charts/opentelemetry-demo/examples/public-hosted-ingress/rendered/component.yaml
puckpuck: fixed

TylerHelmuth (approved):

PR #623, opened 2023-02-01
Add Service for daemonsets as well

jpkrohling:
When running the OTel Demo chart, I realized that I could just change the collector to grab the pod's logs. That requires daemonsets to be used, which caused traces to break, as there are no services created for it. This PR changes the chart by making it create a service also for that case.

I'm not at all familiar with Helm charts, and at first I tried to add another "or" clause to the service template, but that changed way more things than I wanted. I confirmed locally that the service this changes generates is enough to get traces into the collector.

Signed-off-by: Juraci Paixão Kröhling <juraci@kroehling.de>

### DIFF
--- a/charts/opentelemetry-collector/examples/daemonset-collector-logs/rendered/service.yaml
+++ b/charts/opentelemetry-collector/examples/daemonset-collector-logs/rendered/service.yaml
@@ -1,0 +1,46 @@
+---
+# Source: opentelemetry-collector/templates/service.yaml
+apiVersion: v1
+kind: Service
+metadata:
+  name: example-opentelemetry-collector
+  labels:
+    helm.sh/chart: opentelemetry-collector-0.47.0
+    app.kubernetes.io/name: opentelemetry-collector
+    app.kubernetes.io/instance: example
+    app.kubernetes.io/version: "0.70.0"
+    app.kubernetes.io/managed-by: Helm
+    component: agent-collector
+spec:
+  type: ClusterIP
+  ports: 
+    
+    - name: jaeger-compact
+      port: 6831
+      targetPort: 6831
+      protocol: UDP
+    - name: jaeger-grpc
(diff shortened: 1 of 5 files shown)


puckpuck:
Daemonsets shouldn't have services since a service will route to any pod selected, and may not necessarily be the local node.

The use case here, which is to run the Otel demo pointed to a collector deployed as a daemonset instead of a deployment is what we should focus on.

I would like to see a way to abstract the collector's endpoint into its own environment variable (part of the defaults), which then gets reused as part of the Kubernetes env variable expansion for what each demo component needs when specifying the `OTEL_EXPORTER_OTLP*_ENDPOINT` variables.  Then the user only needs to override a single environment variable at the default level to point toward the local K8s node instead.  This might require us to lock onto a port (4317/4318) but should be doable.

jpkrohling:
> Daemonsets shouldn't have services since a service will route to any pod selected, and may not necessarily be the local node.

That's true if you care to reach a local daemonset. This is not the case for the spans, which is why I think a service in front of daemonsets are appropriate for this case.

jpkrohling:
> why not send to the local daemonset?

Using the downward API for the nodeIP? I considered that, but the demo has 10 services, each one would require to be changed to use the downward API *only* if collecting logs from pods is desirable (otherwise a Deployment is used).

puckpuck:
> > why not send to the local daemonset?
> 
> Using the downward API for the nodeIP? I considered that, but the demo has 10 services, each one would require to be changed to use the downward API _only_ if collecting logs from pods is desirable (otherwise a Deployment is used).

I described in detail how we would change the demo's Helm chart to make it work in [this comment](https://github.com/open-telemetry/opentelemetry-helm-charts/issues/624#issuecomment-1415940371). I have a company offsite this week, so I don't think I will get to it, but doing this is at the top of my priority list.

Issue #624, opened 2023-02-01
[demo] - Allow for collector to be deployed in Daemonset mode


puckpuck:
Right now, if you deploy the collector in daemonset mode, the demo will not function since the service will not be created.

The solution requires configuring each demo component to set the `OTEL_EXPORTER_OLTP*_ENDPOINT` for each component to point to the local K8s node name. This requires the user to override environment variables for all components and can be problematic.  Ideally, we configure the environment variables (potential with K8s env variable expansion) so that the collector endpoint is specified once and re-used by all components, allowing the user to make a single change to a default environment variable instead of having to change each component.

puckpuck:
We already expose the node name as an environment variable, so I want to leverage that. What I'm proposing to do is to add a top-level default environment variable like this:

```
    - name: OTEL_COLLECTOR_ENDPOINT
      value: '{{ include "otel-demo.name" . }}-otelcol'
```

Within each demo component, where the collector is specified, we now reference this new default variable with Kubernetes environment variable expansion like this for the adservice (other services would be similar)

```
      - name: OTEL_EXPORTER_OTLP_ENDPOINT
        value: http://$(OTEL_COLLECTOR_ENDPOINT):4317'
```


From the user's point of view, they would set the Collector's configuration options to be Daemonset mode, and pass in an override for the `OTEL_COLLECTOR_ENDPOINT` variable.  So the values file would look something like this:

```
default:
  envOverrides:
      - name: OTEL_COLLECTOR_ENDPOINT
        value: $(OTEL_K8S_NODE_NAME)
        
opentelemetry-collector:
  mode: daemonset
```

ashishkf:
In the demo chart I only see `OTEL_SERVICE_NAME` configured -  https://github.com/open-telemetry/opentelemetry-helm-charts/blob/f1c65af83c666f7d5148416a1e20821df4716886/charts/opentelemetry-demo/values.yaml#L4

There is no configuration for `OTEL_K8S_NODE_NAME` at all anywhere in the repo.

PR #629, opened 2023-02-06
[demo] abstract collector name from demo components

puckpuck:
Fixes: #624 and Closes #623

The PR will abstract the name of the OpenTelemetry Collector away from the components into a higher-level default environment variable `OTEL_COLLECTOR_NAME` pushed into all components. The components will then leverage this variable with Kubernetes environment variable expansion as needed using the `$(OTEL_COLLECTOR_NAME)` syntax.

With this change users can now deploy the collector in daemonset mode, and only change 1 environment variable override to have it working. This also opens up possibilities for uers that have the OpenTelemetry collector installed outside of the demo.

Tested and works with the following specified in a values file
```
default:
  envOverrides:
    - name: OTEL_COLLECTOR_NAME
      value: $(OTEL_K8S_NODE_NAME)

opentelemetry-collector:
  mode: daemonset
```

### DIFF
--- a/charts/opentelemetry-demo/Chart.yaml
+++ b/charts/opentelemetry-demo/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: opentelemetry-demo
-version: 0.18.0
+version: 0.18.1
 description: opentelemetry demo helm chart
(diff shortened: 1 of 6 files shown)


TylerHelmuth (approved):
LGTM

PR #633, opened 2023-02-08
[collector] Add PrometheusRule for spans that failed to be exported

### LINKED ISSUE #632: Prometheus rule for exporter failed spans
Default Prometheus rules was added in #267, but a rule for otelcol_exporter_send_failed_spans was never added.
It seems that the metric was not yet implemented back then: https://github.com/open-telemetry/opentelemetry-helm-charts/pull/268#discussion_r918062076

I'm proposing to add it as it was written in the original issue, as it works just fine with the current collector version.

While I'm at it, I would also like to update the rule descriptions to make it more clear if the rule applies to the receiver, exporter or processor. Currently, the description for dropped spans in both receiver and exporter:
`The otlp is dropping spans at a rate of 596.1 per second`

A small update to include the part of the pipeline like this would make it more explicit:
`The otlp exporter is dropping spans at a rate of 596.1 per second`

I'm attaching a pull request with my proposed changes shortly :)

splitfeed:
* Improve rule description to more easily distinguish between receiver, exporter and processor alerts

Resolves #632

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.48.0
+version: 0.48.1
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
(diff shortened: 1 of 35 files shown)


TylerHelmuth (approved):

dmitryax (approved):

dmitryax:
@splitfeed please rebase

Allex1 (approved):

Issue #635, opened 2023-02-08
[demo] setting serviceAccount in values.yaml does not work


bdekany:
Hi,

The variable `serviceAccount` in the `values.yaml` for the opentelemetry-demo chart does not produce expected outcome.

```shell
helm template --set serviceAccount=otel-sa . | grep serviceAccount
      serviceAccountName: release-name-grafana
      serviceAccountName: release-name-jaeger
      serviceAccountName: release-name-otelcol
      serviceAccountName: release-name-prometheus-server
      serviceAccountName: release-name-grafana-test
```

This is because in the `_objects.tpl` the jinja template use `serviceAccountName`

PR #636, opened 2023-02-08
[demo] Make variable serviceAccount in values.yaml usable

bdekany:
Hi,

With this change, setting `serviceAccount` variable actually generate manifests with `serviceAccountName` properly set.

fix open-telemetry/opentelemetry-helm-charts#635

```shell
helm template --set serviceAccount=otel-sa . | grep serviceAccount
      serviceAccountName: release-name-grafana
      serviceAccountName: release-name-jaeger
      serviceAccountName: release-name-otelcol
      serviceAccountName: release-name-prometheus-server
      serviceAccountName: otel-sa
      serviceAccountName: otel-sa
      serviceAccountName: otel-sa
      serviceAccountName: otel-sa
      serviceAccountName: otel-sa
      serviceAccountName: otel-sa
      serviceAccountName: otel-sa
      serviceAccountName: otel-sa
      serviceAccountName: otel-sa
      serviceAccountName: otel-sa
      serviceAccountName: otel-sa
      serviceAccountName: otel-sa
      serviceAccountName: otel-sa
      serviceAccountName: otel-sa
      serviceAccountName: otel-sa
      serviceAccountName: otel-sa
      serviceAccountName: otel-sa
      serviceAccountName: otel-sa
      serviceAccountName: otel-sa
     serviceAccountName: release-name-grafana-test
```

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -15,6 +15,6 @@
 
 * @open-telemetry/helm-approvers
 
-charts/opentelemetry-collector/ @open-telemetry/helm-approvers @dmitryax @TylerHelmuth
+charts/opentelemetry-collector/ @open-telemetry/helm-approvers @dmitryax @TylerHelmuth @povilasv
 charts/opentelemetry-operator/  @open-telemetry/helm-approvers @Allex1
 charts/opentelemetry-demo/      @open-telemetry/helm-approvers @puckpuck
--- a/.github/workflows/operator-test.yaml
+++ b/.github/workflows/operator-test.yaml
@@ -19,7 +19,7 @@ jobs:
         uses: ./.github/actions/setup
         with:
           create-kind-cluster: "true"
-      
+
       - name: Check CRDs
         run: make check-operator-crds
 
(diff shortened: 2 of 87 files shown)


TylerHelmuth (commented):
@bdekany thanks for fixing this bug.  Please bump the chart's patch number and then run `make generate-examples CHARTS=opentelemetry-demo`.  You'll need to have added all the demo chart's dependencies to be able to generate the examples.

TylerHelmuth (approved):

dmitryax:
Given that `serviceAccount` config is currently broken (not being used in the template), I think it's a good opportunity to bring it to consistency with other helm charts:

```
serviceAccount:
  # Specifies whether a service account should be created
  create: true
  # Annotations to add to the service account
  annotations: {}
  # The name of the service account to use.
  # If not set and create is true, a name is generated using the fullname template
  name: ""
```

Currently, `serviceAccount` is a string in the demo helm chart which will be harder to migrate later if we accept this fix as is

@open-telemetry/helm-approvers WDYT?

TylerHelmuth:
@dmitryax I had the same thought, but figured we could fix the bug in this PR and do the refactor in a separate PR. If @bdekany is interested in doing the refactor now that's fine with me too.

bdekany:
Hi, I'll try to implement this feature.
Should I keep version `0.18.2` as it is considered as a patch or bump to `0.19.0` as it is a change in the `values.yaml` usage ❓

TylerHelmuth:
@bdekany bump to 0.19.0 is good.

bdekany:
@povilasv Thanks. The CI issue was due to helm unable to generate a secret to store release version. I changed to scope from .Values.serviceAccount to .serviceAccount in `_helpers.tpl` and `serviceaccount.yaml`
Both `helm template`and `helm install` work now

TylerHelmuth (approved):

PR #645, opened 2023-02-15
[opentelemetry-collector] Remove default resource values

### LINKED ISSUE #644: [collector] Unable to remove CPU limits when used as a subchart
Using this chart forces one to set a CPU limit due to it not being possible to remove the default keys.

As an example, say I want to have 1 core of cpu requests, but not set a limit. If my values are
```
# values.yaml
  resources:
    requests:
      cpu: 1000m
      memory: 1Gi
    limits:
      cpu: null
      memory: 4Gi
```
this yields the generated manifest:
```
# manifest.yaml
          resources:
            limits:
              cpu: 256m
              memory: 4Gi
            requests:
              cpu: 1000m
              memory: 1Gi
```
i.e., the CPU limit is populated by the default value.

This is a known bug in helm: https://github.com/helm/helm/issues/5184

Can we please comment out the default resources in the value file and leave them as an example?

verejoel:
Fixes #644

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.48.1
+version: 0.49.0
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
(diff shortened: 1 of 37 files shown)


verejoel:
Hey @povilasv, thanks for taking a look. I responded to the comment (it is indeed run as a subchart).

I would still like to discuss whether resources should be set in the base chart. I don't think this is a good practice, as the resource usage is completely dependent on each particular use-case. Resources should be empty by default (leading to a best-effort QoS) and the user should be encouraged to think about their resource requirements, rather than relying on some arbitrary defaults.

dmitryax:
Hi @verejoel, can you please rebase and incorporate the latest suggestions from https://github.com/open-telemetry/opentelemetry-helm-charts/issues/644?

PR #651, opened 2023-02-21
added docs in the helm values.yaml for the init-fs

### LINKED ISSUE #650: Permissions Issue when using Persistence with EBS Volumes
Just wanted to point this out in an issue and potentially submit a PR to document it in the Chart Values that are using EBS volumes / anything permissions-limiting for persistence would highly likely produce an error like below:
```bash
Error: cannot start pipelines: open /var/lib/storage/otc/exporter_zipkin__traces: permission denied; failed to shutdown pipelines: no existing monitoring routine is running

collector server run finished with error: cannot start pipelines: open /var/lib/storage/otc/exporter_zipkin__traces: permission denied; failed to shutdown pipelines: no existing monitoring routine is running
``` 

I think the error is due to the fact that the OTLP container by default is using user `10001` and thus get this permissions denied error.

The solution I have is basically using an `InitContainer` that can handle the filesystem ownership change to our guy (`10001`).

```yaml
# List of init container specs, e.g. for copying a binary to be executed as a lifecycle hook.
initContainers:
  - name: init-fs
    image: busybox:1.36
    command:
      - sh
      - '-c'
      - 'chown -R 10001: /var/lib/storage/otc'
    volumeMounts:
      - name: opentelemetry-collector-data
        mountPath: /var/lib/storage/otc
```

sherifkayad:
closes #650

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.48.2
+version: 0.48.3
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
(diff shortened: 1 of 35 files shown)


TylerHelmuth (approved):
LGTM

dmitryax (approved):

PR #662, opened 2023-03-02
[demo] prometheus honor_labels support

puckpuck:
Adds proper support for `honor_labels` with Prometheus scaped metrics, so the `job` label will be set properly to the service the metric belongs to.  See [this PR](https://github.com/open-telemetry/opentelemetry-demo/pull/743) in the demo repo.


dmitryax (approved):

TylerHelmuth (commented):
LGTM, but needs a patch bump.

TylerHelmuth (approved):

Issue #628, opened 2023-02-06
Consistent way of defining labels/annotations


Allex1:
Currently extra labels/annotations are being injected in different ways across resource types. We should try and keep them consistent while allowing for the `tpl` [function](https://helm.sh/docs/howto/charts_tips_and_tricks/#using-the-tpl-function) to inject helm variables trough values.yaml.

yamagai:
Hello @Allex1, I'm new here and would like to contribute!

The goal is to align the way to specify extra anntation/labels in each resource with the following, right?
```
    {{- range $key, $value := .Values.<resource>.<extraLabels/Annotations> }}
    {{- printf "%s: %s" $key (tpl $value $ | quote) | nindent 4 }}
    {{- end }}
```

Allex1:
@yamagai yes, go for it. Thanks

PR #671, opened 2023-03-08
Define annotations in consistent way

yamagai:
Close https://github.com/open-telemetry/opentelemetry-helm-charts/issues/628.

Unified the way to define annotations as follows. (lables/extraLabels are already unified.)
```
annotations:
  {{- range $key, $value := .Values.annotations }}
  {{- printf "%s: %s" $key (tpl $value $ | quote) | nindent 4 }}
  {{- end }}
```

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.49.1
+version: 0.49.2
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-collector/examples/daemonset-and-deployment/rendered/configmap-agent.yaml
+++ b/charts/opentelemetry-collector/examples/daemonset-and-deployment/rendered/configmap-agent.yaml
@@ -5,7 +5,7 @@ kind: ConfigMap
 metadata:
   name: example-opentelemetry-collector-agent
   labels:
-    helm.sh/chart: opentelemetry-collector-0.49.1
+    helm.sh/chart: opentelemetry-collector-0.49.2
     app.kubernetes.io/name: opentelemetry-collector
     app.kubernetes.io/instance: example
     app.kubernetes.io/version: "0.72.0"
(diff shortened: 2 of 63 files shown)


puckpuck:
Please bump the version on the demo chart as well

yamagai:
Oops, I missed it. Done.

dmitryax:
I agree that we need a consistent way, but why is this one considered a better way? If the list is empty, it'll render 

```yaml
annotations: null
```

Even if k8s can read it, it's not correct IMO

Allex1:
@yamagai I think you should split this pr, one for each chart as it's hard to review.

yamagai:
> why is this one considered a better way?

I understood that the problem with using "with statement" was that the root context in the block would become `.Values.annotations` as mentioned [here](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/627#discussion_r1097642493)

This is also described in helm document: [here](https://helm.sh/docs/chart_template_guide/variables/).

For example, we can use global context in setting annotations like below
```
# values.yaml
annotations:
  hoge: hoge
  fuga: "{{ tpl (.Release.Name) . }}"

# template
annotations:
  {{- range $key, $value := .Values.annotations }}
  {{- printf "%s: %s" $key (tpl $value $ | quote) | nindent 4 }}
  {{- end }}

# output
annotations:
  fuga: "release-name"
  hoge: "hoge"
```

But we cannot use global context in setting annotations like below
```
# values.yaml
annotations:
  hoge: hoge
  fuga: "{{ tpl (.Release.Name) . }}"

# template
{{- with .Values.annotations }}
annotations:
  {{- toYaml . | nindent 4 }}
{{- end }}

# output
annotations:
  fuga: '{{ tpl (.Release.Name) . }}'
  hoge: "hoge"
```

how do you think? @Allex1 @dmitryax 

> it's not correct IMO

I feel you 🧐 
how about the following way? this is the only way I could think of that doesn't use "with statement"...
```
  {{ if .Values.annotations }}
  annotations:
    ....
  {{ end }}
```

yamagai:
> I think you should split this pr, one for each chart as it's hard to review.

Okay.
Once an implementation policy has been decided, I will split this pr 🚀

dmitryax:
Ok, I see the problem now. Wrapping it with the if statements sounds good to me

yamagai:
split this PR like below. So, this PR will close.

https://github.com/open-telemetry/opentelemetry-helm-charts/pull/674
https://github.com/open-telemetry/opentelemetry-helm-charts/pull/675
https://github.com/open-telemetry/opentelemetry-helm-charts/pull/676

PR #672, opened 2023-03-08
[demo] support demo 1.3.1 release

puckpuck:
This supports to recently released 1.3.1 version of the demo.  More importantly, it adds arm64 support so the demo can run locally on M1/M2 macs.


dmitryax:
That's amazing! 🎉

dmitryax (approved):

PR #678, opened 2023-03-10
[operator] - Add verticalpodautoscaler support for manager

Allex1:
Also regenerated schema like described in [HERE](https://www.arthurkoziel.com/validate-helm-chart-values-with-json-schemas/) as it was inconsistent

### DIFF
--- a/charts/opentelemetry-operator/Chart.yaml
+++ b/charts/opentelemetry-operator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.24.2
+version: 0.24.3
 description: OpenTelemetry Operator Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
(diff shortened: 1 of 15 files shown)


dmitryax (approved):

avoidik:
just fyi, this change has introduced a regression to PDB schema validation when PDB is enabled, `maxUnavailable` and `minAvailable` can be either string or integer (the reason why `oneOf` has been there before)

```
Error: values don't meet the specifications of the schema(s) in the following chart(s): opentelemetry-operator: - pdb.maxUnavailable: Invalid type. Expected: string, given: integer
```

Allex1:
@avoidik I don't think it's not realistic to manually update the schema for every pr so I started generating it. While that can break some older behaviour it tends to update the parts which were missed in some prs.
While I understand how this can be annoying, I'm not sure why users would want to declare `maxUnavailable` as string.

avoidik:
> I'm not sure why users would want to declare maxUnavailable as string.

As you may see it's expected integer,  so I did provide an integer, however as per new schema definition it's only a string.

Allex1:
@avoidik That's what I get for commenting from my phone 😓
That can be both integer and string (when using percentages) so I guess the old value was correct ... 
Not yet sure how we should approach this by also generating the schema...
Thanks for surfacing this btw

PR #685, opened 2023-03-14
[demo] Add support for an init container

austinlparker:
This PR adds a very straightforward init container to the feature flag service that blocks until the postgres db is available and listening.


charts/opentelemetry-demo/templates/_objects.tpl
```diff
@@ -70,6 +70,12 @@ spec:
           configMap:
             name: {{ include "otel-demo.name" . }}-{{ .name }}-config
       {{- end }}
+      {{- if .initContainers }}
+      initContainers:
+        - name: {{.name}}-init
+          image: busybox:1.28
```
dmitryax, line 76: Do we need to fix the `busybox` version? Maybe keep it latest to not case security updates etc?

charts/opentelemetry-demo/values.yaml
```diff
@@ -286,6 +286,8 @@ components:
         port: 8081
       initialDelaySeconds: 30
       periodSeconds: 10
+    initContainers:
+      dependency: '{{ include "otel-demo.name" . }}-ffspostgres 5432'
```
dmitryax, line 290: Is there a use case to make it configurable? Maybe just put this value right in the template?

charts/opentelemetry-demo/templates/_objects.tpl
austinlparker: Yeah, latest is fine I'd think. Will update.

charts/opentelemetry-demo/values.yaml
austinlparker: Not sure what you mean -- there's a use case for other containers to init (for example, perhaps things that need redis should wait for redis), but this is kind of a workaround in general that I'd prefer we deprecate eventually.

TylerHelmuth: Instead of `initContainer` being an object that allows defining a dependency, can we instead let the initContainer object be all the parts of the initContainer?  This is how the collector chart works.  In _objects.tpl we can then do something like:

```yaml
{{- if .Values.initContainers }}
initContainers:
  {{- tpl (toYaml .Values.initContainers) . | nindent 2 }}
{{- end }}
```

For the values.yaml we'd update the comment at the start of the services detailing how initContainer can be used and then for featureflagservice we'd supply the default init container that it needs.

This gives us the container we need and also the flexibility for users to add any sort of initContainer they want.  What do you think?

charts/opentelemetry-demo/values.yaml
```diff
@@ -139,6 +139,11 @@ components:
   #   configuration: {}
   # # Kubernetes container health check options
   #   livenessProbe: {}
+  # # Optional init container to run before the pod starts. This is implemented as a netcat check to the port specified.
```
TylerHelmuth, line 142: Do we need the second sentence about netcat?  That feels specific to the featureflagservice.

TylerHelmuth (approved):
1 nit, otherwise LGTM

charts/opentelemetry-demo/values.yaml
austinlparker: good catch, forgot to update that line.

charts/opentelemetry-demo/values.yaml
```diff
@@ -286,6 +291,10 @@ components:
         port: 8081
       initialDelaySeconds: 30
       periodSeconds: 10
+    initContainers:
+    - name: featureflagservice-init
+      image: busybox:latest
+      command: ['sh', '-c', 'until nc -z {{ include "otel-demo.name" . }}-ffspostgres 5432; do echo waiting for postgres; sleep 2; done']
```
TylerHelmuth, line 297: If postgres never comes up is this deployment stuck indefinitely?

austinlparker: yes, but the featureflag service would never come up successfully if postgres never came up, so i don't think that's really a problem.

dmitryax (approved):

Issue #669, opened 2023-03-07
Kubernetes Events preset


povilasv:
Opening an issue for discussion, we could use [k8sobjectsreceiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/k8sobjectsreceiver) to watch Kubernetes Events and add a preset similiar to Kubernetes logs preset.

One caveat the receiver [k8sobjectsreceiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/k8sobjectsreceiver)  is alpha stability.

Thoughts?

Reference: https://github.com/open-telemetry/opentelemetry-helm-charts/issues/381

cc @nicolastakashi

nicolastakashi:
Just a small note!
The `k8seventsreceiver` we're planning to add as a preset as described on the issue as reference is still an ALPHA receiver, maybe should be something we must highlight to the user that may have breaking changes in the future.

TylerHelmuth:
I support adding a preset for `k8sobjectreceiver` with a clear warning, both in the values.yaml and in `NOTES.txt` that the component is alpha.

nicolastakashi:
@povilasv and @TylerHelmuth I'm going to prepare a PR for this.

nicolastakashi:
@povilasv we can close this 😄

PR #687, opened 2023-03-14
[collector] #686 - Ensure ClusterRole is created for kubeletMetrics preset

cjbhaines:
Addresses: https://github.com/open-telemetry/opentelemetry-helm-charts/issues/686

PR #688, opened 2023-03-14
[FEATURE] adding k8sobjects receiver as preset

nicolastakashi:
Adding k8sobjects receiver as a preset to collect Kubernetes Events as agreed on #669

### DIFF
--- a/charts/opentelemetry-collector/templates/clusterrole.yaml
+++ b/charts/opentelemetry-collector/templates/clusterrole.yaml
@@ -1,4 +1,4 @@
-{{- if or (.Values.clusterRole.create) (.Values.presets.kubernetesAttributes.enabled) (.Values.presets.clusterMetrics.enabled) (.Values.presets.kubeletMetrics.enabled) -}}
+{{- if or (.Values.clusterRole.create) (.Values.presets.kubernetesAttributes.enabled) (.Values.presets.clusterMetrics.enabled) (.Values.presets.kubeletMetrics.enabled) (.Values.presets.kubernetesEvents.enabled) -}}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
@@ -41,5 +41,10 @@ rules:
   - apiGroups: [""]
     resources: ["nodes/stats"]
     verbs: ["get", "watch", "list"]
+  {{- end }}
+  {{- if .Values.presets.kubernetesEvents.enabled }}
+  - apiGroups: ["events.k8s.io"]
+    resources: ["events"]
+    verbs: ["watch"]
   {{- end }}
 {{- end }}
--- a/charts/opentelemetry-collector/values.yaml
+++ b/charts/opentelemetry-collector/values.yaml
@@ -40,6 +40,12 @@ presets:
   # Best used with mode = daemonset.
   kubernetesAttributes:
     enabled: false
+  # Configures the collector to collect kubernetes events.
+  # Adds the k8sobject receiver to the logs pipeline
+  # and collects kubernetes events by default.
+  # Best used with mode = deployment or statefulset.
+  kubernetesEvents:
+    enabled: false
   # Configures the Kubernetes Cluster Receiver to collect cluster-level metrics.
   # Adds the k8s_cluster receiver to the metrics pipeline
   # and adds the necessary rules to ClusteRole.
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.51.1
+version: 0.51.2
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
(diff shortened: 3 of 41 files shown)


charts/opentelemetry-collector/templates/clusterrole.yaml
```diff
@@ -42,4 +42,9 @@ rules:
     resources: ["nodes/stats"]
     verbs: ["get", "watch", "list"]
   {{- end }}
+  {{- if .Values.presets.kubernetesObjects.enabled }}
```
cjbhaines, line 45: I think you will need to add this into the overall if statement on line 1. You will also need to add it to the first line of the cluster role binding.

I've addressed a similar issue here: https://github.com/open-telemetry/opentelemetry-helm-charts/pull/687

nicolastakashi: Oh cool!
Got it and fixed thanks @cjbhaines

cjbhaines: LGTM!

Allex1:
@nicolastakashi you need to bump the chart version and then run `make generate-examples` in order for the CI to pass

nicolastakashi:
> @nicolastakashi you need to bump the chart version and then run `make generate-examples` in order for the CI to pass. [GUIDE](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/CONTRIBUTING.md)

Done, I'm not sure if it's right all those files being deleted.

TylerHelmuth (commented):
@nicolastakashi thanks for taking on this work.  I just merged in another collector chart change so you'll need to rebase, update the chart version to `0.50.2` (we can do a patch bump for this change), and then run `make generate-examples CHARTS=opentelemetry-collector`.  Also looks like you changed some stuff in the demo chart, you'll want to revert those changes.

charts/opentelemetry-collector/values.yaml
```diff
@@ -40,6 +40,20 @@ presets:
...
+  kubernetesObjects:
+    enabled: false
+    defaultEventsConfig:
+      mode: 'watch'
+      labelSelector: ""
+      fieldSelector: ""
+      interval: 60m
+      resourceVersion: ""
+      namespaces: []
+      group: 'events.k8s.io'
```
dmitryax, line 56: I'm not sure we should introduce another limited configuration interface layer for k8sobjects. We should allow users to introduce new rules, e.g., `watch` and `list`.

We can have a default the most "common" config, which users will be able to override with:

```yaml
config:
  recievers:
    k8sobjects:
      objects:
         ...
```

TylerHelmuth: @dmitryax @nicolastakashi since this is a preset I would opt for default, out-of-the-box content that can be enabled with `.Values.presets.kubernetesObjects.enabled` and keep the `kubernetesObjects` configuration very simple.  If there are extra scenarios we need to allow, we should try to handle it like we do for `logsCollection`'s scenarios.  Our goal with the preset quick, easy-to-use configuration that provides fast, basic value.

Any complex configuration falls outside the expectations for a preset; in those situations the component should be configured manually.

Are there meaningful values we can use as defaults for this component?

nicolastakashi: IMH the values provided are meaningful and provide a default, out-of-the-box content, the extensibility we have under the `defaultEventsConfig` don't change or include any feature on the existing preset, only changes some basic configs like default mode if it will be watch or pull and so on.

TylerHelmuth: Ok I understand now how this default section interacts with the template in _config.tpl.  Since the user is able to add to/modify a preset via `.Values.config` I would vote we don't need a `defaultEventsConfig` section.  If  `label_selector`, `field_selector`, or `namespaces` are pretty important options for the component maybe it is worth it to include those next to `enable`, but we could hard code the rest.  @dmitryax @povilasv what do you think?

nicolastakashi: Don't think they are very important if we don't specify we will have a default behavior, I'm ok with not having those fields available for overriding preset config.

But let's see what the guys think about 😄

dmitryax: @povilasv  I like the suggestion. And calling preset `kubernetesEvents` makes more sense. Collection of k8s objects in pull mode seems like a far less common use case and can always be configured via `config` section.

I'd keep the preset slim:

```yaml
presets:
  kubernetesEvents:
    enabled: true
```

nicolastakashi: Hey folks!
Another push, can you take a look @dmitryax ?

TylerHelmuth (commented):
Looks great.  Please add a new values.yaml to the `CI` folder so we can test this preset.

TylerHelmuth:
Also needs a rebase (probably bc of the chart version/examples).

charts/opentelemetry-collector/values.yaml
```diff
@@ -40,6 +40,12 @@ presets:
   # Best used with mode = daemonset.
   kubernetesAttributes:
     enabled: false
+  # Configures the collector to collect kubernetes objects.
```
dmitryax, line 43:
```suggestion
  # Configures the collector to collect kubernetes events.
```

dmitryax (approved):
Thank @nicolastakashi !

nicolastakashi:
@povilasv indeed, it was by mistake

dmitryax (approved):

PR #689, opened 2023-03-14
debug operator tests

moh-osman3:
[DO NOT MERGE] This branch is just to debug why operator tests fail (have not been able to run locally)


moh-osman3:
Not needed. Will try to run tests locally instead.

PR #693, opened 2023-03-21
[operator] add a prefix for ordering MutatingWebhookConfiguration

mmorev:
Hi there.

Mutating webhooks are called sequentially in alphabetical order, whereas validating webhooks are called independently in parallel.

In some cases MutatingWebhooks should run in specific order. For example, when using Opentelemetry Operator and [Gatekeeper Mutations](https://open-policy-agent.github.io/gatekeeper/website/docs/mutation), I want to modify securityContext for auto-instrumentation initContainer.

But, because of ordered sequential calling of mutating webhooks, gatekeeper mutation is called first, and does nothing, because auto-instrumentation initContainer is not created yet.

This MR implements optional ordering prefix for MutatingWebhookConfiguration. By default, prefix value is empty, and webhook name doesn't change during upgrade from previous chart versions. When set, prefix value is added to webhook name with a hyphen.

MutationWebhookConfiguration list before:
```
$ kubectl get MutatingWebhookConfiguration
NAME                                        WEBHOOKS
gatekeeper-mutating-webhook-configuration   1
opentelemetry-operator-mutation             3
```
After:
```
$ kubectl get MutatingWebhookConfiguration
NAME                                           WEBHOOKS
10-opentelemetry-operator-mutation             3
20-gatekeeper-mutating-webhook-configuration   1
```

### DIFF
--- a/charts/opentelemetry-operator/values.yaml
+++ b/charts/opentelemetry-operator/values.yaml
@@ -155,7 +155,7 @@ admissionWebhooks:
 
   ## Adds a prefix to the mutating webook name.
   ## This can be used to order this mutating webhook with all your cluster's mutating webhooks.
-  orderPrefix: ""
+  namePrefix: ""
 
   ## Customize webhook timeout duration
   timeoutSeconds: 10
--- a/charts/opentelemetry-operator/values.yaml
+++ b/charts/opentelemetry-operator/values.yaml
@@ -153,6 +153,10 @@ admissionWebhooks:
   failurePolicy: Fail
   secretName: ""
 
+  ## Mutating webhooks are called sequentially. In some cases you would like your MutatingWebhooks
+  ## to run in specific order. In such case, add a prefix to all of them to sort alphabetically.
+  orderPrefix: ""
+
   ## Customize webhook timeout duration
   timeoutSeconds: 10
 
--- a/charts/opentelemetry-operator/values.yaml
+++ b/charts/opentelemetry-operator/values.yaml
@@ -153,8 +153,8 @@ admissionWebhooks:
   failurePolicy: Fail
   secretName: ""
 
-  ## Mutating webhooks are called sequentially. In some cases you would like your MutatingWebhooks
-  ## to run in specific order. In such case, add a prefix to all of them to sort alphabetically.
+  ## Adds a prefix to the mutating webook name.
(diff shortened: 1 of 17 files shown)


charts/opentelemetry-operator/values.yaml
```diff
@@ -153,6 +153,10 @@ admissionWebhooks:
   failurePolicy: Fail
   secretName: ""
 
+  ## Mutating webhooks are called sequentially. In some cases you would like your MutatingWebhooks
+  ## to run in specific order. In such case, add a prefix to all of them to sort alphabetically.
+  orderPrefix: ""
```
TylerHelmuth, line 158:
```suggestion
  prefix: ""
```

Since this field truly is a prefix to the name I don't think we need to include `order` in the name.

charts/opentelemetry-operator/values.yaml
```diff
@@ -153,6 +153,10 @@ admissionWebhooks:
   failurePolicy: Fail
   secretName: ""
 
+  ## Mutating webhooks are called sequentially. In some cases you would like your MutatingWebhooks
+  ## to run in specific order. In such case, add a prefix to all of them to sort alphabetically.
```
TylerHelmuth, line 157:
```suggestion
  ## Adds a prefix to the mutating webook name.
  ## This can be used to order this mutating webhook with all your cluster's mutating webhooks.
```

My suggestion is to call out that the ordering is necessary not because this chart has multiple mutating webhooks but because your cluster might.

TylerHelmuth (commented):
@mmorev thank you for this contribution.  Please run `make generate-examples CHARTS=opentelemetry-operator`.

charts/opentelemetry-operator/values.yaml
mmorev: maybe "namePrefix"?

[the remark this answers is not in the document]
mmorev: done

charts/opentelemetry-operator/values.yaml
mmorev: thanks

Allex1 (approved):
after rebase

mmorev:
rebase done

TylerHelmuth (approved):

mmorev:
Thanks for ultra-fast review and merge!

PR #705, opened 2023-03-24
[demo] remove `observability` config section

### LINKED ISSUE #682: [demo] Do we need `observability` section?
Why do we need a separate section to disable/enable observability components? Maybe better to add `enabled` option to each observability component itself? It'll be in line with how regular components can disabled.

puckpuck:
Fixes #682 

This is a breaking change for anyone that disabled one of the four sub-charts before. UPGRADING.md is updated accordingly.


dmitryax (approved):
Thanks, @puckpuck !

dmitryax:
Please rebase

PR #711, opened 2023-03-29
[opentelemetry-collector] Remove default resource values

povilasv:
Pushing this forward since https://github.com/open-telemetry/opentelemetry-helm-charts/pull/645 went stale :)

Closes https://github.com/open-telemetry/opentelemetry-helm-charts/issues/644

Not sure if I should bump patch or minor version?


TylerHelmuth (commented):
Since this is a breaking change we also need the UPGRADING.md changes.  Also the original PR has some changes to `_pod.tpl`.

PR #715, opened 2023-03-30
[demo]:  label and port.name align istio standard

### LINKED ISSUE #714: opentelemetry-demo svc port.name aligns istio naming standard
move issue   https://github.com/open-telemetry/opentelemetry-demo/issues/815 here

istio requires port name with http- or tcp- prefix.

panpan0000:
fix https://github.com/open-telemetry/opentelemetry-helm-charts/issues/714

    

    1. service port.name to align istio naming standard
    2. adjust pod app.kubernetes.io/name to make istio happy

### DIFF
--- a/charts/opentelemetry-demo/Chart.yaml
+++ b/charts/opentelemetry-demo/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: opentelemetry-demo
-version: 0.20.4
+version: 0.21.0
 description: opentelemetry demo helm chart
 home: https://opentelemetry.io/
 sources:
--- a/charts/opentelemetry-demo/examples/bring-your-own-observability/rendered/component.yaml
+++ b/charts/opentelemetry-demo/examples/bring-your-own-observability/rendered/component.yaml
@@ -5,10 +5,12 @@ kind: Service
 metadata:
   name: example-adservice
   labels:
-    helm.sh/chart: opentelemetry-demo-0.20.4
-    app.kubernetes.io/name: example
+    helm.sh/chart: opentelemetry-demo-0.21.0
+    
+    opentelemetry.io/name: example-adservice
     app.kubernetes.io/instance: example
     app.kubernetes.io/component: adservice
+    app.kubernetes.io/name: example-adservice
     app.kubernetes.io/version: "1.3.1"
     app.kubernetes.io/part-of: opentelemetry-demo
     app.kubernetes.io/managed-by: Helm
@@ -16,12 +18,11 @@ spec:
   type: ClusterIP
   ports:
     - port: 8080
-      name: service
+      name: tcp-service
       targetPort: 8080
   selector:
-    app.kubernetes.io/name: example
-    app.kubernetes.io/instance: example
-    app.kubernetes.io/component: adservice
+    
+    opentelemetry.io/name: example-adservice
 ---
 # Source: opentelemetry-demo/templates/component.yaml
 apiVersion: v1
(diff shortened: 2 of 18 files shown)


panpan0000:
@puckpuck  combined #716 here.

panpan0000:
> Please bump the chart's Minor number and then run `make generate-examples CHARTS=opentelemetry-demo`.

thx.done

charts/opentelemetry-demo/templates/_objects.tpl
```diff
@@ -10,10 +10,12 @@ spec:
   selector:
     matchLabels:
       {{- include "otel-demo.selectorLabels" . | nindent 6 }}
+      app.kubernetes.io/name: {{ include "otel-demo.name" . }}-{{ .name }}
```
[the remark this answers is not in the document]
panpan0000: Many thanks for reminder

TylerHelmuth (commented):
This since this PR changes selectorLabels its considered a breaking change and we need an accompanying UPGRADING.md entry.  

@puckpuck what are your thoughts?

panpan0000:
> This since this PR changes selectorLabels its considered a breaking change and we need an accompanying UPGRADING.md entry.
> 
> @puckpuck what are your thoughts?

Hmm... you are right.. @TylerHelmuth.

it will block direct `helm upgrade` for existing deployments. 

Error will be  
```
 spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{..... "app.kubernetes.io/name":"..."}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
``` 

What's our strategy and solution ?

panpan0000:
> 

Thanks for advice. But a pity that the problem is against the `deployment` instead of the `service`.

puckpuck:
I'm ok with breaking an upgrade path with proper notes in UPGRADING.md.

The Demo shouldn't be used as an actual production workload; thus a short downtime is ok.

puckpuck:
I would prefer to use different selectors for services that contain fewer labels.

charts/opentelemetry-demo/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: opentelemetry-demo
-version: 0.20.0
+version: 0.20.1
```
puckpuck, line 4: Since this is a breaking change, bump the minor version instead of patch.

```suggestion
version: 0.21.0
```

panpan0000:
> I would prefer to use different selectors for services that contain fewer labels.

I've updated the PR.  please help to review again.

with my changes, the new deployment looks like  below (service selector is similar)

```
apiVersion: apps/v1
kind: Deployment 
spec:
  selector:
    matchLabels:     # single selector
      open-telemetry.io/name: example-cartservice        # new label
  template:
    metadata:       # pod metadata below
      labels:
        opentelemetry.io/name: example-cartservice         # new label
        app.kubernetes.io/component: cartservice
        app.kubernetes.io/name: example-cartservice
```

charts/opentelemetry-demo/examples/bring-your-own-observability/rendered/component.yaml
```diff
@@ -5,130 +5,119 @@ kind: Service
 metadata:
   name: example-adservice
   labels:
-    helm.sh/chart: opentelemetry-demo-0.20.2
-    app.kubernetes.io/name: example
-    app.kubernetes.io/instance: example
-    app.kubernetes.io/component: adservice
+    helm.sh/chart: opentelemetry-demo-0.21.0
+    
+    opentelemetry.io/name: example-adservice
```
TylerHelmuth, line 10: since `/instance` and `/component` were moved to the new helper the service labels changed.  I think the service template needs updated like the deployment template was.

puckpuck (approved):
Tested, and everything works great on a fresh install. TYVM!

charts/opentelemetry-demo/examples/bring-your-own-observability/rendered/component.yaml
panpan0000: Thanks @TylerHelmuth .
it's fixed in latest commit as below ( change one line in `charts/opentelemetry-demo/templates/_helpers.tpl`)

```
     {{/*
     Common labels
     */}}
     {{- define "otel-demo.labels" -}}
     ...
     {{ include "otel-demo.selectorLabels" . }}
+  {{ include "otel-demo.workloadLabels" . }}
     ...
     {{- end }}
```

TylerHelmuth (approved):

PR #716, opened 2023-03-30
adjust pod app.kubernetes.io/name to make istio happy

panpan0000:
background:  it will be very glad that the demo works together with istio.

rational:

refer to https://github.com/istio/istio/issues/10724
istio tell workload by `app.kubernetes.io/name` label as first priority, then use `app` label.
for different deployment, the `app.kubernetes.io/name` should be different.

That's why I made this change.

@kebe7jun to help to clarify on istio side if any question.


panpan0000:
> Can you combine this with #715 to get all the Istio stuff into a single PR?

sure. done.

Issue #719, opened 2023-03-31
[operator] pod not rolling pod when cert be updated


JaredTan95:
when cert file content is updated by `helm upgrade`,  any request(CR apply) to the operator  and if the operator pod does not rolling pod or certwatcher does not update TLS in time, it ocuurs errors in operator `manager`:

```bash
TLS handshake error from 100.117.128:48732: remote error: tls: bad cetificate

```

PR #720, opened 2023-03-31
Fix operator tls error when cert file update

JaredTan95:
close https://github.com/open-telemetry/opentelemetry-helm-charts/issues/719

ref helm docs : https://v3.helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments

we can use following annotation to make operator rolling:
```bash
      annotations:
        rollme: {{ randAlphaNum 5 | quote }}
```


Allex1:
@JaredTan95 ci complaining about `make generate-examples`

JaredTan95:
Some wired, my PR already re-rendered by `make generate-examples` and I re-run it again, but there is no difference  with my PR..

JaredTan95:
> my PR was already re-rendered by `make generate-examples` and I re-run it again, but there is no difference with my PR..

Found the reason, the `roleName` annotation will genarate new value in each `helm` command. So,  a hack method is ignore the `roleName` during `make check-examples`. :-(

JaredTan95:
BTW, I will also investigate if the `tls` update interval time is configurable, If so , we don;t need add `roleName` annotation.

charts/opentelemetry-operator/templates/deployment.yaml
```diff
@@ -15,6 +15,7 @@ spec:
   template:
     metadata:
       annotations:
+        rollme: {{ randAlphaNum 5 | quote }}
```
TylerHelmuth, line 18: Does this annotation always need to be preset or is this scenario specific to the way you're managing certificates?  I'm trying to understand if this annotations is needed for everybody or if it could be handled case-by-case by updating the `podAnnotations` configuration.

JaredTan95: oh, your'e right, make `rollme: {{ randAlphaNum 5 | quote }}` as configurable by `podAnnotations`  👍

JaredTan95: I update according to your idea, but  we still need to ignore `roleName ` during `make check-examples` stage.

JaredTan95:
> I like this new configuration. I don't think we need a new example for this tho as the new configuration is straight forward and new examples are costly during development (hello merge conflicts). Can you remove your new example?

ok

Allex1 (approved):

TylerHelmuth (approved):

PR #723, opened 2023-04-01
[collector] Add support for NetworkPolicy

### LINKED ISSUE #722: [opentelemetry-collector] Add Support for NetworkPolicy
I'd like to propose adding [NetworkPolicy](https://kubernetes.io/docs/concepts/services-networking/network-policies/) support to the helm chart.

There are a few different use cases I can see:

* In `DaemonSet` mode, restrict the ability to access the collector to only cluster nodes or specific namespaces.
* In `Deployment`/`StatefulSet` mode, restrict collection to specific namespaces.

This would help a lot in a multi-tenant environment. The primary motivation of having the `NetworkPolicy` in this chart is to keep the object installed/updated along side the collector.

I'm willing to write up a PR with this if the team thinks this would be useful/accepted!

fernferret:
Adds NetworkPolicy support to the collector chart.

Fixes #722

## Questions for reviewers

~~When `networkPolicy` is enabled, I decided to include a default policy of "all namespaces".~~

Per the [discussion below](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/723#issuecomment-1495841280), when `networkPolicy` is enabled with no other arguments, a `NetworkPolicy` object will be created, but traffic will not be restricted until `allowIngressFrom` is set to a `from` clause of a network policy

```yaml
networkPolicy:
  enabled: true
```

Here is an example that would restrict traffic to only allow ingress from pods in the cluster:

```yaml
networkPolicy:
  enabled: true
  allowIngressFrom:
    # Allow traffic from any pod in any namespace, but not external hosts
    - namespaceSelector: {}
```

### DIFF
--- a/charts/opentelemetry-collector/ci/networkpolicy-override-values.yaml
+++ b/charts/opentelemetry-collector/ci/networkpolicy-override-values.yaml
@@ -1,0 +1,30 @@
+mode: daemonset
+resources:
+  limits:
+    cpu: 100m
+    memory: 200M
+
+networkPolicy:
+  enabled: true
+
+  allowIngressFrom:
+    - namespaceSelector: {}
+    - ipBlock:
+        cidr: 127.0.0.1/32
+
+  extraIngressRules:
+    - ports:
+      - port: metrics
+        protocol: TCP
+      from:
+        - ipBlock:
+            cidr: 127.0.0.1/32
+
+  egressRules:
+    - to:
+        - podSelector:
+            matchLabels:
+              app: jaeger
+      ports:
+        - port: 4317
+          protocol: TCP
--- a/charts/opentelemetry-collector/templates/networkpolicy.yaml
+++ b/charts/opentelemetry-collector/templates/networkpolicy.yaml
@@ -1,0 +1,38 @@
+{{- if .Values.networkPolicy.enabled }}
+kind: NetworkPolicy
+apiVersion: networking.k8s.io/v1
+metadata:
+  name: {{ include "opentelemetry-collector.fullname" . }}
+  labels:
+    {{- include "opentelemetry-collector.labels" . | nindent 4 }}
+  {{- if .Values.networkPolicy.annotations }}
+  annotations:
+    {{- range $key, $value := .Values.networkPolicy.annotations }}
+    {{- printf "%s: %s" $key (tpl $value $ | quote) | nindent 4 }}
+    {{- end }}
+  {{- end }}
+spec:
+  podSelector:
+    matchLabels:
+      {{- include "opentelemetry-collector.selectorLabels" . | nindent 6 }}
+      {{- include "opentelemetry-collector.component" . | nindent 6 }}
+  ingress:
+    - ports:
+        {{- range $port := .Values.ports }}
+        {{- if $port.enabled }}
+        - port: {{ $port.containerPort }}
+          protocol: {{ $port.protocol }}
+        {{- end }}
+        {{- end }}
+      {{- if .Values.networkPolicy.allowIngressFrom }}
+      from:
+        {{- toYaml .Values.networkPolicy.allowIngressFrom | nindent 8 }}
+      {{- end }}
+    {{- if .Values.networkPolicy.extraIngressRules }}
+    {{- toYaml .Values.networkPolicy.extraIngressRules | nindent 4 }}
+    {{- end }}
+  {{- if .Values.networkPolicy.egressRules }}
+  egress:
+    {{- toYaml .Values.networkPolicy.egressRules | nindent 4 }}
+  {{- end }}
+{{- end }}
--- a/charts/opentelemetry-collector/values.yaml
+++ b/charts/opentelemetry-collector/values.yaml
@@ -432,3 +432,55 @@ statefulset:
   # volumeClaimTemplates for a statefulset
   volumeClaimTemplates: []
   podManagementPolicy: "Parallel"
+
+networkPolicy:
+  enabled: false
+
+  # Annotations to add to the NetworkPolicy
+  annotations: {}
+
+  # Configure the 'from' clause of the NetworkPolicy.
+  # By default this will restrict traffic to ports enabled for the Collector. If
+  # you wish to further restrict traffic to other hosts or specific namespaces,
+  # see the standard NetworkPolicy 'spec.ingress.from' definition for more info:
+  # https://kubernetes.io/docs/reference/kubernetes-api/policy-resources/network-policy-v1/
+  allowIngressFrom: []
+  # # Allow traffic from any pod in any namespace, but not external hosts
+  # - namespaceSelector: {}
+  # # Allow external access from a specific cidr block
+  # - ipBlock:
+  #     cidr: 192.168.1.64/32
+  # # Allow access from pods in specific namespaces
+  # - namespaceSelector:
+  #     matchExpressions:
+  #       - key: kubernetes.io/metadata.name
+  #         operator: In
+  #         values:
+  #           - "cats"
+  #           - "dogs"
+
+  # Add additional ingress rules to specific ports
+  # Useful to allow external hosts/services to access specific ports
+  # An example is allowing an external prometheus server to scrape metrics
+  #
+  # See the standard NetworkPolicy 'spec.ingress' definition for more info:
+  # https://kubernetes.io/docs/reference/kubernetes-api/policy-resources/network-policy-v1/
+  extraIngressRules: []
+  # - ports:
+  #   - port: metrics
+  #     protocol: TCP
+  #   from:
+  #     - ipBlock:
+  #         cidr: 192.168.1.64/32
+
+  # Restrict egress traffic from the OpenTelemetry collector pod
+  # See the standard NetworkPolicy 'spec.egress' definition for more info:
+  # https://kubernetes.io/docs/reference/kubernetes-api/policy-resources/network-policy-v1/
+  egressRules: []
+  #  - to:
+  #      - namespaceSelector: {}
+  #      - ipBlock:
+  #          cidr: 192.168.10.10/24
+  #    ports:
+  #      - port: 1234
+  #        protocol: TCP
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.52.1
+version: 0.53.0
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-collector/ci/networkpolicy-values.yaml
+++ b/charts/opentelemetry-collector/ci/networkpolicy-values.yaml
@@ -1,0 +1,8 @@
+mode: deployment
+resources:
+  limits:
+    cpu: 100m
+    memory: 200M
+
+networkPolicy:
+  enabled: true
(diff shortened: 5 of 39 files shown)


fernferret:
Whoopsie, fixed the whitespace issues... not sure how those crept in. My local run of `ct` passes now:

```
ct lint --target-branch main --chart-repos opentelemetry-collector=https://open-telemetry.github.io/opentelemetry-helm-charts --chart-repos prometheus=https://prometheus-community.github.io/helm-charts --chart-repos grafana=https://grafana.github.io/helm-charts --chart-repos jaeger=https://jaegertracing.github.io/helm-chart
...
1 chart(s) linted, 0 chart(s) failed
------------------------------------------------------------------------------------------------------------------------
 ✔︎ opentelemetry-collector => (version: "0.52.0", path: "charts/opentelemetry-collector")
------------------------------------------------------------------------------------------------------------------------
All charts linted successfully
```

charts/opentelemetry-collector/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,39 @@
...
+        {{- toYaml .Values.networkPolicy.allowIngressFrom | nindent 8 }}
+      {{- end }}
+    {{- if .Values.networkPolicy.extraIngressRules }}
+    {{- toYaml .Values.networkPolicy.extraIngressRules | nindent 4 }}
+    {{- end }}
+  {{- if .Values.networkPolicy.egressRules }}
+  egress:
+    {{- toYaml .Values.networkPolicy.egressRules | nindent 4 }}
+  {{- end }}
+{{- end }}
```
TylerHelmuth, line 39:
```suggestion
{{- end }}

```

charts/opentelemetry-collector/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,39 @@
...
+spec:
+  podSelector:
+    matchLabels:
+      {{- include "opentelemetry-collector.selectorLabels" . | nindent 6 }}
+      {{- include "opentelemetry-collector.component" . | nindent 6 }}
+  ingress:
+    - ports:
+        {{- range $key, $port := .Values.ports }}
+        {{- if $port.enabled }}
+        # {{ $key }}
```
TylerHelmuth, line 23: Can we remove this comment if not needed?

TylerHelmuth (commented):
I think the decision you laid out in the description makes sense, thanks for the thoughtful write-up.  

Can you add a new test scenario to the CI folder that enables this new capability?

Please rebase and then run `make generate-examples CHARTS=opentelemetry-collector`.

fernferret:
@TylerHelmuth Thank you for the feedback. These are all great suggestions. I have the patches made and rebased and will push very soon.

Before I get this pushed, I did have one question about what you asked for:

> Can you add a new test scenario to the CI folder that enables this new capability?

Is there something else to do other than add new `values.yaml` files to the `ci` dir? I added 2 new files (), one to showcase the default capability, and one with a bit more customization. 

I've not used `ct` before, is there any guidance or documentation on how I could add these or verify the tests are working?

TylerHelmuth:
@fernferret Adding new files, one per test scenario, is all that's needed.

charts/opentelemetry-collector/ci/networkpolicy-override-values.yaml
fernferret, line 1: @TylerHelmuth I apologize for my confusion, but are these the `ci` files you requested be added? I believe these were part of the original PR. If they're not I'm happy to change/move them.

TylerHelmuth: I'm not sure what you mean by "original PR", but yes, these are the test scenarios I'm looking for.

TylerHelmuth: Oh were these part of the first commit and I missed them?

fernferret: > Oh were these part of the first commit and I missed them?

I believe they were, no worries at all I just wanted to make sure I was following all the feedback! I'll get the rebased/`make`'d changes pushed soon!

charts/opentelemetry-collector/templates/networkpolicy.yaml
fernferret: Done, I also removed the `$key` as that's just extra if the comment is gone.

TylerHelmuth:
@povilasv I actually like the idea that we can provide a default policy of all kubernetes pods/hosts and exclude external traffic.  For the collector, especially daemonset mode, its reasonable to expect that the only thing communicating with it is traffic from inside the cluster.  The default value is also very easy to override.  

All that said, I don't have experience using NetworkPolicy myself. If it isn't standard for a chart to provide a default value the we could remove it.  @dmitryax what do you think?

fernferret:
After diving a bit deeper, I agree with @povilasv and think the default values file should contain the following. I also like providing reasonable alternatives as commented out examples.

```yaml
networkPolicy:
  enabled: false

  # Annotations to add to the NetworkPolicy
  annotations: {}

  # Configure the 'from' clause of the NetworkPolicy
  # By default this will restrict traffic to the listed ports only.
  # If you wish to further restrict traffic to other hosts or specific
  # namespaces, see the standard NetworkPolicy 'spec.ingress.from'
  # definition for more info:
  # https://kubernetes.io/docs/reference/kubernetes-api/policy-resources/network-policy-v1/
  allowIngressFrom: []
  # Allow traffic from any pod in any namespace, but not external hosts
  # - namespaceSelector: {}
```

---

There's been some really good discussion and feedback from both of y'all that's caused me to really dig into `NetworkPolicy`.  I'm using them to restrict tenant access to the `DaemonSet`, but I definitely want to make sure this PR breaks as little as possible and works for people!

## About NetworkPolicy

The following 3 bullets (extracted for readability) come from the [K8S API describing the `NetworkPolicy`'s `spec.ingress.from`](https://kubernetes.io/docs/reference/kubernetes-api/policy-resources/network-policy-v1/#NetworkPolicySpec) which when combined paint a much better picture:

* If this field (`spec.ingress.from`) is empty or missing, this rule matches all sources (traffic not restricted by source).
* If this field (`spec.ingress.from`) is present and contains at least on item, this rule allows traffic only if the traffic matches at least one item in the from list.
* If this field (`spec.ingress.ports`) is present and contains at least one item, then this rule allows traffic only if the traffic matches at least one port in the list.

So, given the following policy:

```yaml
spec:
  ingress:
  - ports:
    - port: 4318
      protocol: TCP
    - port: 4317
      protocol: TCP
```

I believe this means that both a pod and an external user, would be able to access the collector on only ports `4317` and `4318`. However, if a malicious actor somehow figured out how to start up another application on another port (let's say `1234`) that port would not be accessible inside or out.

This also means that external pods won't be able to hit the health port `13113` unless it is exposed (which is available to advanced users with `extraIngressRules`), however I think this makes for a sensible default as `13113` is not bound to the Pod IP as far as I see.

Since we're also dealing with `hostPort`s I'd also like the ability to restrict things further (as described in the PR/issue description).

This example would allow only traffic from inside the cluster, thus allowing all pods within to talk to the collector, but not my development laptop, even in DaemonSet mode:

```yaml
spec:
  ingress:
  - ports:
    - port: 4318
      protocol: TCP
    - port: 4317
      protocol: TCP
    from:
    - namespaceSelector: {}
```

And finally, if someone wanted to allow all pods within the cluster **and** a subset of development nodes, which is my use case:

```yaml
spec:
  ingress:
  - ports:
    - port: 4318
      protocol: TCP
    - port: 4317
      protocol: TCP
    from:
    - namespaceSelector: {}
    - ipBlock:
        cidr: 192.168.1.0/24
```

After thinking through and reviewing these 3 use cases, I think the first one, with no default `from` clause as @povilasv suggested, makes the most sense as it does provide some guarantees that no other ports will be exposed.

## Relevant Charts

I looked at the prometheus chart and it actually has a reasonably stock [NetworkPolicy](https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus/templates/network-policy.yaml).

I followed is lineage back to: https://github.com/prometheus-community/helm-charts/commit/ac9c25afcd2268897d586ddbae39f36b82240192.

I would certainly like to see some more options added to the Prometheus chart, but it does point at the fact that:

* network policy came from a place where some portions of it were defined with annotations
* for many deployments, network policy is targeted at ensuring only the known ports are exposed

## Test results

Here are some results from my kubeadm provisioned dev cluster, just so I could verify my own understanding.

<details>
<summary>Show Test Results</summary>

### No NetworkPolicy

Allow traffic from all pods and hosts, to all ports

Ports active: `4318`, `4317`

```
$ curl --connect-timeout 5 http://farnsworth:4318
404 page not found
$ curl --connect-timeout 5 http://farnsworth:4317
curl: (1) Received HTTP/0.9 when not allowed       <-- traffic made it to the pod
```

### Empty NetworkPolicy `from` clause

Allow traffic from all pods and hosts, but only to 4317

Ports active: `4318`, `4317`

```yaml
spec:
  ingress:
  - ports:
    - port: 4318
      protocol: TCP
```

```
$ curl --connect-timeout 5 http://farnsworth:4318
404 page not found
$ curl --connect-timeout 5 http://farnsworth:4317
curl: (28) Connection timeout after 5002 ms
```

### Only allow internal traffic

Allow traffic from all pods to 4318, but no external hosts. No traffic allowed to 4317.

Ports active: `4318`, `4317`

```yaml
spec:
  ingress:
  - ports:
    - port: 4318
      protocol: TCP
    from:
    - namespaceSelector: {}
```

```
$ curl --connect-timeout 5 http://farnsworth:4318
curl: (28) Connection timeout after 5004 ms
$ curl --connect-timeout 5 http://farnsworth:4317
curl: (28) Connection timeout after 5004 ms
```

### Allow internal + development laptop

Allow traffic from all pods and hosts to 4318. No traffic allowed to 4317.

Ports active: `4318`, `4317`

```yaml
spec:
  ingress:
  - ports:
    - port: 4318
      protocol: TCP
    from:
    - namespaceSelector: {}
    - ipBlock:
        cidr: 192.168.1.0/24
```

```
$ curl --connect-timeout 5 http://farnsworth:4318
404 page not found
$ curl --connect-timeout 5 http://farnsworth:4317
curl: (28) Connection timeout after 5003 ms
```

</details>

## Wrap Up

I appreciate y'all continuing the discussion and look forward to your advice.

charts/opentelemetry-collector/values.yaml
```diff
@@ -432,3 +432,56 @@ statefulset:
...
+
+networkPolicy:
+  enabled: false
+
+  # Annotations to add to the NetworkPolicy
+  annotations: {}
+
+  # Configure the 'from' clause of the NetworkPolicy
+  # By default this has no effect in Deployment or StatefulSet modes
+  # In DaemonSet mode, this will restrict access to pods (no external hosts)
```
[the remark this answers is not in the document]
fernferret: Thanks for pointing this out, I really dislike how I phrased this and have proposed a change in the main comment thread.

TylerHelmuth:
@fernferret thanks again for the beautiful writes, it makes reviewing and discussing a new feature so much easier.  I wish all the PRs I reviewed had this level of investment and presentation haha

Your extra digging and explain, along with Prometheus as an example, feel like a good argument for not providing any default `from`.  I support the change suggested by you and @povilasv.

charts/opentelemetry-collector/values.yaml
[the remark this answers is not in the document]
fernferret: This comment has been removed.

fernferret:
I appreciate the feedback!

I've made the requested changes (as well as reworded the comment block within the `networkPolicy` section) and they're pushed for review.

TylerHelmuth (approved):

TylerHelmuth:
@fernferret needs one more rebase (I merged the other PR first)

puckpuck:
Would network policy allow the collector to work with `kubectl port-forward ...` or would we still need to force it to bind to 127.0.0.1?

PR #724, opened 2023-04-01
[doc] Add a note about port-forwarding

fernferret:
Add a note to the UPGRADING document about how to enable port-forwarding after chart release 0.47.0.

This is [related to the discussion](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/603#issuecomment-1483513610) in PR #603

### DIFF
--- a/charts/opentelemetry-collector/UPGRADING.md
+++ b/charts/opentelemetry-collector/UPGRADING.md
@@ -11,6 +11,39 @@ The chart will continue to allow complete configuration of the Collector via the
 
 See [Security Best Practices docummentation](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/security-best-practices.md#safeguards-against-denial-of-service-attacks) for more details.
 
+The new default of binding to the pod IP, rather than `0.0.0.0`, will cause `kubectl port-forward` to fail. If port-forwarding is desired, the following `value.yaml` snippet will allow the Collector bind to `127.0.0.1` inside the pod, in addition to the pod's IP:
+
+```yaml
+config:
+  receivers:
+    jaeger/local:
+      protocols:
+        grpc:
+          endpoint: 127.0.0.1:14250
+        thrift_compact:
+          endpoint: 127.0.0.1:6831
+        thrift_http:
+          endpoint: 127.0.0.1:14268
+    otlp/local:
+      protocols:
+        grpc:
+          endpoint: 127.0.0.1:4317
+        http:
+          endpoint: 127.0.0.1:4318
+    zipkin/local:
+      endpoint: 127.0.0.1:9411
+  service:
+    pipelines:
+      traces:
+        receivers:
+        - otlp
+        - otlp/local
+        - jaeger
+        - jaeger/local
+        - zipkin
+        - zipkin/local
+```
+
 ## 0.40.7 to 0.41.0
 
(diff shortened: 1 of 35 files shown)


fernferret:
Wow I totally didn't realize `UPGRADING.md` was packed in the chart, sorry about the multiple failed runs! I think we should be looking better now!

charts/opentelemetry-collector/UPGRADING.md
```diff
@@ -11,6 +11,39 @@ The chart will continue to allow complete configuration of the Collector via the
...
+  service:
+    pipelines:
+      traces:
+        receivers:
+        - otlp
+        - otlp/local
+        - jaeger
+        - jaeger/local
+        - zipkin
+        - zipkin/local
```
TylerHelmuth, line 44: Instead of adding new receivers lets update the default receivers to keep the example shorter and simpler

TylerHelmuth (commented):
Thanks for updating these docs. Please rebase, bump the chart patch again, and then run `make generate-examples CHARTS=opentelemetry-collector`.

charts/opentelemetry-collector/UPGRADING.md
fernferret: @TylerHelmuth

The main reason I opened this documentation PR was because I had issues using the chart in my lab under both of the following conditions:

* Software running configured to point to the collector inside the cluster (`http://otel-col.otel:4317`)
* Me doing testing on my dev box port-forwarding in to try out otel locally and verify configs were working

I'm happy to simplify this example but I'm struggling to find the wording to convey the ability to have both Pod binding and localhost.

Perhaps we move this config as a full values file to the examples directory and just leave a much simpler note. Something along the lines of:

>The new default of binding to the pod IP, rather than `0.0.0.0`, will cause `kubectl port-forward` to fail. A config is included in the `examples` that will configure the collector to bind to both the Pod IP and `127.0.0.1`.

This would keep the UPGRADING a lot cleaner and provide folks with a complete reference.

TylerHelmuth: I didn't realize the goal was to allow both at the same time.  With that in mind, this is appropriate.

TylerHelmuth (approved):

Allex1 (approved):

charts/opentelemetry-collector/UPGRADING.md
```diff
@@ -11,6 +11,39 @@ The chart will continue to allow complete configuration of the Collector via the
 
 See [Security Best Practices docummentation](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/security-best-practices.md#safeguards-against-denial-of-service-attacks) for more details.
 
+The new default of binding to the pod IP, rather than `0.0.0.0`, will cause `kubectl port-forward` to fail. If port-forwarding is desired, the following `value.yaml` snippet will allow the Collector bind to `127.0.0.1` inside the pod, in addition to the pod's IP:
```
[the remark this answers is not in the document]
TylerHelmuth: "... makes kubectl port-forward to fail" isn't proper.  Lets stick with "will cause" for that part of the sentence.  Also lets keep the "in addition to the pod's IP" snippet.  Without that snippet there is no reason for the example config to add new named receivers.

PR #731, opened 2023-04-06
[demo] Add "app" labels for when integrated with Istio

Leandroschwab:
I added explicit "app" labels to the Kubernetes Helm Deployment , as recommended by Istio. These labels add contextual information to the metrics and telemetry that Istio collects, which will help with visualization and analysis of the data.  Using these labels in conjunction with tools like Kiali and calisti can create nicer visualizations of the system topology, making it easier to understand and analyze the relationships between different components of our system.

ref:
https://istio.io/latest/docs/ops/deployment/requirements/


puckpuck:
I think #715 is addressing this as well

puckpuck:
I'm curious as to why Istio recommends `app` and `version`, instead of the label keys [recommended by Kubernetes](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels) itself.

Leandroschwab:
Hi, I'm investigating further, this change seems to only affect calisti, I'll debug further to see how to get the service and pod visually together in Kiali.

Leandroschwab:
I think I figured out what was wrong with Kiali working. Apparently the labels "app.kubernetes.io/name" and "app" must be the same to display correctly in the graph view.

This image show the result I want to have when viewing the topology. 
kiali-withot-app-label

and with the "app" label

kiali-with-app-label

I will update my fix my commit with this fix.

puckpuck:
So is this a bug in Kiali needing to understand the recommended labels from Kubernetes? 

I'm hesitant to add labels to support a specific project that doesn't conform to the recommended approach.

TylerHelmuth:
> I'm hesitant to add labels to support a specific project that doesn't conform to the recommended approach.
Agreed

PR #733, opened 2023-04-06
[demo] v1.4 release

puckpuck:
Takes in the v1.4 demo release and some changes associated with those images.

- Kafka memory changes and handling of metrics from some components.
- Uses new `connectors` in the collector functionality for spanmetrics


puckpuck:
I can confirm that this PR works great with the 1.4.0-beta release. Installed using the following values

```yaml
default:
  image:
    tag: 1.4.0-beta
```

We expect to release Demo 1.4.0 early next week.

charts/opentelemetry-demo/values.yaml
```diff
@@ -641,16 +644,31 @@ opentelemetry-collector:
...
+          - context: metric
+            statements:
+              - set(description, "Measures the duration of inbound HTTP requests") where name == "http.server.duration"
+      # we can remove when this is fixed: https://github.com/open-telemetry/opentelemetry-demo/issues/826
+      filter:
+        metrics:
+          exclude:
+            match_type: strict
+            metric_names:
+              - queueSize
```
TylerHelmuth, line 659: Let's use the filterprocessor config that uses OTTL since the other config will be deprecated eventually.

TylerHelmuth: It'll be something like

```
processors:
  filter/ottl:
    error_mode: ignore
    metrics:
      metric:
          - 'name == "queueSize"'
```

TylerHelmuth (commented):
~~@puckpuck if this version supports arm64 we can remove the arm64 exclusion in the README.~~

Nevermind that was 1.3

TylerHelmuth (approved):

PR #734, opened 2023-04-06
[demo] add grafana dashboards

puckpuck:
Adds the [Spanmetrics](https://github.com/open-telemetry/opentelemetry-demo/pull/787) and [OpenTelemetry Collector Data Flow](https://github.com/open-telemetry/opentelemetry-demo/pull/797) dashboards from the Demo repo.

Because we have multiple dashboards, and managing them in a single file is problematic, I used the `(.Files.Glob "grafana-dashboards/*.json").AsConfig` construct and store all the dashboard files in the respective subfolder instead.


TylerHelmuth (approved):
Side note: we should replace the spanmetricsprocessor with the spanmetricsconnector.

puckpuck:
> Side note: we should replace the spanmetricsprocessor with the spanmetricsconnector.

We plan to do so after Connectors get out of alpha state and don't require a feature gate to be enabled.

dmitryax:
Going forward, maybe we can automate fetching the dashboards from the demo repo instead copy-pasting them? We solved a similar issue for the operator with `update-operator-crds` make target

PR #735, opened 2023-04-06
[demo] fix kubectl port-forward for collector

puckpuck:
Fixes [this issue](https://github.com/open-telemetry/opentelemetry-demo/issues/794) in the Demo repo

The otlp/http receiver will bind to 127.0.0.1 so that `kubectl port-forward` will work.


charts/opentelemetry-demo/values.yaml
```diff
@@ -610,6 +610,8 @@ opentelemetry-collector:
       otlp:
         protocols:
           http:
+            # Bind to 127.0.0.1 to allow for `kubectl port-forward` to work
+            endpoint: 127.0.0.1:4318
```
TylerHelmuth, line 614: Do we need to modify grpc's endpoint as well?

puckpuck: No. We only ask people to port-forward the HTTP port so you can send telemetry to it via a web browser.

TylerHelmuth (approved):

PR #738, opened 2023-04-11
Add support for adding annotations to the cert-manager cert and issuer objects

jvoravong:
Description:
My company has a use case where adding annotations to the cert and issuer created by the cert-manager would be very useful. This shouldn't affect previous behavior. Let me know if I missed anything.

Testing:
Tested manually against a Kubernetes 1.26 cluster.


Allex1:
@jvoravong can you also update the [values schema](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/values.schema.json)  ? Otherwise lgtm

dmitryax (approved):

Issue #217, opened 2022-05-27
Posibility to use daemonset along with the service.


theodrim:
Hello. I was wondering is it possible to have mode set to `daemonset` and have service available at the same time?
I'm asking about opentelemetry-collector helm chart.

TylerHelmuth:
The charts do not currently support a service with mode=daemonset. What use case did you have in mind?

theodrim:
@TylerHelmuth replacing `node-exporter` but since our cluster using spot instances - it would be reasonable to use daemonset to assume that every new node gets `opentelemetry-collector` deployed.
Unless I'm missing something and metrics still can be scraped even without the service?

TylerHelmuth:
> Unless I'm missing something and metrics still can be scraped even without the service?

I might have interpreted this question incorrectly. Are you referring to scraping the collectors metrics on itself 
 (0.0.0.0:8888/metrics) ?

theodrim:
> Are you referring to scraping the collectors metrics on itself
(0.0.0.0:8888/metrics) ?

Probably. I've initially though that it is possible to have only agent exposing metrics by itself (as having `mode: deployment` and service enabled giving desired results) but now realizing that collector might be required to have external scraping done by prometheus at first and vmagent later.

Is my understanding correct, that having `opentelemetry-agent` with `mode: daemonset` and `opentelemetry-collector` with enabled prometheus receiver as well as pipeline will get desired results?

TylerHelmuth:
I think it would be possible for a deployment instance of the collector to scrape the telemetry of the agent instances, but the current charts set up the collector to scrape itself: https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/examples/daemonset-only/rendered/configmap-agent.yaml. When deployed as a daemonset (or a deployment) the default config for the chart is to have the collector scrape its own metrics, which could then be exported anywhere and anyway you want.

kostya9:
I have another use case which could benefit from service for daemonset.

Imagine that apps could be both in a k8s cluster and not (e.g. in the process of migrating or for some other reason some apps are outside of the cluster). As a developer, I want to send tracing data from inside and outside of the cluster to the collectors.

TylerHelmuth:
> As a developer, I want to send tracing data from inside and outside of the cluster to the collectors.

[...]

PR #743, opened 2023-04-14
fix: Enable Service creation on DaemonSet

achetronic:
**Description:**
This change enable Service creation for all the cases, included DaemonSets

**Use case:**
When OTEL Collector is deployed to collect metrics from /metrics using the prometheusreceiver, it is better for performance to be deployed as a DaemonSet. This way you can configure relabel_configs in a way where each OTEL's pod process only targets on its node.

At the same time, you can configure otlp receiver for the metrics too, in the same OTEL Collector. In this case, we need a Service to point the SDKs, even when it is a daemonset

Deployment mode is not suitable in most cases due to all the pods would scrappe all the metrics, so they all suffer the same pressure and consume much more resources in terms of memory

### DIFF
--- a/charts/opentelemetry-collector/templates/service.yaml
+++ b/charts/opentelemetry-collector/templates/service.yaml
@@ -1,4 +1,3 @@
-{{- if or (eq .Values.mode "deployment") (eq .Values.mode "statefulset") (.Values.ingress.enabled) -}}
 apiVersion: v1
 kind: Service
 metadata:
@@ -30,4 +29,4 @@ spec:
   selector:
     {{- include "opentelemetry-collector.selectorLabels" . | nindent 4 }}
     {{- include "opentelemetry-collector.component" . | nindent 4 }}
-{{- end }}
+


TylerHelmuth:
@achetronic typically pods push their traces metrics and logs to their node's collector instance using the `<Node IP>:<port>` since the chart sets up `hostPort` for the collector daemonset.  See https://github.com/open-telemetry/opentelemetry-helm-charts/issues/217 for more details.  Does this fit your needs?

achetronic:
> @achetronic typically pods push their traces metrics and logs to their node's collector instance using the `<Node IP>:<port>` since the chart sets up `hostPort` for the collector daemonset. See #217 for more details. Does this fit your needs?

Hello @TylerHelmuth thanks for your reply

I think you are refering to this possibility

```
containers:
- env:
  - name: NODE_IP
    valueFrom:
      fieldRef:
         status.hostIP
```

TBH, i think this approach could fit most cases, but the problem is that there are some environments where it is easier to set `OTEL_EXPORTER_OTLP_ENDPOINT` env var as a simple string than adding this snippet. For example on companies with Helm chart generators, where would be needed to upgrade the whole generator to cover this

I think it would be easier to use giving this possibility too, just to have more options to reduce technical debt in a progressive way 

WDYT?

puckpuck:
I feel that adding a Service to a Daemonset is a Kubernetes anti-pattern. Perhaps we can provide better documentation on adding the Node name/ip to each Daemonset and leverage that as part of the Otel exporter variable.

achetronic:
> I feel that adding a Service to a Daemonset is a Kubernetes anti-pattern. Perhaps we can provide better documentation on adding the Node name/ip to each Daemonset and leverage that as part of the Otel exporter variable.

Hello there, I don't think this is an antipattern because they are intended for different things: Service resources are only a way to access pods, no matter the way used to deploy them :)

I know if we add some snipped like the following would do the trick:

```
- name: NODE_IP
   valueFrom:
     fieldRef:
       fieldPath: status.hostIP
   value: http://$(NODE_IP):4317
```

But still think that would be so so helpful to give the other possibility to everyone, because it's easier to add something "fixed" like:

```
OTEL_EXPORTER_OTLP_ENDPOINT: http://open-telemetry-collector.open-telemetry-collector.svc.cluster.local:4317
```

The difference seems to be minimal, but most cases, all the people look for the second before the first trick :)

TylerHelmuth:
I feel both 

```
- name: NODE_IP
   valueFrom:
     fieldRef:
       fieldPath: status.hostIP
   value: http://$(NODE_IP):4317
```

and 

```
  value: http://open-telemetry-collector.open-telemetry-collector.svc.cluster.local:4317
```

are fixed entries that could be copied into any deployment.  Is there a scenario where the downward API cannot be used?

achetronic:
> I feel both
> 
> ```
> - name: NODE_IP
>    valueFrom:
>      fieldRef:
>        fieldPath: status.hostIP
> - name: OTEL_EXPORTER_OTLP_ENDPOINT
>    value: http://$(NODE_IP):4317
> ```
> 
> and
> 
> ```
> - name: OTEL_EXPORTER_OTLP_ENDPOINT
>   value: http://open-telemetry-collector.open-telemetry-collector.svc.cluster.local:4317
> ```
> 
> are fixed entries that could be copied into any deployment. Is there a scenario where the downward API cannot be used?

Hello @TylerHelmuth thanks for your answer

TBH, I tested both solutions and they are both working

Let me explain a situation and discuss it: I have the feeling that having a Service in the middle is more resilient. 
I mean, if each node is scrapping its Prometheus metrics (with Prometheus receiver) and at the same time listens for OTLP metrics, and in some point, one of the pods is simply not working (overloaded or whatever), you are loosing all the metrics for that node. 

This means if the pods are not well balanced across the cluster, you are not only loosing metrics from infra (coming from /metrics, usually) but metrics from entire products too (OTLP nowadays). If your SLOs depends on those metrics for products, it is easier to loose the entire Error Budget

However, if you point yor OTLP to a Service and a DS pod is not available, it will call simply to another pod.
With this approach, and having the DS, can increase the reliability from OTLP side, and decrease the costs from Prometheus receiver resources consumption (because it is configured to scrape only infra metrics on its node). Because from my POV, it does not make sense to have a simple deployment and force all pods to scrape all the metrics all the time, this is the reason to have a DS instead

Anyway, as I told, both approaches are working. It is a matter of reliability on the product side from my POV, but if this is not shared by others, may be it does not make sense to do the change :)

WDYT?

TylerHelmuth:
> I have the feeling that having a Service in the middle is more resilient

I believe this is an incorrect feeling.  Kubernetes guarantees the daemonset will run on every node (lets ignore tolerations, node affinity, etc for this scenario), rescheduling other pods to make sure that happens.  Any resource constraints can be handled by resource requests/limits and would be a similar concern with a deployment.

That said, we have had enough people asking for this feature that we'd like to work out a solution that provides it under certain circumstances.  Let's keep our discussion going in https://github.com/open-telemetry/opentelemetry-helm-charts/issues/749.

achetronic:
Right, let's discuss there first :)

Issue #746, opened 2023-04-17
topologySpreadConstraints should be an array


gjacquet:
In opentelemetry-collector Helm chart, `topologySpreadConstraints` should be an array rather than an object as it is an array in Kubernets API: https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/#topologyspreadconstraints-field

This seems to be causing this error when configuring `topologySpreadConstraints`:

```
ERRO[0000] [generateHelm] : values don't meet the specifications of the schema(s) in the following chart(s):
opentelemetry-collector:
- topologySpreadConstraints: Invalid type. Expected: object, given: array  application=opentelemetry-collector-refinery engineMode=standalone environmentConfig=config-poc.yaml operation=generateApplications region=us-west-2 version=
FATA[0000] error rendering templates: values don't meet the specifications of the schema(s) in the following chart(s):
opentelemetry-collector:
- topologySpreadConstraints: Invalid type. Expected: object, given: array 
```

PR #747, opened 2023-04-17
Set topologySpreadConstraints type to array

gjacquet:
Set type to array in order to match type from Kubernetes API specs: https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/#topologyspreadconstraints-field

Addresses #746

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.53.0
+version: 0.53.1
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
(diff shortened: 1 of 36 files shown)


gjacquet:
@Allex1 done!

TylerHelmuth (commented):
@gjacquet can you provide a render of the new values.yaml in action?  I want to make sure the deployment template doesn't need changed.

gjacquet:
In my case it gave something like this:
```yaml
  topologySpreadConstraints:
  - labelSelector:
      matchLabels:
        app.kubernetes.io/instance: opentelemetry-collector
    maxSkew: 1
    topologyKey: topology.kubernetes.io/zone
    whenUnsatisfiable: ScheduleAnyway
  - labelSelector:
      matchLabels:
        app.kubernetes.io/instance: opentelemetry-collector
    maxSkew: 1
    topologyKey: kubernetes.io/hostname
    whenUnsatisfiable: DoNotSchedule
```

gjacquet:
I would not expect any change to be required as `toYaml` is used to render this value and it works on all sort of types, including lists and objects.
https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/templates/_pod.tpl#L191

PR #748, opened 2023-04-17
[opentelemetry-collector] Add support to generate CRD

hivp:
### Description

This PR adds support to generate the CRD (`kind: OpenTelemetryCollector`) directly from `opentelemetry-collector` chart and use it to deploy so `opentelemetry-operator` is able to detect that deployment an create the proper type (daemonset, deployment, statefulset).

### DIFF
--- a/charts/opentelemetry-collector/templates/_config.tpl
+++ b/charts/opentelemetry-collector/templates/_config.tpl
@@ -80,31 +80,6 @@ Build config file for deployment OpenTelemetry Collector
 {{- tpl (toYaml $config) . }}
 {{- end }}
 
-{{/*
-Build config file for daemonset OpenTelemetry Collector
-*/}}
-{{- define "opentelemetry-collector.statefulsetConfig" -}}
-{{- $values := deepCopy .Values }}
-{{- $data := dict "Values" $values | mustMergeOverwrite (deepCopy .) }}
-{{- $config := include "opentelemetry-collector.baseConfig" $data | fromYaml }}
-{{- if eq (include "opentelemetry-collector.logsCollectionEnabled" .) "true" }}
-{{- $config = (include "opentelemetry-collector.applyLogsCollectionConfig" (dict "Values" $data "config" $config) | fromYaml) }}
-{{- end }}
-{{- if .Values.presets.hostMetrics.enabled }}
-{{- $config = (include "opentelemetry-collector.applyHostMetricsConfig" (dict "Values" $data "config" $config) | fromYaml) }}
-{{- end }}
-{{- if .Values.presets.kubeletMetrics.enabled }}
-{{- $config = (include "opentelemetry-collector.applyKubeletMetricsConfig" (dict "Values" $data "config" $config) | fromYaml) }}
-{{- end }}
-{{- if .Values.presets.kubernetesAttributes.enabled }}
-{{- $config = (include "opentelemetry-collector.applyKubernetesAttributesConfig" (dict "Values" $data "config" $config) | fromYaml) }}
-{{- end }}
-{{- if .Values.presets.clusterMetrics.enabled }}
-{{- $config = (include "opentelemetry-collector.applyClusterMetricsConfig" (dict "Values" $data "config" $config) | fromYaml) }}
-{{- end }}
-{{- tpl (toYaml $config) . }}
-{{- end }}
-
 {{- define "opentelemetry-collector.applyHostMetricsConfig" -}}
 {{- $config := mustMergeOverwrite (include "opentelemetry-collector.hostMetricsConfig" .Values | fromYaml) .config }}
 {{- $_ := set $config.service.pipelines.metrics "receivers" (append $config.service.pipelines.metrics.receivers "hostmetrics" | uniq)  }}
--- a/charts/opentelemetry-collector/templates/_config.tpl
+++ b/charts/opentelemetry-collector/templates/_config.tpl
@@ -80,6 +80,31 @@ Build config file for deployment OpenTelemetry Collector
 {{- tpl (toYaml $config) . }}
 {{- end }}
 
+{{/*
+Build config file for daemonset OpenTelemetry Collector
+*/}}
+{{- define "opentelemetry-collector.statefulsetConfig" -}}
+{{- $values := deepCopy .Values }}
+{{- $data := dict "Values" $values | mustMergeOverwrite (deepCopy .) }}
+{{- $config := include "opentelemetry-collector.baseConfig" $data | fromYaml }}
+{{- if eq (include "opentelemetry-collector.logsCollectionEnabled" .) "true" }}
+{{- $config = (include "opentelemetry-collector.applyLogsCollectionConfig" (dict "Values" $data "config" $config) | fromYaml) }}
+{{- end }}
+{{- if .Values.presets.hostMetrics.enabled }}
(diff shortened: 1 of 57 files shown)


charts/opentelemetry-collector/values.yaml
```diff
@@ -5,6 +5,10 @@
 nameOverride: ""
 fullnameOverride: ""
 
+# Define an OpenTelemetryCollector custom resource
+defineCustomResource: true
```
TylerHelmuth, line 9: This should be false by default.

charts/opentelemetry-collector/templates/cr-collector.yaml
```diff
@@ -0,0 +1,60 @@
...
+  labels:
+    {{- include "opentelemetry-collector.labels" . | nindent 4 }}
+  {{- if .Values.annotations }}
+  annotations:
+    {{- range $key, $value := .Values.annotations }}
+      {{- printf "%s: %s" $key (tpl $value $ | quote) | nindent 4 }}
+      {{- end }}
+  {{- end }}
+spec:
+  mode: {{ .Values.mode }}
```
TylerHelmuth, line 16: To be fully configurable we need to allow `mode` to be `sidecar`.  We should fail in `NOTES.txt` if `mode=sidecar` and `defineCustomResource=false`

charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -80,6 +80,31 @@ Build config file for deployment OpenTelemetry Collector
 {{- tpl (toYaml $config) . }}
 {{- end }}
 
+{{/*
+Build config file for daemonset OpenTelemetry Collector
+*/}}
+{{- define "opentelemetry-collector.statefulsetConfig" -}}
```
TylerHelmuth, line 86: Do we need this?  The `mode=statefulset` configuration uses the deployment config since deployments and statefulsets don't need different collector configuration.

TylerHelmuth (commented):
@hivp this is a pretty exciting addition, but we need to have a discussion about it as there has been lots of discussion in the past on how to do this (open issue here: https://github.com/open-telemetry/opentelemetry-helm-charts/issues/69).  

I think the big question is whether or not we need to have a dedicated chart or the operator's custom resources, the operator chart should be about to create them, or if this chart should be the one to handle `OpenTelemetryCollector` and the `Instrumentation` resource should be handled somewhere else.

I left some review comments that will be beneficial if we chose to move forward with this approach vs a dedicated.

@open-telemetry/helm-approvers please review.

charts/opentelemetry-collector/values.yaml
hivp: Resolved:
https://github.com/nhuray/opentelemetry-helm-charts/commit/28a2ad2d4c46f9808a2102b80a45f84cd92f3c3d

charts/opentelemetry-collector/templates/cr-collector.yaml
```diff
@@ -0,0 +1,60 @@
...
+  config: |
+    {{- if eq .Values.mode "deployment" }}
+    {{- include "opentelemetry-collector.deploymentConfig" . | nindent 4 -}}
+    {{- end }}
+    {{- if eq .Values.mode "daemonset" }}
+    {{- include "opentelemetry-collector.daemonsetConfig" . | nindent 4 -}}
+    {{- end }}
+    {{- if eq .Values.mode "statefulset" }}
+    {{- include "opentelemetry-collector.statefulsetConfig" . | nindent 4 -}}
+    {{- end }}
```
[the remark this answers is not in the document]
hivp: Resolved:
https://github.com/nhuray/opentelemetry-helm-charts/commit/28a2ad2d4c46f9808a2102b80a45f84cd92f3c3d

charts/opentelemetry-collector/templates/_config.tpl
hivp: Resolved:
https://github.com/nhuray/opentelemetry-helm-charts/commit/28a2ad2d4c46f9808a2102b80a45f84cd92f3c3d

charts/opentelemetry-collector/templates/_helpers.tpl
```diff
@@ -129,3 +129,19 @@ Check if logs collection is enabled via deprecated "containerLogs" or "preset.lo
...
+{{- if eq .Values.mode "deployment" }}
+{{- include "opentelemetry-collector.deploymentConfig" . }}
+{{- end }}
+{{- if eq .Values.mode "daemonset" }}
+{{- include "opentelemetry-collector.daemonsetConfig" . }}
+{{- end }}
+{{- if eq .Values.mode "statefulset" }}
+{{- include "opentelemetry-collector.deploymentConfig" . }}
+{{- end }}
+{{- end }}
```
Allex1, line 147: same as https://github.com/open-telemetry/opentelemetry-helm-charts/pull/748/files#r1170208321

Allex1:
@hivp Thanks for this contribution. This is very exciting. I would support adding this functionality to the collector repo and reuse the existing code.
Before we can further discuss/review/test can you please follow our [contribution guide](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/CONTRIBUTING.md) in order to have CI pass and maybe add support for sidecar as well.

charts/opentelemetry-collector/values.yaml
```diff
@@ -5,6 +5,10 @@
 nameOverride: ""
 fullnameOverride: ""
 
+# Define an OpenTelemetryCollector custom resource
+defineCustomResource: false
+
+
 # Valid values are "daemonset", "deployment", and "statefulset".
```
Allex1, line 12: same as https://github.com/open-telemetry/opentelemetry-helm-charts/pull/748/files#r1170208321

hivp:
Hey guys:
I have added support for sidecar: https://github.com/nhuray/opentelemetry-helm-charts/commit/1ea8f05d3ac04e8448c5dd48ff1db43edf5a74e1 . I tested it in my k8s cluster and worked.



And some other changes according to your [contribution](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/CONTRIBUTING.md) guide:  https://github.com/nhuray/opentelemetry-helm-charts/commit/38b74cdf7f7af2d5481a877b02bea4b0ad057965 , https://github.com/nhuray/opentelemetry-helm-charts/commit/5644959c796a81c58768c53f4f7cd884cef685c4

TylerHelmuth:
> reuse the existing code

@Allex1 can you expand on this?  Do you mean if we wanted to install the custom resource as part of the operator chart or dedicated chart we'd reuse this chart as a subchart?

Allex1:
> > reuse the existing code
> 
> @Allex1 can you expand on this? Do you mean if we wanted to install the custom resource as part of the operator chart or dedicated chart we'd reuse this chart as a subchart?

@TylerHelmuth I'm torn between having the CR in the collector chart or bundled in the operator chart like kps. 
I think because kps is focused only on monitoring it makese sense to have cr's bundled together with the operator.
Currently we deploy the otel operator chart once per cluster and the collector multiple times per use case so I'm leaning more towards having the cr in the collector chart.

dmitryax:
Hi folks, can we have an issue to discuss it first before considering merging this? (as we have been doing before) I don't understand why CRD should be part of this chart. It doesn't work without operator, and this chart should not depend on installation of the operator.

dmitryax (changes requested):
At this point, I don't think CRD should be part of this chart. Let's discuss it in an issue first

Issue #750, opened 2023-04-18
[demo] support custom default pod label in pod


Frapschen:
support costum default pod label for istio injection case

puckpuck:
Istio support was [added](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/715) about a month ago. Can you do a `helm repo update` and try again?

Frapschen:
> Istio support was https://github.com/open-telemetry/opentelemetry-helm-charts/pull/715 about a month ago.

https://github.com/open-telemetry/opentelemetry-helm-charts/pull/715 is about making service name align istio's standard, not about adding an istio sidecar inject label, if I'm wrong, please correct me.

PR #751, opened 2023-04-18
[demo] support custom default pod label in pod

Frapschen:
support costum default pod label for istio sidecar injection case

close https://github.com/open-telemetry/opentelemetry-helm-charts/issues/750

### DIFF
--- a/charts/opentelemetry-demo/values.yaml
+++ b/charts/opentelemetry-demo/values.yaml
@@ -50,7 +50,7 @@ default:
   securityContext: {}
   # list of pod labels applied to all components
   podLabels: {}
-#    sidecar.istio.io/inject: true
+  # sidecar.istio.io/inject: true
 serviceAccount:
   # Specifies whether a service account should be created
   create: true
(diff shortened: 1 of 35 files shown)


TylerHelmuth (approved):

charts/opentelemetry-demo/values.yaml
```diff
@@ -48,6 +48,9 @@ default:
     affinity: {}
     tolerations: []
   securityContext: {}
+  # list of pod labels applied to all components
+  podLabels: {}
+  # sidecar.istio.io/inject: true
```
dmitryax, line 53: If you need it for istio, I believe the best practice is using annotations, not labels. I didn't even know that labels supported the same way

puckpuck:
Istio support was https://github.com/open-telemetry/opentelemetry-helm-charts/pull/715 about a month ago. Can you do a helm repo update and try again?

charts/opentelemetry-demo/values.yaml
Frapschen: I hope I can use annotations, but I found this way is [Deprecated](https://istio.io/latest/docs/reference/config/annotations/). [Controlling the injection policy](https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/#controlling-the-injection-policy) tell me to use lables.

TylerHelmuth (approved):

TylerHelmuth:
@puckpuck please take another look

Issue #757, opened 2023-04-21
[operator] Use own self-signed certificate


zehenrique:
A few months back this [PR](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/355) introduced a really nice feature where Helm generates a self-signed certificate automatically if cert-manager is not enabled/available.

Although this helps a lot and covers many use cases, there is no option where we can use our own generated self-signed certificate with the current charts. The only option I see for that is to manually create a secret and the mutation and validation webhooks…

I think it would be useful if we could update this [template ](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook.yaml) to allow `tls.crt`, `tls.key`, and `caBundle ` to be configurable. That way we would be to use our own self-signed certificate instead of the one created by Helm.
If those properties are set, own certificate would be used. If not, Helm auto-generated cert is used. 

Please let me know what you think or if I’m missing something. Thanks guys.

PR #760, opened 2023-04-26
[operator] Allow using own self-signed certificate

zehenrique:
Addresses https://github.com/open-telemetry/opentelemetry-helm-charts/issues/757

### DIFF
--- a/charts/opentelemetry-operator/values.yaml
+++ b/charts/opentelemetry-operator/values.yaml
@@ -172,7 +172,6 @@ admissionWebhooks:
   ## Provide the issuer kind and name to do the cert auth job.
   ## By default, OpenTelemetry Operator will use self-signer issuer.
   certManager:
-    # false means helm will create self-signd cert and secret for you.
     enabled: true
     issuerRef: {}
       # kind:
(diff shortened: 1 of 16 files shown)


zehenrique:
@Allex1 thanks for the heads-up. 
Updated. Please let me know if you think the description added to [guideline](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/760/files#diff-79d57ec79556a5537b31dad906fced0c6b9497973d37b8c7c8974d10c920d3a2R3) should be clearer.

Allex1 (approved):

charts/opentelemetry-operator/values.yaml
```diff
@@ -181,6 +180,18 @@ admissionWebhooks:
     certificateAnnotations: {}
     issuerAnnotations: {}
 
+  ## Enable certificate auto generation if certManager is not enabled.
+  ## If true, Helm will automatically create self-signd cert and secret for you.
+  autoGenerateCert: true
```
TylerHelmuth, line 185: It doesn't feel right for `autoGenerateCert` and `certManager.enabled` to both be true by default.  I know it is safe because these values are only used when `certManager.enabled=false`, but if I was a user reading through this values.yaml file I think it would be confusing.

Can we add more comments in the values.yaml value explaining how these 2 fields interact?

zehenrique: @TylerHelmuth updated. please let me know if you think that isn't clear and I'll try to add more detail. thanks

charts/opentelemetry-operator/values.yaml
```diff
@@ -169,18 +169,37 @@ admissionWebhooks:
...
       # kind:
       # name:
-    # Annotations for the cert and issuer if cert-manager is enabled.
+    ## Annotations for the cert and issuer if cert-manager is enabled.
     certificateAnnotations: {}
     issuerAnnotations: {}
 
+  ## TLS Certificate Option 2: Use Helm to automatically generate self-signed certificate.
+  ## certManager must be disabled and autoGenerateCert must be enabled.
+  ## If true, Helm will automatically create a self-signd cert and secret for you.
```
TylerHelmuth, line 188:
```suggestion
  ## If true and certManager.enabled is false, Helm will automatically create a self-signd cert and secret for you.
```

charts/opentelemetry-operator/values.yaml
TylerHelmuth: This is great, thank you.

TylerHelmuth (approved):

zehenrique:
hey @Allex1 and @TylerHelmuth . sorry to bother but is there anything else blocking the merge? we need this feature so we can move forward. thanks

PR #765, opened 2023-04-27
[opentelemetry-collector] Add extraObjects value

indreek:
Adds extraObject possiblity to chart. Enables to objects not described in chart, for example external-secrets.

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.54.1
+version: 0.55.0
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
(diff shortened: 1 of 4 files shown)


Allex1:
@indreek you need to `make generate-examples` in order for CI to pass

I'm not sure this is the right approach as this makes the chart a bit too ... freeform.
I understand that different environments have varying requirements so instead I would propose to create an internal (to your company/project) wrapper chart that sub-charts this one , in which you could add your specific extra resources .

@TylerHelmuth @dmitryax @povilasv @puckpuck wdyt ?

TylerHelmuth:
@Allex1 I agree with your sentiment.  @indreek please open an issue that details the problem this PR would solved and we can discuss there.

dmitryax:
I don't think this should be included in the helm chart. any extra objects should be created separately

indreek:
Hello @dmitryax @Allex1 

Sorry for delay. I was out of office. 

Whatever is security is must for me. In my case it was missing certificates for transport. There should be either option to generate certs or define them through chart - but there are way too many ways how to solve it. This extra option would give user free hands how to solve it. Yes it means that i might break the chart in my machine, but that's the risk if you go extra mile with your chart.

As I'm still learning I'm not aware of possible downsides what it might create for chart owners.

indreek:
@povilasv  

Do I understand you correctly that it would be acceptable, if this external-secret would be solved with condition like 
https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/templates/hpa.yaml, but it would be limited to single object?

Issue #761, opened 2023-04-26
Use ${env:ENV} style syntax when getting environment variables


mx-psi:
The OpenTelemetry Collector supports both the legacy `${ENV}` syntax as well as the new `${env:ENV}` syntax (enabled by default since v0.62.0, see open-telemetry/opentelemetry-collector@b15179e4da112b8ce2b56c6cd336072f7307c87a, fully working for the use case of this Helm chart since v0.71.0, see open-telemetry/opentelemetry-collector@202c03010605012f00b01c45ff90b1a0afd46c01). The latter is preferred and we have updated examples in the OpenTelemetry Collector repository to use it (see open-telemetry/opentelemetry-collector-contrib/issues/17299).

The Helm chart uses the legacy syntax since #603, see https://github.com/open-telemetry/opentelemetry-helm-charts/blob/7252934f33a67697fd04558da3ce750fc70f25eb/charts/opentelemetry-collector/values.yaml#L82-L96

We should use the new `${env:ENV}` style syntax here instead. Note that this would break users of the Collector < v0.71.0 (not sure how this is handled in the Helm chart, but we may want to wait a bit).

JaredTan95:
I'd like to update this, pls assign it to me :-P

PR #767, opened 2023-04-28
[opentelemetry-collector] update MY_POD_IP with ${env:MY_POD_IP}

JaredTan95:
close https://github.com/open-telemetry/opentelemetry-helm-charts/issues/761

### DIFF
--- a/charts/opentelemetry-collector/UPGRADING.md
+++ b/charts/opentelemetry-collector/UPGRADING.md
@@ -1,9 +1,5 @@
 # Upgrade guidelines
 
-## 0.55.0 to 0.55.1
-
-As of v0.55.1 Collector chart use `${env:ENV}` style syntax when getting environment variables and that $`{env:ENV}` syntax is not supported before collector 0.71. If you upgrade collector chart to v0.55.1, you need to make sure your collector version is after than 0.71 (default is v0.76.1).
-
 ## 0.53.1 to 0.54.0
 
 As of v0.54.0 Collector chart, the default resource limits are removed. If you want to keep old values you can use the following configuration:
--- a/charts/opentelemetry-collector/UPGRADING.md
+++ b/charts/opentelemetry-collector/UPGRADING.md
@@ -1,5 +1,9 @@
 # Upgrade guidelines
 
+## 0.54.2 to 0.55.0
+
+As of v0.55.0 Collector chart use `${env:ENV}` style syntax when getting environment variables and that $`{env:ENV}` syntax is not supported before collector 0.71. If you upgrade collector chart to v0.55.0, you need to make sure your collector version is after than 0.71 (default is v0.75.0).
+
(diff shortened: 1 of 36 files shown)


Allex1:
@JaredTan95 this seems like a breaking change so can you pls bump the minor version of the chart and update the [docs ](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/UPGRADING.md)?

JaredTan95:
> this seems like a breaking change 

Sorry, I don't understand why it's a breaking change. Can you tell me about it? thx~

Allex1:
> Sorry, I don't understand why it's a breaking change. Can you tell me about it? thx~

Reading https://github.com/open-telemetry/opentelemetry-helm-charts/issues/761 seems that `${env:ENV}` syntax is not supported before collector 0.71 . We may have users of this helm chart that overwrite the older version of the collector right ? 
I would make it clear that it will not work in that case. wdyt ?

JaredTan95:
@Allex1 @povilasv thx, got you~

Allex1 (approved):

dmitryax:
> Reading https://github.com/open-telemetry/opentelemetry-helm-charts/issues/761 seems that ${env:ENV} syntax is not supported before collector 0.71 . We may have users of this helm chart that overwrite the older version of the collector right ?

0.71.0 got only a fix that is unrelated to our use cases. `${env:...}` syntax has been supported for long time. I belive it was added in 2001 :) Not really sure we need the upgrade guidelines

dmitryax (approved):

charts/opentelemetry-collector/UPGRADING.md
```diff
@@ -1,5 +1,9 @@
 # Upgrade guidelines
 
+## 0.55.0 to 0.55.1
+
+As of v0.55.1 Collector chart use `${env:ENV}` style syntax when getting environment variables and that $`{env:ENV}` syntax is not supported before collector 0.71. If you upgrade collector chart to v0.55.1, you need to make sure your collector version is after than 0.71 (default is v0.76.1).
+
```
TylerHelmuth, line 6: Based on @dmitryax comment lets remove this.

TylerHelmuth (approved):

charts/opentelemetry-collector/UPGRADING.md
JaredTan95: done

dmitryax (approved):

TylerHelmuth:
@mx-psi thanks for the details.  @JaredTan95 sorry, but would you be willing to submit another PR touching up the UPGRADING doc saying that using chart version `0.55.1` or higher requires a collector version `0.71.0` or higher?

dmitryax:
Oh, @mx-psi is right. I am sorry for causing the confusion!

PR #772, opened 2023-05-02
[collector] bump appVersion to 0.76.3

povilasv:
Bumps the default collector version to 0.76.3, since 0.76.2 and 0.76.1 was retracted 

Currently draft as The 0.76.3 image is not available https://hub.docker.com/r/otel/opentelemetry-collector-contrib/tags


TylerHelmuth:
@povilasv there won't be a 0.76.1 image. Although the Contrib libraries are tagged 0.76.3, the released image was tagged 0.76.1.

PR #782, opened 2023-05-10
Add support for HPA behavior

gjacquet:
We ran into a use case where it would have been helpful to fine tune a little bit HPA behaviour.

This PR will allow to optionally define a behaviour on the `HorizontalPodAutoscaler` resource. See https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/ and https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/#horizontalpodautoscalerspec-v2-autoscaling for reference.

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.57.0
+version: 0.57.1
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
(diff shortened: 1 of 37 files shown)


charts/opentelemetry-collector/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.57.0
+version: 0.58.0
```
TylerHelmuth, line 3: Since this is a backwards-compatible feature add bump the patch version instead of the minor (since we don't have a major to use yet we normally save the minor bumps for more impactful changes.)

TylerHelmuth (commented):
Thanks for adding this.

charts/opentelemetry-collector/Chart.yaml
gjacquet: Done!

TylerHelmuth (approved):

Allex1 (approved):
after rebase

gjacquet:
@Allex1 Rebased!

TylerHelmuth (approved):

PR #784, opened 2023-05-11
feat: add container specific securityContext

Demonsthere:
This PR adds the missing container level securityContext to the deployment. 
This option is required to run the operator inside a restricted namespace

### DIFF
--- a/charts/opentelemetry-operator/values.yaml
+++ b/charts/opentelemetry-operator/values.yaml
@@ -132,11 +132,11 @@ manager:
 
   ## Container specific securityContext
   ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container
-  securityContext:
-    allowPrivilegeEscalation: false
-    capabilities:
-      drop:
-      - ALL
+  securityContext: {}
+    # allowPrivilegeEscalation: false
+    # capabilities:
+    #   drop:
+    #   - ALL
 
 ## Provide OpenTelemetry Operator kube-rbac-proxy container image.
 ##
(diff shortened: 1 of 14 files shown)


charts/opentelemetry-operator/values.yaml
```diff
@@ -130,6 +130,14 @@ manager:
   # Enable manager pod automatically rolling
   rolling: false
 
+  ## Container specific securityContext
+  ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container
+  securityContext:
```
TylerHelmuth, line 135: I don't think we need to provide any default values for with securityContext section. Lets set them to `securityContext: {}`.

TylerHelmuth (commented):
Thanks for this addition.  Please bnump the patch number of the chart and then run `make generate-examples CHARTS=opentelemetry-operator`

Allex1 (approved):

TylerHelmuth (commented):
@Demonsthere I merged a different operator chart PR first so there are merge conflict nows.  Once they are handled, instead of bumping to `0.30.0` lets bump to `0.29.2`.

Demonsthere:
Conflicts have been addressed :)

TylerHelmuth (approved):

Demonsthere:
Hey @TylerHelmuth i see some tests are failing but can’t say I see the reason. It does not look to me like it’s related to my changes, or?

TylerHelmuth:
It is flakiness with the operator tests: https://github.com/open-telemetry/opentelemetry-operator/pull/1733

PR #792, opened 2023-05-23
[collector] provide support for windows

povilasv:
Allow users to set `isWindows` flag to true, which will configure mounts and various presets to use windows paths.

This change will also add a new node selector for all existing users:
```
      nodeSelector:
        kubernetes.io/os: linux
```

AFAIK this shouldn't be breaking change?

Added example with windows:
- https://github.com/open-telemetry/opentelemetry-helm-charts/pull/792/files#diff-903e8d385e1abafd5f68d83210069f94ee41ed1fd832db2523356f4633b067aaR31


TylerHelmuth (commented):
I'm concerned that we don't have a good way to test this setting since all our CI runs on linux.  Were you able to test locally?

charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -89,14 +89,17 @@ Build config file for deployment OpenTelemetry Collector
 {{- define "opentelemetry-collector.hostMetricsConfig" -}}
 receivers:
   hostmetrics:
+    {{- if not .Values.isWindows }}
     root_path: /hostfs
+    {{- end }}
```
dmitryax, line 94: This is still needed for windows, just under different path, right?

PR #796, opened 2023-05-25
[opentelemetry-collector] [chore] Update README to reflect changes to `mode` behavior.

matej-g:
When following the instructions in the README, I'm unable to install the collector Helm chart without further specification of `mode` (i.e. there is no longer a default mode). This change corrects the README and instructs user to specify the mode.

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.59.0
+version: 0.59.1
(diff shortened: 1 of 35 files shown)


TylerHelmuth:
@matej-g even tho it is only a [readme change](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/CONTRIBUTING.md#immutability) you have to bump the chart's patch version and then run `make generate-examples CHARTS=opentelemetry-collector`.  For the charts any change to the chart

matej-g:
@TylerHelmuth got it, thanks, wasn't sure if it requires version bump. I'm on it.

TylerHelmuth (approved):

PR #797, opened 2023-05-25
fix chart failing if internalTrafficPolicy isn't set

RainofTerra:
With 0.59.0 I started getting the following:
```
Error: Failed to render chart: exit status 1: Error: template: opentelemetry-collector/templates/NOTES.txt:44:42: executing "opentelemetry-collector/templates/NOTES.txt" at <eq .Values.service.internalTrafficPolicy "Cluster">: error calling eq: incompatible types for comparison
```

I think this should fix it, by setting a default value for `internalTrafficPolicy`


RainofTerra:
Nevermind, our build pipeline has a version of helm 3 from the before time

PR #807, opened 2023-06-07
Added option to add additional filelog operators.

ckt114:
Added option to add additional filelog operators.


TylerHelmuth:
@ckt114 please update your description or create an issue detailing the problem this PR solves

ckt114:
> @ckt114 please update your description or create an issue detailing the problem this PR solves

Description updated, @TylerHelmuth.

TylerHelmuth:
@ckt114 our stance with the presets is that they should not include any complex configuration.  Instead of adding this to the preset I think that configuring filelogreceiver and the extra operators via `.Values.config` is a better approach.

ckt114:
@TylerHelmuth I agree with you however with Helm one cannot merge arrays. If I define something new below `receviers.filelog.operators` it'll overwrite all the preset that's included in the chart. That's why I put in this PR. Do you agree?

TylerHelmuth:
@ckt114 you are correct.  For your use case you should not use the preset and instead configure the filelogreceiver manually via `.Values.Config` and add any volumes you need via `.Values.extraHostPathMounts`.

TylerHelmuth:
@ckt114 Sorry for the confusion.

I propose that instead of adding this to the operator I propose that you use the `.Values.Config` section to configure a filelogreceiver manually, including in that configuration any operators that you need.  These operators might be the same operators that the preset provides and any extras.

In addition to configuring the filelogreceiver, you may need to configure extra host path mounts to facilitate the filelogreceiver working as expected.  Depending on your needs you may not need to.

The presets are not made to be configurable, they exists to quickly show value, but as use cases grow in complexity their value diminishes.  Once the use case has out grown the preset the solution is to configure the collector manually.

ckt114:
@TylerHelmuth I'm already doing what you suggested but I feel that it's not DRY. I'm copying the exact same thing that's in the chart into my config. However, I do see your point. If you feel that this is not necessary then you can deny this PR. Thanks.

PR #814, opened 2023-06-19
Add ability to set your own common labels for all resources

martinrw:
This change gives us a new optional variable that we can set which allows you to add your own custom labels to all of the resources being deployed. Currently we are able to do this with Pods but not anything else like deployments, config-maps etc.
This is already quite common among many other open source helm charts. It fixes a problem for us where we use these labels for identifying which teams own certain resources which in turn helps us direct prometheus alerts to the correct teams.

### DIFF
--- a/charts/opentelemetry-collector/values.yaml
+++ b/charts/opentelemetry-collector/values.yaml
@@ -4,7 +4,7 @@
 
 nameOverride: ""
 fullnameOverride: ""
-
+commonLabels: {}
 # Valid values are "daemonset", "deployment", and "statefulset".
 mode: ""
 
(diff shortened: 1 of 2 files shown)


TylerHelmuth (commented):
Please add a description to this PR describing what it changes and what problem it solves.

martinrw:
apologies for missing out the description, I've updated it now

charts/opentelemetry-collector/values.yaml
```diff
@@ -4,7 +4,7 @@
 
 nameOverride: ""
 fullnameOverride: ""
-
+commonLabels: {}
```
TylerHelmuth, line 7: Please add a description the explains this field similar to https://github.com/prometheus-community/helm-charts/blob/8b91a619e9407c0e830df7071ee73cd160937b6b/charts/kube-prometheus-stack/values.yaml#L25-L27.

TylerHelmuth (commented):
Please add the new field to the json schema file.  Also, please bump the chart's patch version and then run `make generate-examples CHARTS=opentelemetry-collector`.

Issue #809, opened 2023-06-08
[opentelemetry collector] logCollection preset multiline logs


povilasv:
I would like to work on some kind of support for multi-line logs for log collection.

Currently I have two ideas:

1) Some runtimes like CRI, have a marker to indicate continuation of log line. See example in recombine operator 

![image](https://github.com/open-telemetry/opentelemetry-helm-charts/assets/22289110/17f183a9-4138-4943-9608-4617f3d3e502)

So for CRI we could add this:
```
- type: regex_parser
  regex: '^(?P<timestamp>[^\s]+) (?P<stream>\w+) (?P<logtag>\w) (?P<message>.*)'
- type: recombine
  combine_field: body.message
  combine_with: ""
  is_last_entry: "body.logtag == 'F'"
  overwrite_with: "newest"
```

That's one idea.

2) Another idea came from this - I've played with some app that sends multline configs and with containerd I see:
```
2023-06-08T12:29:28.123760947Z stderr F goroutine 17 [finalizer wait]:
2023-06-08T12:29:28.12378603Z stderr F runtime.gopark(0x40000345a8?, 0x600000000224a8?, 0xa8?, 0xd0?, 0x1?)
2023-06-08T12:29:28.123805571Z stderr F         /usr/local/go/src/runtime/proc.go:381 +0xe0 fp=0x4000034580 sp=0x4000034560 pc=0x422e0
2023-06-08T12:29:28.123824237Z stderr F runtime.runfinq()
2023-06-08T12:29:28.123842029Z stderr F         /usr/local/go/src/runtime/mfinal.go:193 +0x100 fp=0x40000347d0 sp=0x4000034580 pc=0x23af0
2023-06-08T12:29:28.123859236Z stderr F runtime.goexit()
2023-06-08T12:29:28.123876736Z stderr F         /usr/local/go/src/runtime/asm_arm64.s:1172 +0x4 fp=0x40000347d0 sp=0x40000347d0 pc=0x6dba4
2023-06-08T12:29:28.123893944Z stderr F created by runtime.createfing
2023-06-08T12:29:28.123911444Z stderr F         /usr/local/go/src/runtime/mfinal.go:163 +0x80
2023-06-08T12:29:28.123927777Z stderr F
```

All lines marked as single line entries. Which means that 1) idea wouldn't help in my case.

So what I'm thinking is to have an option to configure router and recombine operator to recombine certain logs?

[...]

PR #818, opened 2023-06-26
[opentelemetry-collector] recombine cri multi-line logs

povilasv:
REF: https://github.com/open-telemetry/opentelemetry-helm-charts/issues/809

Tested with an app that produces large logs 

Example log:
```
2023-06-26T08:53:52.757586309Z stderr P <large log message here>
2023-06-26T08:53:52.758090299Z stderr F <continuation of large log message>
```

### DIFF
--- a/charts/opentelemetry-collector/values.yaml
+++ b/charts/opentelemetry-collector/values.yaml
@@ -28,6 +28,9 @@ presets:
     # Enabling this writes checkpoints in /var/lib/otelcol/ host directory.
     # Note this changes collector's user to root, so that it can write to host directory.
     storeCheckpoints: false
+    # The maximum bytes size of the recombined field.
+    # Once the size exceeds the limit, all received entries of the source will be combined and flushed.
+    maxRecombineLogSize: 102400
   # Configures the collector to collect host metrics.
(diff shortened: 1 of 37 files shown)


TylerHelmuth (commented):
@povilasv can you share the input log(s) and output log of this change when applied to a set of logs that needs recombined and set of logs that don't need recombined?

charts/opentelemetry-collector/values.yaml
```diff
@@ -28,6 +28,9 @@ presets:
     # Enabling this writes checkpoints in /var/lib/otelcol/ host directory.
     # Note this changes collector's user to root, so that it can write to host directory.
     storeCheckpoints: false
+    # The maximum bytes size of the recombined field.
+    # Once the size exceeds the limit, all received entries of the source will be combined and flushed.
+    maxRecombineLogSize: 102400
```
dmitryax, line 33: Is it supposed to be 100MB or 100MiB?

dmitryax: nvm. it's 100KiB. I got confused

dmitryax (approved):

TylerHelmuth (approved):

PR #827, opened 2023-07-05
JAC: fix checksum config in statefulset template

jorgeancal:
Hello there 
I've been noticing that each time at work someone changed something in the statefulset they said that they have to force the rollout and I took a look at the chart and I realised that the checksum in the template is pointing to the the wrong confingmap path.  this is the fix

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.62.0
+version: 0.62.1
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-collector/examples/daemonset-and-deployment/rendered/configmap-agent.yaml
+++ b/charts/opentelemetry-collector/examples/daemonset-and-deployment/rendered/configmap-agent.yaml
@@ -5,7 +5,7 @@ kind: ConfigMap
 metadata:
   name: example-opentelemetry-collector-agent
   labels:
-    helm.sh/chart: opentelemetry-collector-0.62.0
+    helm.sh/chart: opentelemetry-collector-0.62.1
     app.kubernetes.io/name: opentelemetry-collector
     app.kubernetes.io/instance: example
     app.kubernetes.io/version: "0.81.0"
--- a/charts/opentelemetry-collector/examples/daemonset-and-deployment/rendered/configmap.yaml
+++ b/charts/opentelemetry-collector/examples/daemonset-and-deployment/rendered/configmap.yaml
@@ -5,7 +5,7 @@ kind: ConfigMap
 metadata:
   name: example-opentelemetry-collector
   labels:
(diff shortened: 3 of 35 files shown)


Allex1:
@jorgeancal thanks for your contribution .Please follow our [guide](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/CONTRIBUTING.md) in order for the ci to pass.

jorgeancal:
sorry my bad the last commit should align the PR with the guide

jorgeancal:
sorry, my bad. this should be good to go now but let me know if I missed something

jorgeancal:
I've updated  it to the latest image that was released just a few hours ago as the guide says. I had a bit of a confusion as you could see but this should be working correctly now

charts/opentelemetry-collector/Chart.yaml
```diff
@@ -11,4 +11,4 @@ maintainers:
   - name: dmitryax
   - name: TylerHelmuth
 icon: https://opentelemetry.io/img/logos/opentelemetry-logo-nav.png
-appVersion: 0.80.0
+appVersion: 0.81.0
```
TylerHelmuth, line 14: Please revert this back to 0.80.0. We like to do the appVersion bumps in a standalone PR.  I also see the Collector chart's Contributing Guide needs some work to be clearer which I'll fix up.

TylerHelmuth (commented):
great catch @jorgeancal thank you!

charts/opentelemetry-collector/Chart.yaml
TylerHelmuth: I got the appVersion bumped to `0.81.0`, so now we need a rebase.  Will need to change the chart's version to `0.62.1` and rerun example generation.

charts/opentelemetry-collector/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.61.2
+version: 0.62.0
```
Allex1, line 3: @jorgeancal this should be 0.62.1 since we just merged the latest collector triggering a minor version bump and your pr is not smth breaking but just a bug fix.

charts/opentelemetry-collector/Chart.yaml
jorgeancal: Sure @TylerHelmuth  there you go

TylerHelmuth:
I tried to beat you to it but you got your updates in before the bump haha

Allex1 (approved):
After addressing comments

charts/opentelemetry-collector/Chart.yaml
jorgeancal: I will change straight away

jorgeancal:
okay I think I've done all the changes

jorgeancal:
@Allex1  @TylerHelmuth  It has been done now sorry for all the commits the rebase did something strange there but in the end everything ended well 😅

TylerHelmuth (approved):

PR #833, opened 2023-07-13
[CHORE] adding ebpf collector

nicolastakashi:
As proposed on this https://github.com/open-telemetry/opentelemetry-ebpf/issues/186, I'm adding the eBPF collector as part of the collector helm chart.


nicolastakashi:
cc @atoulme

atoulme:
Nice! Thanks for this work.

atoulme:
@nicolastakashi what would you think of making this a new top level chart? @TylerHelmuth is there a process to add a new chart?

TylerHelmuth:
@atoulme new charts should be proposed via an Issue.

TylerHelmuth:
@nicolastakashi I do not have the context necessary to review this PR.  Can you create an issue (or can we move the issue linked to this PR) to discuss what problem this PR solves?

dmitryax:
I suggest making it another helm chart. Later, if we want it to be installable along with the collector, it can be set as a subchart. Or, just provide instructions on how to install it with the collector

nicolastakashi:
Going to open another PR, cause I was struggling with conflicts

PR #838, opened 2023-07-19
Add externalTrafficPolicy value for service type LoadBalancer

ahurtaud:
This PR is to implement the `spec.externalTrafficPolicy` for the kubernetes service of type LoadBalancer.
The value can be `Cluster` (default) or `Local`. The difference is well-explained in this official kubernetes documentation:
https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/#preserving-the-client-source-ip

_`.spec.externalTrafficPolicy` - denotes if this Service desires to route external traffic to node-local or cluster-wide endpoints. There are two available options: Cluster (default) and Local. Cluster obscures the client source IP and may cause a second hop to another node, but should have good overall load-spreading. Local preserves the client source IP and avoids a second hop for LoadBalancer and NodePort type Services, but risks potentially imbalanced traffic spreading._

I need this because we are doing some networkPolicies based on incoming source IP traffic, and I must keep the client source IP going through my LoadBalancer.

This PR add a `externalTrafficPolicy` field in the `service` value that you can set if you are using a service.type: LoadBalancer.

If it is not set, the field will not be part of the blueprint and default `Cluster` is applied by kubernetes like before this PR - Backward compatible.

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.62.3
+version: 0.62.4
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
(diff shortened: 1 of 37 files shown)


TylerHelmuth (commented):
@ahurtaud thank you for your contribution.  Please add a description to your PR detailing the problem this PR solves and how it solves it.

TylerHelmuth (approved):

Allex1 (approved):

TylerHelmuth:
@ahurtaud as soon as the easycla issue (https://github.com/communitybridge/easycla/issues/4071) is resolved I'll merge this.

PR #839, opened 2023-07-20
feat: update helm chart to use the new /oltp-http/ endpoint on frontendproxy

chigia001:
As this require new image with 1.5.0 tag to be released, we will need to wait before this can be merged.


julianocosta89 (commented):
LGTM, but we need to wait oteldemo version 1.5 to be released first

Allex1:
@chigia001 Please bump the patch version of the chart and then run `make generate-examples CHARTS=opentelemetry-demo`

Issue #841, opened 2023-07-27
[collector] Cannot disable all ports


TylerHelmuth:
Ran into an issue today where I was installing a collector that didn't need any exposed ports and found out that disabling all the ports in `.Values.ports` causes the chart to not be installable.

PR #842, opened 2023-07-27
[collector] Allow disabling all ports

TylerHelmuth:
Updates the collector chart to allow disabling of all ports.

Closes https://github.com/open-telemetry/opentelemetry-helm-charts/issues/841

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.66.0
+version: 0.65.2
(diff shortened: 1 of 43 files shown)


Allex1 (approved):
after rebase

charts/opentelemetry-collector/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.65.1
+version: 0.66.0
```
dmitryax, line 3: It seems more like a bug fix that should go to a patch version?

TylerHelmuth: Fair, I'll update the version

PR #844, opened 2023-07-27
[collector] Respect pipeline order when enabling kubernetesAttributes

JDemler:
When enabling `kubernetesAttributes`  an ` k8sAttributes` processor is automatically prepended to all available pipelines.
There are situations where one relies on another processor to provide or prepare data for ` k8sAttributes` to work correctly.

This change makes sure that when a `k8sAttributes` -processor was added manually to a pipeline, the pipeline itself is not touched to preserve its order.
As communicated in `UPGRADING.md` this might break existing installations when they implicitly relied on `k8sAttributes` being run first but explicitly configuring it otherwise in the `values.yaml`.


charts/opentelemetry-collector/UPGRADING.md
```diff
@@ -4,6 +4,11 @@ These upgrade guidelines only contain instructions for version upgrades which re
 If the version you want to upgrade to is not listed here, then there is nothing to do for you. 
 Just upgrade and enjoy.
 
+## 0.62 to 0.63
+
+The `kubernetesAttributes` preset now respects order of processors in logs, metrics and traces pipelines.
+This implicitly might break your pipelines if you relied on having the `k8sAttributes` processor rendered as first processor.
```
TylerHelmuth, line 10:
```suggestion
This implicitly might break your pipelines if you relied on having the `k8sAttributes` processor rendered as the first processor but also explicitly listed it in the signal's pipeline somewhere else.
```

TylerHelmuth (approved):
Thanks for this contribution. The breaking change is so obscure that I'm tempted to not even call it out, @dmitryax @povilasv what do you think?

TylerHelmuth:
@povilasv since most of the presets are receivers prepending to the front doesn't affect anything, but for processors I can see the argument for wanting to get the value of the preset while still placing it specifically in the pipeline.  I think this is a good compromise between the preset's value and our goal to not make them configurable since it is configuring the `.Values.config` section instead of the preset.

TylerHelmuth:
@JDemler I merged in an older PR, please rebase.

JDemler:
Thanks! That was a painless process :)

PR #851, opened 2023-08-07
charts/opentelemetry-collector: fix CPU limit

kevinburkesegment:
While "256m" is a valid setting, the intention was likely to allow 4 pods to be scheduled on a single CPU, which we will currently overshoot by 24/1000, since there are only 1000 units in a CPU fraction.

Adjust to "250m", which would allow exactly 4 pods to share a CPU.

See e.g. https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-cpu


TylerHelmuth (commented):
@kevinburkesegment please bump the patch version of the chart and then run `make generate-examples CHARTS=opentelemetry-collector`.

kevinburkesegment:
Done

charts/opentelemetry-collector/values.yaml
```diff
@@ -270,7 +270,8 @@ ports:
 resources: {}
 # resources:
 #   limits:
-#     cpu: 256m
+#     # CPU units are in fractions of 1000; memory in powers of 2
```
TylerHelmuth, line 273:
```suggestion
```

TylerHelmuth (approved):

Allex1 (approved):
lgtm

PR #852, opened 2023-08-09
Add ability to configure replicas for services

martinjt:
This PR adds 2 new settings to the `values.yaml`

- `default.replicas`
  This will configure how many replicas of each of the services should be create
- `components.<service>.replicaOverride`
  This will override the specific service's replicas from the default.

In the supplied `values.yaml` the `default.replicas` value is set to 1 so that it doesn't change for existing users. In addition there are specific services that I didn't think should be automatically scaled based on this setting, for these I've added a `replicasOverride` to their section to keep it at 1, users can obviously override these as they see fit. These are:

- Redis
- Postgres
- Kafka
- Load generator

I'm sure I need to update some change logs, but I'm not sure where, so any guidance would be appreciated.

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.66.2
+version: 0.66.3
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.66.3
+version: 0.66.4
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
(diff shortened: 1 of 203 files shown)


TylerHelmuth (commented):
@martinjt looks great.  Please update the README with the new values.  Also, bump the chart's minor version and run `make generate-examples CHARTS=opentelmetry-demo`

martinjt:
Thanks @TylerHelmuth are there any other stateful services that I missed in the replicasOverride?

TylerHelmuth:
I'd like @puckpuck opinion, buy maybe `featureflagService` and `frontendProxy`.

martinjt:
The Feature Flag service has a postgres database, so I figured that was ok, and I'd assumed the proxy wasn't stateful.

martinjt:
The idea is to scale up all the services at once, and make that easier for the scenario of just "we want more load". which makes it easier than adding an entry for each of them. Also when/if new components are added, they'll immediately get the replicas.

puckpuck (approved):

martinjt:
Sorry, yes, this was in relation to the question about example charts
changing on every release. Want me to PR the removal?

martinjt:
No idea what's going on with that failure, it looks fine, I ran the generate-examples and committed them. Any ideas @puckpuck ?

TylerHelmuth (approved):

TylerHelmuth:
@martinjt looks like you've had a bad merge maybe? Please undo the changes to `./charts/opentelemetry-collector` and `./charts/opentelemetry-operator`. When re-rendering the demo examples use `make generate-examples CHARTS=opentelemetry-demo` so that only the demo examples are rendered.

PR #865, opened 2023-08-17
[Operator] Feature certificate manifest improvements

izazahamed-babji:
This PR addresses issues related to certificates of opentelemetry-operator
1. If group: cert-manager.io is not mentioned in issuerRef, there is a chance that the issuer might not be picked if the cluster has another external CA issuer.
2. In case of not opting for self-signed issuer, you need extraSpecs such as the commented ones in values.yaml which is not possible with the current structure

### DIFF
--- a/charts/opentelemetry-operator/templates/certmanager.yaml
+++ b/charts/opentelemetry-operator/templates/certmanager.yaml
@@ -19,6 +19,7 @@ spec:
     {{- if .Values.admissionWebhooks.certManager.issuerRef }}
     {{- toYaml .Values.admissionWebhooks.certManager.issuerRef | nindent 4 }}
     {{- else }}
+    group: cert-manager.io
     kind: Issuer
     name: {{ template "opentelemetry-operator.fullname" . }}-selfsigned-issuer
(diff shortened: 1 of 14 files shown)


charts/opentelemetry-operator/templates/certmanager.yaml
```diff
@@ -19,13 +19,17 @@ spec:
     {{- if .Values.admissionWebhooks.certManager.issuerRef }}
     {{- toYaml .Values.admissionWebhooks.certManager.issuerRef | nindent 4 }}
     {{- else }}
+    group: cert-manager.io
```
TylerHelmuth, line 22: @izazahamed-babji can you clarify that this is not a breaking change?

izazahamed-babji: @TylerHelmuth I don't think it is a breaking change as it is to make sure that the self signed issuer is picked and more of a patch.

Allex1:
@izazahamed-babji don't know enough to make an opinion on this . Can you please share some docs ?

PR #867, opened 2023-08-18
[collector] Add new options to the k8sattributes preset for extracting labels and annotations

TylerHelmuth:
Adds 2 new boolean flags to the `kubernetesAttributes` preset to enable extraction of all pod labels and pod annotations. The k8sattributesprocessor configuration for this capability is a little difficult to conceive, so this preset removes the need for users to understand the processor's extraction logic and regex capture groups.


TylerHelmuth:
> For me probably raw format is better, since if we prepend `k8s.pod.annotation` we mix two different formats which IMO is worse

Agreed, I definitely prefer the raw value.

PR #871, opened 2023-08-22
[collector] Make it possible to specify namespace when running Helm template

krisztianfekete:
Currently it can be problematic to conveniently support GitOps tools like flux as `helm template` does not add `.Release.Namespace` to manifests as per https://github.com/helm/helm/issues/3553.

This PR fixes this by making it possible to render the the config into a specified namespace. Similar behaviour is present in other OSS charts as well.

E.g. `helm template otel opentelemetry-collector/ --namespace my-ns --set mode=daemonset`, will add `namespace: my-ns` to the rendered resources.

### DIFF
--- a/charts/opentelemetry-collector/templates/ingress.yaml
+++ b/charts/opentelemetry-collector/templates/ingress.yaml
@@ -9,6 +9,7 @@ metadata:
   {{- else }}
   name: {{ include "opentelemetry-collector.fullname" $ }}
   {{- end }}
+  namespace: {{ template "opentelemetry-collector.namespace" . }}
   labels:
     {{- include "opentelemetry-collector.labels" $ | nindent 4 }}
     {{- include "opentelemetry-collector.component" $ | nindent 4 }}
--- a/charts/opentelemetry-collector/templates/ingress.yaml
+++ b/charts/opentelemetry-collector/templates/ingress.yaml
@@ -9,7 +9,7 @@ metadata:
   {{- else }}
   name: {{ include "opentelemetry-collector.fullname" $ }}
   {{- end }}
-  namespace: {{ template "opentelemetry-collector.namespace" . }}
+  namespace: {{ template "opentelemetry-collector.namespace" $ }}
   labels:
     {{- include "opentelemetry-collector.labels" $ | nindent 4 }}
     {{- include "opentelemetry-collector.component" $ | nindent 4 }}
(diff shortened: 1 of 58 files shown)


krisztianfekete:
Thanks @povilasv, just pushed the commit.

krisztianfekete:
Hm, I think it should be good now. Maybe it's failing due to the linter having preconceptions?

TylerHelmuth:
> Similar behaviour is present in other OSS charts as well.

@krisztianfekete can you share some examples of other helm charts implementing this behavior?

krisztianfekete:
@TylerHelmuth, yes 
-  charts in prometheus-community e.g. https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack, or
- some bitnami ones e.g. https://github.com/bitnami/charts/tree/main/bitnami/zookeeper, or
- some Grafana ones e.g. https://github.com/grafana/helm-charts/blob/main/charts/grafana

TylerHelmuth:
@krisztianfekete thanks for sharing.  We like following the patterns set by the prometheus chart, so lets use `namespaceOverride` in our values.yaml

krisztianfekete:
Thanks @TylerHelmuth, I pushed the changes.

krisztianfekete:
@TylerHelmuth, can you please check CI? I am not sure how to resolve this. I'd prefer not to default to any namespace as that might break existing installations, but I am not sure how to pass CI without that.

charts/opentelemetry-collector/templates/ingress.yaml
```diff
@@ -9,6 +9,7 @@ metadata:
   {{- else }}
   name: {{ include "opentelemetry-collector.fullname" $ }}
   {{- end }}
+  namespace: {{ template "opentelemetry-collector.namespace" . }}
```
TylerHelmuth, line 12: @krisztianfekete I believe this is causing your issue.  The ingress template includes a `range` at the top which changes the scope that everything inside works on.  `.` in this instance is not referencing the entire configuration object but a specific subset that does not include `namespaceOverride`.  You'll need to do some manipulation or use `$` (I think, you'll need to look it up) to reference the outer scope.  Also change the other files you change and make sure `.` is the appropriate scope.

krisztianfekete: Excellent, that was it! Thanks @TylerHelmuth. I rebased it too, so now it should be ready to merge!

TylerHelmuth (approved):
LGTM

krisztianfekete:
Thanks! Can we merge this?

PR #885, opened 2023-09-05
[operator] limit prometheus rules to opentelemetrycollector application

abaguas:
The chart deploys alerting rules `ReconcileErrors` and `WorkqueueDepth` to monitor the operator. However, since these rules do not filter on any labels they apply to any operator deployed in the Kubernetes cluster.
To contain this chart only to the Opentelemetry Operator this PRs adds the relevant labels to the PrometheusRules.

No user impact is expected with this change

### DIFF
--- a/charts/opentelemetry-operator/templates/prometheusrule.yaml
+++ b/charts/opentelemetry-operator/templates/prometheusrule.yaml
@@ -23,7 +23,7 @@ spec:
   - name: managerRules
     rules:
     - alert: ReconcileErrors
-      expr: rate(controller_runtime_reconcile_total{result="error"}[5m]) > 0
+      expr: rate(controller_runtime_reconcile_total{controller="opentelemetrycollector",result="error"}[5m]) > 0
       for: 5m
       labels:
         severity: warning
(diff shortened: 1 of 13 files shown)


charts/opentelemetry-operator/templates/prometheusrule.yaml
```diff
@@ -23,15 +23,15 @@ spec:
   - name: managerRules
     rules:
     - alert: ReconcileErrors
-      expr: rate(controller_runtime_reconcile_total{result="error"}[5m]) > 0
+      expr: rate(controller_runtime_reconcile_total{controller="opentelemetrycollector",result="error"}[5m]) > 0
```
TylerHelmuth, line 26: I'm not very familiar with Prometheus syntax, what is determining the value of `name`? Will it change based on the release name?

Allex1: @TylerHelmuth apparently the `name` label for the 2 metrics used in these alerts is hardcoded to `opentelemetrycollector`.

Allex1 (approved):

charts/opentelemetry-operator/templates/prometheusrule.yaml
TylerHelmuth: @Allex1 hard coded in the Operator with those names?

Allex1: @TylerHelmuth yeah, in the metrics endpoint. Validated against my own metrics endpoint (which has a different helm release name), could not find it in the actual operator code.

TylerHelmuth (approved):

PR #887, opened 2023-09-08
Will allow to add extra prometheus rule labels

klubi:
This PR introduces ability to define additional labels to be added to prometheus rules.

Why?
I have an use case, where multiple teams deploy OTLP collectors to their own namespaces, and each team is monitoring collector on their own. Ability to define additional labels will allow those teams to properly route rules using alertmanager.

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.67.0
+version: 0.68.0
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.68.0
+version: 0.67.1
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-collector/values.yaml
+++ b/charts/opentelemetry-collector/values.yaml
@@ -500,6 +500,8 @@ prometheusRule:
   # Create default rules for monitoring the collector
   defaultRules:
     enabled: false
+    # additional labels for rules (to be used with alert manager routing)
+    extraLabels: {}
 
   # additional labels for the PrometheusRule
   extraLabels: {}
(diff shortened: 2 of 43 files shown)


charts/opentelemetry-collector/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.67.0
+version: 0.68.0
```
Allex1, line 3: the version can be 0.67.1 as this is a minor change

Allex1:
@klubi thanks for your contribution. Pls run `make generate-examples CHARTS=opentelemetry-collector` and commit the changes in order for the CI to pass .

Allex1 (approved):

charts/opentelemetry-collector/values.yaml
```diff
@@ -500,6 +500,8 @@ prometheusRule:
   # Create default rules for monitoring the collector
   defaultRules:
     enabled: false
+    # additional labels for rules (to be used with alert manager routing)
+    extraLabels: {}
```
TylerHelmuth, line 504: I am a little torn on this as you could achieve your desired outcome via the `groups` field.  Typically in the charts we like to keep the opinionated, out-of-the-box configuration simple, excluding additional configuration. @dmitryax @Allex1  @povilasv what do you think?

At the minimum we should use a different name since this name conflicts with the other `extraLabels` field which is adding k8s labels.

klubi: If I was to use `groups` I would have to re-write all default rules, for sake of adding some extra labels.

Now that you've mentioned it, `extraLabels` does seem like suboptimal name.
Maybe:
- `extraRuleLabels`
- `ruleLabels`
- `additionalRuleLabels` - this one is used by kube-prometheus-stack chart

I would also have to make it explicit, that `severity` label is pre-defined.

klubi: I get your point, yet there are other big projects out there (like kube-prometheus-stack) that do allow for that, and actually have full blown customization in that area.
It would be great not to ask teams to copy-paste whole huge rule definitions for sake of adding 2-3 labels :/

klubi: Would be nice if this feature was accepted, there is however a workaround in place, so I will be able to achieve my goal. Unfortunately it will be suboptimal, and will require more maintenance, yet it will work :)

klubi:
It's been some time since any activity on this PR.
Can I get a statement around direction of this PR?

TylerHelmuth:
@klubi at this time we are going to stick with our stance that default rules/presets should be highly opinionated and not configurable.  Thank you for bringing this topic to our attention and engaging with us!

PR #892, opened 2023-09-15
update charts to use debug exporter

codeboten:
The logging exporter is being deprecated in favour of the debug exporter. This changes the charts to use that exporter instead.

Note: this will remain in draft until v0.86.0 is released which will be the first release to include the debug exporter.

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.69.0
+version: 0.70.0
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.70.0
+version: 0.69.1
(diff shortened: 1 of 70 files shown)


TylerHelmuth:
We need to bump the collector first since every commit is a release.

TylerHelmuth:
@codeboten we're ready for a rebase.  You'll need to bump the chart version again and then rerun `make generate-examples CHARTS=opentelemetry-collector`.

dmitryax (approved):

TylerHelmuth (commented):
@codeboten I led you astray bc I didn't see you adjusted all the charts.  You'll need to bump each chart version, then run `cd charts/opentelemetry-demo && helm dependencies build` and then run `make generate-examples`.

codeboten:
@TylerHelmuth ran the commands, it reverted the debug exporter back to the logging exporter in a few places... does this mean there's an upstream repo that needs updating as well?

codeboten:
See https://github.com/open-telemetry/opentelemetry-helm-charts/pull/892/commits/f2bb9aecf2b07fa307fcc92587dbaa6e66ad0b03

TylerHelmuth:
Ah I see whats happening. The demo chart depends on the collector chart that is has specified in the `chart.yaml` dependencies section.  That version of the collector hasn't had this change yet.  

For the demo to be right, we'll need to release the collector chart and then update the demo chart.  Let's take the demo chart out of this PR and update it after this is merged.

TylerHelmuth (approved):

PR #899, opened 2023-09-28
(demo)-collector: update prom endpoint

puckpuck:
Updates the prometheus endpoint to use the port number only.

The prior reference to an environment variable for this property was incorrect, using `env.` instead of `env:`. Removing the IP portion of the endpoint altogether makes this configuration easier to maintain.

### DIFF
--- a/charts/opentelemetry-demo/values.yaml
+++ b/charts/opentelemetry-demo/values.yaml
@@ -620,7 +620,7 @@ opentelemetry-collector:
           insecure: true
       # Create an exporter to Prometheus (metrics)
       prometheus:
-        endpoint: "${env.MY_POD_IP}:9464"
+        endpoint: ":9464"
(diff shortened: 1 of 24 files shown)


charts/opentelemetry-demo/values.yaml
```diff
@@ -620,7 +620,7 @@ opentelemetry-collector:
           insecure: true
       # Create an exporter to Prometheus (metrics)
       prometheus:
-        endpoint: "${env.MY_POD_IP}:9464"
+        endpoint: ":9464"
```
TylerHelmuth, line 623: How does this endpoint work? Is the prom chart doing some magic behind the scenes?

puckpuck:
I will close this in favor of a move to the OTLP write receiver built into Prometheus instead.

PR #904, opened 2023-09-29
[collector] Switch to tail mode for logsCollection preset

TylerHelmuth:
The `logsCollection` preset uses `start_at: beginning` in its filelog receiver configuration. This results in a significant amount of data to be ingested on startup, and significant duplicate data on a collector restart. I propose we switch to using `start_at:end` for default k8s log ingestion.

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.69.1
+version: 0.69.2
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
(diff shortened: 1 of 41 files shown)


dmitryax:
> significant duplicate data on a collector restart

This should not be the case with the checkpoint

dmitryax (approved):

TylerHelmuth:
@dmitryax agreed, checkpointing solve all these problems, but I don't feel we can turn it on by default since it requires elevated permissions.

dmitryax:
> checkpointing solve all these problems, but I don't feel we can turn it on by default since it requires elevated permissions.

Does filelog receiver have an option to disable it?

TylerHelmuth:
Yes, checkpoints are not enabled by default in the file log receiver. You must configure a storage extension first and then reference in the file log receiver collector. Our checkpoints preset handles that.

dmitryax:
Do you want to disable checkpointing by default in the preset to not run as root?

TylerHelmuth:
Yes, the preset has that option disabled by default: https://github.com/open-telemetry/opentelemetry-helm-charts/blob/3471a2afe2d5e01d23b4bc02f62ef70077c7dcc7/charts/opentelemetry-collector/values.yaml#L34

dmitryax:
I see. Makes sense 👍

Issue #906, opened 2023-10-03
[collector] add back the logging exporter temporarily


TylerHelmuth:
I realized that when we removed the logging exporter that we broke anyone who was relying on the default configuration when they included it in an exporters list. It is extremely common to add an exporter when using the chart - when people configure the exporters list with the new exporter AND the logging exporter, an upgrade to the latest chart would fail. 

I think we should re-add a logging exporter configuration for now so that either can be used. We should also update the UPGRADING doc. In a future, telegraphed release we can remove the default logging export config.

PR #907, opened 2023-10-03
[collector] re-add the logging exporter config

TylerHelmuth:
We accidentally created a breaking change when removing the logging exporter.  This PR adds back the loggingexporter configuration so that users who are explicitly listing it in an exporter list are not broken.  This gives users a chance to update their values.yaml before we completely remove the logging exporter.

Fixes https://github.com/open-telemetry/opentelemetry-helm-charts/issues/906

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.69.2
+version: 0.69.3
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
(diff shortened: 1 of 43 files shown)


codeboten (approved):

Allex1 (approved):

dmitryax:
What's the reason to define debug or logging exporter in the first place?

I think we should remove both going forward and keep only the components that are being used in the pipeline.

In order to enable the exporter, users would need to define it additionally:

```
config:
  exporters:
    debug: {}
  service:
    pipelines:
      traces:
        exporters:
          - otlp
          - debug
```

It's just two lines more but more explicit. And we wouldn't run into this problem if we didn't have the logging exporter defined

dmitryax:
BTW, we can check if `logging` exporter is used in any of the pipelines and throw a warning in the NOTES.txt

PR #922, opened 2023-10-13
(demo) fix jaeger spm and prometheus config

puckpuck:
References: [this PR](https://github.com/open-telemetry/opentelemetry-demo/pull/1174) from the demo repo

Configures Jaeger to properly support all the changes that have happened to the spanmetrics connector (formerly processor). It also removes unneeded Prometheus scrape configs since all data is now sent using the otlp write receiver.

Support for all the spanmetrics changes in Jaeger is done in the latest released version (1.50), but that version is not set with their Helm chart. An override on the Jaeger image tag is set. This override should be removed when the Jaeger Helm chart is updated with this version.


TylerHelmuth (approved):

MrAlias:
@dependabot rebase

PR #934, opened 2023-10-25
[demo] add replicas support

puckpuck:
Adds support to specify the number of replicas for all components or specifically for individual ones. Allowing to set replicas and various load generator settings allows the demo to be deployed in more complex performance-affected scenarios where observability tooling can be used to diagnose the root cause.

Shared demo components such as Redis, Postgres, and Kafka have a `replicas: 1` setting specified at the component level. All other components do not set/override this value, which will inherit from default (1).

### DIFF
--- a/charts/opentelemetry-demo/README.md
+++ b/charts/opentelemetry-demo/README.md
@@ -82,6 +82,7 @@ the demo
 | `default.image.tag`                    | Demo components image tag (leave blank to use app version)                                | `nil`                                                |
 | `default.image.pullPolicy`             | Demo components image pull policy                                                         | `IfNotPresent`                                       |
 | `default.image.pullSecrets`            | Demo components image pull secrets                                                        | `[]`                                                 |
+| `default.replicas`                     | Number of replicas for each component                                                     | `1`                                                  |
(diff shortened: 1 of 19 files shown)


TylerHelmuth:
@JaredTan95 can you take a look at this?

JaredTan95:
> @JaredTan95 can you take a look at this?

yes, glad to review

charts/opentelemetry-demo/README.md
```diff
@@ -82,6 +82,7 @@ the demo
 | `default.image.tag`                    | Demo components image tag (leave blank to use app version)                                | `nil`                                                |
 | `default.image.pullPolicy`             | Demo components image pull policy                                                         | `IfNotPresent`                                       |
 | `default.image.pullSecrets`            | Demo components image pull secrets                                                        | `[]`                                                 |
+| `default.replicas`                     | Number of replicas for each component                                                     | `1`                                                  |
```
JaredTan95, line 85: How about adding a note that shared  components such as `Redis, Postgres, and Kafka` should be set by `replicas` under `components`?

puckpuck: I set a note at the component level in the docs to mention this

JaredTan95 (approved):
LGTM

TylerHelmuth (approved):

PR #936, opened 2023-10-29
add option to reuse previously auto-generated cert

### LINKED ISSUE #845: [operator] helm recreates certs repetededly
Hello Team.

I use the opentelemetry-operator chart as a sub-chart and I've faced the issue where during helm upgrade I get an error:

```
cannot patch "opentelemetry" with kind OpenTelemetryCollector: 
Internal error occurred: failed calling webhook "mopentelemetrycollector.kb.io": 
failed to call webhook: Post "https://opentelemetry-operator-webhook.test.svc:443/mutate-opentelemetry-io-v1alpha1-opentelemetrycollector?timeout=10s": 
x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "opentelemetry-operator-operator-ca")
```

It looks like a certificate generated by Helm (certManager.enabled=False) is always replaced and the operator needs some time to reload it, during this time is not possible to update OpenTelemetryCollector's (the webhook has the previous cert). Could the chart use the old Secret if exists or maybe this regeneration is intentional? 

I can prepare a PR for it.

Thanks!

tomplus:
Add a new value `autoGenerateCertIfNotExists` to control whether certs has to be regenerated on each run or not.

It fixes #845 

Thanks!


charts/opentelemetry-operator/values.yaml
```diff
@@ -220,6 +220,9 @@ admissionWebhooks:
   ## If true and certManager.enabled is false, Helm will automatically create a self-signd cert and secret for you.
   autoGenerateCert: true
 
+  # Set true to reuse existing cert, autoGenerateCert must be enabled too
+  autoGenerateCertIfNotExists: true
```
JaredTan95, line 224: `autoGenerateCertIfNotExists` makes me feel a little confused, and feel separation with existing `autoGenerateCert`, how about put it as a child in `autoGenerateCert` configuration?

JaredTan95, line 224: The following configuration for reference only. If you apply this comments, you may need to change the logic in `operator-webhook.yaml`

```suggestion
  ## If true and certManager.enabled is false, Helm will automatically create a self-signd cert and secret for you.
  autoGenerateCert: 
    enabled: true
    # use the old secret  auto generated before, applicable to helm upgrade case.
    reuseExisting: true
```

or 

```suggestion
  ## If true and certManager.enabled is false, Helm will automatically create a self-signd cert and secret for you.
  autoGenerateCert: 
    enabled: true
    # true means always re-generate cert
    always: true
```

JaredTan95: The following configuration for reference only. If you apply this comments, you may need to change the logic in `operator-webhook.yaml`

```yaml
  ## If true and certManager.enabled is false, Helm will automatically create a self-signd cert and secret for you.
  autoGenerateCert: 
    enabled: true
    # use the old secret  auto generated before, applicable to helm upgrade case.
    reuseExisting: true
```

or 

```yaml
  ## If true and certManager.enabled is false, Helm will automatically create a self-signd cert and secret for you.
  autoGenerateCert: 
    enabled: true
    # true means always re-generate cert
    always: true
```

tomplus: I like your suggestions, I was worried about making breaking changes but if it's acceptable it's the best option. Thanks.

charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook.yaml
```diff
@@ -1,7 +1,27 @@
 {{- if and (.Values.admissionWebhooks.create) (not .Values.admissionWebhooks.certManager.enabled) }}
 {{- $altNames := list ( printf "%s-webhook.%s" (include "opentelemetry-operator.fullname" .) .Release.Namespace ) ( printf "%s-webhook.%s.svc" (include "opentelemetry-operator.fullname" .) .Release.Namespace ) -}}
-{{- $ca := genCA "opentelemetry-operator-operator-ca" 365 -}}
-{{- $cert := genSignedCert (include "opentelemetry-operator.fullname" .) nil $altNames 365 $ca -}}
+{{- $caCertEnc := "" }}
```
jinja2, line 3: Can we move this to a _helpers.tpl function to keep the manifest less busy?

charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook.yaml
```diff
@@ -1,7 +1,27 @@
...
+{{- $certCrtEnc := "" }}
+{{- $certKeyEnc := "" }}
+{{- if .Values.admissionWebhooks.autoGenerateCert.enabled }}
+{{- $prevSecret := (lookup "v1" "Secret" .Release.Namespace (default (printf "%s-controller-manager-service-cert" (include "opentelemetry-operator.fullname" .)) .Values.admissionWebhooks.secretName )) }}
+{{- $prevHook := (lookup "admissionregistration.k8s.io/v1" "MutatingWebhookConfiguration" .Release.Namespace (print (include "opentelemetry-operator.MutatingWebhookName" . ) "-mutation")) }}
```
jinja2, line 8: I would suggest we add the ca cert to the tls secret (`ca.crt`) instead. We can then get all 3 from the same object with one lookup.

charts/opentelemetry-operator/values.yaml
```diff
@@ -218,7 +218,10 @@ admissionWebhooks:
   ## TLS Certificate Option 2: Use Helm to automatically generate self-signed certificate.
   ## certManager must be disabled and autoGenerateCert must be enabled.
   ## If true and certManager.enabled is false, Helm will automatically create a self-signd cert and secret for you.
-  autoGenerateCert: true
+  autoGenerateCert:
+    enabled: true
+    # true means generate cert if cert not exists only
+    ifNotExists: true
```
jinja2, line 224:
```suggestion
    # If set to true, new webhook key/certificate is generated on helm upgrade.
    recreate: false
```
I think recreate might be more appropriate. I see user wanting to recreate the cert when the original expires in a year (might be ok to increase the validity when we generate the cert btw).

charts/opentelemetry-operator/values.yaml
JaredTan95: If it is consistent with the original logic(re-generate always), I think it is acceptable.

charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook.yaml
tomplus: Agree, moved.

charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook.yaml
tomplus: Done, but I still need to look up to one of the webhook for previously generated ca.

charts/opentelemetry-operator/values.yaml
tomplus: Done.

charts/opentelemetry-operator/values.yaml
```diff
@@ -218,7 +218,10 @@ admissionWebhooks:
   ## TLS Certificate Option 2: Use Helm to automatically generate self-signed certificate.
   ## certManager must be disabled and autoGenerateCert must be enabled.
   ## If true and certManager.enabled is false, Helm will automatically create a self-signd cert and secret for you.
-  autoGenerateCert: true
+  autoGenerateCert:
+    enabled: true
+    # If set to true, new webhook key/certificate is generated on helm upgrade.
+    recreate: false
```
JaredTan95, line 224: Can we keep this true with the original logical here？so we can make no breaking changes.

JaredTan95 (approved):
LGTM

charts/opentelemetry-operator/templates/_helpers.tpl
```diff
@@ -79,3 +79,39 @@ Create an ordered name of the MutatingWebhookConfiguration
...
+{{- else }}
+{{- $altNames := list ( printf "%s-webhook.%s" (include "opentelemetry-operator.fullname" .) .Release.Namespace ) ( printf "%s-webhook.%s.svc" (include "opentelemetry-operator.fullname" .) .Release.Namespace ) -}}
+{{- $ca := genCA "opentelemetry-operator-operator-ca" 365 }}
+{{- $cert := genSignedCert (include "opentelemetry-operator.fullname" .) nil $altNames 365 $ca }}
+{{- $certCrtEnc = b64enc $cert.Cert }}
+{{- $certKeyEnc = b64enc $cert.Key }}
+{{- $caCertEnc = b64enc $ca.Cert }}
+{{- end }}
+{{- else }}
+{{- $certCrtEnc = b64enc .Values.admissionWebhooks.cert_file }}
```
jinja2, line 111: Not directly related to this PR, since this variable/logic is not newly introduced. Is the variable `cert_file` meant to be the file path to cert? If that's the intentional, we should be using `Files.Get` to read the content. Same for the key/ca_file vars. We can fix this in a different issue.

jinja2 (approved):

charts/opentelemetry-operator/templates/_helpers.tpl
tomplus: Yes, I've noticed it too. In values.yaml there are comments that it's a path but it's treated as data.

Allex1 (approved):
lgtm
Not sure why `Unable to validate cosign version: 'v2.0.0'` started failing but seems unrelated to this pr.
@TylerHelmuth can you pls have a look ?

TylerHelmuth:
@tomplus I merged another Operator PR that also fixes the CI issue.  Please rebase and regenerate the examples. Thanks for contributing!

tomplus:
It's rebased.
Thanks to all for reviewing.

charts/opentelemetry-operator/values.yaml
```diff
@@ -221,7 +221,10 @@ admissionWebhooks:
...
```
TylerHelmuth, line 227: If we do this as a breaking change we need to add an entry in UPGRADING.md

charts/opentelemetry-operator/values.schema.json
```diff
@@ -1336,12 +1336,31 @@
                     }]
                 },
                 "autoGenerateCert": {
-                    "type": "boolean",
-                    "default": true,
-                    "title": "The autoGenerateCert schema",
-                    "examples": [
-                        true
-                    ]
+                    "type": "object",
```
TylerHelmuth, line 1339: Have you confirmed that this will cause users who are doing `autoGenerateCert: false` to fail an install?

charts/opentelemetry-operator/values.yaml
tomplus: My mistake, it should be True, fixed.

charts/opentelemetry-operator/values.schema.json
tomplus: Yes, it fails with an error - invalid type, expected object instead of boolean.
I can try to use oneOf in schema definition and detect which version is provided, what do you think?

TylerHelmuth: What you've done here is correct, but we should add an entry to UPGRADING.md that explains how to handle this breaking change since users doing `autoGenerateCert: false` will not be able to upgrade until that update their values.yaml.

tomplus: The same situation is with `autoGenerateCert: true`, it looks like a breaking change for me.

tomplus: I've updated UPGRADING.md

charts/opentelemetry-operator/UPGRADING.md
```diff
@@ -1,5 +1,10 @@
 # Upgrade guidelines
 
+## <0.42.1 to 0.42.1
+
+A type of flag `autoGenerateCert` has been changed, now it is an object with two attributes `enabled` and `recreate`.
+If you previously set `autoGenerateCert` to `true` or `false` you have to set `autoGenerateCert.enabled` accordingly.
```
JaredTan95, line 6:
```suggestion
If you previously set `autoGenerateCert=true`, no changes need. Otherwise, set `autoGenerateCert.enabled=false`.
In addition, you can use `autoGenerateCert.recreate` to decide whether to regenerate the cert or not.
```

tomplus: true is a default value, but if someone explicitly set it, upgrading chart will fail too...

```
helm upgrade my-test open-telemetry/opentelemetry-operator --install --set admissionWebhooks.autoGenerateCert=true

helm upgrade  my-test . --install --reuse-values

- admissionWebhooks.autoGenerateCert: Invalid type. Expected: object, given: boolean
```

setting the same value as a default value is not common so we can ignore it, right?

charts/opentelemetry-operator/values.schema.json
TylerHelmuth: @tomplus true.  If the user explicitly set the value to `autoGenerateCert: true` (which would be unnecessary since it is the default value), they'd experience an error.

charts/opentelemetry-operator/UPGRADING.md
TylerHelmuth: Setting the default value is not common, but some users do choose to include the whole values.yaml file in their config. I like your original text that references both values.

Allex1:
Hi @TylerHelmuth @JaredTan95  can we merge this ? Is it something else that needs adressing?

JaredTan95:
> Hi @TylerHelmuth @JaredTan95 can we merge this ? Is it something else that needs adressing?

I think we can get it in.

JaredTan95:
@tomplus pls resolve conflicts~

tomplus:
@JaredTan95  rebased, version updated

charts/opentelemetry-operator/_helpers.tpl
TylerHelmuth, line 1: Please delete this file - we only want the helpers from the template folder

TylerHelmuth (commented):
Our CI isn't able to test this new change because it always gets a fresh env. Can you provide some test results from you local tests of this feature?

charts/opentelemetry-operator/_helpers.tpl
tomplus: Sorry, accidentally committed.

tomplus:
my test:

1. install chart from my local directory:

```
helm install test-otel-op . -ntest3 --set admissionWebhooks.certManager.enabled=False
```

2. inspect created ca/key/cert (jq to get keys from secret to calc md5 sum)

```
kubectl get secret test-otel-op-opentelemetry-operator-controller-manager-service-cert -n test3 -ojson | jq '.data' | md5sum
b6f72381930de584378baf32ec1c3ab1  -
```

3. update chart

```
helm upgrade test-otel-op . -ntest3 --reuse-values
```

4. verify ca/key/cert - should be different

```
kubectl get secret test-otel-op-opentelemetry-operator-controller-manager-service-cert -n test3 -ojson | jq '.data' | md5sum
0d1225664c7691e59c157b9c38232ae5  -
```

5. upgrade again with option recreate=False

```
helm upgrade test-otel-op . -ntest3 --reuse-values --set admissionWebhooks.autoGenerateCert.recreate=False
```

6. verify certs, should be unchanged (the same checksum)

```
kubectl get secret test-otel-op-opentelemetry-operator-controller-manager-service-cert -n test3 -ojson | jq '.data' | md5sum
0d1225664c7691e59c157b9c38232ae5  -
```

7. and recreate=True again

```
helm upgrade test-otel-op . -ntest3 --reuse-values --set admissionWebhooks.autoGenerateCert.recreate=True
```

8. and I get new ca/cert/key

```
kubect get secret test-otel-op-opentelemetry-operator-controller-manager-service-cert -n test3 -ojson | jq '.data' | md5sum
95cd92880d3dc38ea6565cf1134c319e  -
```

TylerHelmuth (approved):

PR #945, opened 2023-11-09
exclude file logs where we have otlp logs

zeitlinger:
the duplicate logs are quite confusing

### DIFF
--- a/charts/opentelemetry-demo/values.yaml
+++ b/charts/opentelemetry-demo/values.yaml
@@ -612,6 +612,14 @@ opentelemetry-collector:
     opentelemetry_community_demo: "true"
   config:
     receivers:
+      filelog:
+        exclude:
+          - '/var/log/pods/*-otelcol*/opentelemetry-collector/*.log'
+          # logs are sent as OTLP, so we filter out the file logs
+          - '/var/log/pods/*/adservice/*.log'              # java
+          - '/var/log/pods/*/frauddetectionservice/*.log'  # kotlin
+          - '/var/log/pods/*/recommendationservice/*.log'  # python
+          - '/var/log/pods/*/cartservice/*.log'            # dotnet
       otlp:
         protocols:
           http:
--- a/charts/opentelemetry-demo/values.yaml
+++ b/charts/opentelemetry-demo/values.yaml
@@ -620,6 +620,7 @@ opentelemetry-collector:
           - '/var/log/pods/*/frauddetectionservice/*.log'  # kotlin
           - '/var/log/pods/*/recommendationservice/*.log'  # python
           - '/var/log/pods/*/cartservice/*.log'            # dotnet
+          - '/var/log/pods/*/quoteservice/*.log'           # php
       otlp:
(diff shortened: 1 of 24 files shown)


charts/opentelemetry-demo/values.yaml
```diff
@@ -612,6 +612,14 @@ opentelemetry-collector:
     opentelemetry_community_demo: "true"
   config:
     receivers:
+      filelog:
+        exclude:
+          - '/var/log/pods/*-otelcol*/opentelemetry-collector/*.log'
+          # logs are sent as OTLP, so we filter out the file logs
```
TylerHelmuth, line 618: @puckpuck this sounds like a bug with the demo.  Is the demo purposefully emitting the logs both as otlp and stdout?

zeitlinger: I thought this is the intended behavior - but I didn't see a definitive answer in https://opentelemetry.io/docs/specs/otel/logs/

TylerHelmuth: @zeitlinger are you adding logs to the demo yourself? I don't believe the demo does logs over otlp in 1.6

zeitlinger: I have enabled the flag for jvm (adservice, frauddetectionservice), but not for the others (recommendationservice/python, cartservice/dotnet, quoteservice/php)

JaredTan95:
pls run `make generate-examples` to render examples.

charts/opentelemetry-demo/values.yaml
zeitlinger: I would actually propose to add the java flag to enable logging as well - it will be enabled by default with 2.0 (of the java agent)

zeitlinger:
@JaredTan95 can you have another look?

JaredTan95:
I have a question about this, I didn't seem to see the configuration about collection logs in `otel-demo` values. If there is no configuration, why do we need to exclude logs?

JaredTan95:
I have a question about this, I didn't seem to see the configuration about collection logs in `otel-demo` values. If there is no configuration, why do we need to exclude logs? Could you remind me if I missed something :-P

JaredTan95:
I have a question about this, I didn't seem to see the configuration about collection logs in `otel-demo` values. If there is no configuration, why do we need to exclude logs? Could you remind me if I missed something :-P

zeitlinger:
> I have a question about this, I didn't seem to see the configuration about collection logs in `otel-demo` values. If there is no configuration, why do we need to exclude logs? Could you remind me if I missed something :-P

I've excluded logs for applications that are double reported - once for otlp logs and once for logs from the filelog receiver in the collector.

charts/opentelemetry-demo/values.yaml
puckpuck: The demo has started to add pieces to support logging, but it isn't complete yet. Any logging that services do is likely some defaults from an SDK that is not yet configured.

Issue #953, opened 2023-11-15
Cluster Role Binding subjects namespace incorrect when overriding namespace


nittendev:
Hi.

I've observed an issue when deploying the collector chart with an overriden namespace.
The clusterrolebinding gets the wrong subject namespace and incorrect access as a result.

I believe the issue is located here:
`charts/opentelemetry-collector/templates/clusterrolebinding.yaml`

Line 21:
`namespace: {{ .Release.Namespace }}`

I belive? it should be something like:
`namespace: {{ template "opentelemetry-collector.namespace" . }}`

etiennep:
I'd like to work on this.

TylerHelmuth:
@etiennep thank you, but there is already a PR submitted: https://github.com/open-telemetry/opentelemetry-helm-charts/pull/960

PR #960, opened 2023-11-18
[collector] fix clusterrolebinding binding  wrong namespace

JaredTan95:
fix https://github.com/open-telemetry/opentelemetry-helm-charts/issues/953


charts/opentelemetry-collector/examples/namespace-override/values.yaml
TylerHelmuth, line 1: I think this situation is basic enough that we don't need to add a new example for it.

Allex1 (approved):

TylerHelmuth (approved):

PR #962, opened 2023-11-18
fix: #961 for cartservice disable reload on config change

### LINKED ISSUE #961: CrashLoopBackOff in cartservice when fs.inotify.max_user_instances is set to a relatively low number
When deploying the opentelemetry-demo using the Helm chart, the cart service has a watcher responsible for reloading when files are changed. In host systems where the `fs.inotify.max_user_instances` is set to a relatively low value, the container will crash loop with a log message similar to the following:

```
Unhandled exception. System.IO.IOException: The configured user limit (128) on the number of inotify instances has been reached, or the per-process limit on the number of open file
```

aknot242:
Prevent cart service from watching the file system for changes and reloading to resolve #961. See details and rationale in #961.

### DIFF
--- a/charts/opentelemetry-demo/values.yaml
+++ b/charts/opentelemetry-demo/values.yaml
@@ -180,6 +180,8 @@ components:
         value: "8080"
       - name: ASPNETCORE_URLS
         value: http://*:$(CART_SERVICE_PORT)
+      - name: DOTNET_HOSTBUILDER__RELOADCONFIGONCHANGE
+        value: "false"
       - name: FEATURE_FLAG_GRPC_SERVICE_ADDR
         value: '{{ include "otel-demo.name" . }}-featureflagservice:50053'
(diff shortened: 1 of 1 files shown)


joshleecreates (commented):
Should this value maybe be baked into the docker image instead of managed via helm?

aknot242:
If you mean adding the env var to the Dockerfile, I think that could place an undue burden on developers contributing to this service if they are using [development containers](https://containers.dev/) and desire the ability to reload on source code changes.

TylerHelmuth (commented):
I'd like to get @puckpuck take on this as a default value for the chart. If this is not a widespread problem I'd prefer that it be handled by users via their values.yaml

puckpuck:
We should take this discussion over to the Demo repo. We don't need to put it in the Dockerfile, but having it as part of docker-compose and Helm may make sense, which is how this gets deployed.

PR #967, opened 2023-11-22
Add k8s infra monitorig demo example

ChrsMark:
This PR adds a new example for the `opentelemetry-demo` chart.
This example aims to provide a way to enable Kubernetes cluster monitoring by setting the proper collector's components.

Note that only node level metrics can be enabled because `clusterMetrics` requires `deployment` mode instead of `daemonset`. However, it would be nice if we could combine `clusterMetrics` as well so any ideas more than welcome. I guess this would require to edit the chart a bit so as to add an additional collector running as Deployment?

### DIFF
--- a/charts/opentelemetry-demo/Chart.yaml
+++ b/charts/opentelemetry-demo/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: opentelemetry-demo
-version: 0.26.0
+version: 0.27.0
 description: opentelemetry demo helm chart
 home: https://opentelemetry.io/
 sources:
--- a/charts/opentelemetry-demo/Chart.yaml
+++ b/charts/opentelemetry-demo/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: opentelemetry-demo
-version: 0.27.0
+version: 0.26.1
 description: opentelemetry demo helm chart
 home: https://opentelemetry.io/
 sources:
--- a/charts/opentelemetry-demo/examples/README.md
+++ b/charts/opentelemetry-demo/examples/README.md
@@ -6,6 +6,7 @@ Here is a collection of common configurations for the OpenTelemetry demo.  Each 
 - [Bring your own Observability](bring-your-own-observability)
 - [Collector as a Daemonset](collector-as-daemonset)
 - [Custom Environment Variables](custom-environment-variables)
+- [Kubernetes Infrastructure Monitoring](kubernetes-infra-monitoring)
 - [Public Hosted Ingress](public-hosted-ingress)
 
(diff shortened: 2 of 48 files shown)


charts/opentelemetry-demo/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: opentelemetry-demo
-version: 0.26.0
+version: 0.27.0
```
JaredTan95, line 4: This is not a functional PR, I don't think we need to update the minor version, how about `0.26.1`?

JaredTan95 (approved):

ChrsMark:
Fair point @povilasv, but if we think that the demo can also run on k8s I find value into combining traces of the services with the infrastructure metrics from the cluster the services are running on. In this we have a full stack Observability demonstration for Otel.

Allex1 (approved):
This is pretty awesome @ChrsMark .
Thanks

TylerHelmuth (commented):
Since our CI doesn't test using examples, can you confirm @ChrsMark that the new example produces a working instance of the demo?

ChrsMark:
Sure. Here is a complete working example that I used:

```yml
opentelemetry-collector:
  mode: "daemonset"
  presets:
    kubernetesAttributes:
      enabled: true
    kubeletMetrics:
      enabled: true
    hostMetrics:
      enabled: true
    logsCollection:
      enabled: true
      includeCollectorLogs: false
      storeCheckpoints: true

  extraEnvs:
    - name: ELASTIC_APM_ENDPOINT
      valueFrom:
        secretKeyRef:
          name: elastic-secret
          key: elastic_apm_endpoint
    - name: ELASTIC_APM_SECRET_TOKEN
      valueFrom:
        secretKeyRef:
          name: elastic-secret
          key: elastic_apm_secret_token
  config:
    exporters:
      otlp/elastic:
        endpoint: ${ELASTIC_APM_ENDPOINT}
        compression: none
        headers:
          Authorization: Bearer ${ELASTIC_APM_SECRET_TOKEN}
      otlp:
        protocols:
          grpc:
            endpoint: ${MY_POD_IP}:4317
          http:
            cors:
              allowed_origins:
              - http://*
              - https://*
            endpoint: ${MY_POD_IP}:4318
    service:
      pipelines:
        traces:
          receivers: [otlp]
          processors: [batch]
          exporters: [otlp/elastic]
        metrics:
          receivers: [otlp]
          processors: [batch]
          exporters: [otlp/elastic]
        logs:
          receivers: [otlp]
          processors: [batch]
          exporters: [otlp/elastic]
grafana:
  enabled: false

jaeger:
  enabled: false

prometheus:
  enabled: false
```

TylerHelmuth (approved):

PR #975, opened 2023-12-01
Bump collector to 0.90.1

dmitryax:


dmitryax:
I don't think so. It's a bugfix patch. Not another minor version of the collector

TylerHelmuth (approved):

Issue #891, opened 2023-09-13
[opentelemetry-collector] Remove `memory_ballast` extension from default template and replace by `GOMEMLIMIT`


mx-psi:
We want to remove the `memory_ballast` extension in favor of Go 1.19+ `GOMEMLIMIT` environment variable as stated in open-telemetry/opentelemetry-collector/issues/8343.

The first step to verify that we can effectively remove the extension is to change this on the OpenTelemetry Collector Helm chart.
We can replace the extension by a value of `GOMEMLIMIT` that is something like 80-90% of the total memory limit for the Collector container. Once we release this change and wait some time to gather feedback, we can move on with deprecation of the memory ballast extension.

mx-psi:
cc @dmitryax @TylerHelmuth

TylerHelmuth:
@mx-psi @dmitryax i think we should implement this via a feature-gate-like config. 

The first PR can add the option to use the Go setting instead of the extension, but disabled by default. After some time, we can enable it by default, but still allow users who want to use the extension to disable it. Then we'd wait until the extension is deprecated to remove the option.

mx-psi:
> i think we should implement this via a feature-gate-like config.

Have we done anything similar before on the Collector Helm chart? The plan sounds reasonable to me

mx-psi:
@TylerHelmuth should this be a usual configuration option or is there an alternative mechanism in the Helm chart for temporary toggles?

TylerHelmuth:
We dont have anything like feature gates in the chart, we normally do it via a configuration option (or 2).

mx-psi:
> I think the only thing we can do is not set the GOMEMLIMIT env?

I am personally fine with that, if you explicitly unset resource limits then you should be on your own to configure this

TylerHelmuth:
I have started work on this

mx-psi:
How much should we wait until we enable the GOMEMLIMIT option by default? Is something like one month enough?

TylerHelmuth:
@mx-psi is this helm chart feature critical for the decision on whether or not to deprecate the extension or has the decision already been made to deprecate it?

mx-psi:
> @mx-psi is this helm chart feature critical for the decision on whether or not to deprecate the extension or has the decision already been made to deprecate it?

[...]

PR #977, opened 2023-12-04
[collector] enable useGOMEMLIMIT by default

TylerHelmuth:
In order to move closer to deprecating the memory ballast extension, this PR enables the `useGOMEMLIMIT` feature flag by default.

Related to https://github.com/open-telemetry/opentelemetry-helm-charts/issues/891

### DIFF
--- a/charts/opentelemetry-collector/values.yaml
+++ b/charts/opentelemetry-collector/values.yaml
@@ -574,4 +574,4 @@ networkPolicy:
 # In a future release this setting will be enabled by default.
 # See https://github.com/open-telemetry/opentelemetry-helm-charts/issues/891
 # for more details.
-useGOMEMLIMIT: false
+useGOMEMLIMIT: true
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.75.1
+version: 0.76.0
 description: OpenTelemetry Collector Helm chart for Kubernetes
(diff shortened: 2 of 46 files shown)


charts/opentelemetry-collector/values.yaml
```diff
@@ -574,4 +574,4 @@ networkPolicy:
 # In a future release this setting will be enabled by default.
 # See https://github.com/open-telemetry/opentelemetry-helm-charts/issues/891
 # for more details.
-useGOMEMLIMIT: false
+useGOMEMLIMIT: true
```
TylerHelmuth, line 577: Enabling this by default does mean that anyone running an install of this chart without resources set just lost the memory ballast extension AND is not getting any value from `GOMEMLIMIT`.  Since the chart no longer has default resources, anyone who installs the default chart will be affected.  NOTES.txt has warning for both not having any resources AND enabling `useGOMEMLIMIT` without resources.


```
❯ helm install testing ./charts/opentelemetry-collector --set mode=deployment
NAME: testing
LAST DEPLOYED: Mon Dec  4 10:53:38 2023
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
[WARNING] No resource limits or requests were set. Consider setter resource requests and limits for your collector(s) via the `resources` field.

[WARNING] "useGOMEMLIMIT" is enabled but memory limits have not been supplied, which means no GOMEMLIMIT env var was configured but the Memory Ballast Extension was removed. It is highly recommended to only use "useGOMEMLIMIT" when memory limits have been set.
```

dmitryax (approved):
LGTM

Issue #997, opened 2024-01-16
[release]- Otel operator helm release v0.92.0 operator


vasireddy99:
[otel-operator v0.92.0](https://github.com/open-telemetry/opentelemetry-operator/releases/tag/v0.92.0) was released yesterday. Creating this request to release opentelemetry helm chart with v0.92.0 of [operator](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/Chart.yaml#L14) and v0.92.0 collector

PR #998, opened 2024-01-16
[operator] Bump operator version to 0.92.1

TylerHelmuth:
closes https://github.com/open-telemetry/opentelemetry-helm-charts/issues/997

### DIFF
--- a/charts/opentelemetry-operator/Chart.yaml
+++ b/charts/opentelemetry-operator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.45.2
+version: 0.46.0
 description: OpenTelemetry Operator Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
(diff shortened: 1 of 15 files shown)


Allex1 (approved):

TylerHelmuth:
@swiatekm-sumo I suspect [these tests](https://github.com/open-telemetry/opentelemetry-helm-charts/actions/runs/7548490691/job/20550655394?pr=998) are failing because of https://github.com/open-telemetry/opentelemetry-operator/pull/2479.  Do I need to update https://github.com/open-telemetry/opentelemetry-helm-charts/blob/21b8e825c72f038378795ebfda94d5b6139fe26e/.github/workflows/operator-test.yaml#L78 to do https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/.github/workflows/operator-test.yaml to do `make prepare-e2e` to get the right setup?

swiatekm-sumo:
@TylerHelmuth this is actually unrelated to that PR. The actual cause is https://github.com/open-telemetry/opentelemetry-operator/pull/2394, which should've required CRUD permissions on ClusterRoles and Bindings only conditionally, but seems to have unintentionally done so unconditionally here: https://github.com/open-telemetry/opentelemetry-operator/blob/7dc743fe1d406d28a2c1823f707b6d8fd1a32af2/controllers/opentelemetrycollector_controller.go#L145.

Realistically, I think whether the operator should manage RBAC for its resources should be configurable in the Chart, As is, we can either issue a patch release for the operator, or unconditionally add these permissions to the chart's ClusterRole. See #1004 for a demonstration that this actually fixes the tests.

TylerHelmuth:
Since the Operator team is discussing whether or not to release a patch version we'll hold off on this PR.

vasireddy99 (approved):

Issue #822, opened 2023-06-28
Add "retry_on_failure" setting to filelog


djaglowski:
User reports seeing the following error when using the `logsCollection` preset. 

`2023-06-27T20:26:11.105Z	error	consumerretry/logs.go:39	ConsumeLogs() failed. Enable retry_on_failure to slow down reading logs and avoid dropping.	{"kind": "receiver", "name": "filelog", "data_type": "logs", "error": "data refused due to high memory usage"}`

Likely the problem is that the `filelog` receiver is ingesting data at a faster rate than the collector can export. A new [`retry_on_failure`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/filelogreceiver#configuration) configuration was [recently added](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/20864), which allows the the receiver to respond to backpressure by blocking ingesting until data is successfully consumed by the next component.

Additional context: https://cloud-native.slack.com/archives/C01N5UCHTEH/p1687898643893809

TylerHelmuth:
Seems like a good idea.  @djaglowski is the best practice to enable this feature by default?

djaglowski:
> Seems like a good idea. @djaglowski is the best practice to enable this feature by default?

@dmitryax may be better suited to answer this but I think likely we should.

dmitryax:
Yes, I think we should enable it by default.

husnialhamdani:
Hi, Can I work on this?

PR #999, opened 2024-01-16
[collector] Enable `retry_on_failure` for filelog receiver created by logsCollection preset

husnialhamdani:
https://github.com/open-telemetry/opentelemetry-helm-charts/issues/822

Add `retry_on_failure` setting to filelog

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.78.0
+version: 0.78.1
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-collector/examples/daemonset-and-deployment/rendered/configmap-agent.yaml
+++ b/charts/opentelemetry-collector/examples/daemonset-and-deployment/rendered/configmap-agent.yaml
@@ -6,7 +6,7 @@ metadata:
   name: example-opentelemetry-collector-agent
   namespace: default
   labels:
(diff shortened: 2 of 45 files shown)


husnialhamdani:
@TylerHelmuth am I adding on a correct place?

TylerHelmuth:
@husnialhamdani no, you'll want to make the changes here: https://github.com/open-telemetry/opentelemetry-helm-charts/blob/21b8e825c72f038378795ebfda94d5b6139fe26e/charts/opentelemetry-collector/templates/_config.tpl#L190.  We only add the filelog receiver when the `presets.logsCollection` is enabled, so we should not modify the default config.

husnialhamdani:
> @husnialhamdani no, you'll want to make the changes here:
> 
> https://github.com/open-telemetry/opentelemetry-helm-charts/blob/21b8e825c72f038378795ebfda94d5b6139fe26e/charts/opentelemetry-collector/templates/_config.tpl#L190
> 
> . We only add the filelog receiver when the `presets.logsCollection` is enabled, so we should not modify the default config in the values.yaml.

updating

husnialhamdani:
> Please update chart version and generate examples

Will it minor or major version change?,
for generate example, i think we should add only under [this directory](https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-collector/examples/daemonset-collector-logs), isn't it?

Is there any hack script or just render the Helm chart manually?

TylerHelmuth (commented):
@husnialhamdani please regenerate the examples 1 more time as bumping the version affects the examples.

husnialhamdani:
> @husnialhamdani please regenerate the examples 1 more time as bumping the version affects the examples.

done

TylerHelmuth (approved):

PR #1000, opened 2024-01-17
[demo] - fix grafana dashboards

puckpuck:
Updates all dependencies to their latest version (including Grafana) and updates the Grafana definitions. There was also an issue with the Prometheus scrape configuration that prevented the dashboards from working properly.

### DIFF
--- a/charts/opentelemetry-demo/Chart.yaml
+++ b/charts/opentelemetry-demo/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: opentelemetry-demo
-version: 0.27.0
+version: 0.27.1
 description: opentelemetry demo helm chart
 home: https://opentelemetry.io/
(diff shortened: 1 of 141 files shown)


charts/opentelemetry-demo/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: opentelemetry-demo
-version: 0.27.0
```
JaredTan95, line 4: how about bump up to `0.28.0`?

puckpuck: I've typically reserved minor upgrades for when the Demo itself would get bumped a minor version or we had a structural change to this chart itself

Allex1 (approved):

TylerHelmuth (approved):

PR #1007, opened 2024-01-17
[demo] add mountedConfigMaps support

puckpuck:
This feature is in preparation to allow users to set default Feature Flag values, as well as any other default configuration various different components offer.

It introduces a new property at the components level called `mountedConfigMaps` which allows the user to create volume mounts on the components that reference existing config maps, or leverage an inline data sub-property to specify the contents of a config map that will be created by the chart.

Using existing config map:
```yaml
components:
  ffsPostgres:
    mountedConfigMaps:
      - name: flag-values
        mountPath: /docker-entrypoint-initdb.d
        existingConfigMap: my-ffs-initialvalues
```

Using inline configuration data which will be part of created configmap:
```yaml
components:
  ffsPostgres:
    mountedConfigMaps:
      - name: flag-values
        mountPath: /docker-entrypoint-initdb.d
        data:
          30-flagValues.sql: |
            -- SQL UPDATE statements go here
```

This is also a BREAKING CHANGE. This replaces, and augments existing though seldom used functionality from the demo. A note in UPGRADING.md on how to migrate the old `configuration` property to the new `mountedConfigMaps` is included in this PR.


JaredTan95 (approved):

TylerHelmuth (commented):
Please add this new section to the comment in the values.yaml above all the components.  Do you think we should support this in the defaults?

puckpuck:
Updated the comments in the values files. This shouldn't be set at the default level. I don't know what use case we would support needing to mount the same volume and content to each component.

TylerHelmuth (approved):

PR #1014, opened 2024-01-24
Added environment variable and increased memory limit of loadgenerator

jordibisbal8:
In this PR, we aim to add the environment variable `LOCUST_BROWSER_TRAFFIC_ENABLED` and increase the memory limit for the load generation. The main reason is to be able to run playwright in locust and generate web browser traffic. For more information, please have a look at the [corresponding PR](https://github.com/open-telemetry/opentelemetry-demo/pull/1345)

### DIFF
--- a/charts/opentelemetry-demo/Chart.yaml
+++ b/charts/opentelemetry-demo/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: opentelemetry-demo
-version: 0.28.0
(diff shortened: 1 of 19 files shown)


puckpuck:
This change makes sense, but we should wait until after the demo is released with the updated loadgenerator, which will have a new version (image tag) associated with it.

puckpuck:
This PR is going to be addressed as part of #1026

Issue #1006, opened 2024-01-17
opentelemetry-collector.convertMemToMib produces 0 if using a decimal point


snuggie12:
This is fairly easily worked around, but it did cause our HPAs to max out so figured I'd make the ticket in case anyone else runs into it.

We had a memory limit of `1.5Gi` in place and the `atoi` function will convert that to `0`. This is especially bad with `GOMEMLIMIT` since it out of nowhere started using 100% of the CPU.

TylerHelmuth:
You're not wrong, but also I don't have a great solution in mind for this other than don't use decimals. Maybe https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/30289 can help

PR #1015, opened 2024-01-25
[opentelemetry-collector] Fix calculation for gomemlimit

jinja2:
GOMEMLIMIT is set to 0 when the memory limit is floating point value. Similar problem raised in this [issue](https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1006). This PR updates the calculation to work with floats and adds "unit tests" to verify the env var is set correctly.

### DIFF
--- a/.github/actions/setup/action.yaml
+++ b/.github/actions/setup/action.yaml
@@ -33,3 +33,4 @@ runs:
           helm repo add prometheus https://prometheus-community.github.io/helm-charts
           helm repo add grafana https://grafana.github.io/helm-charts
           helm repo add jaeger https://jaegertracing.github.io/helm-charts
+          helm plugin install https://github.com/helm-unittest/helm-unittest.git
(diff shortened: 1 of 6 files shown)


TylerHelmuth (commented):
@jinja2 thanks for doing this.  Your unit tests are enough to ensure me that your math is correct, I don't think we need to permanently check them in and add a new testing framework.

jinja2:
It is pretty handy to test templating logic, but I understand there is an additional overhead to maintaining another test suite.

TylerHelmuth (approved):

ramdaspotale:
Hi @jinja2 - i am trying to upgrade open telemetry collector to use helm chart from 0.72.0 to latest version and it is breaking my helm install.

i dont see mulf function defined elsewhere and also its not built in function too, how it is supposed to work. am i missing anything?

`stderr: Error: parse error at (opentelemetry-collector/templates/_helpers.tpl:178): function \"mulf\" not defined\n",`

PR #1020, opened 2024-01-31
Allow to configure a local and global image registry

aacevedoosorio:
Handy when you have specific security requirements that prevent you from using a public image registry

### DIFF
--- a/charts/opentelemetry-collector/templates/_pod.tpl
+++ b/charts/opentelemetry-collector/templates/_pod.tpl
@@ -24,9 +24,9 @@ containers:
       {{- toYaml .Values.securityContext | nindent 6 }}
       {{- end }}
     {{- if .Values.image.digest }}
-    image: "{{ include "common.image.registry" (dict "image" .Values.image "context" .) }}{{ .Values.image.repository }}@{{ .Values.image.digest }}"
+    image: "{{ (.Values.global).imageRegistry | default "" }}{{ .Values.image.repository }}@{{ .Values.image.digest }}"
     {{- else }}
-    image: "{{ include "common.image.registry" (dict "image" .Values.image "context" .) }}{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+    image: "{{ (.Values.global).imageRegistry | default "" }}{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
     {{- end }}
     imagePullPolicy: {{ .Values.image.pullPolicy }}
 
--- a/charts/opentelemetry-collector/templates/_pod.tpl
+++ b/charts/opentelemetry-collector/templates/_pod.tpl
@@ -24,9 +24,9 @@ containers:
       {{- toYaml .Values.securityContext | nindent 6 }}
       {{- end }}
     {{- if .Values.image.digest }}
-    image: "{{ .Values.image.repository }}@{{ .Values.image.digest }}"
+    image: "{{ include "common.image.registry" (dict "image" .Values.image "context" .) }}/{{ .Values.image.repository }}@{{ .Values.image.digest }}"
     {{- else }}
-    image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+    image: "{{ include "common.image.registry" (dict "image" .Values.image "context" .) }}/{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
     {{- end }}
     imagePullPolicy: {{ .Values.image.pullPolicy }}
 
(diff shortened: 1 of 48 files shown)


JaredTan95:
@aacevedoosorio pls bump up the chart version by following [CONTRIBUTING docs](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/CONTRIBUTING.md)

aacevedoosorio:
> Question, why do we need `global.imageRegistry`? Why can't you set the `.image.registry` directly?
> 
> Also is this standard helm chart convention?

I have found this convention used on bitnami helm charts and by setting the `global.imageRegistry` you automatically set all images to be downloaded from that registry rather than having to override `.image.registry` in several placed. E.g. https://github.com/bitnami/charts/blob/main/bitnami/clickhouse/values.yaml#L15

charts/opentelemetry-collector/values.yaml
```diff
@@ -1,6 +1,10 @@
+
 # Default values for opentelemetry-collector.
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
+global:
+  # global.imageRegistry -- Global Docker image registry
+  imageRegistry: ""
```
TylerHelmuth, line 7: We only have 1 image in this helm chart so I'm not sure what this is providing us.  If you're trying to us this chart as a subchart and the `"additionalProperties": false,` schema is preventing the failure we can update the schema file.  We already have some other common `global` values in there.

aacevedoosorio: Thanks for the suggestion @TylerHelmuth . Indeed I'm trying to use this chart as a subchart and I found the `additionalProperties": false` is not blocking me at all. I just dropped the irrelevant `global.imageRegistry`

charts/opentelemetry-collector/values.yaml
```diff
@@ -163,6 +163,8 @@ config:
           - zipkin
 
 image:
+  # Registry where to find the image
+  registry: docker.io
```
TylerHelmuth, line 167: I think we can continue to get by without any new fields here.  Since this chart only has one image if a user wants to specify a specific registry they can add it to `image.repository`.

For the subchart situation, I think we can update the template to prepend `global.imageRegistry` if it is present.

aacevedoosorio: I applied the suggestion or dropping `image.repository`.
But I left `global.imageRegistry` to follow the bitnami helm charts https://github.com/bitnami/charts/blob/main/bitnami/clickhouse/values.yaml#L15 for my usage of the `otel-collector` chart as a subchart

charts/opentelemetry-collector/values.yaml
```diff
@@ -164,7 +164,7 @@ config:
 
 image:
   # If you want to use the core image `otel/opentelemetry-collector`, you also need to change `command.name` value to `otelcol`.
-  repository: otel/opentelemetry-collector-contrib
+  repository: otel/opentelemetry-collector-contrib2
```
TylerHelmuth, line 167:
```suggestion
  repository: otel/opentelemetry-collector-contrib
```

charts/opentelemetry-collector/values.yaml
TylerHelmuth, line 1: We can remove all changes from this file.

charts/opentelemetry-collector/templates/_pod.tpl
```diff
@@ -24,9 +24,9 @@ containers:
       {{- toYaml .Values.securityContext | nindent 6 }}
       {{- end }}
     {{- if .Values.image.digest }}
-    image: "{{ .Values.image.repository }}@{{ .Values.image.digest }}"
+    image: "{{ include "common.image.registry" (dict "image" .Values.image "context" .) }}{{ .Values.image.repository }}@{{ .Values.image.digest }}"
     {{- else }}
-    image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+    image: "{{ include "common.image.registry" (dict "image" .Values.image "context" .) }}{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
```
TylerHelmuth, line 29: Could this be simplified to 

`{{ .Values.global.imageRegistry | default "" }}`

charts/opentelemetry-collector/values.yaml
```diff
@@ -1,3 +1,4 @@
+
```
TylerHelmuth, line 1:
```suggestion
```

TylerHelmuth (approved):
Thank you!

TylerHelmuth:

aacevedoosorio:
> Please bump the chart patch version and run `make generate-examples CHARTS=opentelemetry-collector`

```
 opentelemetry-collector => (version: "0.80.1", path: "charts/opentelemetry-collector") > chart version not ok. Needs a version bump! 
```
new version bump?

TylerHelmuth:
> new version bump?

Looks like you took care of it

PR #1026, opened 2024-02-02
[demo] - release 1.8.0

puckpuck:
This release adds a new subchart: OpenSearch.

The demo has added Logs support, and it was decided to use OpenSearch for the logging backend. This sets up a single-node OpenSearch cluster as part of the helm chart.

I have a concern that the OpenSearch chart doesn't use the Helm Chart Release name for the service and pod names, nor does it allow us to template inline variables as part of the `nameOverride` or `fullnameOverride` properties. As such, we end up with a hardcoded prefix of `otel-demo` for the OpenSearch pod and service names.


puckpuck:
This one is ready for review.

It also incorporates the changes needed for the currency service, which supersedes #1039

TylerHelmuth (approved):

PR #1027, opened 2024-02-02
[operator] Bump operator to 0.93.0

TylerHelmuth:


jetersen:
What about the collector-contrib inside the operator?
https://github.com/open-telemetry/opentelemetry-helm-charts/blob/5a9ec97a5d52167b7eaa5a641462cdbdb448c505/charts/opentelemetry-operator/values.yaml#L33-L35

TylerHelmuth:
You're right I need to do that. I should update the release notes too

PR #1029, opened 2024-02-05
create featureflags table and populate it

fvlankvelt:
The postgres database used by the featureflagservice is not initialized / migrated any more by ecto.  This PR tries to mimic the behavior of the `docker-compose.yml` of the `opentelemetry-demo` project.


puckpuck:
This is fixed in the current build on main of the demo, where the table is created as part of the postgres image. When that version gets released we will update this Helm chart to work with it.

puckpuck:
This PR is going to be addressed as part of #1026

PR #1033, opened 2024-02-10
[demo] Add seccompProfile property to securityContext

terjesannum:
Specifying seccompProfile in securityContext fail validation, this PR adds seccompProfile to the values schema


TylerHelmuth (commented):
Since we're using the entire `securityContext` object as is, let's remove any specific properties from the schema value and allow additional properties.

TylerHelmuth (approved):

puckpuck (approved):

PR #1034, opened 2024-02-12
Add insecure_skip_verify in kubeletMetrics preset

ChrsMark:
The OTel documentation for the [Kubeletstats Receiver](https://opentelemetry.io/docs/kubernetes/collector/components/#kubeletstats-receiver) component suggests the following config:

```yaml
receivers:
  kubeletstats:
    collection_interval: 10s
    auth_type: 'serviceAccount'
    endpoint: '${env:K8S_NODE_NAME}:10250'
    insecure_skip_verify: true
    metric_groups:
      - node
      - pod
      - container
```

However [kubeletMetrics preset](https://opentelemetry.io/docs/kubernetes/helm/collector/#kubelet-metrics-preset) of the respective Helm Chart does not include the `insecure_skip_verify`.

Thus, using the preset results in not being able to scrape metrics from Kubelet:

```console
2024-02-12T13:06:38.943Z	error	scraperhelper/scrapercontroller.go:200	Error scraping metrics	{"kind": "receiver", "name": "kubeletstats", "data_type": "metrics", "error": "Get \"https://kind-worker2:10250/stats/summary\": tls: failed to verify certificate: x509: certificate signed by unknown authority", "scraper": "kubeletstats"}
``` 

I guess it would be nice to skip the tls verification with the preset as well since this is mostly the case and would also be aligned with what the plain config example suggests.

### DIFF
--- a/charts/opentelemetry-collector/templates/_config.tpl
+++ b/charts/opentelemetry-collector/templates/_config.tpl
@@ -169,6 +169,7 @@ receivers:
     collection_interval: 20s
     auth_type: "serviceAccount"
     endpoint: "${env:K8S_NODE_NAME}:10250"
+    insecure_skip_verify: true
 {{- end }}
 
 {{- define "opentelemetry-collector.applyLogsCollectionConfig" -}}
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.80.1
+version: 0.81.0
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-collector/examples/daemonset-and-deployment/rendered/configmap-agent.yaml
+++ b/charts/opentelemetry-collector/examples/daemonset-and-deployment/rendered/configmap-agent.yaml
@@ -6,7 +6,7 @@ metadata:
   name: example-opentelemetry-collector-agent
   namespace: default
   labels:
-    helm.sh/chart: opentelemetry-collector-0.80.1
+    helm.sh/chart: opentelemetry-collector-0.81.0
     app.kubernetes.io/name: opentelemetry-collector
     app.kubernetes.io/instance: example
     app.kubernetes.io/version: "0.93.0"
--- a/charts/opentelemetry-collector/examples/daemonset-and-deployment/rendered/configmap.yaml
+++ b/charts/opentelemetry-collector/examples/daemonset-and-deployment/rendered/configmap.yaml
@@ -6,7 +6,7 @@ metadata:
   name: example-opentelemetry-collector
   namespace: default
   labels:
-    helm.sh/chart: opentelemetry-collector-0.80.1
+    helm.sh/chart: opentelemetry-collector-0.81.0
     app.kubernetes.io/name: opentelemetry-collector
     app.kubernetes.io/instance: example
     app.kubernetes.io/version: "0.93.0"
--- a/charts/opentelemetry-collector/examples/daemonset-and-deployment/rendered/daemonset.yaml
+++ b/charts/opentelemetry-collector/examples/daemonset-and-deployment/rendered/daemonset.yaml
@@ -6,7 +6,7 @@ metadata:
   name: example-opentelemetry-collector-agent
   namespace: default
   labels:
-    helm.sh/chart: opentelemetry-collector-0.80.1
+    helm.sh/chart: opentelemetry-collector-0.81.0
     app.kubernetes.io/name: opentelemetry-collector
     app.kubernetes.io/instance: example
     app.kubernetes.io/version: "0.93.0"
@@ -23,7 +23,7 @@ spec:
   template:
     metadata:
       annotations:
-        checksum/config: ebb07485a27d8b65d144c469236066f85d4e2c578b0cabf9f4826e3150428c8f
+        checksum/config: 9ca53e5a2d20bd271d191ca1343ee63ffbd60f7bf8e49c33a6ba4d53e8711f04
         
       labels:
         app.kubernetes.io/name: opentelemetry-collector
--- a/charts/opentelemetry-collector/examples/daemonset-and-deployment/rendered/deployment.yaml
+++ b/charts/opentelemetry-collector/examples/daemonset-and-deployment/rendered/deployment.yaml
@@ -6,7 +6,7 @@ metadata:
   name: example-opentelemetry-collector
   namespace: default
   labels:
-    helm.sh/chart: opentelemetry-collector-0.80.1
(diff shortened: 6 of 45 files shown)


charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -169,6 +169,7 @@ receivers:
     collection_interval: 20s
     auth_type: "serviceAccount"
     endpoint: "${env:K8S_NODE_NAME}:10250"
+    insecure_skip_verify: true
```
TylerHelmuth, line 172: I don't think we want to enable this by default since it isn't the recommended behavior for production.  Really the receiver's README should be updated to not set that value. If the cluster's certificates are in order this value is not necessary.  @jinja2 do you agree?

TylerHelmuth: Also @povilasv and @dmitryax

ChrsMark: At the moment the `preset` does not add any of the `ca_file`, `key_file`, `cert_file`, right? So I wonder in which cases the preset is actually useful with the current implementation.

ref: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/27070/files#r1334828771

TylerHelmuth: Someone correct me if I'm wrong, but I thought the receiver was looking in [an expected place for the certs](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/e1403742f39d090bfc3c164f5a295d1783df3fb1/internal/kubelet/client.go#L25-L26) and `ca_file`, `key_file`, and `cert_file` was only necessary if your certs didn't live in that expected place.

Are the certs for a standard production k8s cluster no longer installed at that location?

jinja2: The preset in the chart is using the `serviceAccount` auth type. This auth type does not use the `ca_file`, `key_file`, `cert_file` options at all. These are used for auth type = TLS so basically if you are using client certificates to authenticate. With serviceAccount auth we get the cluster's root CA cert so it can be added to the trust store/cert pool, and this is the `ca.crt` we see along with the token for service account at the well-known path that Tyler linked. The standard practice is to setup kubelet to serve certificate signed by the cluster CA which we have access to already when using serviceaccount auth.

Some setups might not set the k8s node name in the Subject alternative name in the kubelet's server cert. In such cases users might get the `hostname mismatch` error since we use the K8S_NODE_NAME as host here. The referenced [comment]( https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/27070/files#r1334828771) has more details on my findings on some of the more commonly used distros.

I would recommend to not add `insecure_skip_verify: true` as default cause that would be a step back in terms of secure practices. This is the first issue about this setting after the change in behavior went out so it might not be impacting most users. A compromise might be an option under the preset variable?

The error posted in the issue with running in kind cluster can most likely be fixed by adding below patch to the kind cluster's config.
```
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
kubeadmConfigPatches:
  - |
    kind: KubeletConfiguration
    serverTLSBootstrap: true
 ```
 After the cluster is up, you'll probably need to approve the certificate signing request for the kubelet certs by running `kubectl certificate approve {PENDING CSR}`.

Comment (ChrsMark) [reply charts/opentelemetry-collector/templates/_config.tpl:172]: Thank's for the details @jinja2 @TylerHelmuth on how the internals of `serviceAccount` auth type work.

Based on what you shared `insecure_skip_verify: true` does not make a lot of sense to be the default.
I tested the preset on a GKE cluster and works without any issues.
However would that make sense to document these details around the `serviceAccount` type so as to make it clear how it works and what is required in order for this work?

Comment (TylerHelmuth) [reply charts/opentelemetry-collector/templates/_config.tpl:172]: @ChrsMark Yes, let's definitely document it and the possible fix.

**Comment by ChrsMark** [comment]
Closing this in honor of https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/31280 as per the discussions at https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1034#discussion_r1486645971. Let's continue the conversation there.

**Comment by ChrsMark** [comment]
Bumping this up again since I got an error while testing the preset GKE.

Config:

```yaml
mode: daemonset
presets:
  kubeletMetrics:
    enabled: true

config:
  exporters:
    debug:
      verbosity: detailed

  service:
    pipelines:
      metrics:
        receivers: [kubeletstats]
        processors: [batch]
        exporters: [debug]
```

Logs:

```console
2024-07-03T13:16:16.966Z	error	kubeletstatsreceiver@v0.100.0/scraper.go:77	call to /stats/summary endpoint failed	{"kind": "receiver", "name": "kubeletstats", "data_type": "metrics", "error": "Get \"https://gke-otel-chrismark-test-default-pool-b71e7458-hnfn:10250/stats/summary\": tls: failed to verify certificate: x509: certificate is not valid for any names, but wanted to match gke-otel-chrismark-test-default-pool-b71e7458-hnfn"}
github.com/open-telemetry/opentelemetry-collector-contrib/receiver/kubeletstatsreceiver.(*kubletScraper).scrape
	github.com/open-telemetry/opentelemetry-collector-contrib/receiver/kubeletstatsreceiver@v0.100.0/scraper.go:77
go.opentelemetry.io/collector/receiver/scraperhelper.ScrapeFunc.Scrape
	go.opentelemetry.io/collector/receiver@v0.100.0/scraperhelper/scraper.go:20
go.opentelemetry.io/collector/receiver/scraperhelper.(*controller).scrapeMetricsAndReport
	go.opentelemetry.io/collector/receiver@v0.100.0/scraperhelper/scrapercontroller.go:194
go.opentelemetry.io/collector/receiver/scraperhelper.(*controller).startScraping.func1
	go.opentelemetry.io/collector/receiver@v0.100.0/scraperhelper/scrapercontroller.go:169
2024-07-03T13:16:16.967Z	error	scraperhelper/scrapercontroller.go:197	Error scraping metrics	{"kind": "receiver", "name": "kubeletstats", "data_type": "metrics", "error": "Get \"https://gke-otel-chrismark-test-default-pool-b71e7458-hnfn:10250/stats/summary\": tls: failed to verify certificate: x509: certificate is not valid for any names, but wanted to match gke-otel-chrismark-test-default-pool-b71e7458-hnfn", "scraper": "kubeletstats"}
go.opentelemetry.io/collector/receiver/scraperhelper.(*controller).scrapeMetricsAndReport
	go.opentelemetry.io/collector/receiver@v0.100.0/scraperhelper/scrapercontroller.go:197
go.opentelemetry.io/collector/receiver/scraperhelper.(*controller).startScraping.func1
	go.opentelemetry.io/collector/receiver@v0.100.0/scraperhelper/scrapercontroller.go:169
```

Container image:  `"otel/opentelemetry-collector-k8s:0.100.0"`
GKE version: `v1.29.4-gke.1043002`

I don't get these same on EKS, so I guess that sth changed recently on how GKE generates these certs.

PR #1039, opened 2024-02-16
Add OTLP logs to currencyservice

julianocosta89:
Add `VERSION` env var to `currencyservice` as the value is required since: https://github.com/open-telemetry/opentelemetry-demo/pull/1378.

### DIFF
--- a/charts/opentelemetry-demo/Chart.yaml
+++ b/charts/opentelemetry-demo/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: opentelemetry-demo
-version: 0.28.3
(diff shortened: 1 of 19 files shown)


julianocosta89:
@puckpuck could we get this one merged?
Or could we include it to the 1.8 helm chart?

Without `VERSION` env var the `currencyservice` won't work.

puckpuck:
> @puckpuck could we get this one merged? Or could we include it to the 1.8 helm chart?
> 
> Without `VERSION` env var the `currencyservice` won't work.

I incorporated this into the Helm 1.8.0 release PR.

PR #1050, opened 2024-02-23
[operator] new value for k8s cluster domain

jouve:
override k8s default cluster domain (cluster.local)

### DIFF
--- a/charts/opentelemetry-operator/values.yaml
+++ b/charts/opentelemetry-operator/values.yaml
@@ -11,6 +11,9 @@ nameOverride: ""
 ## Reference one or more secrets to be used when pulling images from authenticated repositories.
 imagePullSecrets: []
 
+## Kubernetes cluster domain suffix
+clusterDomain: cluster.local
+
 ## Pod Disruption Budget configuration
 ##
 pdb:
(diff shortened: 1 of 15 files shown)


Allex1:
@jouve thanks for your contribution .
Can you please bump the chart version and run `make generate-examples CHARTS=opentelemetry-operator` so that the CI can pass ?

jouve:
done :)

Allex1 (approved):

charts/opentelemetry-operator/values.yaml
```diff
@@ -11,6 +11,9 @@ nameOverride: ""
 ## Reference one or more secrets to be used when pulling images from authenticated repositories.
 imagePullSecrets: []
 
+## Kubernetes cluster domain suffix
+clusterDomain: cluster.local
```
TylerHelmuth, line 15: I think we should nest this under `certManager` unless such a setting would be useful other places in the chart.

jouve: I've found lots of charts have this as clusterDomain or global.clusterDomain, so it's less surprising there

jouve:
how can I update the values.schema.yaml ?

Allex1:
> how can I update the values.schema.json ?

it's mostly a manual process currently

TylerHelmuth (approved):
Thanks for sticking with this!

PR #1051, opened 2024-02-23
[operator] Bump operator to 0.94.0

TylerHelmuth:


TylerHelmuth:
@open-telemetry/helm-approvers @open-telemetry/operator-approvers I wont have time to look at this for the next 2 weeks.  I believe the issue is only about setting up the tests properly, not the operator or the chart.

Honestly maybe we dont even need to run these tests? We don't do the same for the other charts.

Allex1:
I'll take a look

swiatekm-sumo:
I can take care of this as well, I've done so in the past and I'm familiar with operator E2E tests.

Allex1:
@swiatekm-sumo it's all yours. Thanks

Issue #1056, opened 2024-03-01
[opentelemetry-collector] StatefulSets do not delete PVCs when scaled or deleted


lindeskar:
When creating PVCs through the OTelCol StatefulSet the PVCs are not removed on scale-down or when the StatefulSet is deleted.

This behaviour could be changed by adding `.spec.persistentVolumeClaimRetentionPolicy`.

More information here: https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#persistentvolumeclaim-retention

lindeskar:
#1057 will allow setting the values:

```yaml
statefulset:
  persistentVolumeClaimRetentionPolicy:
    enabled: true
    whenDeleted: Delete
    whenScaled: Delete
```

which resolves this issue.

PR #1057, opened 2024-03-01
[opentelemetry-collector] add persistentVolumeClaimRetentionPolicy in StatefulSet

lindeskar:
Fixes: #1056

- Add `persistentVolumeClaimRetentionPolicy` in StatefulSet template
- Add values to enable and customize `persistentVolumeClaimRetentionPolicy`
- Add new example: "statefulset-with-pvc"


charts/opentelemetry-collector/templates/statefulset.yaml
```diff
@@ -17,6 +17,11 @@ spec:
   podManagementPolicy: {{ .Values.statefulset.podManagementPolicy }}
 {{- if not .Values.autoscaling.enabled }}
   replicas: {{ .Values.replicaCount }}
+{{- end }}
+{{- if and ( semverCompare ">= 1.23-0" .Capabilities.KubeVersion.Version ) ( .Values.statefulset.persistentVolumeClaimRetentionPolicy.enabled )  }}
```
TylerHelmuth, line 21: I'd prefer not to add complexity in the template to check k8s version. A user on a version of Kubernetes that doesn't support `persistentVolumeClaimRetentionPolicy` should not use this feature.  Can you add a comment in the values.yaml that specifies the first k8s version the feature is supported in?

lindeskar: Ok that makes sense.

I have removed the semverCompare and added a comment in values.yaml now.

TylerHelmuth (approved):

PR #1062, opened 2024-03-04
Bump operator to 0.95.0

swiatekm-sumo:
Part of https://github.com/open-telemetry/opentelemetry-operator/issues/2656

I had to add new permissions to the operator ClusterRole due to https://github.com/open-telemetry/opentelemetry-operator/pull/2575. I ended up copying the rules from the operator's kustomize manifest, resulting in a smaller ClusterRole at the end of it.

### DIFF
--- a/.github/workflows/operator-test.yaml
+++ b/.github/workflows/operator-test.yaml
@@ -46,7 +46,7 @@ jobs:
           git clone -b v"$appVersion" --single-branch https://github.com/open-telemetry/opentelemetry-operator.git ./opentelemetry-operator
 
       - name: Install chainsaw
-        uses: kyverno/action-install-chainsaw@v0.1.6
+        uses: kyverno/action-install-chainsaw@v0.1.7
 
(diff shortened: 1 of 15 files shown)


Allex1:
@swiatekm-sumo I think the cluster role needs to be identical to https://github.com/open-telemetry/opentelemetry-operator/blob/main/config/rbac/role.yaml. Also we had some tests for the roles.

swiatekm-sumo:
> @swiatekm-sumo I think the cluster role needs to be identical to https://github.com/open-telemetry/opentelemetry-operator/blob/main/config/rbac/role.yaml. Also we had some tests for the roles.

It *is* identical, just formatted differently. I copied it from the operator, and then chart tooling reformatted it.

swiatekm-sumo:
@Allex1 can we merge this? Operator 0.96.0 is going to be out very soon.

TylerHelmuth (approved):

PR #1065, opened 2024-03-05
[operator] Reject invalid/unknown properties using `additionalProperties`

heyLu:
This can catch misspellings like the one below, which I encountered when
using this Helm chart myself (but the previous validation did not catch it):

    opentelemetry-operator:
    - (root): Additional property kubeRBACPRoxy is not allowed

I was not entirely sure about the version, so I bumped the minor version
as this change may reject charts that worked previously (if they contain
unknown or unspecified fields).

The three other charts already use `additionalProperties: false` in many
places, this adds similar validation to the `opentelemetry-operator` chart
as well: https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-helm-charts+additionalProperties&type=code

### DIFF
--- a/charts/opentelemetry-operator/Chart.yaml
+++ b/charts/opentelemetry-operator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.49.1
+version: 0.50.0
 description: OpenTelemetry Operator Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-operator/UPGRADING.md
+++ b/charts/opentelemetry-operator/UPGRADING.md
@@ -1,5 +1,10 @@
 # Upgrade guidelines
 
+## <0.50.0 to 0.50.0
+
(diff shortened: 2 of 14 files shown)


Allex1:
@heyLu Thanks for your contribution. Can you please bump the chart and run `make generate-examples CHARTS=opentelemetry-operator` to have the CI pass?

heyLu:
> @heyLu Thanks for your contribution. Can you please bump the chart and run `make generate-examples CHARTS=opentelemetry-operator` to have the CI pass?

Done!  Bumped to `0.50.0` as this can still break existing charts with invalid values as mentioned in the description.

heyLu:
@Allex1, could you have another look at this?  I rebased this again to merge cleanly onto `main`.  `ct lint` passes locally, so I think this should be mergeable as-is.

TylerHelmuth (approved):
@dmitryax please take a look

TylerHelmuth:
@Allex1 please take another look

Allex1 (approved):

PR #1096, opened 2024-03-25
[opentelemetry-operator] fix `fullnameOverride`

### LINKED ISSUE #1095: [opentelemetry-operator] values schema does not allow `fullnameOverride`
If you want to use the `fullnameOverride` property, it fails to render the chart, as the `values.schema.json` does not allow to set it.
The property is however referenced in the `templates/_helpers.tpl` (the default when creating a new helm chart skeleton) which then is used on resources (f.e. the `templates/deployment.yaml`):

https://github.com/open-telemetry/opentelemetry-helm-charts/blob/30763a710227621137966281d2a17b55521083d5/charts/opentelemetry-operator/templates/_helpers.tpl#L13-L15

https://github.com/open-telemetry/opentelemetry-helm-charts/blob/30763a710227621137966281d2a17b55521083d5/charts/opentelemetry-operator/templates/deployment.yaml#L7

Error message will show up like this:
```
opentelemetry-operator:
- (root): Additional property fullnameOverride is not allowed
```

I will create a PR for this.

skuethe:
This fixes #1095 

I also added the property to the `values.yaml` to make it more clear that this is available.

```bash
$ helm template my-otel-operator . --set fullnameOverride="fullnameoverride-otel-operator" -s templates/deployment.yaml| yq '.metadata.name'
fullnameoverride-otel-operator
```

### DIFF
--- a/charts/opentelemetry-operator/Chart.yaml
+++ b/charts/opentelemetry-operator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.52.0
+version: 0.52.1
 description: OpenTelemetry Operator Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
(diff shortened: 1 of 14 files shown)


skuethe:
Done @TylerHelmuth

PR #1098, opened 2024-03-26
Allow `WATCH_NAMESPACE` as `manager.env`

janario:
This will allow to add extra `WATCH_NAMESPACE` env var to the manager, since it is a known env var https://github.com/open-telemetry/opentelemetry-operator/blob/main/main.go#L225C41-L225C56

### DIFF
--- a/charts/opentelemetry-operator/values.schema.json
+++ b/charts/opentelemetry-operator/values.schema.json
@@ -621,6 +621,14 @@
                             "examples": [
                                 "true"
                             ]
+                        },
+                        "WATCH_NAMESPACE": {
+                          "type": "string",
+                          "default": "",
+                          "title": "The WATCH_NAMESPACE namespaces",
+                          "examples": [
+                            "my-namespace-1,my-namespace-2"
+                          ]
                         }
                     },
                     "examples": [{
--- a/charts/opentelemetry-operator/values.schema.json
+++ b/charts/opentelemetry-operator/values.schema.json
@@ -609,22 +609,6 @@
                     "type": "object",
                     "default": {},
                     "title": "The env Schema",
-                    "required": [
-                        "ENABLE_WEBHOOKS"
-                    ],
-                    "additionalProperties": {
-                      "type": "string"
-                    },
-                    "properties": {
-                        "ENABLE_WEBHOOKS": {
-                            "type": "string",
-                            "default": "",
-                            "title": "The ENABLE_WEBHOOKS Schema",
-                            "examples": [
-                                "true"
-                            ]
-                        }
-                    },
                     "examples": [{
                         "ENABLE_WEBHOOKS": "true"
                     }]
--- a/charts/opentelemetry-operator/values.schema.json
+++ b/charts/opentelemetry-operator/values.schema.json
@@ -612,7 +612,9 @@
                     "required": [
                         "ENABLE_WEBHOOKS"
                     ],
-                    "additionalProperties": false,
+                    "additionalProperties": {
+                      "type": "string"
+                    },
                     "properties": {
                         "ENABLE_WEBHOOKS": {
                             "type": "string",
@@ -621,14 +623,6 @@
                             "examples": [
                                 "true"
                             ]
-                        },
-                        "WATCH_NAMESPACE": {
-                          "type": "string",
-                          "default": "",
-                          "title": "The WATCH_NAMESPACE namespaces",
-                          "examples": [
-                            "my-namespace-1,my-namespace-2"
-                          ]
                         }
                     },
                     "examples": [{
--- a/charts/opentelemetry-operator/values.schema.json
+++ b/charts/opentelemetry-operator/values.schema.json
@@ -609,6 +609,9 @@
                     "type": "object",
                     "default": {},
                     "title": "The env Schema",
+                    "additionalProperties": {
+                      "type": "string"
+                    },
                     "examples": [{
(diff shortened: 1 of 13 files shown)


charts/opentelemetry-operator/values.schema.json
```diff
@@ -621,6 +621,14 @@
                             "examples": [
                                 "true"
                             ]
```
janario, line 623: Another more flexible option is to allow any with string type
``` 
"additionalProperties": {
                      "type": "string"
                    },
```

that would block me to add wrong values:
```
- manager.env.CUSTOM_ENV: Invalid type. Expected: string, given: integer

- manager.env.CUSTOM_ENV: Invalid type. Expected: string, given: object
 
```

TylerHelmuth: Ya this section shouldnt be restricted at all, you should be able to put whatever env vars you want. The schema shouldn't have any required values and should allow additional properties

janario: changed 👍 

I still kept the type since the `deployment.yaml` expects it to be a key/value

charts/opentelemetry-operator/values.schema.json
```diff
@@ -612,7 +612,9 @@
                     "required": [
                         "ENABLE_WEBHOOKS"
                     ],
-                    "additionalProperties": false,
+                    "additionalProperties": {
+                      "type": "string"
+                    },
```
TylerHelmuth, line 617: Lets not even restrict it like this.  I think the entiure section should be:

```json

                "env": {
                    "type": "object",
                    "examples": [{
                        "ENABLE_WEBHOOKS": "true"
                    }]
                },
```

janario: I've changed it as you shared, but let me know your opinion on the case I mentioned above :-)

janario: I can already make the deployment.yaml env dynamic
```
          {{- if .Values.manager.env }}
          env:
            {{- range $name, $value := .Values.manager.env }}
            - name: {{ $name }}
              {{- if eq (kindOf $value) "map" }}
              {{- tpl ($value | toYaml) $ | nindent 14 }}
              {{- else }}
              value: {{ tpl $value $ | quote -}}
              {{- end }}
            {{- end }}
          {{- end }}
```

```
manager:
  env:
    WATCH_NAMESPACE:
      valueFrom:
        fieldRef:
          fieldPath: metadata.namespace

```

helm template -s templates/deployment.yaml -f values-test.yaml ./ | grep -A 10 env
```
          env:
            - name: ENABLE_WEBHOOKS
              value: "true"
            - name: WATCH_NAMESPACE
              valueFrom:
                fieldRef:
                  fieldPath: metadata.namespace
          image: "ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator:0.96.0"
          name: manager
```

TylerHelmuth: You're right, this field isn't following the same pattern as the collector chart. The schema should restrict it to any number of `key: value` pairs where `key` and `value` are strings

janario: done. that should do it

TylerHelmuth:
@janario can you confirm via local testing that the `manager.env` is usable after this change?

janario:
> @janario can you confirm via local testing that the `manager.env` is usable after this change?

I tested it with the template.

--- 
no custom values:
```
# helm template -s templates/deployment.yaml ./ | grep -B 3 -A 10 env

          env:
            - name: ENABLE_WEBHOOKS
              value: "true"
```

Allowed values:
```
# values-test.yaml
manager:
  env:
    DQUOTE: "my-ns"
    DQUOTE2: "my-ns,my-ns2"
    SQUOTE: 'my-ns'
    SQUOTE2: 'my-ns,my-ns2'
    TEXT: my-ns
    TEXT2: my-ns,my-ns2
    WNUMBER: "1"
    WNUMBER2: '1'
    WBOOL: "true"
    WBOOL2: 'true'
---
# helm template -s templates/deployment.yaml -f values-test.yaml ./ | grep -B 3 -A 25 env
          env:
            - name: DQUOTE
              value: "my-ns"
            - name: DQUOTE2
              value: "my-ns,my-ns2"
            - name: ENABLE_WEBHOOKS
              value: "true"
            - name: SQUOTE
              value: "my-ns"
            - name: SQUOTE2
              value: "my-ns,my-ns2"
            - name: TEXT
              value: "my-ns"
            - name: TEXT2
              value: "my-ns,my-ns2"
            - name: WBOOL
              value: "true"
            - name: WBOOL2
              value: "true"
            - name: WNUMBER
              value: "1"
            - name: WNUMBER2
              value: "1"
```

And invalid ones

```
manager:
  env:
    OBJ: {}
    ARR: []
    NUMBER: 1
    BOOL: true
---
# helm template -s templates/deployment.yaml -f values-test.yaml ./ 
opentelemetry-operator:
- manager.env.ARR: Invalid type. Expected: string, given: array
- manager.env.BOOL: Invalid type. Expected: string, given: boolean
- manager.env.NUMBER: Invalid type. Expected: string, given: integer
- manager.env.OBJ: Invalid type. Expected: string, given: object


```

---

Some other specific case that you would like to test? 🤔

TylerHelmuth (approved):

janario:
Thanks a lot for the quick review and support with it.

We will soon upgrade our chart usage ;-)

PR #1100, opened 2024-03-27
Consistent definition of securityContext schema values

patst:
While updating the opentelemetry-operator helm chart from 0.49.1 -> 0.52.2 we get an error during `helm template`:

```bash
opentelemetry-operator:
- securityContext: Additional property seccompProfile is not allowed
```

Our security context is defined like that:

```yaml
securityContext:
  runAsGroup: 65532
  runAsNonRoot: true
  runAsUser: 65532
  fsGroup: 65532
  seccompProfile:
    type: RuntimeDefault
```
Everything worked up until now. I think the 'issue' was introduced with https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1065.

The securitycontext in the `charts/opentelemetry-operator/values.schema.json` file is different.
After this PR it is the same for all occurences.

An alternative would be to add the missing attributes to the schema.

### DIFF
--- a/charts/opentelemetry-operator/Chart.yaml
+++ b/charts/opentelemetry-operator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.52.2
+version: 0.52.3
 description: OpenTelemetry Operator Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-operator/Chart.yaml
+++ b/charts/opentelemetry-operator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.52.1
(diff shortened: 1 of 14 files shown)


TylerHelmuth (commented):
Please bump the chart's patch version and then run `make generate-examples CHARTS=opentelemetry-operator`

patst:
> Please bump the chart's patch version and then run `make generate-examples CHARTS=opentelemetry-operator`

@TylerHelmuth  thanks for the fast reply. 

Bumbed the chart version and regenerated the examples

TylerHelmuth (approved):

TylerHelmuth:
@patst I merged another operator PR chart first, you'll need to bump and regenerate again

patst:
> @patst I merged another operator PR chart first, you'll need to bump and regenerate again

@TylerHelmuth merged and increased chart version to 0.52.4

PR #1102, opened 2024-03-27
Supplies the example base config needed for core functionality

jaronoff97:
This PR creates the scaffolding needed to create collectors for the chart. The structure of this allows a user to supply a single group of collectors (under the `collectors` key) and override them accordingly. Each collector is layered on top of the default collector configuration (`defaultCRConfig`) and then templated. This also allows users to supply a file for a prometheus scrape configuration as a convenience. For validation, a values schema json has been provided. This PR also creates the necessary cluster role and cluster role bindings for each collector accordingly, including some convenience for the k8s_cluster receiver.


charts/opentelemetry-kube-stack/templates/collector.yaml
```diff
@@ -0,0 +1,241 @@
...
+    valueFrom:
+      fieldRef:
+        apiVersion: v1
+        fieldPath: metadata.namespace
+  - name: OTEL_K8S_POD_NAME
+    valueFrom:
+      fieldRef:
+        apiVersion: v1
+        fieldPath: metadata.name
+  - name: OTEL_K8S_POD_UID
```
Allex1, line 175: Do we want both the pod uid and name? Thinking it might increase cardinality .
Also, do we really need these env vars as long as we're using the resourcedetection processor and OTEL_RESOURCE_ATTRIBUTES which adds the k8s attributes from the app pod? I'm not sure if that works only for sidecar mode though.

Allex1 (approved):
@jaronoff97 left a comment but this is pure gold
Thanks

charts/opentelemetry-kube-stack/templates/collector.yaml
jaronoff97: we can probs get rid of the UID. I think we def need the name for self telemetry purposes.

charts/opentelemetry-kube-stack/values.yaml
```diff
@@ -0,0 +1,335 @@
...
+  #     matchLabels:
+  #       app: my-app
+
+  # Config maps for the collector
+  configmaps: []
+  # - name: config
+  #   mountPath: /etc/config
+
+# Top level field specifying collectors configuration
+collectors: {}
```
TylerHelmuth, line 324: Let's add a comment that helps users understand how they configure this field. [We do this in the demo chart](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/0d569572313bbbde68a09db945e9bbbfde16a582/charts/opentelemetry-demo/values.yaml#L46).  Also, if this is a list, the default should be `[]` I believe.

charts/opentelemetry-kube-stack/templates/collector.yaml
```diff
@@ -0,0 +1,241 @@
...
+    {{- $prom_override = (dict "receivers" (dict "prometheus" (dict "config" (dict "scrape_configs" $merged_prom_scrape_configs)))) }}
+    {{- end }}
+    # merge the existing configuration with our override
+    {{- $new_config := (mergeOverwrite $collector.config $prom_override)}}
+    # write the config as yaml
+    {{- toYaml $new_config | nindent 4 }}
+    {{- else }}
+    # otherwise just template whats there
+    {{- toYaml $collector.config | nindent 4 }}
+    {{- end }}
```
TylerHelmuth, line 49: There is a lot of logic here, lets move some stuff (or all) into templates.

charts/opentelemetry-kube-stack/templates/collector.yaml
```diff
@@ -0,0 +1,241 @@
...
+  {{- if $collector.replicas }}
```
TylerHelmuth, line 50: To reduce the number of if statements, lets set some default values in the `values.yaml` using either kubernetes defaults (like `hostNetwork: false`), and then require the value be set via the json schema.

charts/opentelemetry-kube-stack/values.yaml
jaronoff97: i prefer a map here because it allows for more complex and specific overrides as helm doesn't allow for overriding or modifying array contacts. happy to explain more why this is done

TylerHelmuth: Is it a map where each key is a new collector? Let's make sure the values.yaml makes it easy to understand how to configure this section properly.

charts/opentelemetry-kube-stack/templates/collector.yaml
jaronoff97: actually, the with covers most of these, i can probably get rid of the majority of the ifs

charts/opentelemetry-kube-stack/values.yaml
jaronoff97: yes, that's the idea. can do!

jaronoff97: the values.schema.json should also be type checking the contents for validity as well.

Allex1 (approved):

TylerHelmuth (approved):

PR #1103, opened 2024-03-28
Add support of `ephemeral-storage` resource definition for opentelemetry-collector

skyzy06:
Following the discussion on this issue : [Impossibility to set up `ephemeral-storage` value in the opentemetry-collector values.yaml](https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1104)
As additional values are not allowed, adding explicitly the support of `ephemeral-storage` value in the resources.

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.86.0
+version: 0.86.1
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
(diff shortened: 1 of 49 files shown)


TylerHelmuth:
@skyzy06 please open an issue for use to discuss this new feature and the problem it solves.

charts/opentelemetry-collector/values.yaml
```diff
@@ -287,6 +287,7 @@ resources: {}
 # resources:
 #   limits:
 #     cpu: 250m
+#     ephemeral-storage: 50Mi
```
TylerHelmuth, line 290: Let's not include this comment as most users won't use this.

TylerHelmuth (commented):

TylerHelmuth (approved):

PR #1105, opened 2024-03-28
[collector] Make it possible to mount configMap by chart name

chunter0:
This change adds `configMap.mount` field which can be used to control mounting the configMap by chart name. This new field is helpful when using opentelemetry-collector as a subchart it will allow users the ability to create their own configuration and still have the configMap mounted under the default schema.


TylerHelmuth:
@chunter0 it isn't clear to me how this field is different than `create`.  Can you open an issue to discuss the problem this PR solves?

chunter0:
Opened an issue [here](https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1124) for discussion

charts/opentelemetry-collector/values.yaml
```diff
@@ -80,6 +80,8 @@ presets:
 configMap:
   # Specifies whether a configMap should be created (true by default)
   create: true
+  # Specifies whether the configMap should be mounted by chart name (false by default)
```
TylerHelmuth, line 83: Can you update this comment to more explicitly list out what this setting does

chunter0: Sure do you think the name should be updated as well something like `mountOverride`? @TylerHelmuth

charts/opentelemetry-collector/values.yaml
```diff
@@ -80,6 +80,8 @@ presets:
 configMap:
   # Specifies whether a configMap should be created (true by default)
   create: true
+  # Specifies whether the configMap should be mounted by chart name (false by default) 
+  mount: false
```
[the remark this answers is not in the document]
TylerHelmuth: Actually this will break users who are setting `create` to false and expecting nothing to mount :/ So `create` does need to stay in the `if` statement or we have to treat this as a breaking change.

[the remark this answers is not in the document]
chunter0: Okay keeping it the same

charts/opentelemetry-collector/values.yaml
TylerHelmuth: Let's name it `includeMount`

TylerHelmuth:
@open-telemetry/helm-approvers please review

charts/opentelemetry-collector/values.yaml
rnishtala-sumo: curious, why isn't creating a configMap a sufficient condition for it to be mounted in the deployment/daemonset/statefulset?

chunter0: Mainly for the use case when a user is bringing their own config

jaronoff97:
I commented on [the issue](https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1124#issuecomment-2067047173)

charts/opentelemetry-collector/templates/_pod.tpl
```diff
@@ -148,10 +148,10 @@ initContainers:
 priorityClassName: {{ .Values.priorityClassName | quote }}
 {{- end }}
 volumes:
-  {{- if .Values.configMap.create }}
+  {{- if or .Values.configMap.create .Values.configMap.existingName }}
   - name: {{ include "opentelemetry-collector.lowercase_chartname" . }}-configmap
     configMap:
-      name: {{ include "opentelemetry-collector.fullname" . }}{{ .configmapSuffix }}
+      name: {{ .Values.configMap.existingName | default (printf "%s%s" (include "opentelemetry-collector.fullname" .) (.configmapSuffix)) }}
```
TylerHelmuth, line 154: Lets move this logic into a template so that we can do a simple if/else

charts/opentelemetry-collector/templates/_pod.tpl
```diff
@@ -17,7 +17,7 @@ containers:
       - /{{ .Values.command.name }}
     {{- end }}
     args:
-      {{- if .Values.configMap.create }}
+      {{- if or .Values.configMap.create .Values.configMap.existingName }}
       - --config=/conf/relay.yaml
```
TylerHelmuth, line 21: We need to make is clear in the values.yaml file that if you use `existingName` the confmap must contain relay.

charts/opentelemetry-collector/templates/_pod.tpl
chunter0: Updated c0fd7da296b427a4e2496495a8eb94fae19d82b3 @TylerHelmuth

charts/opentelemetry-collector/templates/_helpers.tpl
```diff
@@ -214,3 +214,15 @@ The capitalization is important for StatefulSet.
 {{- print "StatefulSet" -}}
 {{- end -}}
 {{- end }}
+
+{{/*
+Get ConfigMap name if existingName is defined, otherwise use default name for generated config.
+*/}}
+{{- define "opentelemetry-collector.configName" -}}
+  {{- if .Values.configMap.existingName -}}
+    {{ .Values.configMap.existingName }}
```
TylerHelmuth, line 223:
```suggestion
    {{- .Values.configMap.existingName }}
```

charts/opentelemetry-collector/values.yaml
```diff
@@ -80,6 +80,9 @@ presets:
 configMap:
   # Specifies whether a configMap should be created (true by default)
   create: true
+  # Specifies an existing configMap to be mounted to the pod which should contain an item with key 'relay' and path 'relay.yaml'. 
+  # These settings will be used to properly configure the collector.
```
TylerHelmuth, line 84:
```suggestion
  # Specifies an existing ConfigMap to be mounted to the pod
  # The ConfigMap MUST include the collector configuration via a key named 'relay' or the collector will not start.
```

charts/opentelemetry-collector/values.yaml
```diff
@@ -80,6 +80,9 @@ presets:
 configMap:
   # Specifies whether a configMap should be created (true by default)
   create: true
+  # Specifies an existing ConfigMap to be mounted to the pod
+  # The ConfigMap MUST include the collector configuration via a key named 'relay' or the collector will not start.
+  existingName: ""
```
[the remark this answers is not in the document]
chunter0: IMO `existingName` seems clearer but wondering what others think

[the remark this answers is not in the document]
jaronoff97: I like existingName as it implies that you do not want to create a new configmap but rather reference the existing one everywhere

charts/opentelemetry-collector/values.schema.json
```diff
@@ -131,7 +131,11 @@
         "create": {
           "description": "Specifies whether a configMap should be created (true by default).",
           "type": "boolean"
-        }
+        },
+        "existingName": {
+          "description": "Specifies an existing configMap to be mounted to the pod",
```
jaronoff97, line 136: is there ever a case when you want existingName set to a real value _and_ create true?

jaronoff97 (commented):
one question

charts/opentelemetry-collector/values.schema.json
chunter0: I don't think so. When _create_ is true the name of the ConfigMap is generated via a helper function, meaning it won't come from values.

chunter0: Thats handled here https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1105/files#diff-12c2446fb11e3b1d1120f6b547457ee4dc1420594649d1d6101060548e8a5afbR219

ExistingName field is only used when its not empty @jaronoff97

jaronoff97: yes, i know. I meant it would also be used [here](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/templates/configmap.yaml#L1) to prevent that potential footgun

chunter0: Oh something like this

```
{{- if and (eq .Values.mode "daemonset") (.Values.configMap.create) (not .Values.configMap.existingName) -}}
```

jaronoff97: yep! that way you don't accidentally try to create a configmap that already exists :) NOTE: there are three files where this could happen right now.

chunter0: Updated https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1105/commits/c75f3288566355bf6e6d072105462c6d23b4c5df

jaronoff97 (approved):
looks great :D thank you!

charts/opentelemetry-collector/values.schema.json
TylerHelmuth: This was a good discussion. Instead of doing the check in the templates, we should add a `fail` condition to `NOTES.txt` that fails the install if `create: true` and `existingName` is truthy.

chunter0: Might not need to do that part, this check mainly prevents generating unnecessary resources. Now the check is on the ConfigMap I don't think this case is possible from user perspective right?

TylerHelmuth: Adding the check to `NOTES.txt` prevents users from writing a misleading values.yaml.  It would outlaw

```
configMap:
  create: true
  existingName: "my-custom-config"
```

That values.yaml shouldn't be supported because it is unclear what happens.  The changes you made to the template results in `existingName` being ignored, which saves the user, but I think explicitly failing would be better.  We don't like allowing bad config to be passed to the chart.

chunter0: Makes sense what do you think about this message?

```
{{- if and .Values.configMap.create .Values.configMap.existingName }}
{{ fail "[ERROR] The opentelemetry-collector chart support one of configMap.create or configMap.existingName but not both" }}
{{ end }}
```

Example from helm template

```console
helm template charts/opentelemetry-collector \
  --set configMap.create=true \
  --set configMap.existingName="ok" \
  --set image.repository=gcr.io \
  --set mode=daemonset                                    
Error: execution error at (opentelemetry-collector/templates/NOTES.txt:32:3): [ERROR] The opentelemetry-collector chart support one of configMap.create or configMap.existingName but not both

Use --debug flag to render out invalid YAML
```

chunter0: Updated 1bca1b4eb5b69c6feaceb8eb7043d91679ac7244 @TylerHelmuth

TylerHelmuth (approved):

PR #1106, opened 2024-03-28
[collector] Allow more flexibility around adding volumes/volumeMounts

chunter0:
The changes included here add more flexibilitly to opentelemetry-collector deployments when being consumed as a subchart. Having both `volumes/volumeMounts` and `extraVolumes/extraVolumeMounts` allows the owner of the subchart a space to add pre-defined volumes/volumeMounts. Giving the consumer the freedom to add more volumes/volumeMounts under extraVolumes/extraVolumeMounts fields.

### DIFF
--- a/charts/opentelemetry-collector/values.yaml
+++ b/charts/opentelemetry-collector/values.yaml
@@ -230,7 +230,9 @@ priorityClassName: ""
 
 extraEnvs: []
 extraEnvsFrom: []
+volumes: []
 extraVolumes: []
+volumeMounts: []
 extraVolumeMounts: []
 
 # Configuration for ports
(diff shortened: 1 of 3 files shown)


TylerHelmuth:
@chunter0 please on an issue where we can discuss in depth the problem you're experiencing and how this PR solves it.

charts/opentelemetry-collector/values.yaml
```diff
@@ -230,7 +230,9 @@ priorityClassName: ""
 
 extraEnvs: []
 extraEnvsFrom: []
+volumes: []
```
JaredTan95, line 233: Why do we need repeated `volumes` when we already have `extraVolumes`?

charts/opentelemetry-collector/values.yaml
```diff
@@ -230,7 +230,9 @@ priorityClassName: ""
...
 extraVolumes: []
+volumeMounts: []
```
JaredTan95, line 235: Same as above

chunter0:
I opened https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1168 for discussion

CC: @TylerHelmuth @JaredTan95 @povilasv

PR #1107, opened 2024-03-28
[collector] Allow more flexibility around adding environment variables

chunter0:
The changes included here add more flexibilitly to opentelemetry-collector deployments when being consumed as a subchart. Having `envs` field allows the owner of the subchart a space to add pre-defined environment variables. Giving the consumer the freedom to add more under `extraEnvs` fields.


chunter0:
I opened https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1168 for discussion

CC: @povilasv @JaredTan95

Issue #1108, opened 2024-03-28
[FileLogReceiver] `.Values.namespaceOverride` instead of `.Release.Namespace` should be used to "exclude" Otel-collector log


yingsamsung:
https://github.com/open-telemetry/opentelemetry-helm-charts/blob/0d569572313bbbde68a09db945e9bbbfde16a582/charts/opentelemetry-collector/templates/_config.tpl#L196

Problem:
When `overrideNamespace` is different from `Release.Namespace`, Otel-collector log is not excluded properly.

Repro:
1. install OpenTelemetry-Collector chart with `namespaceOverride: monitoring`, without specify `--namespace` on `helm install`. Note that, open-telemetry-agent pod is installed in `monitoring` namespace in K8S cluster. 
2. check logs under `/var/log/pods`. The folder of collector log is actually `monitoring_opentelemetry-collector-agent-....`, while in ConfigMap,

```
Data
====
relay:
----
receivers:
  filelog:
    exclude:
    - /var/log/pods/default_obv-opentelemetry-collector*_*/opentelemetry-collector/*.log
```
Note that, the `default` release namespace is used, which causes log from Otel collector is still being collected.

TylerHelmuth:
@yingsamsung nice find, would you like to submit a PR?

PR #1109, opened 2024-04-02
[collector] fix otel col's log was not excluded correctly by filelogreceiver

JaredTan95:
fix https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1108


charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -193,7 +193,12 @@ receivers:
     exclude: []
     {{- else }}
     # Exclude collector container's logs. The file format is /var/log/pods/<namespace_name>_<pod_name>_<pod_uid>/<container_name>/<run_id>.log
-    exclude: [ /var/log/pods/{{ .Release.Namespace }}_{{ include "opentelemetry-collector.fullname" . }}*_*/{{ include "opentelemetry-collector.lowercase_chartname" . }}/*.log ]
+    exclude:
+    {{- if .Values.namespaceOverride }}
+    - /var/log/pods/{{ .Values.namespaceOverride }}_{{ include "opentelemetry-collector.fullname" . }}*_*/{{ include "opentelemetry-collector.lowercase_chartname" . }}/*.log
+    {{- else }}
+    - /var/log/pods/{{ .Release.Namespace }}_{{ include "opentelemetry-collector.fullname" . }}*_*/{{ include "opentelemetry-collector.lowercase_chartname" . }}/*.log
+    {{- end }}
```
TylerHelmuth, line 201: instead of an if-else in this template, lets make a new template in `_helpers` that returns `.Values.namespaceOverride` or `.Release.Namespace`.  Then we can do 

```
exclude: [ /var/log/pods/{{ include "opentelemetry-collector.new-template-name }}_{{ include "opentelemetry-collector.fullname" . }}*_*/{{ include "opentelemetry-collector.lowercase_chartname" . }}/*.log ]
```
This keeps the pattern for building the string consistent across all its variable secitons.

charts/opentelemetry-collector/templates/_helpers.tpl
```diff
@@ -163,6 +163,10 @@ Allow the release namespace to be overridden
   {{- end -}}
 {{- end -}}
 
+{{- define "opentelemetry-collector.retrieveNamespace" -}}
```
TylerHelmuth, line 166: I see we already have `opentelemetry-collector.namespace`, can we use that instead?

JaredTan95: I've noticed it before, but I haven't tested it, I'll try to use the existing ones later

JaredTan95: use `opentelemetry-collector.namespace` works as expected.

TylerHelmuth (approved):

PR #1113, opened 2024-04-03
[opentelemetry-operator] bump to 0.97.0

jaronoff97:
completes operator release for 0.97.0

### DIFF
--- a/charts/opentelemetry-operator/values.yaml
+++ b/charts/opentelemetry-operator/values.yaml
@@ -42,7 +42,7 @@ manager:
     tag: ""
   collectorImage:
     repository: otel/opentelemetry-collector-contrib
-    tag: 0.96.0
+    tag: 0.97.0
   opampBridgeImage:
     repository: ""
     tag: ""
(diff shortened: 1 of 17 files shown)


Allex1 (approved):

charts/opentelemetry-operator/values.yaml
```diff
@@ -160,7 +159,8 @@ manager:
 
   ## Container specific securityContext
   ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container
-  securityContext: {}
+  securityContext:
+    {}
```
TylerHelmuth, line 163: I assume this is a linter, but I'm not sure if it is equivalent to `securityContext: {}`.  Can we convert these back?

jaronoff97: yeah this is a linter and is equivalent, i can swap it back...

jaronoff97: i think its doing it because the comment is indented down. I'll fix that too

jaronoff97: similar to this above:
```
    minAllowed: {}
    # cpu: 200m
    # memory: 100Mi
```

PR #1114, opened 2024-04-03
Add instrumentation to chart

### LINKED ISSUE #1078: [opentelemetry-kube-stack] Enable autoinstrumentation creation
### original issue
#562

### context
https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1058

### description
Make it possible to create an Instrumentation CR from the chart given an instrumentation spec. This should also handle the convenience of connecting an enabled collector.

jaronoff97:
Closes #1078 by providing a method for creating a single instrumentation resource

### DIFF
--- a/charts/opentelemetry-kube-stack/templates/instrumentation.yaml
+++ b/charts/opentelemetry-kube-stack/templates/instrumentation.yaml
@@ -1,0 +1,68 @@
+{{- if .Values.instrumentation.enabled }}
+---
+apiVersion: opentelemetry.io/v1alpha1
+kind: Instrumentation
+metadata:
+  name: {{ include "opentelemetry-collector.instrumentation" . }}
+  labels:
+    {{- include "opentelemetry-kube-stack.labels" $ | nindent 4 }}
+    {{- with .Values.instrumentation.labels }}
+    {{- range $key, $value := . }}
+    {{- printf "%s: %s" $key (tpl $value $ | quote) | nindent 4 }}
+    {{- end }}
+    {{- end }}
+  {{- with .Values.instrumentation.annotations }}
+  annotations:
+    {{- range $key, $value := . }}
+    {{- printf "%s: %s" $key (tpl $value $ | quote) | nindent 4 }}
+    {{- end }}
+  {{- end }}
+spec:
+  exporter:
+    endpoint: {{ .Values.instrumentation.exporter.endpoint }}
+  propagators:
+    {{- toYaml .Values.instrumentation.propagators | nindent 4 }}
+  {{- with .Values.instrumentation.sampler }}
+  sampler:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  env:
+    {{- with $.Values.extraEnvs }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+    {{- with .Values.instrumentation.env }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+  {{- with .Values.instrumentation.resource }}
+  resource:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  {{- with .Values.instrumentation.java }}
+  java:
+    {{- toYaml . | nindent 4}}
+  {{- end }}
+  {{- with .Values.instrumentation.nodejs }}
+  nodejs:
+    {{- toYaml . | nindent 4}}
+  {{- end }}
+  {{- with .Values.instrumentation.python }}
+  python:
+    {{- toYaml . | nindent 4}}
+  {{- end }}
+  {{- with .Values.instrumentation.dotnet }}
+  dotnet:
+    {{- toYaml . | nindent 4}}
+  {{- end }}
+  {{- with .Values.instrumentation.go }}
+  go:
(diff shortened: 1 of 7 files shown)


charts/opentelemetry-kube-stack/values.yaml
```diff
@@ -341,3 +341,172 @@ clusterRole:
...
+# Instrumentation configuration
+instrumentation:
+  # Whether instrumentation is enabled or not
+  enabled: false
+  labels: {}
+  annotations: {}
+
+  # Exporter configuration
+  exporter:
+    endpoint: http://collector:55678
```
TylerHelmuth, line 354: This should match the default name of the collector which I think is `demo`.  Let's change it to `collector`.

charts/opentelemetry-kube-stack/values.yaml
```diff
@@ -341,3 +341,172 @@ clusterRole:
...
+  # Propagators configuration
+  propagators:
+    - tracecontext
+    - baggage
+    - b3
+    - b3multi
+    - jaeger
+    - xray
+    - ottrace
+    - none
```
TylerHelmuth, line 371: Should `none` be a default propagator?

charts/opentelemetry-kube-stack/templates/instrumentation.yaml
```diff
@@ -0,0 +1,68 @@
...
+kind: Instrumentation
+metadata:
+  name: {{ include "opentelemetry-collector.instrumentation" . }}
+  labels:
+    {{- include "opentelemetry-kube-stack.labels" $ | nindent 4 }}
+    {{- with .Values.instrumentation.labels }}
+    {{- range $key, $value := . }}
+    {{- printf "%s: %s" $key (tpl $value $ | quote) | nindent 4 }}
+    {{- end }}
+    {{- end }}
```
TylerHelmuth, line 13: I'd prefer we move this into another template.

charts/opentelemetry-kube-stack/templates/instrumentation.yaml
```diff
@@ -0,0 +1,68 @@
...
+  sampler:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  env:
+    {{- with $.Values.extraEnvs }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+    {{- with .Values.instrumentation.env }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
```
TylerHelmuth, line 35: I'd prefer we move this into another template.

charts/opentelemetry-kube-stack/values.yaml
```diff
@@ -341,3 +341,173 @@ clusterRole:
...
+    - tracecontext
+    - baggage
+    - b3
+    - b3multi
+    - jaeger
+    - xray
+    - ottrace
+
+  # Sampler configuration
+  sampler:
```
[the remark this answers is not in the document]
jaronoff97: that's fair! I can disable it for now like i do the rest.

charts/opentelemetry-kube-stack/values.yaml
```diff
@@ -341,3 +341,173 @@ clusterRole:
...
+
+  # Environment variables for instrumentation
+  env: []
+  # - name: ENV_VAR1
+  #   value: value1
+  # - name: ENV_VAR2
+  #   value: value2
+
+  # Java agent configuration
+  java: {}
```
jvoravong, line 387: A chart project I work on ([link](https://github.com/signalfx/splunk-otel-collector-chart/blob/eb8a3285a8d8707de242509baafb7f372643336e/helm-charts/splunk-otel-collector/values.yaml#L1197)) that integrates with the  collector, operator, and opentelemetry.io/v1alpha1 instrumentation as part of our solution specifically uses pinned tag values that are not "latest". 
  - We found it important to explicitly set and document the versions of instrumentation libraries used. This aids in customer understanding and records the tested versions among components.
  - The operator project maintains a list of instrumentation library versions ([versions.txt](https://github.com/open-telemetry/opentelemetry-operator/blob/main/versions.txt)) that tracks values per operator release, this information can be useful to know what versions were tested with the involved components.
  - While users have the option to specify their instrumentation versions, I suggest this chart should define the specific default versions used for instrumentations libraries to avoid breaking changes in environments. I would suggest the default values match the versions of components that were tested to work together.

jvoravong: The current PR might be good enough as is for this topic. It might be worth creating a follow up issue for the work to ensure the used instrumentation version values are in sync with the version of the operator used. If a new ticket is created, I would like to volunteer to work on it.

jaronoff97: yeah! this makes a lot of sense. I would very much appreciate a hand with that as well. I can make the ticket.

jaronoff97: https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1129#issue-2233785704

jvoravong (approved):

Allex1 (approved):

TylerHelmuth (approved):

PR #1118, opened 2024-04-05
Updated image repository to download using ghcr.io instead of docker.io

AjayKumar4:
Updated to image repository to download using ghcr.io instead of docker.io

Due to Docker Hub Rate Limit
https://www.docker.com/increase-rate-limits/


TylerHelmuth:
@AjayKumar4 did you experience a rate limit? I believe the project pays to allow lots of pulls

PR #1123, opened 2024-04-05
fix: Add missing get permission on namespaces when using --create-rbac-permissions

thefirstofthe300:
After further use, I noticed I'm still missing the `get` verb from the `namespace` API group permissions when using the operator to provision RBAC for the k8s_attributes processor.

### DIFF
--- a/charts/opentelemetry-operator/templates/clusterrole.yaml
+++ b/charts/opentelemetry-operator/templates/clusterrole.yaml
@@ -36,6 +36,9 @@ rules:
     resources:
       - namespaces
     verbs:
+      {{- if .Values.manager.createRbacPermissions }}
+      - create
+      {{- end }}
       - list
       - watch
   - apiGroups:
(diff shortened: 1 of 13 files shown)


Allex1 (approved):

charts/opentelemetry-operator/templates/clusterrole.yaml
```diff
@@ -36,6 +36,9 @@ rules:
     resources:
       - namespaces
     verbs:
+      {{- if .Values.manager.createRbacPermissions }}
+      - create
+      {{- end }}
```
TylerHelmuth, line 41: @thefirstofthe300 this feels tangential to the `--create-rbac-permissions`.  I don't see where setting this flag in the operator provides this specific access. Do you have a link to the operator code/docs that is requiring this access?

TylerHelmuth:
@thefirstofthe300 please resolve conflicts

thefirstofthe300:
All the permissions errors appear to have been resolved.

PR #1126, opened 2024-04-08
updated helm chart values to include the host and port and the flagd …

bornav:
added helm chart update to include the changes in this PR https://github.com/open-telemetry/opentelemetry-demo/pull/1486

### DIFF
--- a/charts/opentelemetry-demo/Chart.yaml
+++ b/charts/opentelemetry-demo/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: opentelemetry-demo
-version: 0.30.0
+version: 0.30.1
 description: opentelemetry demo helm chart
 home: https://opentelemetry.io/
 sources:
--- a/charts/opentelemetry-demo/examples/bring-your-own-observability/rendered/component.yaml
+++ b/charts/opentelemetry-demo/examples/bring-your-own-observability/rendered/component.yaml
@@ -5,7 +5,7 @@ kind: Service
 metadata:
   name: example-adservice
   labels:
-    helm.sh/chart: opentelemetry-demo-0.30.0
+    helm.sh/chart: opentelemetry-demo-0.30.1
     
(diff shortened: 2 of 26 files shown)


julianocosta89 (approved):

puckpuck (approved):

bornav:
@puckpuck  Hi, have a question regarding the Lint Charts test that failed
Rrunning locally both the generate-examples and the check-examples with the flag CHARTS=opentelemetry-demo, both make command pass without any errors.
Is the error happening right now expected or should something be reworked?

TylerHelmuth:
@bornav make sure you have all the chart dependency repos added locally and then try `make generate-examples CHARTS=opentelemetry-demo` again.

charts/opentelemetry-demo/examples/bring-your-own-observability/rendered/opensearch/poddisruptionbudget.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 # Source: opentelemetry-demo/charts/opensearch/templates/poddisruptionbudget.yaml
-apiVersion: policy/v1
+apiVersion: policy/v1beta1
```
puckpuck, line 3: this change is peculiar. It seems like your local version of kubectl is for [K8s version 1.21 or earlier](https://github.com/opensearch-project/helm-charts/blob/main/charts/opensearch/templates/poddisruptionbudget.yaml#L2-L6)

Upgrading your local kubectl / helm should fix this, and allow for the linter to pass.

bornav: good catch, have addressed the problem, but what's weird is my kubectl version is GitVersion:"v1.27.11", have also tried with 1.29 but still the same
And was only able to build it without v1beta on a different host

TylerHelmuth (approved):

PR #1058, opened 2024-03-01
Add otel-cloud-stack chart

jaronoff97:
# Why make this?
This PR serves to finally close #562. This PR creates a helm chart that takes some inspiration from the [kube-prometheus-stack](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack#get-helm-repository-info) chart from the prometheus community repo to create a fully-fledged collector starter for Kubernetes. Prometheus services and CRDs are optionally included to allow users a near-drop-in replacement for the prometheus experience. For non-prometheus users, this provides an easy way to get started with opentelemetry's Kubernetes offerings. This is accomplished by utilizing some clever templating for the configuration for the collector, which I found to be easier to reason about than embedding config in a helper tpl.

# Features
* Centralized tracing collector
* Autoinstrumentation
* Infrastructure metrics and logs collector
* standalone kubernetes cluster stats collector
* Allows for drop-in replacement for kube-prometheus-stack with a sharded statefulset collector
* Allows for custom collectors to be set and validated against an exhaustive schema

# Next steps
* end to end testing to verify telemetry appears as expected between versions
* documentation for what to expect OOTB
* documentation for example configs (i include two samples below that are specific for my company)

<details>
<summary>For a config that gets you tracing, infra logs and metrics, kube cluster stats</summary>
<br>

```
clusterName: jacob-kind
opentelemetry-operator:
  enabled: true
collectors:
  tracing:
    env:
      - name: LS_TOKEN
        valueFrom:
          secretKeyRef:
            key: LS_TOKEN
            name: otel-collector-secret
    config:
      exporters:
        otlp:
          endpoint: ingest.lightstep.com:443
          headers:
            "lightstep-access-token": "${LS_TOKEN}"
      service:
        pipelines:
          traces:
            exporters: [otlp]
  daemon:
    env:
      - name: LS_TOKEN
        valueFrom:
          secretKeyRef:
            key: LS_TOKEN
            name: otel-collector-secret
    config:
      exporters:
        otlp:
          endpoint: ingest.lightstep.com:443
          headers:
            "lightstep-access-token": "${LS_TOKEN}"
      service:
        pipelines:
          metrics:
            exporters: [otlp]
          logs:
            exporters: [otlp]
  cluster:
    env:
      - name: LS_TOKEN
        valueFrom:
          secretKeyRef:
            key: LS_TOKEN
            name: otel-collector-secret
    config:
      exporters:
        otlp:
          endpoint: ingest.lightstep.com:443
[...]

Issue #985, opened 2023-12-21
Issue with Opamp after operator upgrade on 0.44.2 version


flenoir:
Hi, 

i'm trying to upgrade operator to version 0.44.2 of helm chart. 

I get some errors regarding opampbridge but couldn't find how to solve it. 

pod logs reports :

```
`{"level":"error","ts":"2023-12-21T13:20:30Z","logger":"controller-runtime.source.EventHandler","msg":"failed to get informer from cache","error":"failed to get API group resources: unable to retrieve the complete list of server APIs: autoscaling/v2: the server could not find the requested resource","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/source.(*Kind).Start.func1.1\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/source/kind.go:68\nk8s.io/apimachinery/pkg/util/wait.loopConditionUntilContext.func2\n\t/home/runner/go/pkg/mod/k8s.io/apimachinery@v0.28.4/pkg/util/wait/loop.go:73\nk8s.io/apimachinery/pkg/util/wait.loopConditionUntilContext\n\t/home/runner/go/pkg/mod/k8s.io/apimachinery@v0.28.4/pkg/util/wait/loop.go:74\nk8s.io/apimachinery/pkg/util/wait.PollUntilContextCancel\n\t/home/runner/go/pkg/mod/k8s.io/apimachinery@v0.28.4/pkg/util/wait/poll.go:33\nsigs.k8s.io/controller-runtime/pkg/internal/source.(*Kind).Start.func1\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/source/kind.go:56"}

`
[...]

PR #1138, opened 2024-04-12
Add Bridge to opentelemetry-kube-stack

### LINKED ISSUE #1079: [opentelemetry-kube-stack] Enable OpAMP Bridge creation
### original issue
#562

### context
https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1058

### description
Make it possible to create an Opamp bridge CR from the chart given an opamp bridge values. This should also handle the convenience of adding the labels to a enabled collectors.

jaronoff97:
Closes #1079 

This PR adds the ability to create an opamp bridge as part of the chart. This creates the bridge and its clusterroles/bindings as well as convenience for adding the necessary labels when applicable for the collector.

### DIFF
--- a/charts/opentelemetry-kube-stack/templates/bridge.yaml
+++ b/charts/opentelemetry-kube-stack/templates/bridge.yaml
@@ -1,0 +1,129 @@
+{{- if .Values.opAMPBridge.enabled }}
+{{- $merged := (dict "Chart" $.Chart "clusterRole" $.Values.clusterRole "opAMPBridge" $.Values.opAMPBridge "Release" $.Release "fullnameOverride" $.Values.fullnameOverride) }}
+---
+apiVersion: opentelemetry.io/v1alpha1
+kind: OpAMPBridge
+metadata:
+  name: {{ include "opentelemetry-opamp-bridge.fullname" $merged }}
+  labels:
+    {{- include "opentelemetry-kube-stack.labels" $ | nindent 4 }}
+    {{- include "opentelemetry-kube-stack.renderkv" .Values.opAMPBridge.labels | nindent 4 }}
+    {{- with .Values.opAMPBridge.annotations }}
+    annotations:
+      {{- include "opentelemetry-kube-stack.renderkv" . | nindent 4 }}
+    {{- end }}
+spec:
+  endpoint: {{ required "opamp endpoint required" $.Values.opAMPBridge.endpoint }}
+  {{- with $.Values.opAMPBridge.headers }}
+  headers:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  {{- with $.Values.opAMPBridge.capabilities }}
+  capabilities:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  {{- with $.Values.opAMPBridge.componentsAllowed }}
+  componentsAllowed:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  {{- with $.Values.opAMPBridge.nodeSelector }}
+  nodeSelector:
+  {{- toYaml . | nindent 4}}
+  {{- end }}
+  {{- if $.Values.opAMPBridge.replicas }}
+  replicas: {{ $.Values.opAMPBridge.replicas }}
+  {{- end }}
+  {{- if $.Values.opAMPBridge.serviceAccount }}
+  serviceAccount: {{ $.Values.opAMPBridge.serviceAccount }}
+  {{- end }}
+  {{- if $.Values.opAMPBridge.image.digest }}
+  image: "{{ $.Values.opAMPBridge.image.repository }}@{{ $.Values.opAMPBridge.image.digest }}"
+  {{- else }}
+  image: "{{ $.Values.opAMPBridge.image.repository }}:{{ $.Values.opAMPBridge.image.tag | default $.Chart.AppVersion }}"
+  {{- end }}
+  {{- if $.Values.opAMPBridge.upgradeStrategy }}
+  upgradeStrategy: {{ $.Values.opAMPBridge.upgradeStrategy }}
+  {{- end }}
+  {{- if $.Values.opAMPBridge.imagePullPolicy }}
+  imagePullPolicy: {{ $.Values.opAMPBridge.imagePullPolicy }}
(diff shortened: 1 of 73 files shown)


JaredTan95:
Seeing this PR, I thought of a digression. As we all know, helm won't apply the crd in the `chart crds` directory when upgrading an chart. When we upgrade the chart version with the `chart/crds` directory changes, there will always be breaking changes or some additional crd apply. (Related issue: https://github.com/open-telemetry/opentelemetry-helm-charts/issues/985, https://github.com/open-telemetry/opentelemetry-operator/issues/2314)

I see [this](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1114) and this PR both include crd content, will there be the same problem?

jaronoff97:
the way around this is to include crds in the templates directory, it's a hack but is kind of the only thing that can be done to resolve this effectively. You can see an example of me doing this to handle updates [here](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1058/files#diff-8886138c8d3273ca370ceef51e06d7e56cd4e847510d1494c214f6eaac9ec0f4) though idk if that's how we want to do all of this. Either way, we should probably discuss this not in this PR but rather in an issue or in slack

JaredTan95:
Yes, I actually know that there will be no problem with this PR, but I think it is a little strange, maybe it is the hack way you said. But overall for me, lgtm.

charts/opentelemetry-kube-stack/values.yaml
```diff
@@ -511,3 +511,157 @@ instrumentation:
   #   limits:
   #     memory: "128Mi"
   #     cpu: "500m"
+
+# Bridge cluster role configuration
+bridgeClusterRole:
```
JaredTan95, line 516: Should it be the same as `opAMPBridge.enabled`?

jaronoff97: ah like rename this to be opAMPBridgeClusterRole? Probably... good call!

charts/opentelemetry-kube-stack/values.yaml
```diff
@@ -511,3 +511,157 @@ instrumentation:
   #   limits:
   #     memory: "128Mi"
   #     cpu: "500m"
+
+# Bridge cluster role configuration
+opAMPBridgeClusterRole:
+  # Whether the bridge cluster role is enabled or not
+  enabled: true
```
TylerHelmuth, line 518: Should this be enabled by default?

charts/opentelemetry-kube-stack/values.yaml
```diff
@@ -511,3 +511,157 @@ instrumentation:
   #   limits:
   #     memory: "128Mi"
   #     cpu: "500m"
+
+# Bridge cluster role configuration
```
TylerHelmuth, line 515: Can you expand this comment.  I'm not very familiar with the bridge

charts/opentelemetry-kube-stack/values.yaml
```diff
@@ -511,3 +511,157 @@ instrumentation:
...
+opAMPBridgeClusterRole:
```
TylerHelmuth, line 516: Should this be nested under `opAMPBridge?

charts/opentelemetry-kube-stack/values.yaml
```diff
@@ -511,3 +511,157 @@ instrumentation:
...
+  # Whether the bridge cluster role is enabled or not
+  enabled: true
+
+  # Annotations for the bridge cluster role
+  annotations: {}
+
+  # Rules for the bridge cluster role
+  rules: []
+
+# OpAMP bridge configuration
```
TylerHelmuth, line 526: Can you update this comment to explain more about OpAMP?

charts/opentelemetry-kube-stack/values.yaml
```diff
@@ -511,3 +511,157 @@ instrumentation:
...
+    AcceptsOtherConnectionSettings: true
+    AcceptsRemoteConfig: true
+    AcceptsRestartCommand: true
+    ReportsEffectiveConfig: true
+    ReportsHealth: true
+    ReportsOwnLogs: true
+    ReportsOwnMetrics: true
+    ReportsOwnTraces: true
+    ReportsRemoteConfig: true
+    ReportsStatus: true
```
TylerHelmuth, line 556: Can you provide comments or a link to the docs for each of these capabilities?

charts/opentelemetry-kube-stack/values.yaml
```diff
@@ -511,3 +511,157 @@ instrumentation:
...
+  resources:
+    limits:
+      cpu: "250m"
+      memory: "256Mi"
+    requests:
+      cpu: "250m"
+      memory: "256Mi"
+
+  # Number of replicas for OpAMP bridge
+  replicas: 1
```
TylerHelmuth, line 579: Do we need to support more than 1 replicas?

charts/opentelemetry-kube-stack/values.yaml
jaronoff97: I'm going to move this under the bridge at which point it being enabled makes sense as the bridge will be default disabled.

charts/opentelemetry-kube-stack/values.yaml
jaronoff97: not right now... removing!

charts/opentelemetry-kube-stack/templates/bridge.yaml
```diff
@@ -0,0 +1,127 @@
...
+  {{- end }}
+  {{- end }}
+  {{- if $.Values.opAMPBridge.ports }}
+  ports:
+  {{- with $.Values.opAMPBridge.ports }}
+  {{- toYaml . | nindent 4 }}
+  {{- end }}
+  {{- end }}
+  env:
+  {{- include "opentelemetry-kube-stack.renderenvs" (dict "extraEnvs" $.Values.extraEnvs "env" .Values.opAMPBridge.env) | nindent 4 }}
```
TylerHelmuth, line 101: Lets put the static `valueFrom` env vars first so that users can refernce them in `extraEnvs`.

TylerHelmuth (approved):

PR #1145, opened 2024-04-16
[Operator] Allow setting of manager service account name, default value if not used

### LINKED ISSUE #1144: [OpenTelemetry-Operator] Enable setting name for manager serviceAccount
We should be able to set the name for the ServiceAccount that the manager runs as. Making a PR with a suggestion

Gholie:
We should be able to set the name of the service account used. If no service account name is set it defaults to `opentelemetry-operator`.

This also helps with confusion when seeing the values that are in `values.yaml`, specifically this part:
```yaml
  # -- Create the manager ServiceAccount
  serviceAccount:
    create: true
    annotations: {}
    # name: nameOverride
```
Using existing schema uncommenting `# name:` will break the chart

Resolves #1144

### DIFF
--- a/charts/opentelemetry-operator/values.schema.json
+++ b/charts/opentelemetry-operator/values.schema.json
@@ -644,7 +644,7 @@
                         },
                         "name": {
                             "type": "string",
-                            "default": "opentelemetry-operator",
+                            "default": "",
                             "title": "The name of the service account",
                             "examples": [
                                 "opentelemetry-operator"
--- a/charts/opentelemetry-operator/values.schema.json
+++ b/charts/opentelemetry-operator/values.schema.json
@@ -641,6 +641,14 @@
                             "required": [],
                             "properties": {},
                             "examples": [{}]
+                        },
+                        "name": {
+                            "type": "string",
+                            "default": "opentelemetry-operator",
+                            "title": "The name of the service account",
+                            "examples": [
+                                "opentelemetry-operator"
+                            ]
                         }
                     },
                     "examples": [{
--- a/charts/opentelemetry-operator/values.yaml
+++ b/charts/opentelemetry-operator/values.yaml
@@ -92,7 +92,8 @@ manager:
   serviceAccount:
     create: true
     annotations: {}
-    # name: nameOverride
+    ## Provide a name in place of value set in nameOverride or its default
+    name: ""
 
   ## Enable ServiceMonitor for Prometheus metrics scrape
   serviceMonitor:
--- a/charts/opentelemetry-operator/values.yaml
+++ b/charts/opentelemetry-operator/values.yaml
@@ -92,7 +92,7 @@ manager:
   serviceAccount:
     create: true
     annotations: {}
-    ## Provide a name in place of value set in nameOverride or its default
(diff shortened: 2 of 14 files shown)


charts/opentelemetry-operator/values.schema.json
```diff
@@ -641,6 +641,14 @@
                             "required": [],
                             "properties": {},
                             "examples": [{}]
+                        },
+                        "name": {
```
JaredTan95, line 645: The relevant parts of `values.yaml`  also need to be specified

JaredTan95: and pls following `Contributing.md` guide to polish this PR, in general, you need to bump up `operator` chart version and `CHARTS=opentelemetry-collector make generate-examples  `

JaredTan95 (changes requested):

charts/opentelemetry-operator/values.schema.json
Gholie: Will do a pass! :) I upped chart version from 0.55.1 -> 0.55.2. The `values.yaml` already had it specified, just not included in the schema

Gholie:
@JaredTan95 sorry for the bad first implementation, I believe it should be better now

charts/opentelemetry-operator/values.schema.json
Gholie: Ran the `make`, removed comment from `# name:` in `values.yaml`, added some comments to the value

charts/opentelemetry-operator/values.yaml
```diff
@@ -92,7 +92,8 @@ manager:
   serviceAccount:
     create: true
     annotations: {}
-    # name: nameOverride
+    ## Provide a name in place of value set in nameOverride or its default
```
jaronoff97, line 95:
```suggestion
    ## Override the default name of the serviceaccount (the name of your installation)
```

jaronoff97 (approved):

TylerHelmuth:
@JaredTan95 please take another look

JaredTan95 (approved):

Allex1 (approved):

TylerHelmuth (commented):
@Gholie can you confirm your tests locally with this change? I think you actually need to change https://github.com/open-telemetry/opentelemetry-helm-charts/blob/6ca3468af6713dad90d5dcfc20f1efe9023df2af/charts/opentelemetry-operator/templates/_helpers.tpl#L57 for this to work properly

Gholie:
> @Gholie can you confirm your tests locally with this change? I think you actually need to change https://github.com/open-telemetry/opentelemetry-helm-charts/blob/6ca3468af6713dad90d5dcfc20f1efe9023df2af/charts/opentelemetry-operator/templates/_helpers.tpl#L57 for this to work properly

I tested the following:
Left name empty '', SA name was 'opentelemetry-operator'
Set a specific name, and service account name was the one I set.

I can do some more tests tomorrow

TylerHelmuth:
Oh I misread the template, it should work as expected.  Thank you for sharing your tests

Issue #1135, opened 2024-04-11
[collector] use the otelcol-k8s distribution of the collector


TylerHelmuth:
The OpenTelemetry Collector has a new Kubernetes-specific distribution: https://github.com/open-telemetry/opentelemetry-collector-releases/tree/main/distributions/otelcol-k8s.

We should update the collector chart to use this distribution instead of Contrib since it is smaller and[ using Contrib is against Collector SIG best practices](https://github.com/open-telemetry/opentelemetry-collector-releases/tree/main/distributions/otelcol-contrib#recommendation).

This would be a significant breaking change since we use Contrib today and the K8s distro contains many less components.  Our default config wouldn't break, but users who are expecting certain receivers to be present that aren't included in the k8s distro will break.

We'll need to handled this via a feature gate, similar to how we handled GOMEMLIMIT:

1. Introduce a new root-level field that, when set, makes the k8s distribution the used image. Also add a NOTES.txt warning if not using the flag
2. Much later (TBD) switch the flag to be "on" by default.
3. Much later again, remove the flag.

Since the collector image can be directly set, maybe steps 2 and 3 can become "switch the default image" with clear instructions how to set the image/command back to contrib.

@open-telemetry/helm-approvers how long should we wait between steps 1 and 2? 3 months?  6 months?

dmitryax:
I'm not sure why we need a config option (FG) just to control another option, `image.repository`. If users don't like the change, they can easily set `image.repository` back to `otel/opentelemetry-collector-contrib`. We just need to clearly document the change in the release notes or upgrade guidelines. I don't think we need a feature gate here.

TylerHelmuth:
The reason I don't want to do it in one breaking change is that the  breaking change won't happen on install but after the collector tried to start up. 

For breaking changes that fail on install it is a lot easier to write out a meaningful error message and guide the user towards what changed. In this instance, the chart will install fine but the collector won't start, and there isn't any way for use to indicate that it is bc of the new image, it will be the collector's error message. 

I am worried about the end-user experience of doing an upgrade and then the collector doesn't startup.

dmitryax:
We can have an intermediate step with no default `image.repository`. That will fail on install/upgrade unless users specify an image explicitly. We can recommend the new image in the failure message.

TylerHelmuth:
I'm good with that. And I think it should be its own PR, not the 0.98.0 bump.

PR #1148, opened 2024-04-16
[collector] Remove default collector image

TylerHelmuth:
Works towards https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1135

This PR removes the default `image.repository` and `command.name` from the collector chart so that installs/upgrades fail for users depending on the previously set `contrib` image.

Adds a new section to UPGRADING.md to guide users through this failure.

### DIFF
--- a/charts/opentelemetry-collector/README.md
+++ b/charts/opentelemetry-collector/README.md
@@ -19,7 +19,7 @@ helm repo add open-telemetry https://open-telemetry.github.io/opentelemetry-helm
 To install the chart with the release name my-opentelemetry-collector, run the following command:
 
 ```console
-helm install my-opentelemetry-collector open-telemetry/opentelemetry-collector --set mode=<value>
+helm install my-opentelemetry-collector open-telemetry/opentelemetry-collector --set mode=<value> --set image.repository= "otel/opentelemetry-collector-k8s" --set command.name="otelcol-k8s"
 ```
 
 Where the `mode` value needs to be set to one of `daemonset`, `deployment` or `statefulset`.
@@ -87,7 +87,7 @@ The collector can be used to collect logs sent to standard output by Kubernetes 
 This feature is disabled by default. It has the following requirements:
 
 - It needs agent collector to be deployed.
-- It requires the [Filelog receiver](https://opentelemetry.io/docs/kubernetes/collector/components/#filelog-receiver) to be included in the collector, such as [contrib](https://github.com/open-telemetry/opentelemetry-collector-releases/tree/main/distributions/otelcol-contrib) version of the collector image.
+- It requires the [Filelog receiver](https://opentelemetry.io/docs/kubernetes/collector/components/#filelog-receiver) to be included in the collector, such as [k8s](https://github.com/open-telemetry/opentelemetry-collector-releases/tree/main/distributions/otelcol-k8s) version of the collector image.
 
 To enable this feature, set the  `presets.logsCollection.enabled` property to `true`.
 Here is an example `values.yaml`:
@@ -147,7 +147,7 @@ The collector can be configured to add Kubernetes metadata, such as pod name and
 
 This feature is disabled by default. It has the following requirements:
 
-- It requires the [Kubernetes Attributes processor](https://opentelemetry.io/docs/kubernetes/collector/components/#kubernetes-attributes-processor) to be included in the collector, such as [contrib](https://github.com/open-telemetry/opentelemetry-collector-releases/tree/main/distributions/otelcol-contrib) version of the collector image.
+- It requires the [Kubernetes Attributes processor](https://opentelemetry.io/docs/kubernetes/collector/components/#kubernetes-attributes-processor) to be included in the collector, such as [k8s](https://github.com/open-telemetry/opentelemetry-collector-releases/tree/main/distributions/otelcol-k8s) version of the collector image.
 
 To enable this feature, set the  `presets.kubernetesAttributes.enabled` property to `true`.
 Here is an example `values.yaml`:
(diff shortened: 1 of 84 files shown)


charts/opentelemetry-collector/README.md
```diff
@@ -19,7 +19,7 @@ helm repo add open-telemetry https://open-telemetry.github.io/opentelemetry-helm
 To install the chart with the release name my-opentelemetry-collector, run the following command:
 
 ```console
-helm install my-opentelemetry-collector open-telemetry/opentelemetry-collector --set mode=<value>
+helm install my-opentelemetry-collector open-telemetry/opentelemetry-collector --set mode=<value> --set image.repository= "otel/opentelemetry-collector-k8s" --set command.name="otelcol-k8s"
```
dmitryax, line 22: Oh, we have a different name for the command. I've been always finding this annoying. I wonder if we should just set `otelcol` for all the images... What you think? If you think it makes sense, we can create an issue in the collector repo and bring it to the SIG for discussion. We can do it gradually and support both old and new names for some time.

dmitryax (approved):

charts/opentelemetry-collector/README.md
TylerHelmuth: Yes definitely.  I don't see any value to having a unique command in each distro

dmitryax: Submitted https://github.com/open-telemetry/opentelemetry-collector-releases/issues/528

charts/opentelemetry-collector/UPGRADING.md
```diff
@@ -16,6 +16,34 @@ created but the Memory Ballast Extension will still be removed.
 Depending on the progress made in [Issue 891](https://github.com/open-telemetry/opentelemetry-helm-charts/issues/891),
 the use of `GOMEMLIMIT` may completely replace the Memory Ballast Extension in the future.
 
+## 0.87.2 to 0.88.0
+
+As part of working towards using the [OpenTelemetry Collector Kubernetes Distro](https://github.com/open-telemetry/opentelemetry-collector-releases/tree/main/distributions/otelcol-k8s) by default, the chart now requires users to explicitly set an image repository and command name. If you are already explicitly setting an image repository and command name this breaking change does not affect you.
```
jaronoff97, line 21: this may warrant a warning call out. i.e.
```

> [!WARNING]  
> Critical content demanding immediate user attention due to potential risks.
```

charts/opentelemetry-collector/UPGRADING.md
```diff
@@ -16,6 +16,34 @@ created but the Memory Ballast Extension will still be removed.
...
+
+If you are using a OpenTelemetry Community distribution of the Collector we recommend you use `otel/opentelemetry-collector-k8s`, but carefully review the [components included in this distribution](https://github.com/open-telemetry/opentelemetry-collector-releases/blob/main/distributions/otelcol-k8s/manifest.yaml) to make sure it includes all the components you use in your configuration. In the future this distribution will become the default image used for the chart.
+
+You can use the OpenTelemetry Collector Kubernetes Distro by adding these lines to your values.yaml:
+
+```yaml
+image:
+  repository: "otel/opentelemetry-collector-k8s"
+
+command:
```
jaronoff97, line 31: is command required to be set? I would opt for not setting it if we can avoid it.

jaronoff97: [from the docs](https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#define-a-command-and-arguments-when-you-create-a-pod): command corresponds to entrypoint, and if someone is using a collector distro, each one of them should have an accurate entrypoint

TylerHelmuth: Since we set `command` in the container spec instead of `args` we have to set a command.  

If we want to take advantage of `ENTRYPOINT` in the containers, which I think is a good idea, we'll need to switch to using `args`, which impacts how `command.extraArgs` and `command.name` works for users who are overriding it with their own image or args.

jaronoff97: feels like a good follow up to me, I don't think it's required here

TylerHelmuth: I'll try to make a separate PR that handles that transition as I think I'll improve this PR's transition.

TylerHelmuth: Created https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1149

charts/opentelemetry-collector/examples/deployment-otlp-traces/rendered/deployment.yaml
```diff
@@ -40,11 +40,11 @@ spec:
       containers:
         - name: opentelemetry-collector
           command:
-            - /otelcol-contrib
+            - /otelcol-k8s
```
rnishtala-sumo, line 43: why was this called `otelcol-contrib` previously?

charts/opentelemetry-collector/UPGRADING.md
```diff
@@ -16,6 +16,34 @@ created but the Memory Ballast Extension will still be removed.
 Depending on the progress made in [Issue 891](https://github.com/open-telemetry/opentelemetry-helm-charts/issues/891),
 the use of `GOMEMLIMIT` may completely replace the Memory Ballast Extension in the future.
 
+## 0.87.2 to 0.88.0
+
+As part of working towards using the [OpenTelemetry Collector Kubernetes Distro](https://github.com/open-telemetry/opentelemetry-collector-releases/tree/main/distributions/otelcol-k8s) by default, the chart now requires users to explicitly set an image repository and command name. If you are already explicitly setting an image repository and command name this breaking change does not affect you.
+
+If you are using a OpenTelemetry Community distribution of the Collector we recommend you use `otel/opentelemetry-collector-k8s`, but carefully review the [components included in this distribution](https://github.com/open-telemetry/opentelemetry-collector-releases/blob/main/distributions/otelcol-k8s/manifest.yaml) to make sure it includes all the components you use in your configuration. In the future this distribution will become the default image used for the chart.
```
rnishtala-sumo, line 23: why are we looking to move away from the contrib distribution in the helm chart?

jaronoff97: contrib really isn't meant to be run in a production setting ([doc](https://github.com/open-telemetry/opentelemetry-collector-releases/tree/main/distributions/otelcol-contrib#recommendation)). Contrib contains every component with stability greater than or equal to alpha, which means it probably has a lot of stuff the average user doesn't need. This opens up a potential attack vector if there's a vuln in one of said components that users are unknowingly pulling in. It also means that the overall image size is much larger than you may need which would result in worse first-pull latency. There are a few other reasons that are documented around, but these are the two biggest.

charts/opentelemetry-collector/UPGRADING.md
TylerHelmuth: Updated to make command name optional again

rnishtala-sumo (approved):

dmitryax (approved):

PR #1156, opened 2024-04-22
Add kafkaQueueProblems featureflag to flagd json & update env vars

EislM0203:
Add `kafkaQueueProblems` featureflag to flagd file and added `FLAGD_HOST` & `FLAGD_PORT` environment variables. Also adjusted resource limits of the frauddetection service since it kept dying.
See upstream PR here: open-telemetry/opentelemetry-demo#1528

PR #1161, opened 2024-04-30
[demo] -Set flagd memory limit to 50Mi

puckpuck:
This replaces #1131 and fixes [issue 1536](https://github.com/open-telemetry/opentelemetry-demo/issues/1536) in the demo repo.

Increases the memory limit for FlagD to be 50Mi instead of 20Mi

This also bumps the Flagd version the latest release (0.10.1)


charts/opentelemetry-demo/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: opentelemetry-demo
-version: 0.30.3
+version: 0.30.5
```
jaronoff97, line 4: this should be `0.30.4`

charts/opentelemetry-demo/values.yaml
```diff
@@ -527,15 +527,15 @@ components:
     enabled: true
     imageOverride:
       repository: "ghcr.io/open-feature/flagd"
-      tag: "v0.9.0"
+      tag: "v0.10.1"
```
jaronoff97, line 530: can you mention the tag bump in the description?

charts/opentelemetry-demo/Chart.yaml
puckpuck: I set it to 0.30.5 in anticipation of #1156 getting merged.

charts/opentelemetry-demo/values.yaml
puckpuck: Done

JaredTan95 (approved):

TylerHelmuth (approved):

PR #1173, opened 2024-05-07
[demo] Increase Kafka memory limit and bump charts

julianocosta89:
This PR increases memory limit from Kafka to `600 Mi`, otherwise the pod is going into `OOMKill`.

I've also bumped versions of the dependency charts.
This required an update on the values file:

```
opentelemetry-collector:
  image:
    repository: "otel/opentelemetry-collector-contrib"

[...]

jaeger:
  storage:
    type: memory
```


puckpuck:
I also found we have to give Kafka more memory since moving to the community image. However, even with this config, I still had to increase the -Xmx and -Xms to 400M each as well.

julianocosta89:
@puckpuck applied your suggestion.

Can we get that merged?

puckpuck (approved):

TylerHelmuth (approved):

PR #1175, opened 2024-05-08
Move otel operator CRDs to templates

### LINKED ISSUE #1167: Collector CRD 0.99.0 requires templated conversion webhook
Created from https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1164 

The conversion webhook defined in the collector CRD should be templated because it references the operator webhook service which can be dynamically set:

```bash
helm install --namespace=my-opentelemetry-operator-system --create-namespace my-opentelemetry-operator ./charts/opentelemetry-operator
```

```yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  annotations:
    cert-manager.io/inject-ca-from: opentelemetry-operator-system/opentelemetry-operator-serving-cert
    controller-gen.kubebuilder.io/version: v0.14.0
  name: opentelemetrycollectors.opentelemetry.io
spec:
  conversion:
    strategy: Webhook
    webhook:
      clientConfig:
        service:
          name: {missing my-}opentelemetry-operator-webhook
          namespace: {missing my}opentelemetry-operator-system
          path: /convert
      conversionReviewVersions:
      - v1alpha1
      - v1beta1
```

swiatekm-sumo:
Partially resolves #1167. See the issue for more details. I ended up patching both the webhook name and the certificate name in addition to the namespace in the CRD.

I verified the upgrade procedure manually.


swiatekm-sumo:
Looks like the failing test is flaky, it passes locally for me.

charts/opentelemetry-operator/UPGRADING.md
```diff
@@ -1,5 +1,25 @@
...
+```bash
+RELEASE_NAME=my-opentelemetry-operator
+RELEASE_NAMESPACE=opentelemetry-operator-system
+kubectl annotate crds instrumentations.opentelemetry.io opentelemetrycollectors.opentelemetry.io opampbridges.opentelemetry.io \
+  meta.helm.sh/release-name=${RELEASE_NAME} \
+  meta.helm.sh/release-namespace=${RELEASE_NAMESPACE}
+kubectl label crds instrumentations.opentelemetry.io opentelemetrycollectors.opentelemetry.io opampbridges.opentelemetry.io app.kubernetes.io/managed-by=Helm
+```
+
+You can also delete the CRDs and let Helm recreate them, but doing so will also delete any Custom Resources in your cluster.
```
TylerHelmuth, line 21: Ok now I see why you preferred the label/annotate solution

TylerHelmuth (commented):
Similar to how managing the CRDs via the `crd` directory is optional I think installing the CRDs via the template should be optional as well.  We should add a new section to the values.yaml that allows controlling when the templated CRDs are installed.

TylerHelmuth:
You will need to bump the minor version of the chart and regenerate the examples via `make generate-examples CHARTS=opentelemetry-operator`.

swiatekm-sumo:
> Similar to how managing the CRDs via the crd directory is optional I think installing the CRDs via the template should be optional as well. We should add a new section to the values.yaml that allows controlling when the templated CRDs are installed.

Does a top-level `installCrds` option sound good? Or do we want something more elaborate?

TylerHelmuth:
Since we're allowing templating, it feels possible someone will want to set a specific value, so I think a section called `crds` or `installCRDs` with an `enabled` field is good.

dmitryax (approved):
LGTM

jaronoff97:
+1 to tyler's recommendation

charts/opentelemetry-operator/UPGRADING.md
```diff
@@ -1,5 +1,25 @@
...
+## 0.56.0 to 0.57.0
+
+This Chart now installs CRDs as templates. OpenTelemetry Operator is rolling out a new version of the OpenTelemetryCollector CRD. For information
+about this change, see the [following document](https://github.com/open-telemetry/opentelemetry-operator/blob/main/docs/crd-changelog.md#opentelemetrycollectoropentelemetryiov1beta1).
+
+This change includes a conversion webhook, which needs to reference a namespaced webhook Service, and therefore needs to include the release namespace. See [the following issue](https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1167) for more information on the CRD change.
+
+As a result, manual steps are necessary to convince Helm to manage existing CRDs. This involves adding some annotations and labels. Set `RELEASE_NAME` and `RELEASE_NAMESPACE` to the values you're using for your Helm release, respectively:
+
+```bash
```
jaronoff97, line 12: we should specify that this should be done _prior_ to upgrade

charts/opentelemetry-operator/templates/crds/crd-opentelemetry.io_opampbridges.yaml
```diff
@@ -1,9 +1,12 @@
----
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
+    cert-manager.io/inject-ca-from: opentelemetry-operator-system/opentelemetry-operator-serving-cert
```
jaronoff97, line 5: do we want to use `{{ .Release.Namespace }}` here?

swiatekm-sumo: Hm, we should, although this doesn't actually matter if there's no conversion webhook.

jaronoff97: yeah so let's just remove this for now.

charts/opentelemetry-operator/UPGRADING.md
swiatekm-sumo: You can't actually upgrade unless you do this, Helm will complain and fail.

jaronoff97: yeah all the more reason to say that very loudly in the doc ha

charts/opentelemetry-operator/templates/crds/crd-opentelemetry.io_opampbridges.yaml
swiatekm-sumo: Ended up being easier to patch it instead.

swiatekm-sumo:
Added a `crds` section to the values file, with a `create: true` option. This makes it consistent with other create-stuff options that currently exist.

It's increasingly annoying to automatically convert CRDs published by the operator into the form the Helm Chart wants. I added a make function to make this easier, but at some point we may want to look for a more holistic solution here.

@TylerHelmuth can you have another look? Also @pavolloffay.

charts/opentelemetry-operator/values.yaml
```diff
@@ -264,6 +264,11 @@ admissionWebhooks:
   ## Secret labels
   secretLabels: {}
 
+## Install CRDS with the right webhook settings
+## These are installed as templates, so they will clash with existing CRDs in your cluster
```
TylerHelmuth, line 268:
```suggestion
## These are installed as templates, so they will clash with existing OpenTelemetry Operator CRDs in your cluster that are not already managed by the helm chart. See https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/UPGRADING.md#0560-to-0570 for more details
```

Makefile
```diff
@@ -55,23 +56,30 @@ check-examples:
...
 	fi; \
+
+define get-crd
+@curl -s -o $(1) $(2)
+@sed -i 's#opentelemetry-operator-system/opentelemetry-operator-serving-cert#{{ printf "%s/%s-serving-cert" .Release.Namespace (include "opentelemetry-operator.fullname" .) }}#g' $(1)
+@sed -i "s/opentelemetry-operator-system/{{ .Release.Namespace }}/g" $(1)
+@sed -i 's/opentelemetry-operator-webhook-service/{{ template "opentelemetry-operator.fullname" . }}-webhook/g' $(1)
+@sed -i '1s/^/{{- if .Values.crds.create }}\n/' $(1)
+@echo '{{- end }}' >> $(1)
+endef
```
jaronoff97, line 85: 🧙 good stuff

charts/opentelemetry-operator/UPGRADING.md
```diff
@@ -1,5 +1,28 @@
 # Upgrade guidelines
 
+## 0.56.0 to 0.57.0
```
TylerHelmuth, line 3: We need to add a note here and in the readme that if you were using `--skip-crds` that you now need to set `crds.create=false`

jaronoff97 (approved):
🎉

charts/opentelemetry-operator/UPGRADING.md
swiatekm-sumo: Done.

swiatekm-sumo:
I think something I'm missing here is that this change only works with cert-manager enabled. I should inject the `caBundle` directly when using a Secret. Please don't merge this until then, I'll get to it tomorrow morning.

TylerHelmuth:
Currently the chart has a cert-manager requirement by default, but supports other options. We definitely need the CRDs to work without the cert manager like they were before. I'm ok with the CI continuing to use cert manager for now since that is our default requirement

swiatekm-sumo:
Thinking about it a bit more, that change isn't necessary in this PR, as there's no conversion webhooks defined, and the CRDs don't need a caBundle. So we can merge this as-is and solve this problem in the actual upgrade do 0.99.0.

swiatekm-sumo:
I've added running e2e tests without certmanager to the CI. We can remove it later if you want, but I'd like to have confidence everything works right in both cases for the CRD migration.

charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook-with-cert-manager.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: admissionregistration.k8s.io/v1
 kind: MutatingWebhookConfiguration
 metadata:
   annotations:
-    cert-manager.io/inject-ca-from: {{ printf "%s/%s-serving-cert" .Release.Namespace (include "opentelemetry-operator.fullname" .) }}
+    cert-manager.io/inject-ca-from: {{ include "opentelemetry-operator.webhookCertAnnotation" . }}
```
TylerHelmuth, line 6: Can we omit this annotation entirely if `.Values.admissionWebhooks.certManager.enabled=false` instead of setting it to `none`?

charts/opentelemetry-operator/UPGRADING.md
TylerHelmuth: Can you add a note to the README as well that this chart does not support `--skip-crds` and instead needs `crds.create=false`,

charts/opentelemetry-operator/README.md
```diff
@@ -34,6 +34,9 @@ _See [helm repo](https://helm.sh/docs/helm/helm_repo/) for command documentation
 
 ## Install Chart
 
+> [!NOTE]  
+> This Chart uses templated CRDs, and therefore does not support `--skip-crds`. Set `crds.create=false` instead.
```
TylerHelmuth, line 38:
```suggestion
> This Chart uses templated CRDs, and therefore does not support `--skip-crds`. Use `crds.create=false` instead if you do not want the chart to install the OpenTelemetry Operator's CRDs.
```

PR #1183, opened 2024-05-13
[operator] Add operator labels to crds

TylerHelmuth:
Fixes an issue where the standard labels were not being added tot he templated CRDs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,6 +80,7 @@ define get-crd
 @sed -i 's#opentelemetry-operator-system/opentelemetry-operator-serving-cert#{{ include "opentelemetry-operator.webhookCertAnnotation" . }}#g' $(1)
 @sed -i "s/opentelemetry-operator-system/{{ .Release.Namespace }}/g" $(1)
 @sed -i 's/opentelemetry-operator-webhook-service/{{ template "opentelemetry-operator.fullname" . }}-webhook/g' $(1)
+@sed -i 's/labels:/labels:\n    {{- include "opentelemetry-operator.labels" . | nindent 4 }}/' $(1)
 @sed -i '1s/^/{{- if .Values.crds.create }}\n/' $(1)
 @echo '{{- end }}' >> $(1)
 endef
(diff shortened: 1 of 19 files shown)


jaronoff97 (approved):

swiatekm-sumo:
Why is this necessary? We don't have these in the operator release bundle.

TylerHelmuth:
@swiatekm-sumo I am 99% sure that if these labels aren't added then the installed CRDs won't be upgradable bc they wont have the `managed-by` [label and others](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/4845736e68736dd11122c2ed75f6213e9324b260/charts/opentelemetry-operator/templates/_helpers.tpl#L36-L45).

TylerHelmuth:
Also someone please give me the right sed link to add to the makefile lol

swiatekm-sumo:
> Also someone please give me the right sed link to add to the makefile lol

You mean on MacOS? Pretty sure installing gsed via brew works.

swiatekm-sumo:
> @swiatekm-sumo I am 99% sure that if these labels aren't added then the installed CRDs won't be upgradable bc they wont have the `managed-by` [label and others](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/4845736e68736dd11122c2ed75f6213e9324b260/charts/opentelemetry-operator/templates/_helpers.tpl#L36-L45).

I remember testing this and it worked. Helm should add any labels it needs for release tracking on its own. I can re-test if you want.

TylerHelmuth:
@swiatekm-sumo ya that would be good, my local testing showed otherwise.

TylerHelmuth:
so it looks like we get 

```
Labels:       app.kubernetes.io/managed-by=Helm
              app.kubernetes.io/name=opentelemetry-operator
Annotations:  controller-gen.kubebuilder.io/version: v0.14.0
              meta.helm.sh/release-name: testing
              meta.helm.sh/release-namespace: default
```

Which means we are missing `app.kubernetes.io/instance: {{ .Release.Name }}` but it an uninstall of the chart does remove the CRDs, so I was wrong.

Issue #1189, opened 2024-05-17
[Operator] Delete collector image version from the values.yaml


jdiegosierra:
Hello,

I noticed that currently _.Chart.AppVersion_ is the default value for the opentelemetry-operator.image
https://github.com/open-telemetry/opentelemetry-helm-charts/blob/2f4e7a8f960366b961d621c9bc7c6ae389129af3/charts/opentelemetry-operator/templates/_helpers.tpl#L150

Keeping in mind that the operator and the collector should have the same version https://github.com/open-telemetry/opentelemetry-operator?tab=readme-ov-file#opentelemetry-operator-vs-opentelemetry-collector. 

Why not using appVersion as default for collectorImage as well?

Or simply why the values.yaml has the collector image set? 
https://github.com/open-telemetry/opentelemetry-helm-charts/blob/2f4e7a8f960366b961d621c9bc7c6ae389129af3/charts/opentelemetry-operator/values.yaml#L45

If I'm not wrong, if we don't set the  _--collector-image flag_ https://github.com/open-telemetry/opentelemetry-helm-charts/blob/2f4e7a8f960366b961d621c9bc7c6ae389129af3/charts/opentelemetry-operator/templates/deployment.yaml#L51
The operator will take care of deploying a collector with the same version as itself, isn't it? So why is needed to set a value for the collector by default?

TylerHelmuth:
> The operator will take care of deploying a collector with the same version as itself, isn't it? So why is needed to set a value for the collector by default?

I'd rather do this instead of reuse `appVersion`.

jdiegosierra:
> > The operator will take care of deploying a collector with the same version as itself, isn't it? So why is needed to set a value for the collector by default?
> 
> I'd rather do this instead of reuse `appVersion`.

Yeah, probably it makes more sense. So there is no reason to have a default value in the values.yaml, right? Do you agree to delete the default value?

TylerHelmuth:
Yes I think that is a safe, un-breaking change that will make operator upgrades easier

jdiegosierra:
I'm going to check if the operator works as expected in case of not using the flag `--collector-image`

jdiegosierra:
> I'm going to check if the operator works as expected in case of not using the flag `--collector-image`

Seems that it works as expected, I deployed the operator setting an empty tag:
```yaml
collectorImage:
    repository: ghcr.io/otel/opentelemetry-collector-contrib
    tag: ""
```
[...]

PR #1196, opened 2024-05-21
[Operator] Optional tag for collector

jdiegosierra:
Resolves https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1189

### DIFF
--- a/charts/opentelemetry-operator/templates/deployment.yaml
+++ b/charts/opentelemetry-operator/templates/deployment.yaml
@@ -47,10 +47,12 @@ spec:
             {{- if or .Values.admissionWebhooks.create .Values.admissionWebhooks.secretName }}
             - --webhook-port={{ .Values.manager.ports.webhookPort }}
             {{- end }}
-            {{- if and .Values.manager.collectorImage.repository .Values.manager.collectorImage.tag }}
-            - --collector-image={{ .Values.manager.collectorImage.repository }}:{{ .Values.manager.collectorImage.tag }}
-            {{- else if .Values.manager.collectorImage.repository }}
-            - --collector-image={{ .Values.manager.collectorImage.repository }}
+            {{- if .Values.manager.collectorImage.repository -}}
+            {{- $collectorImageTag := "" -}}
+            {{- if .Values.manager.collectorImage.tag -}}
+            {{- $collectorImageTag = print ":" .Values.manager.collectorImage.tag -}}
+            {{- end }}
+            - --collector-image={{ .Values.manager.collectorImage.repository }}{{ $collectorImageTag }}
             {{- end }}
             {{- if and .Values.manager.opampBridgeImage.repository .Values.manager.opampBridgeImage.tag }}
             - --operator-opamp-bridge-image={{ .Values.manager.opampBridgeImage.repository }}:{{ .Values.manager.opampBridgeImage.tag }}
--- a/charts/opentelemetry-operator/templates/deployment.yaml
+++ b/charts/opentelemetry-operator/templates/deployment.yaml
@@ -49,6 +49,8 @@ spec:
             {{- end }}
             {{- if and .Values.manager.collectorImage.repository .Values.manager.collectorImage.tag }}
             - --collector-image={{ .Values.manager.collectorImage.repository }}:{{ .Values.manager.collectorImage.tag }}
+            {{- else if .Values.manager.collectorImage.repository }}
+            - --collector-image={{ .Values.manager.collectorImage.repository }}
             {{- end }}
             {{- if and .Values.manager.opampBridgeImage.repository .Values.manager.opampBridgeImage.tag }}
             - --operator-opamp-bridge-image={{ .Values.manager.opampBridgeImage.repository }}:{{ .Values.manager.opampBridgeImage.tag }}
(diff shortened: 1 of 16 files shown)


jaronoff97 (approved):

charts/opentelemetry-operator/templates/deployment.yaml
```diff
@@ -49,6 +49,8 @@ spec:
             {{- end }}
             {{- if and .Values.manager.collectorImage.repository .Values.manager.collectorImage.tag }}
             - --collector-image={{ .Values.manager.collectorImage.repository }}:{{ .Values.manager.collectorImage.tag }}
+            {{- else if .Values.manager.collectorImage.repository }}
+            - --collector-image={{ .Values.manager.collectorImage.repository }}
```
TylerHelmuth, line 53: Let's move this and the above logic into a helper template so we don't have to do the same condition twice.

jdiegosierra: Do you mean, something like this?
```
{{- define "opentelemetry-operator.collectorImage" -}}
  {{- if and .Values.manager.collectorImage.repository .Values.manager.collectorImage.tag }}
    - --collector-image={{ .Values.manager.collectorImage.repository }}:{{ .Values.manager.collectorImage.tag }}
  {{- else if .Values.manager.collectorImage.repository }}
    - --collector-image={{ .Values.manager.collectorImage.repository }}
  {{- end -}}
{{- end }}
```

TylerHelmuth: I believe we can do something like:

```
{{ - $tag = "" }}
{{ if .Values.manager.collectorImage.tag }}
{{ - $tag = ":" + .Values.manager.collectorImage.tag }}
{{- end }}
- --collector-image={{ .Values.manager.collectorImage.repository }}{{ $tag }}
```

TylerHelmuth, line 53: @jdiegosierra in the situation that we don't pass in a tag, are you sure the tag that ends up being uses is the default collector version from the operator and not the latest version of the set image repository?

jdiegosierra: I tried with the latest version, so let me check it with a different one

jdiegosierra: @TylerHelmuth It works as expected:
`appVersion: 0.100.1 `
```yaml
manager:
  image:
    repository: ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator
    tag: "0.99.0"
  collectorImage:
    repository: ""
    tag: ""
```
```bash
k describe pod opentelemetry-operator-79cbd68444-wdpl9 | grep Image: 
    Image:         ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator:0.99.0
```
```bash
k describe pod simplest-collector-54bb659675-99d7g | grep Image:
    Image:         ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector:0.99.0
```

jdiegosierra: BTW, despite you are forcing the users to add a collector repository. I would add an if condition here so just in case you remove the requirement in the future, it can avoid a bug in case you forget to add the if condition here later.
Check it out:

```
{{- if .Values.manager.collectorImage.repository -}}
{{- $collectorImageTag := "" -}}
{{- if .Values.manager.collectorImage.tag -}}
{{- $collectorImageTag = print ":" .Values.manager.collectorImage.tag -}}
{{- end }}
- --collector-image={{ .Values.manager.collectorImage.repository }}{{ $collectorImageTag }}
{{- end }}
```

TylerHelmuth: What about when you set and image repository and not a tag?

jdiegosierra: In that scenario, the operator assumes the collector image is the latest one. If we want to have both components with the same version in that specific scenario, I guess we can pick the version from the operator.

```yaml
# Source: opentelemetry-operator/templates/deployment.yaml
  - --collector-image=otel/opentelemetry-collector
image: "ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator:0.99.0"
```

jdiegosierra: Anyway, that situation can not happen right now because it is mandatory to have both fields set:
https://github.com/open-telemetry/opentelemetry-helm-charts/blob/ef0e1ac4f645cdbb9bd0108c76b1ed69e418430c/charts/opentelemetry-operator/templates/deployment.yaml#L50

jdiegosierra:
BTW, what do you think about changing this alert with a "warning"?
https://github.com/open-telemetry/opentelemetry-helm-charts/blob/ef0e1ac4f645cdbb9bd0108c76b1ed69e418430c/charts/opentelemetry-operator/templates/NOTES.txt#L1-L3
If the user doesn't set a tag, the image will not take effect, and setting only the repository field as mandatory might create confusion:
https://github.com/open-telemetry/opentelemetry-helm-charts/blob/ef0e1ac4f645cdbb9bd0108c76b1ed69e418430c/charts/opentelemetry-operator/templates/deployment.yaml#L50-L53

Issue #677, opened 2023-03-09
Consider adding a clusterrole for custom resources with aggregate-to-view


andrewdinunzio:
It would be nice if one of the helm charts (preferably a new dedicated helm chart just for CRDs) installed a ClusterRole that looks like this:

```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: otel-view
  labels:
    rbac.authorization.k8s.io/aggregate-to-view: "true"
rules:
- apiGroups:
  - opentelemetry.io
  resources:
  - '*'
  verbs:
  - get
  - list
  - watch
```

This way, users don't need to make their own RBAC permissions to view these resources, and anyone with "view" access can view them. Similar for the other aggregations. Any secrets should be stored in Secrets anyway, so I don't see an issue personally with giving anyone who can view a Deployment the ability to view an OpenTelemetryCollector or Instrumentation too.

Allex1:
@andrewdinunzio it seems your issue has 2 proposals in it :

1. Make a separate chart just for cr's which is something that makes sense to me, would you be able to work on that ?

3. Add a read only cluster role for the cr's so that everyone can at least see their configs
To me this does not really makes sense as Kubernetes clusters vary wildly between companies/vendors. Imagine single tenant vs multi tenant k8s clusters, where even ro for all users at the cluster level is undesirable. Maintaining all these use cases would require a lot of time.
Instead, we provide flags to disable the existing role/clusterrole(which include all rules you may need) and add your custom rbac by sub-charting this in your env.

andrewdinunzio:
I don't have much experience with multi-tenant clusters to know much about that but I figured maybe in that case that 1) they would probably be isolated by namespace and within their tenant namespace they would bind a normal namespaced `Role` to the `view` `ClusterRole` just for that namespace. Then they'd have view permissions in their namespace only. I'm actually not sure if that would work that way though. Also figured if someone can already view Deployments, Services, etc. (i.e. they have the `view` role) I didn't see why they wouldn't also be able to see OTel custom resources too.

W.r.t. a crds chart, I can certainly help with that when I find time, though I never used operator sdk before so will need to familiarize myself with that and how it could be automated with the releases, etc.

Allex1:
> they would probably be isolated by namespace and within their tenant namespace they would bind a normal namespaced Role to the view ClusterRole just for that namespace. 

I see a few use cases:
[...]

PR #1203, opened 2024-05-28
Create isolated chart for operator CRDs

jaronoff97:
Creates an isolated chart for operator CRDs per https://github.com/open-telemetry/opentelemetry-helm-charts/issues/677

### DIFF
--- a/charts/opentelemetry-operator-crds/.helmignore
+++ b/charts/opentelemetry-operator-crds/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*.orig
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+.vscode/
--- a/charts/opentelemetry-operator-crds/Chart.yaml
+++ b/charts/opentelemetry-operator-crds/Chart.yaml
@@ -1,0 +1,13 @@
+apiVersion: v2
+name: opentelemetry-operator-crds
+version: 0.0.1
+description: |
+  This chart contains only the CRDs needed for installation for the opentelemetry operator
+type: application
+home: https://opentelemetry.io/
+sources:
+  - https://github.com/open-telemetry/opentelemetry-operator
+maintainers:
+  - name: jaronoff97
+icon: https://raw.githubusercontent.com/cncf/artwork/a718fa97fffec1b9fd14147682e9e3ac0c8817cb/projects/opentelemetry/icon/color/opentelemetry-icon-color.png
+appVersion: 0.100.0
--- a/charts/opentelemetry-operator-crds/README.md
+++ b/charts/opentelemetry-operator-crds/README.md
@@ -1,0 +1,5 @@
+# OpenTelemetry Operator CRDs
+
+This chart is meant to be used as a subchart for the operator and kube-stack.
+
+[Read more here](https://github.com/open-telemetry/opentelemetry-helm-charts/issues/677)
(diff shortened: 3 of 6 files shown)


andrewdinunzio:
Not to be a pain but just wanted to mention the issue with the `crds` directory that helm provides in that you cannot add new CRDs (or even new versions of CRDs) if you use it. Upgrades won't add them.

There are 3 ways I know of that you can handle CRDs, 2 of which have issues:
1. Use the `crds` directory. This has the issue mentioned above.
2. Put the CRDs in the templates. This has the issue with the stale discovery cache mentioned in the other issue.
3. Put the CRDs in the templates of a **separate** helm chart. This works if you install the CRDs chart **first**, but runs into the issues from number 2 if you try to use the CRDs helm chart as a dependency of another chart.

jaronoff97:
@andrewdinunzio this is a good point, I am opting for simplifying _only_ the installation process and eventually centralizing instructions for upgrading CRDs as prometheus [does here](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack#from-58x-to-59x). I am recommending we do it this way because managing the CRDs via helm has a huge amount of thorns as we discussed in the issue. i.e. we simplify the installation process and give _very_ specific upgrading instructions for CRDs. I have also been struggling with helm and CRDs for a while now and I don't think any of these solutions is great.

andrewdinunzio:
Gotcha. Personally I'm not a fan of manual steps as part of upgrades as it makes it harder to keep things up to date and doesn't fit well with CI/CD pipelines, but so far I've had no issues with just pulling the CRDs out and `kubectl apply`ing them before helm installing the main chart (setting the value to disable CRDs in the helm install). It gets trickier now that the CRDs are templated, but I'll figure something out, probably just `helm template` it first then pull the CRDs out.

jaronoff97:
@andrewdinunzio definitely – that's exactly why i want to get them out of templates. This chart will _only_ work when a user doesn't want the conversion webhook. I think I will leave it to power-users like yourself to decide how to handle automatic upgrades. Hopefully helm will eventually improve this lifecycle 🤞

JaredTan95:
> Not to be a pain but just wanted to mention the issue with the `crds` directory that helm provides in that you cannot add new CRDs (or even new versions of CRDs) if you use it. Upgrades won't add them.
> 
> There are 3 ways I know of that you can handle CRDs, 2 of which have issues:
> 
> 1. Use the `crds` directory. This has the issue mentioned above.
> 2. Put the CRDs in the templates. This has the issue with the stale discovery cache mentioned in the other issue.
> 3. Put the CRDs in the templates of a **separate** helm chart. This works if you install the CRDs chart **first**, but runs into the issues from number 2 if you try to use the CRDs helm chart as a dependency of another chart.
> 
> Possible that I'm missing something, but I've been struggling with helm and CRDs for a while and haven't found any possible option for installing CRDs and CRs in one helm install/upgrade in a way that is reliable.

I'd like to share my approach here.

Again, we put the crds directory in the `crds` directory (it doesn't matter what directory we put in for this solution) and mount the contents of the `crds` directory into a `job` with `"helm.sh/hook": pre-install,pre-upgrade`  annotations added, the pre-hook-install/upgrade `job` will exec  `kubectl apply -- server-side-f /crds --force-conflicts` before helm install or upgrade executes.

TylerHelmuth:
@JaredTan95 have you used that technique for other CRDs already? Do you know of any other helm charts doing that?

JaredTan95:
> @JaredTan95 have you used that technique for other CRDs already? Do you know of any other helm charts doing that?

- [Kubean: cluster lifecycle management toolchains based on kubespray and other cluster LCM engine.](https://github.com/kubean-io/kubean/blob/main/charts/kubean/templates/prehook-job.yml)
- [clusterpedia: It is an encyclopedia of multi-cluster to synchronize, search for, and simply control multi-cluster resources.](https://github.com/clusterpedia-io/clusterpedia-helm/blob/main/charts/clusterpedia-core/templates/pre-hook-crds-configmap.yaml)
- I'll keep you updated over the next few days

jaronoff97:
CLosing this in favor of #1214

PR #1206, opened 2024-05-30
[collector] add node ip to kubeletstats

alextricity25:
In some scenarios, it's desirable to use the node IP instead of the node name to gather kubelet metrics. This commit makes it possible for the user to specify the `presets.kubeletMetrics.useNodeIp` flag so that the node ip is used as an endpoint instead of the node name.

Somewhat resovles https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/22843

### DIFF
--- a/charts/opentelemetry-collector/templates/_pod.tpl
+++ b/charts/opentelemetry-collector/templates/_pod.tpl
@@ -56,6 +56,10 @@ containers:
         valueFrom:
           fieldRef:
             fieldPath: spec.nodeName
+      - name: K8S_NODE_IP
+        valueFrom:
+          fieldRef:
+            fieldPath: status.hostIp
       {{- end }}
       {{- if and (.Values.useGOMEMLIMIT) ((((.Values.resources).limits).memory))  }}
       - name: GOMEMLIMIT
--- a/charts/opentelemetry-collector/templates/_pod.tpl
+++ b/charts/opentelemetry-collector/templates/_pod.tpl
@@ -59,7 +59,7 @@ containers:
       - name: K8S_NODE_IP
         valueFrom:
           fieldRef:
-            fieldPath: status.hostIp
+            fieldPath: spec.hostIp
       {{- end }}
       {{- if and (.Values.useGOMEMLIMIT) ((((.Values.resources).limits).memory))  }}
       - name: GOMEMLIMIT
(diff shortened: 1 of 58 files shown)


TylerHelmuth:
Instead of making this an option I'd be ok with switching to using Node IP in the preset instead of Node Name.

I'm actually surprised we're using node name instead of node ip as I'd expect IP to be more universally usable.  We've seen issues like the one linked where IP is the solution, we could just use it all the time.  I don't think it would break users.

@dmitryax @pavolloffay @jinja2 what do you think?

jinja2:
Yeah should be okay to switch to node ip. It might be a breaking change for setups in which kubelet is not serving a certificate with the IP listed in subject alt name. Users will need to add the `insecure_skip_verify: true` config for such situations.

TylerHelmuth:
> It might be a breaking change for setups in which kubelet is not serving a certificate with the IP listed in subject alt name
@jinja2 how common would this be?

alextricity25:
@TylerHelmuth Done! Latest commit removes useNodeIp and now exclusively uses K8S_NODE_IP for the kubeMetrics endpoint.

TylerHelmuth (commented):
Please bump the minor version of the chart and run `make generate-examples CHARTS=opentelemetry-collector`

TylerHelmuth:
> Error: INSTALLATION FAILED: DaemonSet.apps "opentelemetry-collector-z9u7pzlfno-agent" is invalid: spec.template.spec.containers[0].env[2].valueFrom.fieldRef.fieldPath: Invalid value: "spec.hostIp": error converting fieldPath: field label not supported: spec.hostIp

@alextricity25 looks like something with your use of the downward api is incorrect

charts/opentelemetry-collector/templates/_pod.tpl
```diff
@@ -56,6 +56,10 @@ containers:
         valueFrom:
           fieldRef:
             fieldPath: spec.nodeName
+      - name: K8S_NODE_IP
+        valueFrom:
+          fieldRef:
+            fieldPath: status.hostIp
```
ChrsMark, line 62:
```suggestion
            fieldPath: status.hostIP
```

ChrsMark:
@alextricity25 you will need to rebase as well :).

TylerHelmuth:
@legoGoat feel free to open a new PR

PR #1166, opened 2024-05-02
Add Presets for opentelemetry-kube-stack

jaronoff97:
Closes #1081 

This PR adds in full presets and default collectors for the chart.

TODO:
* ~generate examples~
* ~add some better validations~
  * This will be done in a followup https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1191

PR #1214, opened 2024-06-05
Explicitly embed CRDs for installation ONLY in the OKS

jaronoff97:
Embeds the CRDs for installation for the opentelemetry-kube-stack chart without handling upgrades for now. This brings us to parity with what the kube-prom-stack chart does. This also unblocks #1166. In the future we should investigate the approach @JaredTan95 describes [here](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1203#issuecomment-2150336985).


charts/opentelemetry-kube-stack/values.yaml
```diff
@@ -21,6 +26,11 @@ opentelemetry-operator:
     # Setting this allows for an installation of the otel operator at the same time as the custom resources it manages.
     failurePolicy: "Ignore"
 
+  # This is disabled by default, as doing so creates a race condition with helm.
+  # https://github.com/open-telemetry/opentelemetry-helm-charts/issues/677
```
TylerHelmuth, line 30: We should add that we do not recommend users never set this value to true when using the OKS chart.

TylerHelmuth (commented):
@jaronoff97 don't we need to template the CRDs in the nested chart in order to make the conversion patch work?

jaronoff97:
@TylerHelmuth I probably need to call this out explicitly, but this is for the world where the conversion webhook is fully unnecessary because we will only be applying v1beta1 CRs. We can do this because the chart is also creating the collectors.

charts/opentelemetry-kube-stack/charts/crds/README.md
```diff
@@ -4,6 +4,9 @@ This chart contains the CRDs for _*installation*_ only right now for the opentel
 
 This approach is inspired by the kube-prometheus-stack approach which you can see discussion on [here](https://github.com/prometheus-community/helm-charts/issues/3548).
 
+> [!NOTE]
+> This chart explicitly _does not_ support the conversion webhook that is currently in the opentelemetry-operator chart. This is because the opentelemetry-kube-stack chart will only work with v1beta1 CRDs. If you wish to use v1alpha1 CRDs to install collectors, you must disable the operator installation as part of the opentelemetry-kube-stack and install the operator separately.
```
TylerHelmuth, line 8: I think you'd also need to install the CRDs separately too. Probably just best to strictly limit the chart to v1beta1 and provide no support for v1alpha1

TylerHelmuth (approved):

JaredTan95 (approved):

PR #1223, opened 2024-06-12
[operator] allow setting custom recommender name

t3mi:
It's possible for multiple recommenders with different configurations to be present in single cluster. This change allows VPA configuration to be updated by custom recommender if one was set, otherwise - default recommender is used.

### DIFF
--- a/charts/opentelemetry-operator/values.schema.json
+++ b/charts/opentelemetry-operator/values.schema.json
@@ -947,6 +947,24 @@
                             "properties": {},
                             "examples": [{}]
                         },
+                        "recommenders": {
+                            "type": "array",
+                            "default": [],
+                            "title": "The recommenders Schema",
+                            "maxItems": 1,
+                            "items": {
+                                "type": "object",
+                                "properties": {
+                                    "name": {
+                                        "type": "string"
+                                    }
+                                },
+                                "additionalProperties": false
+                            },
+                            "examples": [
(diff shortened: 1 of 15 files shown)


charts/opentelemetry-operator/values.schema.json
```diff
@@ -947,6 +947,24 @@
...
+                                "type": "object",
+                                "properties": {
+                                    "name": {
+                                        "type": "string"
+                                    }
+                                },
+                                "additionalProperties": false
+                            },
+                            "examples": [
+                                []
```
jaronoff97, line 965: can you add an example for this?

charts/opentelemetry-operator/values.yaml
```diff
@@ -140,6 +140,13 @@ manager:
   # Enable vertical pod autoscaler support for the manager
   verticalPodAutoscaler:
     enabled: false
+
+    # Recommender responsible for generating recommendation for the object.
+    # List should be empty (then the default recommender will generate the recommendation)
+    # or contain exactly one recommender.
+    # recommenders:
+    # - name: custom-recommender-performance
```
jaronoff97, line 148: Given the VPA's recommenders list can only have a single item does it make sense for us to expose this field as a list instead of just a string to set? i.e. if `len(recommender.name) > 0` set it, otherwise don't?

Issue #1224, opened 2024-06-15
Align open-telemetry-operator Helm Chart with Deprecated create-rbac-permission flag


SergK:
**Is your feature request related to a problem? Please describe.**
The `create-rbac-permission` flag is deprecated as per discussions in [open-telemetry/opentelemetry-operator#2588](https://github.com/open-telemetry/opentelemetry-operator/issues/2588) and changes in [open-telemetry/opentelemetry-operator#2787](https://github.com/open-telemetry/opentelemetry-operator/pull/2787). This deprecation impacts the `open-telemetry-operator` helm chart, specifically the `createRbacPermissions` flag, which could affect the functionality of `k8sattributes`.

**Describe the solution you'd like**
To align the helm chart with the deprecation of the `create-rbac-permission` flag, it is essential to ensure the `SERVICE_ACCOUNT_NAME` environment variable is available inside the pod, as indicated by the source code in [open-telemetry/opentelemetry-operator/internal/autodetect/rbac/check.go#L29](https://github.com/open-telemetry/opentelemetry-operator/blob/v0.102.0/internal/autodetect/rbac/check.go#L29) and the manager configuration [open-telemetry/opentelemetry-operator/config/manager/manager.yaml#L55](https://github.com/open-telemetry/opentelemetry-operator/blob/v0.102.0/config/manager/manager.yaml#L55).

**Describe alternatives you've considered**
N/A

**Additional context**
Ensuring the `SERVICE_ACCOUNT_NAME` environment variable is set correctly is crucial for the functionality of `k8sattributes` following the deprecation of the `create-rbac-permission` flag.

PR #1225, opened 2024-06-15
[operator] Align to `create-rbac-permission` flag deprecation

SergK:
Ensure SERVICE_ACCOUNT_NAME env variable is set in pods to align with the deprecation of the `create-rbac-permission flag`

Addressing functionality issues with k8sattributes.

Closes: #1224

### DIFF
--- a/charts/opentelemetry-operator/README.md
+++ b/charts/opentelemetry-operator/README.md
@@ -108,6 +108,18 @@ $ helm show values open-telemetry/opentelemetry-operator
 
 When using this chart as a subchart, you may want to unset certain default values. Since Helm v3.13 values handling is improved and null can now consistently be used to remove values (e.g. to remove the default CPU limits).
 
+### Role-based Access Control (RBAC) Configuration
+
+The OpenTelemetry Collector requires specific RBAC permissions to function correctly, especially when using the `k8sattributesprocessor`. Depending on your deployment's scope, you may need to configure Cluster-scoped or Namespace-scoped RBAC permissions.
+
+- **Cluster-scoped RBAC**: Necessary if the collector is to receive telemetry from across multiple namespaces. This setup requires `get`, `watch`, and `list` permissions on `pods`, `namespaces`, and `nodes`, plus `replicasets` if using deployment-related attributes.
+
+- **Namespace-scoped RBAC**: Suitable for collecting telemetry within a specific namespace. This requires setting up a `Role` and `RoleBinding` to grant access to `pods` and `replicasets` within the target namespace. This setup limits the collector's access to resources within the specified namespace only.
+
+**Important**: The `manager.createRbacPermissions` flag in the Helm chart values should be set to `false` if you are manually configuring RBAC permissions for the collector. Manual configuration allows for more granular control over the permissions granted to the OpenTelemetry Collector, ensuring it has exactly the access it needs based on your specific deployment requirements. Conversely, setting `manager.createRbacPermissions` to `true` will allow the operator to automatically configure RBAC for your collectors.
+
+For detailed instructions and examples on configuring RBAC permissions, please refer to the [official documentation](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/k8sattributesprocessor/README.md).
+
 ## Install OpenTelemetry Collector
 
 _See [OpenTelemetry website](https://opentelemetry.io/docs/collector/) for more details about the Collector_
--- a/charts/opentelemetry-operator/templates/deployment.yaml
+++ b/charts/opentelemetry-operator/templates/deployment.yaml
@@ -77,19 +77,24 @@ spec:
             {{- if .Values.manager.featureGates }}
             - --feature-gates={{ .Values.manager.featureGates }}
             {{- end }}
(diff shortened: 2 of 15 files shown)


SergK:
Closes: #1224

charts/opentelemetry-operator/templates/deployment.yaml
```diff
@@ -74,16 +74,16 @@ spec:
...
             - --feature-gates={{ .Values.manager.featureGates }}
             {{- end }}
-            {{- if .Values.manager.createRbacPermissions }}
-            - --create-rbac-permissions
-            {{- end }}
             {{-  if .Values.manager.extraArgs  }}
             {{- .Values.manager.extraArgs | toYaml | nindent 12 }}
             {{-  end  }}
           command:
-            - /manager
```
TylerHelmuth, line 84: Why was the command removed?

charts/opentelemetry-operator/templates/deployment.yaml
```diff
@@ -74,16 +74,16 @@ spec:
             {{- if .Values.manager.featureGates }}
             - --feature-gates={{ .Values.manager.featureGates }}
             {{- end }}
-            {{- if .Values.manager.createRbacPermissions }}
```
TylerHelmuth, line 77: Please add a case to `NOTES.txt` that fails if `manager.createRbacPermissions` is set

charts/opentelemetry-operator/templates/deployment.yaml
SergK: sorry, returning back...

charts/opentelemetry-operator/templates/deployment.yaml
TylerHelmuth: @SergK I thought that this was the only place `createRbacPermissions` was used but I was wrong.  Since we still use the field in clusterrole.yaml your are correct that we dont need a failure case.

Your readme additions would be great.

SergK: @TylerHelmuth I have updated the README.md file. Please let me know if any changes are needed in the section or anything else.

charts/opentelemetry-operator/templates/deployment.yaml
```diff
@@ -74,16 +74,17 @@ spec:
...
-            - --create-rbac-permissions
-            {{- end }}
             {{-  if .Values.manager.extraArgs  }}
             {{- .Values.manager.extraArgs | toYaml | nindent 12 }}
             {{-  end  }}
           command:
             - /manager
-          {{- if .Values.manager.env }}
           env:
+            - name: SERVICE_ACCOUNT_NAME
```
TylerHelmuth, line 83: instead of `manager.createRbacPermissions` setting the RBAC flag, should we only add this env var when `manager.createRbacPermissions: true`?

SergK: Put it under condition, but it works because of `env` is required based on the schema, otherwise we need to redesign this part with env variables.

charts/opentelemetry-operator/templates/deployment.yaml
```diff
@@ -89,6 +86,12 @@ spec:
               value: {{ $value | quote -}}
             {{- end }}
           {{- end }}
+          {{- if .Values.manager.createRbacPermissions}}
```
TylerHelmuth, line 89: We need to make sure the `env` header is present when .Values.manager.createRbacPermissions is true but .Values.manager.env has no values

charts/opentelemetry-operator/README.md
```diff
@@ -108,6 +108,18 @@ $ helm show values open-telemetry/opentelemetry-operator
...
+
+- **Cluster-scoped RBAC**: Necessary if the collector is to receive telemetry from across multiple namespaces. This setup requires `get`, `watch`, and `list` permissions on `pods`, `namespaces`, and `nodes`, plus `replicasets` if using deployment-related attributes.
+
+- **Namespace-scoped RBAC**: Suitable for collecting telemetry within a specific namespace. This requires setting up a `Role` and `RoleBinding` to grant access to `pods` and `replicasets` within the target namespace. This setup limits the collector's access to resources within the specified namespace only.
+
+**Important**: The `manager.createRbacPermissions` flag in the Helm chart values should be set to `false` if you are manually configuring RBAC permissions for the collector. Manual configuration allows for more granular control over the permissions granted to the OpenTelemetry Collector, ensuring it has exactly the access it needs based on your specific deployment requirements.
```
jaronoff97, line 119: I would include here that setting `manager. createRbacPermissions` will allow the operator to automatically configure RBAC for your collectors in addition to the converse.

charts/opentelemetry-operator/templates/deployment.yaml
jaronoff97: +1 another option would be to append this env var to the `.Values.manager.env` list if its not present already

charts/opentelemetry-operator/README.md
SergK: done

jaronoff97 (approved):

TylerHelmuth (approved):

PR #1226, opened 2024-06-16
[operator] Add extra labels/annotations on default alerts

Allex1:
This is needed for routing alerts to 3rd party systems via labels and using custom annotations for alertmanager templates.

### DIFF
--- a/charts/opentelemetry-operator/templates/prometheusrule.yaml
+++ b/charts/opentelemetry-operator/templates/prometheusrule.yaml
@@ -11,7 +11,7 @@ metadata:
     {{- printf "%s: %s" $key (tpl $value $ | quote) | nindent 4 }}
     {{- end }}
   annotations:
-    {{- range $key, $value := .Values.manager.serviceMonitor.annotations }}
+    {{- range $key, $value := .Values.manager.prometheusRule.annotations }}
     {{- printf "%s: %s" $key (tpl $value $ | quote) | nindent 4 }}
     {{- end }}
 spec:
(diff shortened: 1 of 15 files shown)


charts/opentelemetry-operator/values.yaml
```diff
@@ -121,9 +121,13 @@ manager:
     # Create default rules for monitoring the manager
     defaultRules:
       enabled: false
-    # additional labels on the PrometheusRule
+      ## Additional labels for PrometheusRule alerts
+      additionalRuleLabels: {}
+      ## Additional annotations for PrometheusRule alerts
+      additionalRuleAnnotations: {}
+    # additional labels on the PrometheusRule object
```
TylerHelmuth, line 128: Can this be handled via `prometheusRule.groups` instead?

Allex1: `prometheusRule.groups` allows defining new custom alerts on which we can already add any labels/annotations.
This for the [existing](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/4d3d0548f409b2da6d737228f1f67a3f5602b21d/charts/opentelemetry-operator/templates/prometheusrule.yaml#L23) alerts included with the chart.

TylerHelmuth: I'm wondering if these rules should follow the same rules as presets in the collector chart or if they should be configurable like the defaults in the demo chart.

In theory you could disable the default rules and recreate the exact rules, including labels, via the groups field right?

Allex1: > In theory you could disable the default rules and recreate the exact rules, including labels, via the groups field right?

Absolutely, but then you would need to manage them separately from the chart ... which is undesirable for us.

TylerHelmuth: What do you mean by manage them separately from the chart?

Allex1: I meant keeping them into the values an not using the default templates would require some work on our side vs being able to label the existing templated alerts.
This may cause divergence from the main chart in the future as more alerts are being added/changed.
I see the same setup in [kube-prometheus-stack](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/values.yaml#L97C3-L97C23)
Any concerns for merging this ?

Allex1: @TylerHelmuth can we merge this ?

TylerHelmuth (approved):
@jaronoff97 please review

charts/opentelemetry-operator/templates/prometheusrule.yaml
```diff
@@ -11,7 +11,7 @@ metadata:
     {{- printf "%s: %s" $key (tpl $value $ | quote) | nindent 4 }}
     {{- end }}
   annotations:
-    {{- range $key, $value := .Values.manager.serviceMonitor.annotations }}
+    {{- range $key, $value := .Values.manager.prometheusRule.annotations }}
```
jaronoff97, line 14: this is a bug fix right?

jaronoff97 (approved):

PR #1249, opened 2024-07-08
[operator] support cutom autocert-certPeriod time by days

JaredTan95:
As otel-operator matures, the default 365 days will be at risk when users choose to deploy the operator by autoGenerateCert. Therefore, this PR is intended to provide the ability to configure the expiration time.

I performed local verification in the following way:

1. using custom values to install `operator-chart`, the same with `examples/cutstom-autoGenerateCert-period/values.yaml`
```bash
cd charts/opentelemetry-operator
helm install opentelemetry-operator . -f examples/cutstom-autoGenerateCert-period/values.yaml
```
```yaml
manager:
  collectorImage:
    repository: "otel/opentelemetry-collector-k8s"

admissionWebhooks:
  certManager:
    enabled: false
  autoGenerateCert:
    enabled: true
    # If set to true, new webhook key/certificate is generated on helm upgrade.
    recreate: false
    # Cert period time in years. The default is 1(365 days).
    certPeriodDays: 200
```

3. get `tls.cert`
```bash
kubectl get secret opentelemetry-operator-controller-manager-service-cert -o jsonpath='{.data.tls\.crt}' | base64 --decode > tls.crt
```

4. use `openssl` to check the period time of tls.crt
```bash
openssl x509 -in tls.crt -noout -dates 
```

and I got this:
```txt
notBefore=Jul  9 02:57:00 2024 GMT
notAfter=Jan 25 02:57:00 2025 GMT # 👈 This is in line with my expectations
```


JaredTan95:
sad, https://github.com/open-telemetry/opentelemetry-helm-charts/actions/runs/9841098833/job/27166919581?pr=1249#step:5:200

Since each CI will re-generated `cutstom-autoGenerateCert-period` example, can not be consistent, remove `cutstom-autoGenerateCert-period`?

charts/opentelemetry-operator/values.schema.json
```diff
@@ -1571,11 +1571,20 @@
                             "examples": [
                                 true
                             ]
+                        },
+                        "certPeriodYear": {
```
jaronoff97, line 1575: why are we using years here if we're just going to convert this to days? IMO we should only use days

charts/opentelemetry-operator/examples/cutstom-autoGenerateCert-period/values.yaml
jaronoff97, line 1: the name of this example should be `custom-auto-generate-cert-period` IMO

jaronoff97: also, do we need to have a full example for setting this single variable?

JaredTan95: I think `examples/cutstom-autoGenerateCert-period/values.yaml ` content is the full example

charts/opentelemetry-operator/values.schema.json
jaronoff97: IMO we should stick with days as its more granular, ex: i have expirations set to 290 days for some of my company.

charts/opentelemetry-operator/examples/cutstom-autoGenerateCert-period/values.yaml
jaronoff97: yes, i'm not sure we need to make a whole new full example for this, i was mentioning that the naming with caps is a bit odd here too

charts/opentelemetry-operator/values.schema.json
JaredTan95: oh, day more flexible, I will update it.

JaredTan95:
> sad, https://github.com/open-telemetry/opentelemetry-helm-charts/actions/runs/9841098833/job/27166919581?pr=1249#step:5:200
> 
> Since each CI will re-generated `cutstom-autoGenerateCert-period` example, can not be consistent, remove `cutstom-autoGenerateCert-period`?

@jaronoff97 hi, any idea about this?

jaronoff97:
@JaredTan95 what do you get if you run the make generate locally?

JaredTan95:
every time run ``

> @JaredTan95 what do you get if you run the make generate locally?

Similarly, the certificate is regenerated each time with `CHARTS=opentelemetry-operator make generate-examples`

jaronoff97:
@JaredTan95 seems like we have failures [here ](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1251)too same thing.

jaronoff97:
ohhh i see the problem now... hmmm.....

TylerHelmuth:
Let's not add the new problematic example. The scenario is pretty straight forward (as in the values.yaml file isn't common or interesting) so I don't think we need a new example for it.

TylerHelmuth (approved):

PR #1264, opened 2024-07-18
Add Prometheus CRDs to otel-kube-stack

jaronoff97:
Adds the prometheus CRDs to the chart which will be used for the kube-prometheus-stack

PR #1265, opened 2024-07-18
Add KubeStateMetrics and PrometheusNodeExporter charts to OKS

jaronoff97:
To be merged after #1264

### DIFF
--- a/charts/opentelemetry-kube-stack/values.yaml
+++ b/charts/opentelemetry-kube-stack/values.yaml
@@ -891,3 +891,175 @@ opAMPBridge:
 
     # Rules for the bridge cluster role
     rules: []
+
+############################
+# Prometheus Configuration #
+# (optional)               #
+############################
+# This configuration sections allows for a direct replacement of the kube-prometheus-stack
+# chart where the collector scrapes the same metrics as the default prometheus installation.
+
+## Component scraping kube state metrics
+##
+kubeStateMetrics:
+  enabled: false
+
+## Configuration for kube-state-metrics subchart
+##
+kube-state-metrics:
+  namespaceOverride: ""
+  rbac:
+    create: true
+  releaseLabel: true
+  prometheus:
+    monitor:
+      enabled: true
+
+      ## Scrape interval. If not set, the Prometheus default scrape interval is used.
+      ##
+      interval: ""
+
+      ## SampleLimit defines per-scrape limit on number of scraped samples that will be accepted.
+      ##
+      sampleLimit: 0
+
+      ## TargetLimit defines a limit on the number of scraped targets that will be accepted.
+      ##
+      targetLimit: 0
+
+      ## Per-scrape limit on number of labels that will be accepted for a sample. Only valid in Prometheus versions 2.27.0 and newer.
+      ##
+      labelLimit: 0
+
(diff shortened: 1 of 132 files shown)


charts/opentelemetry-kube-stack/values.yaml
```diff
@@ -891,3 +891,175 @@ opAMPBridge:
...
+# Prometheus Configuration #
+# (optional)               #
+############################
+# This configuration sections allows for a direct replacement of the kube-prometheus-stack
+# chart where the collector scrapes the same metrics as the default prometheus installation.
+
+## Component scraping kube state metrics
+##
+kubeStateMetrics:
+  enabled: false
```
TylerHelmuth, line 905: Are we able to move `enabled` under `kube-state-metrics` or do they have a json schema that wont allow it?

charts/opentelemetry-kube-stack/values.yaml
```diff
@@ -891,3 +891,175 @@ opAMPBridge:
...
+      #   targetLabel: nodename
+      #   replacement: $1
+      #   action: replace
+
+  selfMonitor:
+    enabled: false
+
+## Deploy node exporter as a daemonset to all nodes
+##
+nodeExporter:
```
TylerHelmuth, line 978: What is the node exporter? Can you add more details in the comments

charts/opentelemetry-kube-stack/values.yaml
```diff
@@ -891,3 +891,175 @@ opAMPBridge:
...
+# This configuration sections allows for a direct replacement of the kube-prometheus-stack
+# chart where the collector scrapes the same metrics as the default prometheus installation.
+
+## Component scraping kube state metrics
+##
+kubeStateMetrics:
+  enabled: false
+
+## Configuration for kube-state-metrics subchart
+##
```
TylerHelmuth, line 908: Lets link to the kube state metrics chart.

charts/opentelemetry-kube-stack/values.yaml
TylerHelmuth: If this is for the `prometheus-node-exporter`, can the configs be combined?

TylerHelmuth (commented):
Lets link in the comments to any subchart's readmes.

charts/opentelemetry-kube-stack/values.yaml
jaronoff97: that may work, this is mostly for kube-prom-stack compatibility which matches this.

charts/opentelemetry-kube-stack/values.yaml
jaronoff97: same as above

charts/opentelemetry-kube-stack/values.yaml
TylerHelmuth: Like you could copy from your values.yaml into this one?

jaronoff97: [yes](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/values.yaml#L2024-L2027), i think there's some value in doing that for compatibility, but I think this is overall cleaner...

TylerHelmuth: Ok ya I'm fine with keeping it 100% compatible.

TylerHelmuth (approved):

PR #1266, opened 2024-07-19
[collector] Only use internalTrafficPolicy: if on Kubernetes vesrion 1.26 or later

Uncle-Justice:
following this [PR](https://github.com/grafana/alloy/pull/1213/files), we have to make limits based on k8s version. 

Otherwise we will fail to install collector when k8s version is low


Uncle-Justice:
sorry for CLA problem, i will create another one.

PR #1267, opened 2024-07-19
[collector] Only use internalTrafficPolicy: if on Kubernetes vesrion 1.26 or later

Uncle-Justice:
following this [PR](https://github.com/grafana/alloy/pull/1213/files), we have to make limits based on k8s version. 

Otherwise we will fail to install collector when k8s version is low

### DIFF
--- a/charts/opentelemetry-collector/templates/service.yaml
+++ b/charts/opentelemetry-collector/templates/service.yaml
@@ -35,7 +35,9 @@ spec:
   selector:
     {{- include "opentelemetry-collector.selectorLabels" . | nindent 4 }}
     {{- include "opentelemetry-collector.component" . | nindent 4 }}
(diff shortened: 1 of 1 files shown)


TylerHelmuth:
@Uncle-Justice you linked to a PR that is in a non-opentelemetry organization.  I'm not sure what the issue is that you're trying to fix.  We try not to add kubernetes-specific-version logic to the chart as it makes it a little harder to maintain and normally we can avoid it.  Can you please open an issue to describe your problem in more detail?

bixu:
@Uncle-Justice, was this change in reference to the K8s versioning mentioned [here](https://kubernetes.io/docs/concepts/services-networking/service-traffic-policy/#using-service-internal-traffic-policy)?

Issue #1160, opened 2024-04-30
make it possuble to just supply a full config


trondhindenes:
I'm finding it incredibly difficult to set up a custom collector config, because I need to disable the things I don't want. I can't just supply port values if I need non-standard ones either. 

Please consider making the config easier by providing an option to just supply a full config and list of ports

JaredTan95:
have you tried `configMap.existingName`?

TylerHelmuth:
You should be able to do

```yaml
ports:
  custom:
    enabled: true
    containerPort: 9999
    servicePort: 9999
    hostPort: 9999
    protocol: TCP
```

to add custom ports to both the container and service.  `ports` is a map so it will merge cleanly.

TylerHelmuth:
Can you be more specific about what is difficult to configure?  You can remove default components using helm's `null` feature such as 

```yaml
config:
  receivers:
    jaeger: null
    prometheus: null
    zipkin: null
  service:
    pipelines:
      traces:
        receivers:
          - otlp
      metrics: null
      logs: null
```

Full example: https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/examples/deployment-otlp-traces/values.yaml

boxidau:
+1 on this issue. Having to null out the default config requires merging the nulled structure into our config which seems a little messy. Changes to the default config would then be breaking changes on chart updates.

Having a second alternative key `customConfig` (or similar) in values where a user can define their entire config would be much easier to deal with in cases where a user has an already well formed config they'd like to use.

The other option of an externally managed ConfigMap is okay and gives the flexibility but loses the [auto-reload functionality on config updates](https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1194)

lerminou:
@TylerHelmuth , it seems to not work with null inside the config directly:
https://helm.sh/docs/chart_template_guide/values_files/#deleting-a-default-key

If I try on my own, following the 'otel-trace only' https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-collector/examples/deployment-otlp-traces

I tried:
```
config:
  receivers:
    jaeger: null
    prometheus: null
    zipkin: null
  service:
    pipelines:
      traces:
        receivers:
          - otlp
      metrics: null
      logs: null
```

or more specific:
```
    service:
      pipelines:
        traces:
          exporters:
            - otlp
          receivers:
[...]

PR #1301, opened 2024-08-09
[collector] Provide alternateConfig option for use in subcharts to workaround null issue

TylerHelmuth:
Since there is no fix for https://github.com/helm/helm/pull/12879 yet, this provides an alternate way to provide configuration with the collector.

This new field should really only be used when the collector is used as a subchart.

Closes https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1160


charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -1,6 +1,14 @@
+{{- define "opentelemetry-collector.var_dump" -}}
```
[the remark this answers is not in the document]
TylerHelmuth: Yup

charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -56,7 +60,7 @@ Build config file for deployment OpenTelemetry Collector
 {{- end }}
 
 {{- define "opentelemetry-collector.applyHostMetricsConfig" -}}
-{{- $config := mustMergeOverwrite (include "opentelemetry-collector.hostMetricsConfig" .Values | fromYaml) .config }}
+{{- $config := mustMergeOverwrite (dict "service" (dict "pipelines" (dict "metrics" (dict "receivers" list)))) (include "opentelemetry-collector.hostMetricsConfig" .Values | fromYaml) .config }}
```
jaronoff97, line 63: why did these need to change?

TylerHelmuth: Helm can't handle appending to empty lists.  This additional merge ensures the required list is present. 

This previously wasn't an issue bc the default config defined lists for each pipeline component type, but with the new alternative config, it would be possible to define a logs pipeline, only set the processors and exporters lists, and then enable the `logsCollection` preset.  In order for the preset to insert the filelogreceiver, it also needs to addd the list.

jaronoff97 (approved):

charts/opentelemetry-collector/values.yaml
```diff
@@ -169,6 +169,23 @@ config:
...
 
+# Helm currently has an issue (https://github.com/helm/helm/pull/12879) when using null to remove
+# default configuration from a subchart. The result is that you cannot remove default configuration
+# from `config`, such as a specific recevier or a specific pipeline, when the chart is used as a
+# subchart.
+#
+# Until the helm bug is fixed, this field is provided as an alternative when using this chart as a subchart.
+# It is not recommended to use this field when installing the chart directly.
+#
+# When not empty, `alternateConfig`will be used to set the collector's configuration. It has NO default
```
dmitryax, line 180:
```suggestion
# When not empty, `alternateConfig` will be used to set the collector's configuration. It has NO default
```

charts/opentelemetry-collector/values.yaml
```diff
@@ -169,6 +169,23 @@ config:
...
+# values and IS NOT MERGED with config. Any configuration provided via `config` will be ignored when
+# `alternateConfig` is set. You MUST provide your own collector configuration.
+#
+# Reminder that the healthcheck extension (or something else that provides the same functionality) is required.
+#
+# Components configured by presets will be injected in the same way they are for `config`.
+alternateConfig: {}
```
dmitryax, line 187: Maybe `configOverride`?

charts/opentelemetry-collector/templates/_config.tpl
dmitryax: Why don't we just leave the alternateConfig (or configOverride) as is and say that's the end result and no presets are applied? It'd make things easy for us and users I think.  Do you think it's pretty common that users would want their own config + presets?

charts/opentelemetry-collector/values.yaml
TylerHelmuth: I started with that name, but I didn't want anyone to think it was merging/overwriting values from .Values.config

charts/opentelemetry-collector/templates/_config.tpl
TylerHelmuth: I included the feature bc it is a common use case to enable presets when using the chart as a subchart.

dmitryax (approved):
LGTM

booyaa:
Good news, the upstream fix should land in Helm 3.17.1 🎉 https://github.com/helm/helm/pull/12879#issuecomment-2616724718

Issue #1194, opened 2024-05-21
[collector] No collector reload when using external configmap


boxidau:
Hello 👋 we're using the collector chart with an external config. This is being done in order to utilize templating for the config since the collector chart is invoked as a sub-chart. 

When specifying an externally managed ConfigMap to use as config for a collector, it would be great if the deployment checksum annotation could use a helm lookup function to read in the externally referenced configmap and generate a sha.

This way the deployment will rollout restart on deployment of a new config. In our specific case the externally managed config map is managed by the parent chart so will also cause a re-render of the deployment (and thus lookup) during helm upgrade.

JaredTan95:
it can be closed as it fixed by https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1302?

PR #1302, opened 2024-08-13
[collector] Fix: Enable external configmap checksum

bixu:
Fixes "[collector] No collector reload when using external configmap" (#1194)

This PR ships a new config option, `configMap.existingPath`, which is unset by default. When enabled, it takes a relative path to a custom ConfigMap that can be bundled in a wrapper chart. A new helper template calculates a checksum annotation from the custom ConfigMap file.


charts/opentelemetry-collector/templates/deployment.yaml
```diff
@@ -32,7 +32,11 @@ spec:
   template:
     metadata:
       annotations:
+        {{- if .Values.configMap.existingTemplateFile }}
+        checksum/config: {{ include (print $.Template.BasePath "/" .Values.configMap.existingTemplateFile ) . | sha256sum }}
+        {{- else -}}
         checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
+        {{- end }}
```
TylerHelmuth, line 39: Can you add a new template in _helpers instead to contain this logic? I expect we'd also need it for daemonset and statefulset.

bixu:
@TylerHelmuth, let me know if the latest changes look acceptable. Here's what local templating testing looked like:

```zsh
# Default annotation generation:
for mode in deployment statefulset daemonset; do helm template charts/opentelemetry-collector --set image.repository=foo/bar --set mode=$mode | grep -C4 checksum; done                                                                                                                          
    type: RollingUpdate
  template:
    metadata:
      annotations:
        checksum/config: 89aeb754fa82c7bf672b61bfdc214ca9b52e0f6b097051ad96bc02ac41a7f4c8
        
      labels:
        app.kubernetes.io/instance: release-name
    type: RollingUpdate
  template:
    metadata:
      annotations:
        checksum/config: 854a7b2382807eb76ebdd6803db814b4947bbcc94a00b6d4d549c648a60ae0ff
        
      labels:
        app.kubernetes.io/instance: release-name
    type: RollingUpdate
  template:
    metadata:
      annotations:
        checksum/config: a712714b718b8bef015d8a889e51ad4d9869e9dd096dbabd7dbf88477bcc316e
        
      labels:
        app.kubernetes.io/instance: release-name
        
# Custom annotation generation
for mode in deployment statefulset daemonset; do helm template charts/opentelemetry-collector --set image.repository=foo/bar --set mode=$mode --set configMap.create=false --set configMap.externalName=user-config --set configMap.Template=configmap-user-config.yaml  | grep -C4 checksum; done
    type: RollingUpdate
  template:
    metadata:
      annotations:
        checksum/config: 01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546b
        
      labels:
        app.kubernetes.io/instance: release-name
    type: RollingUpdate
  template:
    metadata:
      annotations:
        checksum/config: 01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546b
        
      labels:
        app.kubernetes.io/instance: release-name
    type: RollingUpdate
  template:
    metadata:
      annotations:
        checksum/config: 01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546b
        
      labels:
        app.kubernetes.io/instance: release-name
```

charts/opentelemetry-collector/values.yaml
```diff
@@ -83,6 +83,9 @@ configMap:
   # Specifies an existing ConfigMap to be mounted to the pod
   # The ConfigMap MUST include the collector configuration via a key named 'relay' or the collector will not start.
   existingName: ""
+  # Must be specified when configMap.create = false to ensure deployment restart when the configMap changes, using
+  # a checksum.config annotation.
+  configTemplate: ""
```
TylerHelmuth, line 88: What does this new field do? In your previous commit it looked like we could get away with an If-Else statement. I was hoping we could move that logic into the helper.

TylerHelmuth (commented):
It looks like the demo examples got removed.  If you only want to generate the collector examples you can do `make generate-examples CHARTS=opentlemetry-collector`.

charts/opentelemetry-collector/values.yaml
bixu: I was trying to be explicit since the name of the template file might not match the name of the ConfigMap in k8s.

But if we prefer to be implicit to reduce the amount of config, I'm okay with that. I'll try to re-use the `existingName` so we can see how that feels.

bixu: Is [this](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1302/commits/9d65a686207b9a2bf021a1ce689ff180d79c769f) closer to what we want?

bixu:
> It looks like the demo examples got removed. If you only want to generate the collector examples you can do `make generate-examples CHARTS=opentlemetry-collector`.

I've [re-run](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1302/commits/9ba9f66ebe5424c380e5f94b1789966846cc4d43) the example generator for the Collector after reverting the previous generation (which seemed to be the cause of deleted files). The diff is much smaller now.

I'm not sure how to best handle merge conflicts from upstream around generated examples, though -- tried a few merge flows locally but was not happy with the results.

TylerHelmuth:
@bixu the best way to handle merging the examples is to take mains, bump the chart version to the appropriate value, and regenerate.

charts/opentelemetry-collector/templates/_helpers.tpl
```diff
@@ -226,3 +226,19 @@ Get ConfigMap name if existingName is defined, otherwise use default name for ge
   {{- end -}}
 {{- end }}
 
+{{/*
+Create ConfigMap checksum annotation if configMap.configTemplate is defined, otherwise use default file for generated config.
+*/}}
+{{- define "opentelemetry-collector.configTemplateChecksumAnnotation" -}}
+  {{- if .Values.configMap.existingName -}}
+  checksum/config: {{ include (print $.Template.BasePath "/" .Values.configMap.existingName ".yaml" ) . | sha256sum }}
```
TylerHelmuth, line 234: If a user has `.Values.configMap.existingName = my-favorite-config.yaml` will an extra `.yaml` be added?

charts/opentelemetry-collector/values.yaml
```diff
@@ -82,8 +82,12 @@ configMap:
   create: true
   # Specifies an existing ConfigMap to be mounted to the pod
   # The ConfigMap MUST include the collector configuration via a key named 'relay' or the collector will not start.
+  # If bundling this external ConfigMap via a wrapper chart, the name of the
+  # ConfigMap file must be in the form '<existingName_value>.yaml' to allow
+  # the chart to calculate a checksum annotation.
```
TylerHelmuth, line 87: What happens if the form is wrong, will the chart fail to install?

charts/opentelemetry-collector/templates/_helpers.tpl
bixu: Yes. I just realized it will work better to just `include` the external template and pass to `cksum` here.

charts/opentelemetry-collector/values.yaml
bixu: It doesn't look like it's possible to feed template contents to Helm except by reading a file, so I [brought back](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1302/commits/15445ca23b07ca198b20f988b24f6c60253107e1) the config key to set the file path.

charts/opentelemetry-collector/values.schema.json
```diff
@@ -135,6 +135,10 @@
         "existingName": {
           "description": "Specifies an existing configMap to be mounted to the pod",
           "type": "string"
+        },
+        "configTemplate": {
```
TylerHelmuth, line 139: This should be renamed to `existingPath`

charts/opentelemetry-collector/values.yaml
```diff
@@ -83,6 +83,8 @@ configMap:
   # Specifies an existing ConfigMap to be mounted to the pod
   # The ConfigMap MUST include the collector configuration via a key named 'relay' or the collector will not start.
   existingName: ""
+  # Specifies relative path to the existing ConfigMap template file
```
TylerHelmuth, line 86: Let's explain that a user should set this value when they want the collector to change bc the configmap changed.

charts/opentelemetry-collector/values.schema.json
bixu: Good catch.

charts/opentelemetry-collector/values.yaml
bixu: Closing the loop on your original question, we're protecting the user from a failed install by avoiding setting the new option in the default `values.yaml`. A user will have to explicitly enable this key to trigger checksumming of a custom template file.

charts/opentelemetry-collector/values.yaml
```diff
@@ -83,6 +83,9 @@ configMap:
   # Specifies an existing ConfigMap to be mounted to the pod
   # The ConfigMap MUST include the collector configuration via a key named 'relay' or the collector will not start.
   existingName: ""
+  # Specifies the relative path to custom ConfigMap template file. This option MUST be used when bundling a custom
+  # ConfigMap template, as it enables pod restart via a template checksum annotation.
```
TylerHelmuth, line 87:
```suggestion
  # Specifies the relative path to custom ConfigMap template file. This option SHOULD be used when bundling a custom
  # ConfigMap template, as it enables pod restart via a template checksum annotation.
```

TylerHelmuth: I view this as a SHOULD because you can install the chart without this field set. You'll miss out on the pod restart feature, but it is still possible to use the chart.

bixu: Good call.

TylerHelmuth (approved):

jaronoff97 (approved):

onematchfox:
Hey @bixu, 

Are you actively using this feature and if so, how? I am trying to deploy this chart as a subchart whilst using `existingPath` but can't seem to find a working configuration and keep getting an error like:

```
foo/charts/opentelemetry-collector/templates/_helpers.tpl:234:22: executing "opentelemetry-collector.configTemplateChecksumAnnotation" at <include (print $.Template.BasePath "/" .Values.configMap.existingPath) .>: error calling include: template: no template "foo/charts/opentelemetry-collector/templates/../../../../otel/config.yaml" associated with template "gotpl"
``` 

According to the [docs](https://helm.sh/docs/chart_template_guide/accessing_files/) 

> Files outside of a Helm application [subchart](https://helm.sh/docs/chart_template_guide/subcharts_and_globals/), including those of the parent, cannot be accessed

so it doesn't look like what I'm trying to do is possible. 

I noticed that in the description you mentioned "can be bundled in a wrapper chart". What exactly is a "wrapper chart". Perhaps I'm approaching this wrong?

bixu:
> Hey @bixu,
> 
> Are you actively using this feature and if so, how? I am trying to deploy this chart as a subchart whilst using `existingPath` but can't seem to find a working configuration and keep getting an error like:
> 
> ```
> foo/charts/opentelemetry-collector/templates/_helpers.tpl:234:22: executing "opentelemetry-collector.configTemplateChecksumAnnotation" at <include (print $.Template.BasePath "/" .Values.configMap.existingPath) .>: error calling include: template: no template "foo/charts/opentelemetry-collector/templates/../../../../otel/config.yaml" associated with template "gotpl"
> ```
> 
> According to the [docs](https://helm.sh/docs/chart_template_guide/accessing_files/)
> 
> > Files outside of a Helm application [subchart](https://helm.sh/docs/chart_template_guide/subcharts_and_globals/), including those of the parent, cannot be accessed
> 
> so it doesn't look like what I'm trying to do is possible.
> 
> I noticed that in the description you mentioned "can be bundled in a wrapper chart". What exactly is a "wrapper chart". Perhaps I'm approaching this wrong?

@onematchfox, I actually ran in to the same issue the other day, but hadn't gotten around to thinking about a fix. I'm afraid this config option is broken and I'd like to remove it, or do you have a better idea?

onematchfox:
> or do you have a better idea?

Nope, sorry, I don't. I was hoping that I was just approaching it wrong.

bixu:
> > or do you have a better idea?
> 
> Nope, sorry, I don't. I was hoping that I was just approaching it wrong.

@onematchfox, are you in the CNCF slack for the Collector? I'm `@bixu` [there](https://cloud-native.slack.com/archives/C01N6P7KR6W); would love to chat more about your use-case and what our options are.

onematchfox:
> @onematchfox, are you in the CNCF slack for the Collector? I'm `@bixu` [there](https://cloud-native.slack.com/archives/C01N6P7KR6W); would love to chat more about your use-case and what our options are.

I am. Will drop you a msg a bit later this afternoon when I have some spare time.

charts/opentelemetry-collector/templates/_helpers.tpl
```diff
@@ -226,3 +226,19 @@ Get ConfigMap name if existingName is defined, otherwise use default name for ge
   {{- end -}}
 {{- end }}
 
+{{/*
+Create ConfigMap checksum annotation if configMap.existingPath is defined, otherwise use default templates
+*/}}
+{{- define "opentelemetry-collector.configTemplateChecksumAnnotation" -}}
+  {{- if .Values.configMap.existingPath -}}
```
csapinoso, line 233: This feature doesn't work, the path will never resolve because the custom template will not exist in the under the base path of the subchart

PR #1305, opened 2024-08-13
Add uninstall hook for the kube-stack chart

jaronoff97:
This adds an uninstall hook which makes sure that we get rid of the CRs before we uninstall the operator

### DIFF
--- a/charts/opentelemetry-kube-stack/templates/hooks.yaml
+++ b/charts/opentelemetry-kube-stack/templates/hooks.yaml
@@ -1,0 +1,56 @@
+{{- if .Values.cleanupJob.enabled }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: delete-resources-sa
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: delete-resources-role
+rules:
+  - apiGroups:
+      - opentelemetry.io
+    resources:
+      - instrumentations
+      - opampbridges
+      - opentelemetrycollectors
+    verbs:
+      - get
+      - list
+      - delete
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: delete-resources-rolebinding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: delete-resources-role
+subjects:
+  - kind: ServiceAccount
+    name: delete-resources-sa
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ include "opentelemetry-kube-stack.name" . }}-pre-delete-job
+  annotations:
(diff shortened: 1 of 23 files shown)


charts/opentelemetry-kube-stack/templates/hooks.yaml
```diff
@@ -0,0 +1,56 @@
+{{- if .Values.cleanupJob.enabled }}
+apiVersion: v1
+kind: ServiceAccount
```
TylerHelmuth, line 3: Do the `ServiceAccount`, `Role`, and `RoleBinding` need something to make sure they get cleaned up on `helm uninstalled` as well?

charts/opentelemetry-kube-stack/values.yaml
```diff
@@ -9,6 +9,10 @@ clusterName: ""
 # Extra environment variables to add to each collector, bridge and instrumentation
 extraEnvs: []
 
+# Enables a cleanup job to make sure the CRs are uninstalled before the operator
+cleanupJob:
+  enabled: true
```
TylerHelmuth, line 14: Let's add a comment that enabling this is necessary for `helm uninstall` to work.  Users should really never be disabling this.

TylerHelmuth (commented):
[I dont love that the hook's resources stick around after the helm chart is uninstalled.](https://helm.sh/docs/topics/charts_hooks/#hook-resources-are-not-managed-with-corresponding-releases)

charts/opentelemetry-kube-stack/templates/hooks.yaml
jaronoff97: argh yeah they probably need more hooks 🙃

jaronoff97: I hate helm hooks.

TylerHelmuth:
@jaronoff97 can we reuse the prom live stack charts solution, whatever it is?

.github/workflows/kube-stack-test.yaml
```diff
@@ -28,5 +28,7 @@ jobs:
           kubectl wait --timeout=5m --for=condition=available deployment cert-manager -n cert-manager
           kubectl wait --timeout=5m --for=condition=available deployment cert-manager-webhook -n cert-manager
 
-      - name: Run chart-testing (install)
-        run: ct install --charts charts/opentelemetry-kube-stack
+      - name: Run Helm Install
+        run: |
+          helm dep build charts/opentelemetry-kube-stack
+          helm install opentelemetry-kube-stack charts/opentelemetry-kube-stack
```
TylerHelmuth, line 34: This is fine for now, but once we add a `ci` folder we'll want to go back to `ct install`.

TylerHelmuth: Ya we'll have to solve that later somehow

TylerHelmuth (approved):

Issue #964, opened 2023-11-20
Add CI scenarios for eBPF Chart


TylerHelmuth:
Currently the ebpf chart has no CI tests.  We should add a test for the default values.yaml at least, but it would be best if we could have more comprehensive coverage.

hyfj44255:
I would like to have a try if this is still needed, thanks!

nicolastakashi:
> Hi @TylerHelmuth I would like to have a try if this is still needed, thanks!

Go for it @hyfj44255 🙏

nicolastakashi:
Hey @hyfj44255, yes exactly like that

hyfj44255:
Hi @nicolastakashi I created a pr https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1318 would it be ok if you take a look when you got chance? thanks

vinod827:
This issue should be closed since its [PR](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1310) getting merged

marcschaeferger:
Can be closed, see above comment. Closed by #1310

jaronoff97:
Closed by #1310

PR #1310, opened 2024-08-20
test: Add CI scenarios for eBPF Chart. #964

hyfj44255:
# Description
Currently the ebpf chart has no CI tests. we should add a test for the default values.yaml at least, but it would be best if we could have more comprehensive coverage.

Fixes #(issue)
https://github.com/open-telemetry/opentelemetry-helm-charts/issues/964

## Type of change
- [X] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?
Go to Contrib repo root directory.
```
cd opentelemetry-helm-charts
```

Create a kind cluster same as the one in repo's Github action
```
kind create cluster --name ebpf-cluster --config ./.github/kind-1.24.yaml

```
Run CT to preform the test
```
ct install --charts charts/opentelemetry-ebpf
```


charts/opentelemetry-ebpf/ci/enabled-log-debug-values.yaml
```diff
@@ -0,0 +1,13 @@
...
+
+# enable --log-console
+log:
+  console: true
+  level: trace
+
+debug:
+  enabled: true
+  storeMinidump: true
+  sendUnplannedExitMetric: true
```
TylerHelmuth, line 13:
```suggestion
  sendUnplannedExitMetric: true

```

charts/opentelemetry-ebpf/ci/requests-limits-values.yaml
```diff
@@ -0,0 +1,63 @@
...
+        memory: "256Mi"
+        cpu: "300m"
+  watcher:
+    resources:
+      requests:
+        memory: "128Mi"
+        cpu: "200m"
+      limits:
+        memory: "256Mi"
+        cpu: "300m"
```
TylerHelmuth, line 63:
```suggestion
        cpu: "300m"

```

charts/opentelemetry-ebpf/templates/k8s-collector-deployment.yaml
```diff
@@ -119,7 +119,7 @@ spec:
           - name: "EBPF_NET_CRASH_METRIC_HOST"
             value: {{ .Values.endpoint.address }}
           - name: "EBPF_NET_CRASH_METRIC_PORT"
-            value: {{ .Values.endpoint.port }}
```
nicolastakashi, line 122: Let's take this changes to another PR wdyt?

hyfj44255: Thanks for review,I'd agree let me move this out, one pr should do one thing

hyfj44255:
Hi @nicolastakashi I've updated the pr according to you suggestion would it be ok if you take a look again?

TylerHelmuth (approved):

PR #1334, opened 2024-09-06
Update examples for v1beta1

### LINKED ISSUE #1333: opentelemetry.io/v1alpha1 is deprecated, yet still appears in operator chart readme
The [README for the operator helm chart](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/README.md) shows examples of installing the collector. All of these examples use `opentelemetry.io/v1alpha1` despite the fact that this version is deprecated. I believe `v1beta1` is the preferred version, so these examples should be updated to use `opentelemetry.io/v1beta1`.

ethanmdavidson:
fixes #1333

I believe the main change is the use of [structured config](https://github.com/open-telemetry/opentelemetry-operator/blob/main/docs/crd-changelog.md#structured-configuration). I've updated the readme based on that upgrade example, but haven't actually applied the updated examples to verify that they are valid.

### DIFF
--- a/charts/opentelemetry-operator/Chart.yaml
+++ b/charts/opentelemetry-operator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.68.2
+version: 0.69.0
 description: OpenTelemetry Operator Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-operator/Chart.yaml
+++ b/charts/opentelemetry-operator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.69.0
+version: 0.68.3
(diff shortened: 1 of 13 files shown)


jaronoff97 (commented):
this LGTM! would you mind bumping the minor version for the operator and re-running `make generate-examples CHARTS=opentelemetry-operator`?

ethanmdavidson:
@jaronoff97 not sure if I understood you correctly, but I think I've done what you asked.

charts/opentelemetry-operator/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.68.2
```
jaronoff97, line 3: only the minor version here `0.68.3`

jaronoff97 (approved):

TylerHelmuth (approved):

PR #1337, opened 2024-09-09
move prometheus exporter config to use readers

codeboten:
This moves the configuration to use metric readers configuration instead of the deprecated `address` configuration.


charts/opentelemetry-collector/values.yaml
```diff
@@ -137,7 +137,12 @@ config:
   service:
     telemetry:
       metrics:
-        address: ${env:MY_POD_IP}:8888
+        readers:
+          - pull:
+              exporter:
+                prometheus:
+                  host: ${env:MY_POD_IP}
+                  port: 8888
```
TylerHelmuth, line 145: This will be equivalent for anyone who is not configuring `config.service.telemetry.metrics.address` but what will happen if someone has done something like `--set config.service.telemetry.metrics.address=${env:MY_POD_IP}:9090`?

codeboten: there would be 2 prometheus ports exposed in that case

TylerHelmuth: We should introduce a template in `config` that will build the reader yaml from `.Values.config.service.telemetry.metrics.address`.

codeboten:
Can someone reopen this please 😬

Issue #1152, opened 2024-04-19
[opentelemetry-collector] Allow to configure startup probe


ElfoLiNk:
Due to compaction on startup when using a persisted queue with file storage, the liveness probe then terminates the collector, causing it to enter a CrashLoopBackoff state in this scenario.

Allowing the setup of a proper startup probe will prevent the need for significant changes to the liveness probe or its disabling.

https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/32456

lindeskar:
Created #1351 to add `startupProbe` to the collector Pods.

PR #1351, opened 2024-09-18
feat: add startupProbe in collector pods

lindeskar:
## Description

Add `startupProbe` in opentelemetry-collector Pods.

It's basically a copy-paste of stuff from `livenessProbe`.

Fixes: #1152

--

## Background

We were hit by the same issue as #1152; long startup times due to compaction - leading to OTelCol Pods being killed before compaction was completed - leading to another round of the same.

Adding a `startupProbe` with a higher `failureThreshold` should be a nice fix.

> _Sometimes, you have to deal with applications that require additional startup time on their first initialization. In such cases, it can be tricky to set up liveness probe parameters without compromising the fast response to deadlocks that motivated such a probe. The solution is to set up a startup probe with the same command, HTTP or TCP check, with a `failureThreshold * periodSeconds` long enough to cover the worst case startup time._
From: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes

### DIFF
--- a/charts/opentelemetry-collector/templates/_pod.tpl
+++ b/charts/opentelemetry-collector/templates/_pod.tpl
@@ -110,6 +110,25 @@ containers:
       httpGet:
         path: {{ .Values.readinessProbe.httpGet.path }}
         port: {{ .Values.readinessProbe.httpGet.port }}
+    startupProbe:
+      {{- if .Values.startupProbe.initialDelaySeconds | empty | not }}
+      initialDelaySeconds: {{ .Values.startupProbe.initialDelaySeconds }}
+      {{- end }}
+      {{- if .Values.startupProbe.periodSeconds | empty | not }}
+      periodSeconds: {{ .Values.startupProbe.periodSeconds }}
+      {{- end }}
+      {{- if .Values.startupProbe.timeoutSeconds | empty | not }}
+      timeoutSeconds: {{ .Values.startupProbe.timeoutSeconds }}
+      {{- end }}
+      {{- if .Values.startupProbe.failureThreshold | empty | not }}
+      failureThreshold: {{ .Values.startupProbe.failureThreshold }}
+      {{- end }}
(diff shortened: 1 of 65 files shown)


charts/opentelemetry-collector/templates/_pod.tpl
```diff
@@ -110,6 +110,25 @@ containers:
       httpGet:
         path: {{ .Values.readinessProbe.httpGet.path }}
         port: {{ .Values.readinessProbe.httpGet.port }}
+    startupProbe:
```
TylerHelmuth, line 113: To make this an addative feature, lets only set `startupProbe` if `.Values.startupProbe` is set and lets update the values.yaml to not set `startupProbe` by default.

lindeskar: Done!

Do you want me to add `startupProbe` values to an existing / new example?

charts/opentelemetry-collector/values.yaml
```diff
@@ -442,6 +442,25 @@ readinessProbe:
     port: 13133
     path: /
 
+# startup probe configuration
+# Ref: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/
+##
+startupProbe:
+  {}
```
lindeskar, line 449: No idea why the formatter wants to put this on a new line. Only happens when the commented text is below..

charts/opentelemetry-collector/templates/_pod.tpl
TylerHelmuth: No, it is simple enough that we don't need to.

charts/opentelemetry-collector/values.yaml
TylerHelmuth, line 449:
```suggestion
startupProbe: {}
```

TylerHelmuth (approved):

Issue #1352, opened 2024-09-20
[opentelemetry-collector] does not add "file_storage" in service.extensions when storeCheckpoints is true


oshamin:
version opentelemetry-collector helm - 0.106.0
version opentelemetry-collector-contrib - 0.109.0

Error when starting the container:

```
2024-09-20T08:36:38.451Z	info	service@v0.109.0/service.go:290	Shutdown complete.
Error: cannot start pipelines: storage client: storage extension 'file_storage' not found; failed to shutdown pipelines: no existing monitoring routine is running
2024/09/20 08:36:38 collector server run finished with error: cannot start pipelines: storage client: storage extension 'file_storage' not found; failed to shutdown pipelines: no existing monitoring routine is running
```

Yaml value file to test:

```
mode: "daemonset"

presets:
  logsCollection:
    enabled: true
    storeCheckpoints: true

image:
  repository: "otel/opentelemetry-collector-k8s"

command:
  name: "otelcol-k8s"
```

The output ConfigMap does not contains **file_storage** reference in the service.extensions section for that reason otel-collector exists with the aforementioned error.

ChrsMark:
I could reproduce it. Thank's for reporting this @oshamin!

@open-telemetry/helm-approvers could it be that the issue was introduced with https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1301?

ChrsMark:
So I did some digging into it and it seems that because we check at https://github.com/open-telemetry/opentelemetry-helm-charts/blob/a927332fab6ba2ea679dfc00a9e84a931c7fd4ee/charts/opentelemetry-collector/templates/NOTES.txt#L22-L25 for the `health_check` extension to always be present we don't really actually need that line at at https://github.com/open-telemetry/opentelemetry-helm-charts/blob/b8d11b63916bc40cd1aa2875a880f3995978d434/charts/opentelemetry-collector/templates/_config.tpl#L148.

However, if we still want to keep it we need to change the name of the internal config variable to sth else during the initialization.

PR #1356, opened 2024-09-23
[collector] Fix how extensions list is created

ChrsMark:
As described at https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1352#issuecomment-2368190174 the initialization of the `service.extensions` list needs to use a different variable name than the top level one. Otherwise the internal only variable will be extended with no actual impact outside of its block.

@TylerHelmuth please verify this since you worked on https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1301 initially. 

Fixes https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1352

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.106.0
+version: 0.106.1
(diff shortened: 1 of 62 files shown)


TylerHelmuth (approved):
I verified this works locally with alternateConfig and `.Values.logsCollection.storeCheckpoints`

PR #1364, opened 2024-10-02
feat[opentelemetry-collector]: add support for multiline recombine from container parser

paulojmdias:
The current `container` operator cannot handle multiline messages like Java stack traces.

This PR adds the needed recombine operator to make it work. Probably in the future, this makes sense to be moved into the container operator itself. 

```
2024-10-02 13:23:27.062 ERROR a.b.c.SolrConnection.dummyFunction ZURL:zookeeper:2181   Exception deleting users for team[123456789] org.apache.solr.common.SolrException: Could not find a healthy node to handle the request.
	at org.apache.solr.client.solrj.impl.CloudSolrClient.sendRequest(CloudSolrClient.java:1085)
	at org.apache.solr.client.solrj.impl.CloudSolrClient.requestWithRetryOnStaleState(CloudSolrClient.java:871)
	at org.apache.solr.client.solrj.impl.CloudSolrClient.requestWithRetryOnStaleState(CloudSolrClient.java:954)
	at org.apache.solr.client.solrj.impl.CloudSolrClient.requestWithRetryOnStaleState(CloudSolrClient.java:954)
	at org.apache.solr.client.solrj.impl.CloudSolrClient.requestWithRetryOnStaleState(CloudSolrClient.java:954)
	at org.apache.solr.client.solrj.impl.CloudSolrClient.requestWithRetryOnStaleState(CloudSolrClient.java:954)
	at org.apache.solr.client.solrj.impl.CloudSolrClient.requestWithRetryOnStaleState(CloudSolrClient.java:954)
	at org.apache.solr.client.solrj.impl.CloudSolrClient.request(CloudSolrClient.java:807)
	at org.apache.solr.client.solrj.SolrRequest.process(SolrRequest.java:150)
	at org.apache.solr.client.solrj.SolrRequest.process(SolrRequest.java:167)
...
```

### DIFF
--- a/charts/opentelemetry-collector/templates/_config.tpl
+++ b/charts/opentelemetry-collector/templates/_config.tpl
@@ -179,6 +179,13 @@ receivers:
       - type: container
         id: container-parser
         max_log_size: {{ $.Values.presets.logsCollection.maxRecombineLogSize }}
+      # combine multiline logs (e.g. stack traces)
+      - id: recombine-multiline
+        type: recombine
+        combine_field: body
+        is_first_entry: body matches "^[^\\s]"
+        source_identifier: attributes["log.file.path"]
+        max_log_size: {{ $.Values.presets.logsCollection.maxRecombineLogSize }}
 {{- end }}
 
 {{- define "opentelemetry-collector.applyKubernetesAttributesConfig" -}}
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.107.0
+version: 0.108.0
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-collector/examples/daemonset-collector-logs/rendered/configmap-agent.yaml
+++ b/charts/opentelemetry-collector/examples/daemonset-collector-logs/rendered/configmap-agent.yaml
@@ -11,7 +11,7 @@ metadata:
     app.kubernetes.io/instance: example
     app.kubernetes.io/version: "0.110.0"
     app.kubernetes.io/managed-by: Helm
-    
+
 data:
   relay: |
     exporters:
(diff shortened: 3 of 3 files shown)


charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -179,6 +179,13 @@ receivers:
       - type: container
         id: container-parser
         max_log_size: {{ $.Values.presets.logsCollection.maxRecombineLogSize }}
+      # combine multiline logs (e.g. stack traces)
+      - id: recombine-multiline
+        type: recombine
+        combine_field: body
+        is_first_entry: body matches "^[^\\s]"
+        source_identifier: attributes["log.file.path"]
+        max_log_size: {{ $.Values.presets.logsCollection.maxRecombineLogSize }}
```
TylerHelmuth, line 188: What happens to logs that don't need recombined? Does this add performance overhead?

paulojmdias: Regarding resources usage, we don't see any performance penalty. Below you can see the average CPU and memory usage before and after this add.

Before: 
image

image

After:

image

image

Is there any other metric that we can take an eye?

Does it make sense to add this using a feature toggle on values.yaml instead of force use it?

TylerHelmuth: Since this is a preset config, I'm trying to understand if `recombine` is universally helpful and wont impact any users who don't need it.  

If it negatively affects users who don't need to recombine logs, then it would be better for you to configure the filelog's operators manually instead of including it automatically in the preset's config.

paulojmdias: Technically isn't seems so, but I understand your point.

I still see value on this recombine or dynamically support custom operator additions to the Kubernetes filelog receiver without copying paste everything and lose future updates.

Let me also give you some possibilities to think about it:
- Add a feature toggle on values to add this block on this preset config
- Add a new hash on values with this new value as default to be used and anyone can make it empty if they want to disable
- Add a new hash on values to support custom operators for this Kubernetes filelog receiver with an empty value as default

Let me know what so you think about it

PR #1367, opened 2024-10-04
[demo] - update to 1.12 release including new flagd UI

klucsik:
For deploying https://github.com/open-telemetry/opentelemetry-demo/pull/1725.
The flagd-ui needs to be in the smae container with flagd so they can share an emptydir volume which containes the jsonfile for flags. flgad-ui can write it, and flagd will pick it up instanly and distributes the new flag changes.
I'm not very proficient with helm, please provide some input if this approach is fine or we can do this more neatly.
I added just a object block for the schema, would be neater to validate the entire inner block, but this is well beyond my jsonschema understanding 😃


puckpuck:
@klucsik I updated your branch to work in a way that feels more natural with the rest of how the demo is setup.

This changes to using an array for the `sidecarContainers` property. It also adds a new property called `additionalVolumes` at the component level, which can be used to define volumes referenced from any of the pod's container volume mounts. I also removed some schema restrictions around defining both init and sidecar containers.

charts/opentelemetry-demo/values.yaml
```diff
@@ -586,40 +590,38 @@ components:
       - "file:./etc/flagd/demo.flagd.json"
     mountedEmptyDirs:
       - name: config-rw
-        mountPath: /etcd/flagd
```
klucsik: whoops, nice catch! thanks

puckpuck:
I can install the chart in this PR. I can also run `ct` against it locally on my system and it works just fine, but our CI tests are failing. I could use some help to understand why our CI fails in GHA but not locally.

The differences I can see are:

- helm version (3.16.2 locally vs 3.9.4 in GHA)
- ct version (3.11.0 locally vs 3.10.1 in GHA)
- environment (MacOS locally vs Linux in GHA)

TylerHelmuth:
I can help investigate this week

TylerHelmuth:
I believe the issue is using `tpl` inside `range`: https://github.com/helm/helm/issues/5979.  @puckpuck @klucsik can you update the `Setup` step in `.github/workflows/lint.yaml` and `.github/workflows/demo-test.yaml` to use helm `3.14.4`

.github/actions/setup/action.yaml
```diff
@@ -8,7 +8,7 @@ inputs:
   helm-version:
     description: "The helm version to be used when setting up"
     required: false
-    default: "v3.9.4"
+    default: "v3.14.4"
```
TylerHelmuth, line 11: I'd prefer not to update the default for all charts since the others can still use 3.9 (and I know people are still using it).  Can you since update the test demo yaml to set this value?

puckpuck: I changed this to set the version in demo-test.yaml instead, and reverted to 3.9.4 in the shared setup action.

TylerHelmuth (approved):

TylerHelmuth:
@puckpuck I'll wait to merge until you approve

puckpuck (approved):

PR #1382, opened 2024-10-14
[opentelemetry-kube-stack] Adding additional labels 

suresh-target:
This is to enable the option to add labels to the components such as servicemonitor.


jaronoff97:
Right now it looks like you're only changing the pattern for the labels, i'm not exactly sure how this is solving the issue... would you mind explaining what I'm missing here? Thank you!

suresh-target:
> Right now it looks like you're only changing the pattern for the labels, i'm not exactly sure how this is solving the issue... would you mind explaining what I'm missing here? Thank you!

@jaronoff97 My plan to is to use separate collector for the controlplane metrics, for that I need to label serviceMonitor so that I would be able use that label in TA to filter the serviceMonitor lookup. This PR is specifically to fix the following error that I get while adding the label to servieMonitor components.

```
- kubeEtcd.serviceMonitor.additionalLabels: Additional property component is not allowed
- kubeApiServer.serviceMonitor.additionalLabels: Additional property component is not allowed
- kubeControllerManager.serviceMonitor.additionalLabels: Additional property component is not allowed
- kubeScheduler.serviceMonitor.additionalLabels: Additional property component is not allowed

```

suresh-target:
> Would this also be resolved by setting additional properties to be true? Also, can you please bump the patch version to 0.3.1 and re-run `make generate-examples`?

@jaronoff97 Yes, that is also possible.. I will modify the change to  update the additional properties to true

charts/opentelemetry-kube-stack/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-kube-stack
-version: 0.3.0
+version: 0.3.1
```
jaronoff97, line 3: it looks like the linter is still failing because of some whitespacing issues? Did your linter/formatter automatically remove extra spacing potentially?

jaronoff97: if not, i can try running this locally and see what happens.

suresh-target: @jaronoff97  I'm closing this PR and created another one https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1388

Issue #1372, opened 2024-10-07
Missing controlplane metrics: opentelemetry-kube-stack-0.2.1


suresh-target:
It is observed that the metrics of controlplane components such as  apiserver, etcd are missing when we use the TargetAllocator with the help of ServiceMonitor CRDs 
 
 **helm char**t - opentelemetry-kube-stack-0.2.1

values.yaml configuration

```
kubernetesServiceMonitors:
  enabled: true
kubeApiServer:
  enabled: true
kubeEtcd:
  enabled: true


```
Looks like we may need a separate TargetAllocator instance to scrape controlplane metrics as the per-node strategy currently ignores targets without a Node, like control plane components.

Notes: 

1. serviceMonitor.additionalLabels also need to be enabled to add labels to ServiceMonitor instances so that TA instance can use them to filter a specific set of Servicemonitors.
2. This was tested by adding a missing label manually to both apiserver and etcd services to overcome the issue mentioned [here](https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1371)

**Additional observations**

The following curl command can be executed to confirm that it returns empty target list 

```
# curl http://opentelemetry-kube-stack-daemon-targetallocator.kube-system.svc.cluster.local/jobs/serviceMonitor%2Fkube-system%2Fopentelemetry-kube-stack-kube-etcd%2F0/targets

{"opentelemetry-kube-stack-daemon-collector-dpst2":{"_link":"/jobs/serviceMonitor%2Fkube-system%2Fopentelemetry-kube-stack-kube-etcd%2F0/targets?collector_id=opentelemetry-kube-stack-daemon-collector-dpst2","targets":[]},"opentelemetry-kube-stack-daemon-collector-d7qsq":{"_link":"/jobs/serviceMonitor%2Fkube-system%2Fopentelemetry-kube-stack-kube-etcd%2F0/targets?collector_id=opentelemetry-kube-stack-daemon-collector-d7qsq","targets":[]},"opentelemetry-kube-stack-daemon-collector-pm5qq":{"_link":"/jobs/serviceMonitor%2Fkube-system%2Fopentelemetry-kube-stack-kube-etcd%2F0/targets?collector_id=opentelemetry-kube-stack-daemon-collector-pm5qq","targets":[]}}

```

jaronoff97:
[...]

PR #1388, opened 2024-10-17
[opentelemetry-kube-stack] Adding additional labels

suresh-target:
This is to enable the option to add labels to the components such as servicemonitor.

While using a separate collector for the controlplane metrics, we need to add a label to serviceMonitors that we are planning to lookup separately (Later TA can us this label to filter the serviceMonitor lookup). This PR is specifically to fix the following error that we get while adding the label to serviceMonitor components.

```
- kubeEtcd.serviceMonitor.additionalLabels: Additional property component is not allowed
- kubeApiServer.serviceMonitor.additionalLabels: Additional property component is not allowed
- kubeControllerManager.serviceMonitor.additionalLabels: Additional property component is not allowed
- kubeScheduler.serviceMonitor.additionalLabels: Additional property component is not allowed

```
Refer the following issue for the related discussions, https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1372

### DIFF
--- a/charts/opentelemetry-kube-stack/Chart.yaml
+++ b/charts/opentelemetry-kube-stack/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-kube-stack
-version: 0.3.0
+version: 0.3.1
 description: |
   OpenTelemetry Quickstart chart for Kubernetes.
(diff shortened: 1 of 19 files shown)


suresh-target:
@jaronoff97 -  Please review

jaronoff97 (approved):

PR #1391, opened 2024-10-17
[opentelemetry-operator] Fix Collector CRD referencing disabled webhook

ElderMael:
When setting the value of `admissionWebhooks.create` to `false`, the webhook service is not created. This is fine but the Collector CRD still references this service in the `spec.conversion` section and this makes kubernetes complain:

```shell
$ k get opentelemetrycollectors.opentelemetry.io                                                                                                                                   
Error from server: conversion webhook for opentelemetry.io/v1alpha1, Kind=OpenTelemetryCollector failed: Post "https://xxxx-opentelemetry-operator-webhook.nxd.svc:443/convert?timeout=30s": service "xxx-opentelemetry-operator-webhook" not found
```

If admissionWebhooks are disabled, then the CRD should not reference the service created and by consequence disable the entire migration section.


jaronoff97 (commented):
can you bump the patch version of the chart and then run `make generate-examples`

ElderMael:
> can you bump the patch version of the chart and then run `make generate-examples`

Done, this modified and deleted a lot of files for some reason. Am I doing something wrong?

TylerHelmuth:
@ElderMael try `make generate-examples CHARTS=opentelemetry-operator`

jaronoff97 (approved):

ElderMael:
I ran the failing tests on Mac but seems they were designed for Linux only. This change in fact does affect the CRDs so the `diff` command fails with an exit code of `1`. 

Trying to sort out the URLs as this makes the CRDs omit the `spec.conversion` section thus the diff.

ElderMael:
Added a couple of lines to the make task to remove the newly added lines. This should fix the tests.

TylerHelmuth (approved):

ElderMael:
@TylerHelmuth thanks a lot for the help here. I appreciate maintainers being active ❤️

PR #1400, opened 2024-10-29
Allow to split the image registry and the repository

aacevedoosorio:
The current helm chart allows to set a `global.imageRegistry` in combination with the repository configured. But it would be nice as well to split the repository into the default registry for charts that use the open-telemetry chart as a subchart and have a default predefined repository.


TylerHelmuth (commented):

TylerHelmuth (commented):
We don't need a new example, but can you share the output of the image string being set properly when using the new field and the global field?

aacevedoosorio:
> We don't need a new example, but can you share the output of the image string being set properly when using the new field and the global field?

I ended up creating the examples. I think they are valuable

TylerHelmuth (commented):
@aacevedoosorio the new examples give me confidence that this process works, but I don't want to merge this PR with them.  Using these options in the values.yaml is straight forward for an end-user even if the logic to build the string is kinda complex.

aacevedoosorio:
> @aacevedoosorio the new examples give me confidence that this process works, but I don't want to merge this PR with them. Using these options in the values.yaml is straight forward for an end-user even if the logic to build the string is kinda complex.

I get your point. Let me revert those examples.

TylerHelmuth (approved):

TylerHelmuth:
@aacevedoosorio needs a rebase and then we'll be good.

PR #1402, opened 2024-10-30
Add support for setting additional annotations on the CRDs

aubm:
This is especially useful since CRDs are now installed through templates. Because of this change that was introduced in 0.57.0, now running `helm uninstall` of the operator chart will delete the CRDs as well, which is a destructive operation.

The workaround is to leverage the special helm.sh/resource-policy=keep annotation, see https://helm.sh/docs/howto/charts_tips_and_tricks/#tell-helm-not-to-uninstall-a-resource

One currently open issue though, is that the annotation needs to be set by Helm itself, see https://github.com/helm/helm/issues/10890 So running kubectl annotate on the CRDs as a post-install step doesn't work.

### DIFF
--- a/charts/opentelemetry-operator/Chart.yaml
+++ b/charts/opentelemetry-operator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.72.0
+version: 0.72.1
 description: OpenTelemetry Operator Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-operator/conf/crds/crd-opentelemetry.io_opampbridges.yaml
+++ b/charts/opentelemetry-operator/conf/crds/crd-opentelemetry.io_opampbridges.yaml
@@ -5,6 +5,9 @@ metadata:
   annotations:
     cert-manager.io/inject-ca-from: {{ include "opentelemetry-operator.webhookCertAnnotation" . }}
     controller-gen.kubebuilder.io/version: v0.16.1
+    {{- range $key, $value := .Values.crds.additionalAnnotations }}
+    {{- printf "%s: %s" $key (tpl $value $ | quote) | nindent 4 }}
+    {{- end }}
   creationTimestamp: null
   labels:
     app.kubernetes.io/name: opentelemetry-operator
--- a/charts/opentelemetry-operator/conf/crds/crd-opentelemetrycollector.yaml
+++ b/charts/opentelemetry-operator/conf/crds/crd-opentelemetrycollector.yaml
@@ -5,6 +5,9 @@ metadata:
   annotations:
     cert-manager.io/inject-ca-from: {{ include "opentelemetry-operator.webhookCertAnnotation" . }}
     controller-gen.kubebuilder.io/version: v0.16.1
+    {{- range $key, $value := .Values.crds.additionalAnnotations }}
+    {{- printf "%s: %s" $key (tpl $value $ | quote) | nindent 4 }}
+    {{- end }}
   creationTimestamp: null
   labels:
     app.kubernetes.io/name: opentelemetry-operator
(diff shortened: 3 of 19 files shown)


charts/opentelemetry-operator/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.71.2
+version: 0.72.0
```
jaronoff97, line 3: you can change this to only bump the patch version

aubm: done

aubm: Not sure how to fix the _Test Operator Charts_ jobs.
Any suggestion?

jaronoff97:
i believe you need to run `make update-operator-crds`

aubm:
> i believe you need to run `make update-operator-crds`

But for some reason, doing so reverts the very changes I want to bring.

jaronoff97:
argh, i think this is because of the sed in the makefile... @swiatekm would you be able to help out?

swiatekm:
I'm not convinced letting you keep the CRDs via Helm magic is something we should support in this Chart. If this is what you actually want, you should disable CRD installation in this Chart and either use https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-kube-stack/charts/crds or manage them on your own.

aubm:
> I'm not convinced letting you keep the CRDs via Helm magic is something we should support in this Chart. If this is what you actually want, you should disable CRD installation in this Chart and either use https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-kube-stack/charts/crds or manage them on your own.

Well there is value in leveraging what is already managed by the opentelemetry-operator chart, that is embedding the CRDs at the version that presumably matches the operator's, and handling the webhook configuration.

I could indeed reproduce that (I actually already had my own Kubernetes job to handle the CRDs upgrades), but I don't really see the point now that it's something directly supported, especially since I expect https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-kube-stack/charts/crds/crds to eventually follow the same pattern to properly configure the webhook.

Beside, from this chart's point of view, it's just supporting custom annotations on the CRDs, not advertising a particular use case, we could probably find others where that comes in handy.

Let me know what you think.

swiatekm:
> > I'm not convinced letting you keep the CRDs via Helm magic is something we should support in this Chart. If this is what you actually want, you should disable CRD installation in this Chart and either use https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-kube-stack/charts/crds or manage them on your own.
> 
> Well there is value in leveraging what is already managed by the opentelemetry-operator chart, that is embedding the CRDs at the version that presumably matches the operator's, and handling the webhook configuration.
> 
> Beside, from this chart's point of view, it's just supporting custom annotations on the CRDs, not advertising a particular use case, we could probably find others where that comes in handy.
> 
> Let me know what you think.

The annotations themselves are fine, although adding templating to the CRDs is a pain from a tooling point of view. If you want this to work, see this [Makefile macro](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/9f20e3ff5bcb08c031580fa24c4465a89526eaf8/Makefile#L85), and ensure it adds the annotation bits.

But more broadly, wanting to do these kinds of adjustments is usually a sign that you should be managing the CRDs yourself. It's unfortunate that the K8s ecosystem in general and Helm more specifically, don't give us good tools to do this automatically, but that won't change in the near future. At the very least, be aware that we don't test for this scenario.

> I could indeed reproduce that (I actually already had my own Kubernetes job to handle the CRDs upgrades), but I don't really see the point now that it's something directly supported, especially since I expect https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-kube-stack/charts/crds/crds to eventually follow the same pattern to properly configure the webhook.

For the record, the opentelemetry-kube-stack chart won't follow the same pattern. One of the reasons it does things differently is so you can pick which method you prefer. I don't think this is clearly documented anywhere, but it's our intent going forward. Managing CRDs with Helm is a bit of a crapshoot in general, and the best we can really do is let users pick their poison.

aubm:
Got you 👌
I ended up handling things on my side, taking inspiration from what was done in this repo.
Feel free to close this PR!

PR #1404, opened 2024-10-31
feat: Add TargetAllocator MTLS configuration to operator chart

thefirstofthe300:
The OTEL operator recently added support for using MTLS to secure the HTTP requests made from the collectors and the Target Allocator.

This PR adds support for enabling that feature gate using the directions found in the added documentation [here](https://github.com/open-telemetry/opentelemetry-operator/commit/b038590fe645d55993ea6f7b4c962a9f33067ba0#diff-824bc26d00bea6ed15b127f31d687b220f72ebe9ff4a40a4b9d0171be7d961f8R214-R248).


jaronoff97:
it looks like something went very wrong with generating examples 😅

thefirstofthe300:
> it looks like something went very wrong with generating examples 😅

I promise it wasn't me. :sweat_smile:

If it makes any difference, I'm running on a linux laptop so I wonder if there's some sort of tool with a different API. I know I always get bit by sed when dealing with Mac vs Linux environments. Same goes for Docker bind mounts.

thefirstofthe300:
Found the issue. Turns out that if values schema validation fails template rendering explodes.

charts/opentelemetry-operator/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.71.2
```
jaronoff97, line 3: only bump the patch version please, we'll bump the minor version and app version in a separate PR.

charts/opentelemetry-operator/values.yaml
```diff
@@ -186,6 +186,11 @@ manager:
   #   drop:
   #   - ALL
 
+  # Configuration for enabling the `operator.targetallocator.mtls` feature gate.
+  targetAllocator:
+    mtls:
+      enabled: false
```
jaronoff97, line 192: IMO we don't need to nest this so deeply and could probably just add a section for featuregates to enable and disable more explicitly.

charts/opentelemetry-operator/Chart.yaml
thefirstofthe300: What will happen if a feature gate that doesn't exist in a specific version is enabled? Will the manager fail?

charts/opentelemetry-operator/values.yaml
thefirstofthe300: I'm thinking about how this might work. Currently, feature gates are explicitly defined using a string value. I'd likely want to convert the string to an object that we can use to define feature gates. Long-term, that seems to be the best idea as it allows the chart to handle feature gates like this one that require extra configuration.

That does mean that we're going to need to bump the chart version though. If we're following semver strictly, we'd need to bump the major version of the chart.

Was that what you had in mind?

charts/opentelemetry-operator/Chart.yaml
jaronoff97: the chart was just bumped so the gate should exist now

charts/opentelemetry-operator/values.yaml
jaronoff97: i'm talking specifically about what's exposed to the user i.e.
```yaml
featureGates:
  targetAllocatorMtls:
    enabled: false # default
```

Where featuregates is a map with no required variables. I'm not sure why we'd need to bump the major version of the chart for that?

thefirstofthe300: I can definitely do this, but I'm talking about bumping the major version of the chart because it is a breaking change to the values file. The current `featureGates` value takes a simple string. We'd be changing that string to an object. For people who don't set the featureGates, it'd be fine but for people do, they'd need to use the new way.

jaronoff97: ahh i see what you mean, in that case we could just do a string contains with a helper template? Or we could just do like `featureGateMap` and deprecate the other method?

jaronoff97: sounds good to me!

thefirstofthe300: Updated. The expected behavior is that if the `featureGatesMap` property is not empty, the featureGates string will be ignored. No one should be using that value currently so by default, updating the chart should do nothing for those people. If they decide to move to the featureGatesMap, it should take precedence as it is designed to ensure all other dependencies are met.

How does a value deprecation get signaled?

jaronoff97: just via a comment usually. you can also include it in the upgrade.md file :)

thefirstofthe300: I updated both the comment and added a note to the UPGRADING.md file. I think this PR is ready for another review.

charts/opentelemetry-operator/values.schema.json
```diff
@@ -512,9 +512,25 @@
                 "featureGates": {
                     "type": "string",
                     "default": "",
+                    "title": "The featureGates to enable",
+                    "examples": ""
+                },
+                "featureGatesMap": {
```
jaronoff97, line 518: i believe the default is true, but we should explicitly set `additionalProperties` for validation.

charts/opentelemetry-operator/values.yaml
```diff
@@ -74,7 +74,12 @@ manager:
   # Prefix a gate with '-' to disable support.
   # Prefixing a gate with '+' or no prefix will enable support.
   # A full list of valid identifiers can be found here: https://github.com/open-telemetry/opentelemetry-operator/blob/main/pkg/featuregate/featuregate.go
+  # NOTE: the featureGates value is deprected and will be replaced by featureGatesMap in the future.
```
jaronoff97, line 77:
```suggestion
  # NOTE: the featureGates value is deprecated and will be replaced by featureGatesMap in the future.
```

charts/opentelemetry-operator/templates/clusterrole.yaml
```diff
@@ -235,6 +235,24 @@ rules:
       - patch
       - update
       - watch
+  {{- if and .Values.manager.featureGatesMap }}
```
jaronoff97, line 238: is the and doing anything here?

jaronoff97:
also be sure to re-run `make generate-examples CHARTS=opentelemetry-operator`

charts/opentelemetry-operator/values.yaml
thefirstofthe300: Done

charts/opentelemetry-operator/templates/clusterrole.yaml
thefirstofthe300: Fixed.

charts/opentelemetry-operator/values.schema.json
thefirstofthe300: Fixed.

jaronoff97:
hmmm it seems that there's a diff in the CRDs... not 100% sure why though. I wonder if it has to do with us checking the wrong version of them?

thefirstofthe300:
> hmmm it seems that there's a diff in the CRDs... not 100% sure why though. I wonder if it has to do with us checking the wrong version of them?

I had to rerun the update CRD target and that seems to have cleared things up.

jaronoff97 (approved):

charts/opentelemetry-operator/templates/_helpers.tpl
```diff
@@ -150,3 +150,9 @@ The image to use for opentelemetry-operator.
 {{- define "opentelemetry-operator.image" -}}
 {{- printf "%s:%s" .Values.manager.image.repository (default .Chart.AppVersion .Values.manager.image.tag) }}
 {{- end }}
+
+{{- define "opentelemetry-operator.featureGatesMap" -}}
```
TylerHelmuth, line 154: If we are deprecating `.Values.manager.featureGates` then this helper needs to be able to handle `.Values.manager.featureGatesMap` containing multiple feature gates.

charts/opentelemetry-operator/templates/deployment.yaml
```diff
@@ -74,8 +74,10 @@ spec:
             {{- if and .Values.manager.autoInstrumentationImage.apacheHttpd.repository .Values.manager.autoInstrumentationImage.apacheHttpd.tag }}
             - --auto-instrumentation-apache-httpd-image={{ .Values.manager.autoInstrumentationImage.apacheHttpd.repository }}:{{ .Values.manager.autoInstrumentationImage.apacheHttpd.tag }}
             {{- end }}
-            {{- if .Values.manager.featureGates }}
-            - --feature-gates={{ .Values.manager.featureGates }}
+            {{- if and .Values.manager.featureGatesMap }}
+            - {{ include "opentelemetry-operator.featureGatesMap" . }}
```
TylerHelmuth, line 78: Including the `-` here will cause problems if someone adds multiple values to the map.  I'd like to see all this logic moved into a help that handles whether to use the `featureGatesMap` or `featureGates`.

charts/opentelemetry-operator/UPGRADING.md
```diff
@@ -1,5 +1,9 @@
 # Upgrade guidelines
 
+## 0.72.0 to 0.72.1
+
+Prior to 0.72.1, feature gates could be enabled via the `manager.featureGates` property. As feature gates may require extra configuration to work properly, e.g. deploying extra permissions on the ClusterRole, the chart has been updated to make use of the `manager.featureGatesMap` property which allows the chart to smartly configure feature gates. If the `manager.featureGatesMap` property is set, the old `manager.featureGates` property will be ignored.
```
TylerHelmuth, line 5: Lets add an entry to NOTES.txt to promote that `featureGates` is deprecated and users should use `featureGatesMap`.

Also a failure scenario for users configuring both `featureGates` and `featureGatesMap`. I don't want to deal with merging them so lets allow only 1 to be set.

charts/opentelemetry-operator/templates/deployment.yaml
```diff
@@ -74,8 +74,10 @@ spec:
             {{- if and .Values.manager.autoInstrumentationImage.apacheHttpd.repository .Values.manager.autoInstrumentationImage.apacheHttpd.tag }}
             - --auto-instrumentation-apache-httpd-image={{ .Values.manager.autoInstrumentationImage.apacheHttpd.repository }}:{{ .Values.manager.autoInstrumentationImage.apacheHttpd.tag }}
             {{- end }}
-            {{- if .Values.manager.featureGates }}
-            - --feature-gates={{ .Values.manager.featureGates }}
+            {{- if and .Values.manager.featureGatesMap }}
```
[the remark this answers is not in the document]
thefirstofthe300: Fixed.

charts/opentelemetry-operator/templates/deployment.yaml
thefirstofthe300: In writing the helper, I approached the problem by generating the string that goes into the `--feature-gates` and kept the `--feature-gates=` in the deployment template.

charts/opentelemetry-operator/UPGRADING.md
thefirstofthe300: Done

TylerHelmuth (commented):
@thefirstofthe300 can you add a new example that uses the new map.

thefirstofthe300:
> @thefirstofthe300 can you add a new example that uses the new map.

Added.

thefirstofthe300:
@TylerHelmuth Can I get another review?

PR #1406, opened 2024-11-04
Upgrade operator version to 112

jaronoff97:


Makefile
```diff
@@ -92,3 +92,19 @@ define get-crd
 @sed -i 's#\(.*\)- v1beta1#&\n{{- end }}#' $(1)
 @echo '{{- end }}' >> $(1)
 endef
+
+# get-crd doesn't work on macos because of sed incompatibility
+# when upgrading the CRDs on MacOS, swap usage of get-crd to this macro
+define get-crd-macos
```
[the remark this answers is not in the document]
TylerHelmuth: Every time I bump the operator version I actidentally commit my `gsed` changes lol

charts/opentelemetry-operator/values.yaml
```diff
@@ -42,7 +42,7 @@ manager:
     tag: ""
   collectorImage:
     repository: ""
-    tag: 0.111.0
+    tag: 0.112.0
```
TylerHelmuth, line 45: FYI there is a bug with collector 0.112.0 and the hostmetricsreceiver on k8s: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/35990

Makefile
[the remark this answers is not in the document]
jaronoff97: yeah oop :( so this one is for us!

charts/opentelemetry-operator/values.yaml
jaronoff97: argh, i assume there won't be a patch for that right? Should i call this out in the release notes or something?

TylerHelmuth: the fix will be in v0.113.0

jaronoff97: kk we may as well skip then too...

TylerHelmuth: Is there a downside to updating this PR to use v0.113.0 of the collector when it releases this week?

jaronoff97: nope, we can do that.

TylerHelmuth:
replaced by https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1410 since @jaronoff97 is on PTO

PR #1408, opened 2024-11-05
Deploy PDB for statefulset as well as deployment

Infogrid-MattBuckland:
I see no reason not to create a podDisruptionBudget resource when deploying as a statefulset. I believe that daemonsets do not currently support PDBs so have not permitted that one at this time.

I've tested this with deployment and statefulset modes and it behaves as expected.


Infogrid-MattBuckland:
> @Infogrid-MattBuckland you need to rebase

On it...

Infogrid-MattBuckland:
Think I did it right. Cheers.

PR #1414, opened 2024-11-13
fix: Configure feature gates appropriately

thefirstofthe300:
The operator does not like feature gates being quoted.

Feature gates are also a flag based on presence. If the flag is present, it is understood to be true, absence leaves the default value.


TylerHelmuth (approved):

thefirstofthe300:
@TylerHelmuth Can this be merged now?

PR #1416, opened 2024-11-16
fix(collector): actually enable the deployment of PrometheusRule when using PodMonitor

### LINKED ISSUE #1379: [collector] PrometheusRule is not created when using PodMonitor
The condition here:

https://github.com/open-telemetry/opentelemetry-helm-charts/blob/3a6fff3f065dafcdcf7041885c5d0d1bf61a1167/charts/opentelemetry-collector/templates/prometheusrule.yaml#L1

means the PrometheusRule resource is not created when using PodMonitor instead of ServiceMonitor (makes sense for `mode: daemonset`).

Can change the condition to check for both values?

crutonjohn:
When fixing this logic regressions were introduced that would prevent the PrometheusRule from rendering in some very unique monitoring scenarios.

I have checked this with as many permutations that I can manually.

If a user wants to render the PrometheusRule, they should be able to so long as they have enabled ServiceMonitor or PodMonitor, and in the case of the ServiceMonitor they have set `Values.service.enabled` == `true`.

closes #1379

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.110.3
+version: 0.110.4
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-collector/examples/alternate-config/rendered/clusterrole.yaml
+++ b/charts/opentelemetry-collector/examples/alternate-config/rendered/clusterrole.yaml
@@ -5,7 +5,7 @@ kind: ClusterRole
 metadata:
   name: example-opentelemetry-collector
   labels:
-    helm.sh/chart: opentelemetry-collector-0.110.3
+    helm.sh/chart: opentelemetry-collector-0.110.4
     app.kubernetes.io/name: opentelemetry-collector
     app.kubernetes.io/instance: example
     app.kubernetes.io/version: "0.114.0"
--- a/charts/opentelemetry-collector/examples/alternate-config/rendered/clusterrolebinding.yaml
+++ b/charts/opentelemetry-collector/examples/alternate-config/rendered/clusterrolebinding.yaml
@@ -5,7 +5,7 @@ kind: ClusterRoleBinding
 metadata:
   name: example-opentelemetry-collector
   labels:
-    helm.sh/chart: opentelemetry-collector-0.110.3
+    helm.sh/chart: opentelemetry-collector-0.110.4
     app.kubernetes.io/name: opentelemetry-collector
     app.kubernetes.io/instance: example
     app.kubernetes.io/version: "0.114.0"
--- a/charts/opentelemetry-collector/examples/alternate-config/rendered/configmap.yaml
+++ b/charts/opentelemetry-collector/examples/alternate-config/rendered/configmap.yaml
@@ -6,7 +6,7 @@ metadata:
   name: example-opentelemetry-collector
   namespace: default
   labels:
-    helm.sh/chart: opentelemetry-collector-0.110.3
+    helm.sh/chart: opentelemetry-collector-0.110.4
     app.kubernetes.io/name: opentelemetry-collector
     app.kubernetes.io/instance: example
     app.kubernetes.io/version: "0.114.0"
--- a/charts/opentelemetry-collector/examples/alternate-config/rendered/deployment.yaml
+++ b/charts/opentelemetry-collector/examples/alternate-config/rendered/deployment.yaml
@@ -6,7 +6,7 @@ metadata:
   name: example-opentelemetry-collector
   namespace: default
   labels:
-    helm.sh/chart: opentelemetry-collector-0.110.3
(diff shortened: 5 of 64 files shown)


crutonjohn:
i didn't realize at first, but i think i introduced a regression initially in https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1415. i think testing should be expanded to validate that the output matches the input of values for these specific scenarios, but is somewhat obfuscated by the helper [here](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/f7704926b88c97e4adbf4e9966d95a74f08facd2/charts/opentelemetry-collector/templates/_helpers.tpl#L131-L141) which can obfuscate user input a little, but i understand why it exists.

charts/opentelemetry-collector/templates/prometheusrule.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if and .Values.prometheusRule.enabled .Values.serviceMonitor.enabled  }}
+{{- if and .Values.prometheusRule.enabled (or (eq (include "opentelemetry-collector.serviceEnabled" .) "true") (or .Values.serviceMonitor.enabled .Values.podMonitor.enabled)) }}
```
TylerHelmuth, line 1: > If a user wants to render the PrometheusRule, they should be able to so long as they have enabled ServiceMonitor or PodMonitor, and in the case of the ServiceMonitor they have set Values.service.enabled == true

@crutonjohn this sentence reads as if the condition should be

`if (or .Values.podMonitor.enabled (and .Values.serviceMonitor.enabled (eq (include "opentelemetry-collector.serviceEnabled" .) "true")))`

charts/opentelemetry-collector/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.109.0
+version: 0.109.1
```
TylerHelmuth, line 3: run `make generate-examples CHARTS=opentelemetry-collector` to generate the examples.

charts/opentelemetry-collector/templates/prometheusrule.yaml
crutonjohn: in hindsight i probably shouldn't have tried to gotmpl that late in the day 😅 

i think i mostly have it figured out:
```suggestion
{{- if and .Values.prometheusRule.enabled (or .Values.podMonitor.enabled (and (eq (include "opentelemetry-collector.serviceEnabled" .) "true") .Values.serviceMonitor.enabled)) }}
```

however, with this logic a user could use these values:
```yaml
mode: "deployment"
image:
  repository: "otel/opentelemetry-collector-k8s"
prometheusRule:
  enabled: true
ports:
  metrics:
    enabled: true
podMonitor:
  enabled: false
serviceMonitor:
  enabled: false
service:
  enabled: false
```
and believe they are going to get a PrometheusRule, but instead get no PrometheusRule/ServiceMonitor/PodMonitor because neither service nor pod monitors are enabled.

would it be considered acceptable UX to prevent the user from getting a PrometheusRule CR when they explicitly say they want one? or should we remain somewhat opinionated in that you should only be deploying a PrometheusRule CR when you have a Pod/ServiceMonitor getting deployed in tandem with said PrometheusRule?

happy to accommodate either way, i just don't know what's best for the project.

TylerHelmuth: Hmm thats a good point.

Is there any harm in creating a prometheusRule even if pod/service monitor don't exist? What will happen if something tries to use the Prometheus rule but the monitors dont exist?

crutonjohn: i'd have to do some verification, but iirc a PrometheusRule is just what it sounds like. it would effectively create the rule(s) contained within the file. there is some nuance with how prometheus operator is deployed that can affect how the PrometheusRule can "land" on the cluster...for example if you are running multiple prometheus instances in the same cluster scraping different targets. the principle is still the same.

the rules in the collector chart are generic enough (read: have low granularity due to promql label selector omission) that there could be a scenario wherein a metric like [otelcol_exporter_send_failed_log_records](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/6f674b102585a158dbacccecfc2c12a3bbc5a120/charts/opentelemetry-collector/templates/prometheusrule.yaml#L85) exists in the TSDB from a source that isn't _our_ deployed collector (or the user has this chart deployed multiple times for different situations), users could hypothetically be getting alerts for something unrelated to our collector deployment.

this would be kind of a "perfect storm" type scenario to be fair, but definitely possible. i've seen some pretty funky prometheus operator setups before. i think the crux of the issue is that there are tons of potential deployment scenarios that we couldn't possibly account for every one with the logic within the chart. part of the responsibility is on the end user being familiar with what the chart's behavior enough to know "hey i don't need to enable the PrometheusRule if i'm not going to be using the Service/PodMonitor". could add some docs somewhere, idk where the best place would be tho.

sorry, not trying to be a headache.

TylerHelmuth: It feels like we should just honor the `enabled/disabled` setting and expect users to hold it correctly.

Issue #1377, opened 2024-10-10
[collector] kubernetesAttributes preset ClusterRole does not include "nodes"


lindeskar:
The [kubernetesAttributes preset](https://opentelemetry.io/docs/kubernetes/helm/collector/#kubernetes-attributes-preset) does not include "nodes" in it's ClusterRole, leading to errors when trying to access Node metadata:

> `User "system:serviceaccount:opentelemetry-collector:otelcol-opentelemetry-collector" cannot list resource "nodes" in API group "" at the cluster scope`

Example chart values causing the error:

```yaml
mode: "daemonset"
image:
  repository: otel/opentelemetry-collector-contrib
config:
  processors:
    k8sattributes:
      extract:
        labels:
          - { tag_name: zone, key: "topology.kubernetes.io/zone", from: node }
presets:
  logsCollection:
    enabled: true
  kubernetesAttributes:
    enabled: true
```

This addition to values fixes the problem:

```yaml
clusterRole:
  rules:
    - apiGroups: [""]
      resources: ["nodes"]
      verbs: ["get", "list", "watch"]
```

Is this by design? I think we could update the preset documentation, or simply add "nodes" to the template:

https://github.com/open-telemetry/opentelemetry-helm-charts/blob/694c38d175d72fe8e3bc6cc5c182728615f72cbd/charts/opentelemetry-collector/templates/clusterrole.yaml#L18-L28

--

From the k8sattributesprocessor [README](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/k8sattributesprocessor/README.md#cluster-scoped-rbac):

> _When using `k8s.node.uid` or extracting metadata from node, the processor needs get, watch and list permissions for `nodes` resources._

TylerHelmuth:
@lindeskar you are correct, the preset only sets the RBAC to meet the permission requirements of the k8sattributes processor as the preset defines it.  Since you added additional scope (extracting node labels) you must also supply the additional RBAC for that feature.

We do this to ensure the preset configures the minimum permissions necessary. We do not want to give more permissions that the components need as that is a security risk.

lindeskar:
Thanks. I agree that's a reasonable default.

Then I wish for a comment about this in:

https://github.com/open-telemetry/opentelemetry-helm-charts/blob/694c38d175d72fe8e3bc6cc5c182728615f72cbd/charts/opentelemetry-collector/values.yaml#L45-L47

and:

[...]

PR #1417, opened 2024-11-16
docs(collector): clarify behavior of presets.kubernetesAttributes

crutonjohn:
the preset applies both the k8sattributes processor as well as minimum viable RBAC rules for the collector, which should be made clear to the user.

just tackling some low-hanging fruit in my post-kubecon hotel exhaustion.

addresses #1377


lindeskar:
I'm adding these changes in #1492

PR #1423, opened 2024-11-21
[demo] update memory limits

puckpuck:
Updates memory limits to be inline with changes specified in this Demo [PR](https://github.com/open-telemetry/opentelemetry-demo/pull/1786)

### DIFF
--- a/charts/opentelemetry-demo/Chart.lock
+++ b/charts/opentelemetry-demo/Chart.lock
@@ -1,18 +1,18 @@
 dependencies:
 - name: opentelemetry-collector
   repository: https://open-telemetry.github.io/opentelemetry-helm-charts
-  version: 0.109.0
+  version: 0.107.0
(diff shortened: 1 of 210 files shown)


charts/opentelemetry-demo/Chart.lock
```diff
@@ -1,18 +1,18 @@
 dependencies:
 - name: opentelemetry-collector
   repository: https://open-telemetry.github.io/opentelemetry-helm-charts
-  version: 0.109.0
+  version: 0.107.0
```
TylerHelmuth, line 4: @puckpuck do you want to decrease these versions?

puckpuck: Thanks for catching this. Certainly did not intend to decrease these.

TylerHelmuth (approved):

Issue #1434, opened 2024-11-28
[operator] operator is attempting to grant RBAC permissions not currently held


grandwizard28:
#### Description

The operator starts failing with the following errors::
```
is attempting to grant RBAC permissions not currently held: {APIGroups:[""], Resources:["nodes/proxy"], Verbs:["get"]}
is attempting to grant RBAC permissions not currently held: {APIGroups:[""], Resources:["nodes/stats"], Verbs:["get"]}
```

My guess is this is has something to do with the configuration of my collector:
```
config:
  extensions:
    k8s_observer/pods:
    health_check:
    file_storage/filelog:
  receivers:
    kubeletstats:
    hostmetrics:
    otlp/grpc:
    receiver_creator/prometheus:
      watch_observers: [k8s_observer/pods]
    receiver_creator/filelog:
      watch_observers: [k8s_observer/pods]
  processors:
    resourcedetection:
      detectors:
        - env
        - gcp
        - k8snode
      k8snode:
        node_from_env_var: K8S_NODE_NAME
        auth_type: serviceAccount
      timeout: 2s
      override: true
    k8sattributes:
      extract:
        metadata:
          - "k8s.namespace.name"
          - "k8s.deployment.name"
          - "k8s.statefulset.name"
          - "k8s.daemonset.name"
          - "k8s.cronjob.name"
          - "k8s.job.name"
          - "k8s.node.name"
          - "k8s.node.uid"
          - "k8s.pod.name"
          - "k8s.pod.uid"
          - "k8s.pod.start_time"
          - "k8s.container.name"
          - "container.image.name"
          - "container.image.tag"
      filter:
        node_from_env_var: K8S_NODE_NAME
      passthrough: false
      pod_association:
        - sources:
          - from: resource_attribute
            name: k8s.pod.ip
        - sources:
          - from: resource_attribute
            name: k8s.pod.uid
        - sources:
          - from: connection
```
I'm not sure which processor/receiver/extension is asking for these permissions. Can we add these permissions to the helm chart?

The fix is to add the above roles in the clusterrole by running:
```
kubectl --context=gke_signozsso_us-central1-c_mgmt edit clusterrole/opentelemetry-operator-manager
```

### Versions
Helm Chart: **0.74.2**
Collector Image: **0.113.0**

grandwizard28:
I'm happy to raise a fix for this!

PR #1436, opened 2024-11-29
fix(opentelemetry-operator): add get on nodes/proxy and nodes/state to the clusterrole

grandwizard28:
Fixes https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1434

### DIFF
--- a/charts/opentelemetry-operator/templates/clusterrole.yaml
+++ b/charts/opentelemetry-operator/templates/clusterrole.yaml
@@ -106,6 +106,18 @@ rules:
       - list
       - watch
   - apiGroups:
+      - ""
+    resources:
+      - nodes/proxy
+    verbs:
+      - get
+  - apiGroups:
+      - ""
+    resources:
+      - nodes/stats
+    verbs:
+      - get
+  - apiGroups:
       - config.openshift.io
     resources:
       - infrastructures


jaronoff97:
@grandwizard28 thanks for submitting this! can you please bump the patch version of the operator chart and then run `make generate-examples CHARTS=opentelemetry-operator`

grandwizard28:
Done @jaronoff97

charts/opentelemetry-operator/templates/clusterrole.yaml
```diff
@@ -105,6 +105,18 @@ rules:
...
+    resources:
+      - nodes/proxy
+    verbs:
+      - get
+  - apiGroups:
+      - ""
+    resources:
+      - nodes/stats
+    verbs:
+      - get
```
TylerHelmuth, line 119: @jaronoff97 if it is the collector that needs the permissions (I believe this is kubeletstatsreceiver or k8sattributesprocessor), why does the operator need them?

jaronoff97: I think this comes from the operator's RBAC magic that allows it to create the necessary RBAC for the collector (getting this from [the issue](https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1434)) I think as a result this permission should only be in the section [here](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1436/files#diff-306d86f5a6db61d4fd383f2e0a05720ed2b0124de0daade5276559f83142dc37R76) which contains the permissions for operator RBAC magics

akasprzok:
Running into the same issue. Anything I can do to help here?

akasprzok:
> @akasprzok for now you should be able to just modify the clusterrole for the operator to include these permissions (or create a new temporary clusterrole with these permissions and bind it to the operator's SA)

Yep, that works just fine.

charts/opentelemetry-operator/templates/clusterrole.yaml
grandwizard28: Shall I move it inside the createRbacPermissions block then?

grandwizard28: Any updates here?

JaredTan95:
@grandwizard28 hi, conflicts needs to be solved before PR merge.

grandwizard28:
Hi @JaredTan95,
Thanks for replying! There's a pending question in the PR comment which needs to be answered.

How can I get that answer?

charts/opentelemetry-operator/templates/clusterrole.yaml
Allex1: @grandwizard28 you should yes imo and maybe move the whole block at the end just bellow `featureGatesMap`

Allex1:
@grandwizard28 I think you should and pls bump version afterwards.
@JaredTan95 @jaronoff97 once this is merged I think we should move other permissions like the openshift ones under the same block or create a feature flag for all non vanilla k8s ones . wdyt ?

JaredTan95:
as https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1536 and https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1546 contains this PR changes, ce be closed?

TylerHelmuth:
@JaredTan95 I don't see the `nodes/proxy` or `nodes/stats` entries in main so I believe this is still desired.

grandwizard28:
Hi @TylerHelmuth,
Shall I rebase and get this ready?

TylerHelmuth:
@grandwizard28 yes

Allex1:
Hi, any activity on this ? Seems we don't want it to become stale

Allex1:
Hi, took over this pr as it was becoming stale. Can we merge in the current form ? @jaronoff97  @TylerHelmuth  @dmitryax

TylerHelmuth (approved):

jaronoff97 (approved):

PR #1438, opened 2024-12-01
[demo] Remove duplicate envvar on frontendproxy

lambdanis:
OTEL_SERVICE_NAME envvar is set on all components as one of envvars listed in default.env, to the value of app.kubernetes.io/component label. Recently 51993de76504 ("Bump dependencies and fix grafana dashboards") duplicated OTEL_SERVICE_NAME on the frontendproxy deployment by setting it explicitly to frontend-proxy. A duplicated envvar causes an error when upgrading the opentelemetry-demo chart:

     Deployment.apps "opentelemetry-demo-frontendproxy" is invalid:
     spec.template.spec.containers[0].env[0].valueFrom: Invalid value: "": may
     not be specified when `value` is not empty

My understanding is that the duplicated envvar was introduced in #1422 as a follow-up to open-telemetry/opentelemetry-demo#1768 that introduced it in docker-compose setup of opentelemetry-demo? But I don't have context on why it was needed there.

cc @julianocosta89 @puckpuck as authors of the PRs.

### DIFF
--- a/charts/opentelemetry-demo/Chart.yaml
+++ b/charts/opentelemetry-demo/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: opentelemetry-demo
-version: 0.33.4
+version: 0.33.5
 description: opentelemetry demo helm chart
 home: https://opentelemetry.io/
 sources:
--- a/charts/opentelemetry-demo/examples/bring-your-own-observability/rendered/component.yaml
+++ b/charts/opentelemetry-demo/examples/bring-your-own-observability/rendered/component.yaml
@@ -1221,8 +1221,6 @@ spec:
               value: "4317"
             - name: OTEL_COLLECTOR_PORT_HTTP
               value: "4318"
-            - name: OTEL_SERVICE_NAME
-              value: frontend-proxy
             - name: OTEL_RESOURCE_ATTRIBUTES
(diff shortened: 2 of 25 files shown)


julianocosta89:
Hello @lambdanis 👋🏽 

Thanks for sending the PR, but the location is wrong.
The k8s manifests in the Demo repo are generated after the Helm chart is updated.
So, in this case, we need to send a PR to the Helm repo and remove the wrongly added `OTEL_SERVICE_NAME` env var.

https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-demo/values.yaml#L404-L405

Thanks again for spotting it.

julianocosta89:
I've sent the PR to fix the Helm chart here: https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1439

julianocosta89:
Thanks for spotting it @lambdanis!

To update the Helm chart you also need to bump the minor version of the chart (https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-demo/Chart.yaml#L4), and then run:

```
make generate-examples CHARTS=opentelemetry-demo
```

lambdanis:
Done, thanks for review @julianocosta89

julianocosta89 (approved):
LGTM

dmitryax (approved):

PR #1439, opened 2024-12-01
Remove duplicated service name from frontend proxy

julianocosta89:
I've wrongly added the env var `SERVICE_NAME` to the `frontendProxy` service in #1422.
The `SERVICE_NAME` is populated via default env var: https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-demo/values.yaml#L5-L9.

This was reported by @lambdanis here: https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1438


julianocosta89:
Duplicated.

PR #1443, opened 2024-12-04
[collector] `configMap.existingName` support tpl

JaredTan95:
This is a widely use case, e.g. The user may deploy multiple sets of otel col through helm chart, and all of them use the existing external configmap, and the name of the configmap may be associated with the chart `release name`, so we can automatically associate it in this way. Especially in the case of otel helm chart as a subchart.


TylerHelmuth:
I think I know why you want this, but can you add something to the description

TylerHelmuth (approved):

PR #1445, opened 2024-12-05
fix(collector): Whitespace/newline issues regarding additionalLabels

brdotv2:
When providing values to the `additionalLabels` config object, they are added inline with the last label defined in `opentelemetry-collector.labels` in `_helpers.tpl` which makes the deployment of the Chart fail.

```yaml
# values.yaml
global:
  test: templated-value

mode: deployment

additionalLabels:
  testLabel: "{{ .Values.global.test }}"
  someLabel: "someValue"

```
Current output:

```yaml
---
# Source: opentelemetry-collector/templates/serviceaccount.yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: release-name-opentelemetry-collector
  namespace: default
  labels:
    helm.sh/chart: opentelemetry-collector-0.110.4
    app.kubernetes.io/instance: release-name
    app.kubernetes.io/version: "0.114.0"
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/component: standalone-collectorsomeLabel: someValue     <------ issue
    testLabel: 'templated-value'
---
```

By wrapping the `include` in an `if` statement, the labels are printed out correctly with the expected newline.

```yaml
---
# Source: opentelemetry-collector/templates/serviceaccount.yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: release-name-opentelemetry-collector
  namespace: default
  labels:
    helm.sh/chart: opentelemetry-collector-0.110.4
    app.kubernetes.io/instance: release-name
    app.kubernetes.io/version: "0.114.0"
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/component: standalone-collector
    someLabel: someValue
    testLabel: 'templated-value'
---
```

### DIFF
--- a/charts/opentelemetry-collector/templates/_helpers.tpl
+++ b/charts/opentelemetry-collector/templates/_helpers.tpl
@@ -69,7 +69,7 @@ app.kubernetes.io/component: agent-collector
 app.kubernetes.io/component: statefulset-collector
 {{- end -}}
 {{- if .Values.additionalLabels }}
-{{ include "opentelemetry-collector.additionalLabels" . }}
+{{ tpl (.Values.additionalLabels | toYaml) . }}
 {{- end }}
 {{- end }}
 
@@ -116,12 +116,6 @@ Create the name of the clusterRoleBinding to use
 {{- define "opentelemetry-collector.podLabels" -}}
 {{- if .Values.podLabels }}
 {{- tpl (.Values.podLabels | toYaml) . }}
-{{- end }}
-{{- end }}
-
-{{- define "opentelemetry-collector.additionalLabels" -}}
-{{- if .Values.additionalLabels }}
-{{- tpl (.Values.additionalLabels | toYaml) . }}
 {{- end }}
 {{- end }}
 
--- a/charts/opentelemetry-collector/templates/_helpers.tpl
+++ b/charts/opentelemetry-collector/templates/_helpers.tpl
@@ -68,7 +68,9 @@ app.kubernetes.io/component: agent-collector
 {{- if eq .Values.mode "statefulset" }}
 app.kubernetes.io/component: statefulset-collector
 {{- end -}}
+{{- if .Values.additionalLabels }}
 {{ include "opentelemetry-collector.additionalLabels" . }}
+{{- end }}
 {{- end }}
 
 {{/*
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.110.4
+version: 0.110.5
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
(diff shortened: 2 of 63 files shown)


charts/opentelemetry-collector/templates/_helpers.tpl
```diff
@@ -68,8 +68,10 @@ app.kubernetes.io/component: agent-collector
 {{- if eq .Values.mode "statefulset" }}
 app.kubernetes.io/component: statefulset-collector
 {{- end -}}
+{{- if .Values.additionalLabels }}
```
TylerHelmuth, line 71: opentelemetry-collector.additionalLabels already includes an if statement so I suspect the actual issue is that this `include` doesn't have a leading `-`.

Can you test with `{{- include "opentelemetry-collector.additionalLabels" . }}`

brdotv2: ups, didn't see the other `if` statement 🤦🏾
 
But a leading `-` doesn't fix it. From what i understand, the ´{{- ´ is supposed to remove any whitespace from the left, so it would do the opposite.

To avoid the duplicate the duplicate `if`, I can add  `| nindent 0` to either `{{- tpl (.Values.additionalLabels | toYaml) . }}` or  `{{ include "opentelemetry-collector.additionalLabels" . }}` which works just fine, but looks more like a "hack"

I can also remove the leading ´-´ from `{{ tpl (.Values.podLabels | toYaml) . }}`, which also works, but is not consistent with the other usages of `{{- tpl() }}` in the file. 

I'm not a Go/Helm template expert, so this is what i got :)

Thanks for review btw! 🙏🏾

TylerHelmuth: Looking through the chart, we only use `opentelemetry-collector.additionalLabels` inside `opentelemetry-collector.labels`.  Lets remove `opentelemetry-collector.additionalLabels` and move its logic directly into 
opentelemetry-collector.labels`. It should look like

```
{{/*
Common labels
*/}}
{{- define "opentelemetry-collector.labels" -}}
helm.sh/chart: {{ include "opentelemetry-collector.chart" . }}
{{ include "opentelemetry-collector.selectorLabels" . }}
{{- if .Chart.AppVersion }}
app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
{{- end }}
app.kubernetes.io/managed-by: {{ .Release.Service }}
{{- if eq .Values.mode "deployment" }}
app.kubernetes.io/component: standalone-collector
{{- end -}}
{{- if eq .Values.mode "daemonset" }}
app.kubernetes.io/component: agent-collector
{{- end -}}
{{- if eq .Values.mode "statefulset" }}
app.kubernetes.io/component: statefulset-collector
{{- end -}}
{{- if .Values.additionalLabels }}
{{ tpl (.Values.additionalLabels | toYaml) . }}
{{- end }}
{{- end }}
```

brdotv2:
> Please run `make generate-examples CHARTS=opentelemetry-collector` and commit the changes

i have, but there are no changes. Do you maybe want me to update some of the example `values.yaml` files to include some values for `additionalLabels`?

TylerHelmuth:
@brdotv2 we've been merging other PRs but if you can get the patch version bumped and examples generated one more time I'll merge this next

jaronoff97 (approved):

PR #1448, opened 2024-12-05
Release operator 0.114.0

araiu:


charts/opentelemetry-operator/values.yaml
```diff
@@ -42,7 +42,7 @@ manager:
     tag: ""
   collectorImage:
     repository: ""
-    tag: 0.113.0
+    tag: 0.114.1
```
TylerHelmuth, line 45:
```suggestion
    tag: 0.114.0
```

charts/opentelemetry-operator/templates/clusterrole.yaml
```diff
@@ -162,6 +163,19 @@ rules:
...
+      - delete
+      - get
+      - list
+      - patch
+      - update
+      - watch
+  - apiGroups:
+      - opentelemetry.io
+    resources:
+      - targetallocators
```
TylerHelmuth, line 178: What in the new operator release needs these permissions?

charts/opentelemetry-operator/templates/clusterrole.yaml
```diff
@@ -210,6 +224,14 @@ rules:
...
       - patch
       - update
+  - apiGroups:
+    - opentelemetry.io
+    resources:
+    - targetallocators/status
+    verbs:
+    - get
+    - patch
+    - update
```
TylerHelmuth, line 234: What in the new operator release needs these permissions?

araiu: Those were added in Generate only [TargetAllocator CR from Collector CR #3402](https://github.com/open-telemetry/opentelemetry-operator/pull/3402/files), but looking at it now, these might be required only `targetAllocatorCR` flag is set

TylerHelmuth: In that case we should make them conditional in the chart if we can.

charts/opentelemetry-operator/values.yaml
araiu: Version updated in 3c7f10d1e2af474b803714428e141a6cce09e8fb

charts/opentelemetry-operator/templates/clusterrole.yaml
araiu: Agree, the permissions will be set if `.Values.manager.featureGatesMap.operator.targetAllocatorCR` is enabled

charts/opentelemetry-operator/Chart.yaml
```diff
@@ -12,4 +12,4 @@ maintainers:
   - name: jaronoff97
   - name: TylerHelmuth
 icon: https://raw.githubusercontent.com/cncf/artwork/a718fa97fffec1b9fd14147682e9e3ac0c8817cb/projects/opentelemetry/icon/color/opentelemetry-icon-color.png
-appVersion: 0.113.0
+appVersion: 0.114.0
```
jaronoff97, line 15: @araiu we also just pushed a patch for 0.114.0 to version 0.114.1 which we should use for the _operator_ app version here

araiu: That was the initial version used, but followed @TylerHelmuth 's suggestion (https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1448#discussion_r1871778845).
I agree to push to `0.114.1` to also get the fix from https://github.com/open-telemetry/opentelemetry-operator/issues/3463

.github/workflows/operator-test.yaml
```diff
@@ -58,7 +58,8 @@ jobs:
         run: |
           helm install --namespace=opentelemetry-operator-system --create-namespace my-opentelemetry-operator ./charts/opentelemetry-operator \
             --set 'manager.extraArgs[0]="--enable-go-instrumentation"' \
-            --set "manager.collectorImage.repository=otel/opentelemetry-collector-k8s"
+            --set "manager.collectorImage.repository=otel/opentelemetry-collector-k8s" \
+            --set "manager.podLabels.control-plane=controller-manager"
```
jaronoff97, line 62: OOC why do we need these labels now?

araiu: [The tests were failing ](https://github.com/open-telemetry/opentelemetry-helm-charts/actions/runs/12193933388/job/34016968409#step:13:116) due to https://github.com/open-telemetry/opentelemetry-operator/pull/3486 which let to `e2e/operator-restart` delete step not matching the operator pod.
The alternative was to add the `Values.manager.podLabels.control-plane` label, but I don't think it makes sense to have this label in the default values.

charts/opentelemetry-operator/Chart.yaml
TylerHelmuth: Ya that was the [collector version](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1448#discussion_r1871778845). There isn't a 0.114.1 of the collector.

araiu: Ah, got mixed with the versions.
Pushed a new comit with the correct `0.114.1` for the operator and `0.114.0` for the collector

TylerHelmuth (approved):

PR #1449, opened 2024-12-05
fix(prometheusrule): simplify deployment logic

### LINKED ISSUE #1379: [collector] PrometheusRule is not created when using PodMonitor

crutonjohn:
if the user specifies that they want to deploy the PrometheusRule, it should be allowed without the interference of other tangential logic such as the status of service monitors or pod monitors

i accidentally force pushed and broke the previous PR, sorry about that. previous discussion: https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1416

fix #1379

cc: @TylerHelmuth (sorry again)

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.110.4
+version: 0.110.5
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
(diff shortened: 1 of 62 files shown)


TylerHelmuth (approved):

JaredTan95 (approved):

jaronoff97 (approved):

jaronoff97:
Thank you for your contribution! :Bow:

Issue #1453, opened 2024-12-06
[opentelemetry-kube-stack]: Prometheus presets not working due to templating


cabrinha:
In the promethus section at the bottom of the values file, only these two options are working:
```yaml
      kubernetesServiceMonitors:
        enabled: true
      kubelet:
        enabled: true
```

All the other options error when enabled:
```
Unable to save changes: application spec for opentelemetry-kube-stack is invalid: InvalidSpecError: Unable to generate manifests in : rpc error: code = Unknown desc = `helm template . --name-template opentelemetry-kube-stack --namespace ns --kube-version 1.26 --values /tmp/96b29881-e6f8-4b5f-a515-b1444bff2595/opentelemetry-kube-stack/values.yaml --values /tmp/d5a1357e-cd52-4b22-a365-57da814a6775 <api versions removed> --include-crds` failed exit status 1: Error: template: opentelemetry-kube-stack/templates/exporters/kube-scheduler/servicemonitor.yaml:19:6: executing "opentelemetry-kube-stack/templates/exporters/kube-scheduler/servicemonitor.yaml" at <include "servicemonitor.scrapeLimits" .Values.kubeScheduler.serviceMonitor>: error calling include: template: no template "servicemonitor.scrapeLimits" associated with template "gotpl" Use --debug flag to render out invalid YAML
```

jaronoff97:
I will look into this!

marcschaeferger:
Can be closed. Implemented by #1455, which added the missing ServiceMonitor scrape-limit helper and explicitly solves this issue.

PR #1455, opened 2024-12-06
[opentelemetry-kube-stack] Add the helper for servicemonitor.scrapeLi…

cabrinha:
Solves https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1453

### DIFF
--- a/charts/opentelemetry-kube-stack/Chart.yaml
+++ b/charts/opentelemetry-kube-stack/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-kube-stack
-version: 0.3.6
+version: 0.3.7
 description: |
   OpenTelemetry Quickstart chart for Kubernetes.
   Installs an operator and collector for an easy way to get started with Kubernetes observability.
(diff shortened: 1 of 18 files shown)


jaronoff97 (approved):
this looks good to me! thank you :bow: I'm wondering if we should also change an example to enable the kubelet so we can see the effect?

charts/opentelemetry-kube-stack/templates/_helpers.tpl
```diff
@@ -286,3 +286,22 @@ Helpers for prometheus servicemonitors
   {{- $userValue := index . 3 -}}
   {{- include "opentelemetry-kube-stack.kubeVersionDefaultValue" (list $values ">= 1.23-0" $insecure $secure $userValue) -}}
 {{- end -}}
+
+{{/* Sets default scrape limits for servicemonitor */}}
+{{- define "servicemonitor.scrapeLimits" -}}
```
TylerHelmuth, line 291: We should update the name to match our naming scheme: `opentelemetry-kube-stack.servicemonitor.scrapeLimits`

jaronoff97 (approved):

jaronoff97:
@cabrinha thank you very much for your contribution! :bow:

PR #1458, opened 2024-12-09
feat(opentelemetry-operator): Adding extraEnv

jtyr:
This PR is adding the possibility to specify extra environment variable definitions that are not simple name/value pairs, for example when `valueFrom` should be used.

### DIFF
--- a/charts/opentelemetry-operator/Chart.yaml
+++ b/charts/opentelemetry-operator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.77.0
+version: 0.78.0
 description: OpenTelemetry Operator Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-operator/values.yaml
+++ b/charts/opentelemetry-operator/values.yaml
@@ -104,6 +104,14 @@ manager:
   env:
     ENABLE_WEBHOOKS: "true"
 
+  # Extra definitions of environment variables.
+  extraEnv: []
+    # - name: GOMEMLIMIT
+    #   valueFrom:
+    #     resourceFieldRef:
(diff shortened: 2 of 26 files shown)


dmitryax:
We use `extraEnvs` in the other helm charts here. Let's keep it consistent. Also, I think we should deprecate `env` now

charts/opentelemetry-operator/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.75.0
+version: 0.76.0
```
jaronoff97, line 3: please bump the patch version, not the minor version. Once you've done that, please run `make generate-examples CHARTS=opentelemetry-operator`

jaronoff97 (commented):
Agreed, we should probably deprecate env, can you update the comment for env and recommend moving to extraEnvs?

charts/opentelemetry-operator/Chart.yaml
jaronoff97: @jtyr it looks like there's no patch version bump anymore, can you do that and re-run the make generate when you can please?

jtyr:
@jaronoff97 Please could you review this PR?

Allex1 (approved):

jaronoff97 (approved):

charts/opentelemetry-operator/values.yaml
```diff
@@ -104,6 +104,14 @@ manager:
   env:
```
jaronoff97, line 104: @jtyr Minor request, can you please update the docstring here and recommend that people use extraEnvs instead of env please?

jtyr: I have added a short deprecation notice. Please review it.

TylerHelmuth (approved):

PR #1461, opened 2024-12-11
fix(opentelemetry-operator): set version label with image tag if provided

jdiegosierra:
If `.Values.manager.image.tag` is provided with a different version other than `appVersion`, `label app.kubernetes.io/version` still shows `appVersion`.


charts/opentelemetry-operator/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.60.0
+version: 0.61.0
```
jaronoff97, line 3: please bump the patch version and not the minor version, the minor version is for when we do an operator release :) thank you!

charts/opentelemetry-operator/conf/crds/crd-opentelemetrycollector.yaml
```diff
@@ -1,9 +1,8 @@
-{{- if .Values.crds.create }}
```
jaronoff97, line 1: why is this changing?

charts/opentelemetry-operator/templates/_helpers.tpl
```diff
@@ -36,8 +36,8 @@ Common labels
 {{- define "opentelemetry-operator.labels" -}}
 helm.sh/chart: {{ include "opentelemetry-operator.chart" . }}
 {{ include "opentelemetry-operator.selectorLabels" . }}
-{{- if .Chart.AppVersion }}
-app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+{{- if or .Chart.AppVersion .Values.manager.image.tag }}
+app.kubernetes.io/version: {{ (default .Chart.AppVersion .Values.manager.image.tag) | quote }}
```
jaronoff97, line 40: Maybe we should make a separate helper template for this

charts/opentelemetry-operator/values.yaml
```diff
@@ -42,7 +42,7 @@ manager:
     tag: ""
   collectorImage:
     repository: ""
-    tag: 0.100.0
+    tag: 0.100.1
```
jaronoff97, line 45: it seems like this actually got misaligned and we should be bumping this to match the operator version.

charts/opentelemetry-operator/templates/_helpers.tpl
```diff
@@ -30,15 +30,20 @@ Create chart name and version as used by the chart label.
...
 {{/*
 Common labels
 */}}
 {{- define "opentelemetry-operator.labels" -}}
 helm.sh/chart: {{ include "opentelemetry-operator.chart" . }}
 {{ include "opentelemetry-operator.selectorLabels" . }}
-{{- if .Chart.AppVersion }}
-app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
-{{- end }}
+app.kubernetes.io/version: {{ include "opentelemetry-operator.appVersion" . | quote }}
```
jdiegosierra, line 46: `.Chart.AppVersion` will be always defined so its not needed

charts/opentelemetry-operator/templates/_helpers.tpl
```diff
@@ -30,15 +30,20 @@ Create chart name and version as used by the chart label.
...
+{{ default .Chart.AppVersion .Values.manager.image.tag }}
+{{- end }}
+
 {{/*
 Common labels
 */}}
 {{- define "opentelemetry-operator.labels" -}}
 helm.sh/chart: {{ include "opentelemetry-operator.chart" . }}
 {{ include "opentelemetry-operator.selectorLabels" . }}
-{{- if .Chart.AppVersion }}
```
jdiegosierra, line 39: `.Chart.AppVersion` will be always defined so its not needed

charts/opentelemetry-operator/conf/crds/crd-opentelemetrycollector.yaml
jdiegosierra: I forgot to sync my branch with master

jaronoff97 (approved):

TylerHelmuth (approved):

sathieu:
This breaks using digest in tag:

one or more objects failed to apply, reason: Secret "opentelemetry-operator-controller-manager-service-cert" is invalid: [metadata.labels: Invalid value: "v0.116.0@sha256:11edfad24ead5f48140eb12aea642a2716ca0cd9e63eae3851f39c4477f3c2c2": must be no more than 63 characters, metadata.labels: Invalid value: "v0.116.0@sha256:11edfad24ead5f48140eb12aea642a2716ca0cd9e63eae3851f39c4477f3c2c2": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')].

PR #1462, opened 2024-12-12
[demo] update prometheus otlp config

puckpuck:
Updates Prometheus config for OTLP based on this [guide](https://prometheus.io/docs/guides/opentelemetry/) provided in Prometheus docs.

Also fixes this [issue](https://github.com/open-telemetry/opentelemetry-demo/issues/1816) in the demo repo.

### DIFF
--- a/charts/opentelemetry-demo/Chart.yaml
+++ b/charts/opentelemetry-demo/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: opentelemetry-demo
-version: 0.33.7
+version: 0.33.8
 description: opentelemetry demo helm chart
 home: https://opentelemetry.io/
 sources:
--- a/charts/opentelemetry-demo/examples/bring-your-own-observability/rendered/component.yaml
+++ b/charts/opentelemetry-demo/examples/bring-your-own-observability/rendered/component.yaml
@@ -5,7 +5,7 @@ kind: Service
 metadata:
   name: example-adservice
   labels:
-    helm.sh/chart: opentelemetry-demo-0.33.7
+    helm.sh/chart: opentelemetry-demo-0.33.8
     
     opentelemetry.io/name: example-adservice
     app.kubernetes.io/instance: example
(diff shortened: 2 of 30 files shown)


julianocosta89 (approved):
LGTM!

Should we also think about adding `translation_strategy`?

```
otlp:
  # Ingest OTLP data keeping UTF-8 characters in metric/label names.
  translation_strategy: NoUTF8EscapingWithSuffixes
```

https://prometheus.io/docs/guides/opentelemetry/#utf-8

I'd go for a new PR though, as all dashboards will break 🫠

puckpuck:
> LGTM!
> 
> Should we also think about adding `translation_strategy`?
> 
> ```
> otlp:
>   # Ingest OTLP data keeping UTF-8 characters in metric/label names.
>   translation_strategy: NoUTF8EscapingWithSuffixes
> ```
> 
> https://prometheus.io/docs/guides/opentelemetry/#utf-8
> 
> I'd go for a new PR though, as all dashboards will break 🫠

I also considered adding this but realized it would break all dashboards, so I opted not to. This is absolutely something we should do in a future PR.

TylerHelmuth (approved):

Ron-Norman:
Just install chart version: 0.33.8 and it dashboards are working better but several dashboards have errors or no data. I am trying to determine what the default state should be. Build this out as training environment and want to get fully functional before we proceed to break things on out own :) 

What should be working by default in a default config install? My assumption is the default install should have all metrics enabled and all dashboard panels working. 

Demo Dashboard
- No data for Spanmetrics panels 
- Application logs, both panels have errors 

OpenTelemetry Collector
- Errors on Data Flows (Error: id field is required for nodes data frame) 
- Receivers panels, not data 
- No Data for CPU Usage, Service Instance Count, Uptime by Service Instance, Service Instance Details

Opentelemetry Collector Data Flow
- Prometheus scrape section panels are all working
- Most of the other panels are "no data" or error state

Spanmetrics Demo Dashboard
- all seem to be working other than "no data" on "op 7 Highest Endpoint Latencies Mean Over Range "

PR #1464, opened 2024-12-13
Update kubelet_scrape_config example

jaronoff97:
The previous example was broken because of an escaping issue.


charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/hooks.yaml
```diff
@@ -1,47 +1,5 @@
 ---
 # Source: opentelemetry-kube-stack/templates/hooks.yaml
-apiVersion: v1
```
jaronoff97, line 3: @TylerHelmuth idk why, but these got deleted from my local make run, despite a manual helm template not removing them.

jaronoff97: i think my helm version was outdated and mad at me! 🙇

PR #1478, opened 2024-12-28
Release operator 0.115.0

swiatekm:
We're behind on operator releases. 0.116.0 is out, and the helm chart is still at 0.114.1. Updating to 0.115.0 first, will follow up with 0.116.0.

### DIFF
--- a/charts/opentelemetry-operator/values.yaml
+++ b/charts/opentelemetry-operator/values.yaml
@@ -42,7 +42,7 @@ manager:
     tag: ""
   collectorImage:
     repository: ""
-    tag: 0.114.0
+    tag: 0.115.1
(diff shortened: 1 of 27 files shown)


charts/opentelemetry-operator/values.yaml
```diff
@@ -42,7 +42,7 @@ manager:
     tag: ""
   collectorImage:
     repository: ""
-    tag: 0.114.0
+    tag: 0.115.0
```
jaronoff97, line 45: i think we need to use the [115.1 image](https://github.com/open-telemetry/opentelemetry-collector-releases/releases/tag/v0.115.1) here

jaronoff97 (approved):

PR #1482, opened 2025-01-03
[charts/opentelemetry-demo] Use helm namespace for service.namespace in default helm values

henry-maier:
Currently, service.namespace always gets set to opentelemetry-demo.

This results in confusing metric labels when deploying the chart into a namespace that is not opentelemetry-demo. With this change, service.namespace will be set to the namespace the chart is deployed in.

Should I also bump the chart patch or minor version here?


puckpuck:
This follows the [semantic conventions](https://opentelemetry.io/docs/specs/semconv/resource/#service), which state that `service.namespace` is to represent the group of related services. This is not meant to be the Kubernetes namespace which should be captured as well with the `k8s.namespace.name` attribute.

henry-maier:
Closing per https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1482#issuecomment-2581566175, as the current setup is actually intended.

PR #1508, opened 2025-01-24
fix: remove null creationTimestamp from CRDs to pass schema validation

gazal-k:
`creationTimestamp` would be generated. Setting null does not seem to provide any benefit. It does fail schema validation using tools like https://github.com/yannh/kubeconform. So, remove the field.

### DIFF
--- a/charts/opentelemetry-operator/conf/crds/crd-opentelemetry.io_opampbridges.yaml
+++ b/charts/opentelemetry-operator/conf/crds/crd-opentelemetry.io_opampbridges.yaml
@@ -5,7 +5,6 @@ metadata:
   annotations:
     cert-manager.io/inject-ca-from: {{ include "opentelemetry-operator.webhookCertAnnotation" . }}
     controller-gen.kubebuilder.io/version: v0.16.5
-  creationTimestamp: null
   labels:
     app.kubernetes.io/name: opentelemetry-operator
   name: opampbridges.opentelemetry.io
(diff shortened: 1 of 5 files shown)


dmitryax (approved):

gazal-k:
Looks like the CRDs should be updated in https://github.com/open-telemetry/opentelemetry-operator/tree/main/bundle/community/manifests first 🤔 Shall create a PR for that.

dmitryax:
@gazal-k, yes. If you have a chance, please submit a PR upsteam

gazal-k:
https://github.com/open-telemetry/opentelemetry-operator/tree/main/bundle/ seems to be all generated. It seems to be using `kustomize`, it's not just `creationTimestamp`, but the generated CRDs and other resources seem to have `status` field too, which is puzzling 🤔

e.g:
https://github.com/open-telemetry/opentelemetry-operator/blob/1980f0877e5cff8e41ff3eafafe4c57133d7c899/bundle/community/manifests/opentelemetry.io_instrumentations.yaml#L1886-L1891

```yaml
status:
  acceptedNames:
    kind: ""
    plural: ""
  conditions: null
  storedVersions: null
```

TylerHelmuth (changes requested):
We copy these from the opentelemetry-operator repo. We do a little manipulation via our Makefile command to make them compliant for helm. 

@jaronoff97 @swiatekm can you confirm if removing this field via our make command would be safe?

swiatekm:
It'd be best to fix this in the operator repo. Not sure how easy that is, as we use the kubebuilder toolchain to generate these CRD manifests, and I don't know if they can be made to cooperate.

In the meantime, I don't think removing these from the Helm Chart will have any effect, so feel free to do so if you want.

PR #1512, opened 2025-01-30
Updated documentation for external scrape_configs_file

### LINKED ISSUE #1331: [opentelemetry-kube-stack] `scrape_configs_file` is basically useless
Hi,
Trying out different features in the `opentelemetry-kube-stack` helm chart, and came over the `scrape_configs_file` attribute in the `collectors.collectorName` section. This is basically useless unless the user wants to use the default scrape config that comes with the helm chart `deamon_scrape_configs.yaml` or the user clones the chart and installes the helm chart locally. When installing the helm chart using the remote repository it is not possible to include "external" scrape config files into the helm structure (because of helm...). This is also the case when the chart is used as a subchart, and the scrape config files exists in the parent chart. Quoting the [helm docs](https://helm.sh/docs/chart_template_guide/accessing_files/):

> "Files outside of a Helm application [subchart](https://helm.sh/docs/chart_template_guide/subcharts_and_globals/), including those of the parent, cannot be accessed"

this is a limitation with the `.Files` object in Helm.

meswapnilk:
Closes #1331

### DIFF
--- a/charts/opentelemetry-kube-stack/Chart.yaml
+++ b/charts/opentelemetry-kube-stack/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-kube-stack
-version: 0.3.11
+version: 0.3.12
 description: |
   OpenTelemetry Quickstart chart for Kubernetes.
   Installs an operator and collector for an easy way to get started with Kubernetes observability.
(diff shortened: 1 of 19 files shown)


jaronoff97 (approved):

jaronoff97:
@meswapnilk thank you for your contribution!

PR #1517, opened 2025-02-07
update schema to allow prestop sleep action in lifecycle hook

dschaaff:
https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/3960-pod-lifecycle-sleep-action/README.md

In order to avoid network errors during deployments or pod interruptions, a pre stop sleep is often used to allow time for the async endpoint removal process to complete. Kubernetes supports doing the sleep natively now, removing the need for an exec based sleep command. The values schema for the chart was blocking the use of this feature.

### DIFF
--- a/charts/opentelemetry-collector/values.schema.json
+++ b/charts/opentelemetry-collector/values.schema.json
@@ -484,14 +484,32 @@
                   "type": "string"
                 }
               }
+            },
+            "sleep": {
+              "type": "object",
+              "additionalProperties": false,
+              "properties": {
+                "seconds": {
+                  "type": "integer"
+                }
+              }
             }
           },
           "oneOf": [
             {
-              "required": ["exec"]
+              "required": [
+                "exec"
+              ]
             },
             {
-              "required": ["httpGet"]
+              "required": [
+                "httpGet"
+              ]
+            },
+            {
+              "required": [
+                "sleep"
+              ]
             }
           ]
         }
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.117.0
+version: 0.117.1
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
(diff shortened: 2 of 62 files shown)


charts/opentelemetry-collector/ci/deployment-values.yaml
```diff
@@ -31,3 +31,8 @@ ingress:
         - path: /
           pathType: Prefix
           port: 4318
+
+lifecycleHooks:
+  preStop:
+    sleep:
+      seconds: 10
```
TylerHelmuth, line 38: This should be an optional feature that users can opt into

dschaaff: Makes sense, updated the file.

charts/opentelemetry-collector/values.schema.json
TylerHelmuth, line 1: Defining this in the schema isn't enough. It would allow the values.yaml file to take the new parameters, but the chart won't use them. You'll need to implement some conditional logic in the appropriate files in the `template` folder to achieve want you want.

gjacquet: I think this is fine, the lifecycle configuration is passed directly to `toYaml`. It is only the schema validation preventing us to use `sleep` at the moment: https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/templates/_pod.tpl#L77

dschaaff: @gjacquet Yep. I tested it with the values and it works as expected. It is only the values schema preventing using this. I am currently disabling schema validation for my own use.

TylerHelmuth:
@dschaaff needs another round of merge conflict resolution, but then I'll make this the next collector chart pr merged.

dschaaff:
> @dschaaff needs another round of merge conflict resolution, but then I'll make this the next collector chart pr merged.

@TylerHelmuth Thanks! I resolved all of the conflicts. Do you want to me bump the chart version again or leave that out?

TylerHelmuth:
@dschaaff all merges into main that modify a chart need a chart version bump. Please bump the patch version then regenerate the examples.

TylerHelmuth (approved):

PR #1521, opened 2025-02-12
Add missing variables for endpoint credentials

### LINKED ISSUE #1469: missing variables and doc for endpoint credentials
https://github.com/open-telemetry/opentelemetry-operator/blob/main/cmd/otel-allocator/README.md#service--pod-monitor-endpoint-credentials

The pre-reqs described in that doc are not sufficient. The feature was not working and eventually I noticed in the following in the logs:

`Cert-Manager is not available to the operator, skipping adding to scheme.`

I do indeed have `cert-manager` installed so the log was a bit perplexing. After digging through the code I found this: https://github.com/open-telemetry/opentelemetry-operator/blob/main/internal/autodetect/autodetectutils/utils.go which detects the service account name and *only* does so via env var. This should probably be documented in the above link.

In addition is may be good to have more seamless integration with the helm chart to autoset the env vars.

Alternatively, it may be easier to have the code decode the `token` file and pull the data out similar to reading the data from the `namespace` file.

```yaml
manager:
  env:
    # these are required to make cert-manager detection work properly
    SERVICE_ACCOUNT_NAME: {{ .Release.Name }}
    NAMESPACE: {{ .Release.Namespace }}
```

meswapnilk:
Closes #1469

### DIFF
--- a/charts/opentelemetry-operator/templates/deployment.yaml
+++ b/charts/opentelemetry-operator/templates/deployment.yaml
@@ -93,6 +93,11 @@ spec:
             {{- with .Values.manager.extraEnvs }}
               {{- . | toYaml | nindent 12 }}
             {{- end }}
+            # These are required to make cert-manager detection work properly
+            - name: SERVICE_ACCOUNT_NAME
+              value: {{ .Release.Name }}
+            - name: NAMESPACE
+              value: {{ .Release.Namespace }}
             {{- if .Values.manager.createRbacPermissions }}
             - name: SERVICE_ACCOUNT_NAME
               valueFrom:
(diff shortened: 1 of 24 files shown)


charts/opentelemetry-operator/templates/deployment.yaml
```diff
@@ -93,6 +93,11 @@ spec:
             {{- with .Values.manager.extraEnvs }}
               {{- . | toYaml | nindent 12 }}
             {{- end }}
+            # These are required to make cert-manager detection work properly
+            - name: SERVICE_ACCOUNT_NAME
+              value: {{ .Release.Name }}
+            - name: NAMESPACE
+              value: {{ .Release.Namespace }}
```
TylerHelmuth, line 100: Lets add these only when we know someone is using cert-manager.

PR #1525, opened 2025-02-14
[kube-stack] Add scrapeconfig and probe selectors to targetAllocator prometheusCR schema

lhp-nemlig:
Also adds the scrapeconfigs and probes crds from prometheus-operator 0.74.0.

Now that opentelemetry-operator has been bumped, it should be possible to use the scrapeConfigSelector and probeSelector features of the targetallocator in this chart.


jaronoff97 (approved):
yep, thank you!

Issue #1529, opened 2025-02-17
[Draft] Fix some certificate bugs for webhooks


jvoravong:
Description:
The operator validation webhook can incorrectly use a cert-manager-generated certificate (if a cert-manager exists within the cluster) even when a Helm-generated or user-supplied certificate is specified (admissionWebhooks.certManager.enabled=false). This was likely a slight oversight that just has not been caught yet.

Impact: 
cert-manager can override the intended certificates to be used or cause Helm install/upgrade failures for the operator.

jvoravong:
Fixed with #1530. Also split out a second cert issue into a separate ticket.

PR #1530, opened 2025-02-17
[operator] Fix bug in the operator validation webhook using the wrong certificate

jvoravong:
Related Issue: https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1529
Description:
- Issue: The operator validation webhook can incorrectly use a cert-manager-generated certificate (if a cert-manager exists within the cluster) even when a Helm-generated or user-supplied certificate is specified (`admissionWebhooks.certManager.enabled=false`). This was likely a slight oversight that just has not been caught yet.
- Impact: cert-manager overriding the intended certificates to be used or causing Helm install/upgrade failures for the operator.
- Fix: Ensures that the validation webhook respects the designated certificate source by not including the cert-manager annotation inappropriately, the same way the [mutating webhook does](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/28c3708993c0ff1e57ec1e5f24040e8cd345987a/charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook.yaml#L31).

### DIFF
--- a/charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook.yaml
+++ b/charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook.yaml
@@ -133,8 +133,6 @@ webhooks:
 apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingWebhookConfiguration
 metadata:
-  annotations:
-    cert-manager.io/inject-ca-from: {{ include "opentelemetry-operator.webhookCertAnnotation" . }}
   labels:
     {{- include "opentelemetry-operator.labels" . | nindent 4 }}
     app.kubernetes.io/component: webhook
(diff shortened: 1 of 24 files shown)


charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook.yaml
```diff
@@ -133,8 +133,6 @@ webhooks:
 apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingWebhookConfiguration
 metadata:
-  annotations:
-    cert-manager.io/inject-ca-from: {{ include "opentelemetry-operator.webhookCertAnnotation" . }}
```
jvoravong, line 137: [Please note the sister mutating webhook](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/b8f1228a5718865adf840855e255dd840a282a90/charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook.yaml#L31) in this same file specifically excludes the `cert-manager.io/inject-ca-from` annotation so cert-manager doesn't try to auto-inject a certificate it generated.

TylerHelmuth (approved):

jaronoff97 (approved):

PR #1531, opened 2025-02-17
[Draft] [operator] Fix bug in the operator secret when using a helm hook

jvoravong:
Related Issue: [#1529](https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1529)
Description:
- Issue: When `admissionWebhooks.certManager.enabled=false`, a Helm `pre-install/pre-upgrade` hook deploys the required secret. However, this provides no clear benefit and introduces several downsides:
  - Helm Lifecycle Management: Objects deployed via Helm hooks are not managed by the Helm lifecycle, leading to orphaned secrets on chart uninstallation. Causing potential future installations to fail. It also blocks users from migrating to or from certificate providers like cert-manager.
  - Conflicts with a`dmissionWebhooks.autoGenerateCert.recreate` (still validating): This value regenerates the certificate on install/update but conflicts with the Helm hook behavior.
  - Unnecessary Complexity: Helm already ensures proper installation ordering, making the pre-install hook redundant. ([Reference](https://helm.sh/docs/intro/using_helm/#helm-install-installing-a-package))
  - I can make it so the disabling of the helm hooks with the secret are optional via the values.yaml file, but currently I think this would just be bloating the values.yaml file when we can just remove the hooks from the templates. Taking suggestions from others on this if you have them.
- Fix: Remove the "helm.sh/hook": "pre-install,pre-upgrade" annotation from the operator webhook secret to align with standard Helm behavior.
- Follow up: After some final validations and input from this community, if we decide to move forward I'll add changes for CI/CD changes to pass (lint, render examples, bump the chart version). I have other fixes in flight and will keep this changeset small at the start.

When migrating from a cert-manager cert to a Helm generated or user supplied certificate you can see an error like such.
Upgrading from certmanager to helm generated cert
```
Error: Received unexpected error:
  pre-upgrade hooks failed: warning: Hook pre-upgrade splunk-otel-collector/charts/operator/templates/admission-webhooks/operator-webhook.yaml failed: 1 error occurred: * secrets "sock-operator-controller-manager-service-cert" already exists
```

### DIFF
--- a/charts/opentelemetry-operator/Chart.yaml
+++ b/charts/opentelemetry-operator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.83.0
+version: 0.84.0
 description: OpenTelemetry Operator Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-operator/UPGRADING.md
+++ b/charts/opentelemetry-operator/UPGRADING.md
@@ -1,5 +1,53 @@
 # Upgrade guidelines
 
+## 0.83.0 to 0.84.0
+
+Prior to 0.84.0, this chart always used Helm hooks to pre-install the TLS secret.
+This behavior has changed to make hook usage optional and may impact users supplying their own certificates.
+- **If you do not set custom values for `admissionWebhooks.secretAnnotations`, this upgrade does not
+affect you.** You can ignore these steps.
+- Impact by TLS Certificate Provider Configuration
+  - Cert-Manager Generated Certificate (_"default chart behavior"_)
+    - No impact. No action required.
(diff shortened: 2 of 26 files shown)


jvoravong:
I think this PR is still valid. But, I'll create a commit to make this change optional so the default values will still result in the same default behavior as before (for rendered templates).

Reasons:
- Not all K8s distributions easily support using Helm hooks. If we do not provide a way to disable using the Helm hook to create the secret, users will not be able to deploy this chart using a Helm generated cert or a user supplied cert to environments like EKS that use EKS add-ons to install this chart.
- Helm install ordering isn't a problem here for the most part, secrets will always be installed before the operator related resources are deployed.
- If we stick with using Helm hooks to pre-install the secret, Helm delete operations will not delete these secrets properly.
- If we don't use a Helm hook to install the secret, it can make it much easier for users to migrate between cert providers (cert-manager, Helm template, customer supplied) for upgrades and down grades.

PR #1560, opened 2025-03-05
[otel-operator] Conform to PSS Restricted

jenrik:
Modified the opentelemetry-operator chart to conform with the [Kubernetes Pod Security Standard's Restricted profile](https://kubernetes.io/docs/concepts/security/pod-security-standards/)

- Adds the ability to set `securityContext` on test containers
- Configures reduced privileges for opentelemetry-operator by default


jenrik:
There seems to be something wrong with EasyCLA, i get the following message when trying to get to the CLA:

_The Individual CLA option is not enabled for this project. Please to your administrator to enable the Individual CLA option for this project._

charts/opentelemetry-operator/values.yaml
```diff
@@ -204,11 +204,14 @@ manager:
 
   ## Container specific securityContext
   ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container
-  securityContext: {}
-  # allowPrivilegeEscalation: false
-  # capabilities:
-  #   drop:
-  #   - ALL
+  securityContext:
```
TylerHelmuth, line 207: I believe this would be considered a breaking change and we dont want to do that. I agree that we can make the template/test files use a security context, but we should not set a default value here.

jenrik: Can you be a little more specific on what you believe setting the `securityContext` breaks, because it is not immediately obvious to me. Existing installation should continue work as the operator does not need does extra privileges or capabilities. I have intentionally not made the change for the collector as some of the collector components does need elevated privileges.

TylerHelmuth: > Existing installation should continue work as the operator does not need does extra privileges or capabilities.

@jaronoff97 can you verify this is true. @jenrik have you confirmed the operator installs and operates correctly with this security context?

When I see changes to default values in our values.yaml I normally consider them breaking changes until we can prove that they aren't. We are very careful with how we change default values as it can cause unexpected problems for users performing a `helm upgrade`.

Allex1: This should probably work fine or get caught by our tests in the CI. @jenrik can you pls [generate examples](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/CONTRIBUTING.md) so we can run tests and have CI pass?

jenrik: @TylerHelmuth I have deployed my modified chart to my personal Kubernetes cluster and found no breakage.

@Allex1 generation of examples done

charts/opentelemetry-collector/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.117.3
+version: 0.118.0
```
Allex1, line 3: @jenrik we should not change anything in this pr outside the operator chart. Can you pls delete this as it seems stale.

Allex1 (approved):
@jenrik LGTM but pls delete collector refference and rebase so we can have CI pass

jenrik:
@Allex1 I finished the rebase and coincidentally fixed collector version change

TylerHelmuth (approved):
needs another `make generate-examples CHARTS=opentelemetry-operator`

iblancasa (approved):

yuriolisa (approved):

jenrik:
@TylerHelmuth I have completed the regeneration of examples, and merged the latest changes from main

Allex1:
@TylerHelmuth @jaronoff97 @dmitryax should we merge this ?

jaronoff97 (approved):
Okay after reading this over a bunch and doing a test i think everything should be fine. Let's monitor very closely post release and be sure there are no user issues.

PR #1564, opened 2025-03-10
chore(Collector):! Move to new Telemetry setup

Syndlex:
Endpoint config is deprecated. Move to new Reader Setup

I think this can be a breaking change in some setups. Because when address is manually defined there are 2 Port Bindings on Port.

### DIFF
--- a/charts/opentelemetry-collector/values.yaml
+++ b/charts/opentelemetry-collector/values.yaml
@@ -138,7 +138,12 @@ config:
   service:
     telemetry:
       metrics:
-        address: ${env:MY_POD_IP}:8888
+      readers:
+        - pull:
+            exporter:
+              prometheus:
+                host: '${env:MY_POD_IP}'
+                port: 8888
     extensions:
       - health_check
     pipelines:


jaronoff97:
@Syndlex thanks for opening this! we have an existing PR to handle breaking changes here https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1541

Syndlex:
So this is not needed?

TylerHelmuth:
@jaronoff97 my PR is to switch from Prometheus to OTLP. But in the meantime this PR is useful for switching us from the `address` field to a reader.

Related: https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1337

jaronoff97:
@Syndlex I'm going to take on reviewing this PR. Per Tyler: v0.123.0 of the collector includes a breaking change for `service.telemetry.metrics.address`. I'd love to shepherd this through, let me know if you need any assistance. As far as the approach goes. I think it's fine to split this into two PRs: change the default to use the new telemetry setup, error if we see the old setup. To make this PR mergable you'll need to bump the patch version of the PR by one and run `make generate-examples CHARTS=opentelemetry-collector`

jaronoff97:
actually, per some discussion with tyler, i think we should try to copy over the previous format to the new one, if the new one isn't set. i.e.
```
service:
  telemetry:
    metrics:
      readers:
        - pull:
            exporter:
              prometheus:
                host: <set host from metrics.address>
                port: <set port from metrics.address>
```

and then add a warning to the chart if they're using the old format

jaronoff97:
@Syndlex are you stable able / interested in working on this? If not, i can take it on.

jaronoff97:
@Syndlex no problem. I think what we want to do is make it so that we check if someone is setting the prometheus address configuration via the old setup, and if that's the case we should warn them and replace it with the new method of configuration.

i.e.:
```
service:
  telemetry:
    metrics:
      address: ${env:MY_POD_IP}:8888
```

gets translated to:
```
service:
  telemetry:
    metrics:
      readers:
        - pull:
            exporter:
              prometheus:
                host: ${env:MY_POD_IP}
                port: 8888
```

charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -26,6 +26,18 @@ metrics:
             {{- end }}
 {{- end }}
 
+{{- define "opentelemetry-collector.metrics.prometheus" -}}
+{{/* First replace env: with env! so we can split the host with the port and replace it back later */}}
+{{- $address:= .Values.config.service.telemetry.metrics.address | replace "{env:" "{env!" | split ":" }}
```
jaronoff97, line 31: cleverrrrr, i wish we could more easily test helm out here

jaronoff97: im trying to decide if this is better than checking for the index of the last : and then doing a substr on that though. I think this approach should always work, but it's maybe not as clear? Not sure... @TylerHelmuth thoughts?

charts/opentelemetry-collector/templates/NOTES.txt
```diff
@@ -59,3 +59,7 @@
 {{- if or (and (.Values.internalTelemetryViaOTLP.logs.enabled) (not (or .Values.internalTelemetryViaOTLP.endpoint .Values.internalTelemetryViaOTLP.logs.endpoint))) }}
 {{ fail "[ERROR] internalTelemetryViaOTLP.endpoint or .Values.internalTelemetryViaOTLP.logs.endpoint must be set when internalTelemetryViaOTLP.logs.enabled is true" }}
 {{ end }}
+
+{{- if .Values.config.service.telemetry.metrics }}
+[WARNING] config.service.telemetry.metrics should not be used and will be removed in the future. Use internalTelemetryViaOTLP instead"
```
jaronoff97, line 64: maybe we should make it clear here that we're rewriting this here and we eventually will not?

charts/opentelemetry-collector/templates/_config.tpl
Syndlex: Im not fully understand how `regexFindAll` will help here. Because it will return the substrings not the index.
As far as i understand it would return `["${env","address}","8888"]` or `["address","8888"]`
so reacting on it would be stange.

I had the idear of getting the last index of `:` to. But there is no `findLast` and `index` is not working on strings. Not even `reverse` is working on strings...

charts/opentelemetry-collector/templates/NOTES.txt
Syndlex: Maybe ripping of the band aid not supporting the old format at all and throwing a "Compile error" with fail would be simpler and not as complicated in the long run.

charts/opentelemetry-collector/templates/_config.tpl
TylerHelmuth: This logic is temporary so I am fine with this approach. Looking through the string functions I don't see another way to do this cleanly.

TylerHelmuth (commented):
Lets also add a new scenario in the `ci` folder to test our logic

charts/opentelemetry-collector/values.yaml
```diff
@@ -173,6 +173,7 @@ config:
     telemetry:
       metrics:
           # if internalTelemetryViaOTLP.metrics.enabled = true, this address will be removed and an otel go sdk will be configured to emit otlp
+          # Deprecated you should use internalTelemetryViaOTLP
```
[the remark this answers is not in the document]
Syndlex: Done

charts/opentelemetry-collector/templates/NOTES.txt
```diff
@@ -59,3 +59,7 @@
 {{- if or (and (.Values.internalTelemetryViaOTLP.logs.enabled) (not (or .Values.internalTelemetryViaOTLP.endpoint .Values.internalTelemetryViaOTLP.logs.endpoint))) }}
 {{ fail "[ERROR] internalTelemetryViaOTLP.endpoint or .Values.internalTelemetryViaOTLP.logs.endpoint must be set when internalTelemetryViaOTLP.logs.enabled is true" }}
 {{ end }}
+
+{{- if .Values.config.service.telemetry.metrics.address }}
+[WARNING] config.service.telemetry.metrics.address should not be used and will be removed in the future. Use internalTelemetryViaOTLP instead"
```
TylerHelmuth, line 64:
```suggestion
[WARNING] config.service.telemetry.metrics.address should not be used and will be removed in the future. Use internalTelemetryViaOTLP or config.service.telemetry.metrics.readers instead"
```

charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -53,8 +63,14 @@ logs:
 {{- $_ := set $config.service.pipelines.metrics "receivers" (mustWithout $config.service.pipelines.metrics.receivers "prometheus") }}
 {{- $_ := unset $config.service.telemetry.metrics "address" }}
```
TylerHelmuth, line 64: The logic in the IF above also needs adjusted to unset the new default prometheus reader when internalTelemetryViaOTLP.metrics is enabled

charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -53,8 +63,14 @@ logs:
...
 {{- $_ := set $config.service "telemetry" (mustMerge $config.service.telemetry (include "opentelemetry-collector.otelsdkotlp.metrics" . | fromYaml)) }}
+{{- else if .Values.config.service.telemetry.metrics.address}}
+{{/* First replace env: with env! so we can split the host with the port and replace it back later */}}
+{{- $address:= .Values.config.service.telemetry.metrics.address | replace "{env:" "{env!" | split ":" }}
+{{- $_ := unset $config.service.telemetry.metrics "address" }}
+{{- $_ := set $config.service "telemetry" (mustMerge $config.service.telemetry (include "opentelemetry-collector.metrics.prometheus" $address | fromYaml)) }}
 {{- end }}
 {{- if .Values.internalTelemetryViaOTLP.logs.enabled }}
+{{- $_ := unset $config.service.telemetry.metrics "address" }}
```
TylerHelmuth, line 73: I don't believe we need to mess with the metrics address in this logs scenario.

charts/opentelemetry-collector/templates/_config.tpl
Syndlex: When I generate the examples for internalTelemetryViaOTLP i don't get any changes in the example. 

Shouldn't this be the default behavior? 

Do you mean: `internalTelemetryViaOTLP.metrics.enabled` and `.Values.config.service.telemetry.metrics.address` should be working on the same time?

charts/opentelemetry-collector/templates/_config.tpl
Syndlex: sry slipped in :(

charts/opentelemetry-collector/templates/_config.tpl
TylerHelmuth: I mean that when `internalTelemetryViaOTLP.metrics.enabled=true` that 

```yaml
  service:
    telemetry:
      metrics:
        readers:
          - pull:
              exporter:
                prometheus:
                  host: ${env:MY_POD_IP}
                  port: 8888
```
should be replaced by the otlp exporter for metrics. In the generated example you can see the otlp export is now missing.

TylerHelmuth (commented):
@Syndlex the chart's minor version needs bumped then regenerate the examples and we'll be good to go

jaronoff97 (approved):

jaronoff97:
Thank you for your contribution @Syndlex :bow:

PR #1563, opened 2025-03-10
[demo] - load-generator Adapt OpenFeature OFREP instead of IN_PROCESS

aepfli:
OpenFeature supports multiple different modes.
The newest implementation relies on threads and other python internal, which seem to be monkeypatched by locust.
Due to this, the IN_PROCESS mode is not working properly anymore and is causing issues.

Instead we suggest to switch to the OFREP (OpenFeature Remote Evaluation Protocol) for this usecase. This is a simple http call which will return all evaluate flags, and will regularly poll for changes.

This pr contains the helm changes matching https://github.com/open-telemetry/opentelemetry-demo/pull/2114 - but my helm knowledge is limited, please let me know, if i am missing anything here.

PR #1571, opened 2025-03-14
[demo] increase flagd-ui memory limits

rlankfo:
flagd kept getting OOMkilled in my kind cluster.

See https://github.com/open-telemetry/opentelemetry-demo/pull/2120 as well.

### DIFF
--- a/charts/opentelemetry-demo/Chart.yaml
+++ b/charts/opentelemetry-demo/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: opentelemetry-demo
-version: 0.36.3
+version: 0.36.4
 description: opentelemetry demo helm chart
 home: https://opentelemetry.io/
 sources:
--- a/charts/opentelemetry-demo/values.yaml
+++ b/charts/opentelemetry-demo/values.yaml
@@ -614,7 +614,7 @@ components:
             value: http://$(OTEL_COLLECTOR_NAME):4318
         resources:
           limits:
-            memory: 75Mi
+            memory: 100Mi
         volumeMounts:
           - name: config-rw
             mountPath: /app/data
--- a/charts/opentelemetry-demo/examples/bring-your-own-observability/rendered/component.yaml
+++ b/charts/opentelemetry-demo/examples/bring-your-own-observability/rendered/component.yaml
@@ -5,7 +5,7 @@ kind: Service
 metadata:
   name: ad
   labels:
-    helm.sh/chart: opentelemetry-demo-0.36.3
+    helm.sh/chart: opentelemetry-demo-0.36.4
     
     opentelemetry.io/name: ad
     app.kubernetes.io/instance: example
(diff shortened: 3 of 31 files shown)


charts/opentelemetry-demo/values.yaml
```diff
@@ -614,7 +614,7 @@ components:
             value: http://$(OTEL_COLLECTOR_NAME):4318
         resources:
           limits:
-            memory: 75Mi
+            memory: 160Mi
```
puckpuck, line 617: This should be 100Mi to match what is in the Demo as per this [PR](https://github.com/open-telemetry/opentelemetry-demo/pull/2120)

```suggestion
            memory: 100Mi
```

puckpuck (changes requested):

charts/opentelemetry-demo/values.yaml
rlankfo: @puckpuck I was still getting OOMs in k8s at 100Mi

rlankfo: image

rlankfo: 100Mi to 160Mi might have been more a jump than necessary -- I just didn't want to have to keep dealing with it at that moment. 😄 

It seems stable for me at 120Mi as well after more testing. Does this value sound reasonable?

puckpuck: I'm interested to know what kind of hardware you are running the Demo on to get OOM still? Under normal conditions flagd-ui should consume less than 60M

Also this field needs to be the same value as what we have in the Demo repo so we don't drift away in configuration. Right now the Demo repo has this value at 100M

rlankfo: Let's set it to 100M and I can update my own config to increase it locally. I'm using an arm Mac (m3 max).

jvoravong:
Bumping this PR to raise visibility and priority to unblock other PRs.

puckpuck (approved):

puckpuck:
This is ready to merge and needed so we can also merge #1563 and finalize the Demo 2.0.2 release.

julianocosta89:
@open-telemetry/helm-maintainers may we get this one merged?

charts/opentelemetry-demo/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: opentelemetry-demo
-version: 0.36.1
+version: 0.36.2
```
jvoravong, line 4: Lately, the merging order has been tricky for the demo chart updates. One of my changes was merged in first and used version [0.63.3](https://github.com/open-telemetry/opentelemetry-helm-charts/releases/tag/opentelemetry-demo-0.36.3) (which was unexpected but manageable). @rlankfo can you bump this to 0.36.4 and generate the examples? This way we get the correct latest tag applied to the correct chart release, we would skip 0.36.2.
```suggestion
version: 0.36.4
```

rlankfo: done! thanks @jvoravong

jaronoff97:
merging to unblock! thanks for your patience here!

PR #1588, opened 2025-03-20
[operator] always include SERVICE_ACCOUNT_NAME env var

ItielOlenick:
The environment variable SERVICE_ACCOUNT_NAME is needed by the operator in order to auto detect permissions.
Up till now it was set only if `manager.createRbacPermissions=true` but it is also needed to auto detect cert manager permissions for enabling the mTLS communication between the TA and the Collector.

This moves the environment variable to always be present so the operator can preform the necessary checks.

closes https://github.com/open-telemetry/opentelemetry-operator/issues/3811

### DIFF
--- a/charts/opentelemetry-operator/templates/deployment.yaml
+++ b/charts/opentelemetry-operator/templates/deployment.yaml
@@ -84,8 +84,12 @@ spec:
             {{-  end  }}
           command:
             - /manager
-          {{- if or .Values.manager.env .Values.manager.extraEnvs .Values.manager.createRbacPermissions }}
           env:
+            - name: SERVICE_ACCOUNT_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.serviceAccountName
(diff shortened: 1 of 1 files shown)


jaronoff97 (commented):
@ItielOlenick thank you for the fix! Can you also bump the patch version of the chart and run `make generate-examples CHARTS=opentelemetry-operator`?

ItielOlenick:
@jaronoff97 Sure thing!

charts/opentelemetry-operator/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.83.0
+version: 0.84.0
```
jaronoff97, line 3: This should be the patch `0.83.1` be sure to run the make after this

ItielOlenick: Got it.
How about now?

jaronoff97 (approved):
Thank you @ItielOlenick 🙇

PR #1592, opened 2025-03-24
feat(opentelemetry-operator): Removing required properties in JSON schema

jtyr:
This PR is making some of the top level properties not required as they hold no value by default and they do not require any value to be set. Not having these properties as non-required makes it possible to set the property to `null` or `{}` value that resets any value values set for example in a wrapping chart.


jtyr:
@Allex1 Please could you review this PR?

charts/opentelemetry-operator/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.84.0
```
jaronoff97, line 3: I think we can just bump the patch version instead of the minor version here.

Allex1 (approved):
@jtyr this sounds reasonable, pls rebase first

jaronoff97 (approved):

jaronoff97:
thanks for your contribution!! :D 🎉

PR #1601, opened 2025-04-01
[operator] Fix bug where the operator can be scheduled on windows nodes

jvoravong:
- Fix bug where the operator can be scheduled on windows nodes and cause pull image errors. The operator docker image doesn't support windows nodes.
- Tested locally via Minikube by deploying the operator to a Linux node.
- https://github.com/open-telemetry/opentelemetry-operator/issues/642

### DIFF
--- a/charts/opentelemetry-operator/values.yaml
+++ b/charts/opentelemetry-operator/values.yaml
@@ -339,7 +339,8 @@ clusterRole:
 
 affinity: {}
 tolerations: []
-nodeSelector: {}
+nodeSelector:
+  kubernetes.io/os: linux
 topologySpreadConstraints: []
 hostNetwork: false
 
--- a/charts/opentelemetry-operator/Chart.yaml
+++ b/charts/opentelemetry-operator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.84.1
+version: 0.84.3
 description: OpenTelemetry Operator Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-operator/Chart.yaml
+++ b/charts/opentelemetry-operator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.86.0
+version: 0.86.1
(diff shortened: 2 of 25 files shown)


charts/opentelemetry-operator/values.yaml
```diff
@@ -339,7 +339,8 @@ clusterRole:
 
 affinity: {}
 tolerations: []
-nodeSelector: {}
+nodeSelector:
+  kubernetes.io/os: linux
```
jaronoff97, line 343: this feels like maybe too big an assumption IMO. although it should always be the case that a node has this, i'm not sure it's a full requirement that it does. From the docs it's stated to be a [well known label](https://kubernetes.io/docs/reference/labels-annotations-taints/#kubernetes-io-os), but i'm not sure that means it's always present. Is there something more concrete that ensures we can rely on this always?

jaronoff97: definitely possible i'm incorrect here, and happy to be proven wrong.

jinja2: The os label is added by the kubelet to the node object when it is being readied for registration with the API - [ref](https://github.com/kubernetes/kubernetes/blob/v1.32.0/pkg/kubelet/kubelet_node_status.go#L309). Imo, it should always exists unless using some non-std kubelet.

If we are still hesitant about the nodeselector, another option would be to set a best effort affinity rule like following

```
spec:
  affinity:
    nodeAffinity:
      preferredDuringSchedulingIgnoredDuringExecution:
      - weight: 1
        preference:
          matchExpressions:
          - key: kubernetes.io/os
            operator: In
            values:
            - linux
```

jaronoff97: ah thats the exact ref i need to make me a bit more confident here, thanks for pointing me there!

jaronoff97 (approved):

Allex1 (approved):

jvoravong:
@Allex1 The last reviewer jinja2 is out sick the start of this week. She's on my team and she would approve these changes if she had working working hours. I think this PR is good as is to merge today if we want, or we can wait for Jina's review later this week. I just don't want to hold up other operator PRs if possible.

jinja2 (approved):

jaronoff97:
Np! Looking good to me, thank you! :bow:

PR #1605, opened 2025-04-02
[opentelemetry-kube-stack] bump operator and crds charts

rogercoll:
- Bumps operator chart dependency to 0.84.1
- Updates operator crds


charts/opentelemetry-kube-stack/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-kube-stack
-version: 0.4.5
+version: 0.4.6
```
jaronoff97, line 3: this should be a minor version change given we're updating the operator here.

rogercoll: makes sense to me, fixed https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1605/commits/d950f8e7967f4b13d76a56108f52b0ca223aaf89

jaronoff97 (approved):

Issue #1616, opened 2025-04-07
[opentelemetry-operator] Option to disable automountServiceAccountToken


jinja2:
Kubernetes security benchmarks recommend to have automounting of service account tokens disabled by default. The operator chart at present does not provide an option to disable the `automountServiceAccountToken` on service account and workload objects.

For example, when deploying the operator to an AKS cluster with gatekeeper/defender running, we see below security issues ([link](https://www.azadvertizer.net/azpolicyadvertizer/kubernetes_block-automount-token.html) to the policy for further details)

![Image](https://github.com/user-attachments/assets/7fdb8d38-99df-46e9-9c20-4906adae24fb)

To solve this issue, I would like to submit a PR which allows users to change the value of the `automountServiceAccountToken` option. Additionally, with automountServiceAccountToken = false, for the operator pod we'd have to manually mount service account token as a projected volume just like how k8s does it when the automount option is enabled. I am considering adding an option to the `manager` called `explicitMountServiceAccountToken` which when set to true (default will be false to preserve current behavior), the chart disables automount and adds the required volume/volumeMount for the kube api creds.

PR #1617, opened 2025-04-08
feat(opentelemetry-operator): Add options for mounting service account token

jinja2:
This PR addresses issue [#1616 ](https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1616)

The PR adds 2 new options -
* explicitMountServiceAccountToken - When enabled, operator pod spec will have `automountServiceAccountToken` set to `false`, and the token will instead be explicitly mounted as volume. The projected volume config added to the chart reflects the default used by k8s when automounting is enabled as seen [here](https://github.com/kubernetes/kubernetes/blob/v1.32.3/plugin/pkg/admission/serviceaccount/admission.go#L485C1-L523C3) and the expiration period value is from the default [here](https://github.com/kubernetes/kubernetes/blob/v1.32.3/pkg/serviceaccount/claims.go#L39).
* manager.serviceAccount.automountServiceAccountToken - Sets the automountServiceAccountToken config of service account.

The default behavior remains the same with these changes.

I also tested these changes in an AKS cluster which had the Microsoft defender enabled with the addon `azure-policy` running. With the default config the operator pod can not be created with below error -

image

But when run with `--set explicitMountServiceAccountToken=true`, the operator pod comes up without an admission policy issues.


charts/opentelemetry-operator/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.84.2
+version: 0.85.0
```
jaronoff97, line 3: this should be a patch version change

charts/opentelemetry-operator/values.yaml
```diff
@@ -354,6 +357,12 @@ securityContext:
   runAsUser: 65532
   fsGroup: 65532
 
+# -- Determines how the operator pod mounts the service account token.
+# When set to true, automounting of the service account token is disabled, and the token is manually mounted
+# using a volume and volumeMounts in the pod spec.
+# This is useful for security hardening, as it prevents the pod from automatically having access to the service account token unless explicitly mounted.
+explicitMountServiceAccountToken: false
```
jaronoff97, line 364: is there a case when this would be true and `automountServiceAccountToken` would be true? From my understanding i think it would be that these are mutually exclusive?

charts/opentelemetry-operator/Chart.yaml
jinja2: done

charts/opentelemetry-operator/values.yaml
jaronoff97: IMO we should keep `automountServiceAccountToken` and remove this flag then. should we also allow a user to supply their own mount rather than inferring it? I would assume anyone setting this false would know to do this.

jinja2: > IMO we should keep automountServiceAccountToken and remove this flag then
Ack, yeah I'll update the flag. My main reason for not keeping the flag same as `automountServiceAccountToken` was to make it clear to the user that there chart config will do more than just updating the value of this pod spec option.

> should we also allow a user to supply their own mount rather than inferring it?
I think the chart adding the vol/mounts is better for user experience. The alternative would require a user to add `volumes` to the pod and `volumeMounts` for both the manager and kube-rbac-proxy containers in addition to setting `automountServiceAccountToken` to false. The chart at present does not have any option to add extra volumes and mounts, so we'll be adding these.

If you prefer the more do-it-yourself approach, I am ok updating the chart accordingly. Lmk.

jaronoff97: Yeah let's just have the chart do it then

jinja2: update the flag name, thanks!

charts/opentelemetry-operator/values.yaml
```diff
@@ -354,6 +357,12 @@ securityContext:
   runAsUser: 65532
   fsGroup: 65532
 
+# -- Determines how the operator pod mounts the service account token.
+# When set to false, automounting of the service account token is disabled, and the token is manually mounted
+# using a volume and volumeMounts in the pod spec by the chart.
+# This is useful for security hardening, as it prevents the pod from automatically having access to the service account token unless explicitly mounted.
+automountServiceAccountToken: true
```
jaronoff97, line 364: do we need both of these?

jinja2: I wouldn't say we "need" two options for service account vs workload. I was going for more flexibility for any advanced use cases. But unifying both object's setting with one config for simplicity would also be okay.

jaronoff97: Yeah i think we should just have one to keep it easy IMO

jinja2: done!

Allex1:
@jinja2 can you please rebase/resolve conflicts ?

jinja2:
@Allex1 conflicts have been fixed, thanks!

jinja2:
Updated the version chart version as 0.86.1 was already merged

TylerHelmuth (approved):

PR #1630, opened 2025-04-15
fix(collector): add support for service trafficDistribution

sjoukedv:
Add support for https://kubernetes.io/docs/concepts/services-networking/service/#traffic-distribution


charts/opentelemetry-collector/values.yaml
```diff
@@ -510,6 +510,7 @@ service:
   # enabled: true
 
   type: ClusterIP
+  # trafficDistribution: PreferClose
```
TylerHelmuth, line 513: Can you check if the json schema file needs updated to allow this new key

TylerHelmuth (commented):
Please merge in the latest changes from main, bump the chart's patch version, and then run `make generate-examples CHARTS=opentelemetry-collector`

charts/opentelemetry-collector/values.yaml
sjoukedv: You're right this was missing, added.

sjoukedv:
> Please merge in the latest changes from main, bump the chart's patch version, and then run `make generate-examples CHARTS=opentelemetry-collector`

Done, can you review it again?

Issue #1642, opened 2025-04-20
K8s attributes labels disappeared in latest chart


dima-sh-papaya:
Hi, we're using the [opentelemetry-collector](https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-collector) chart with K8s attributes enabled,
```
presets:
  kubernetesAttributes:
    enabled: true
```

after updating to latest version, the extra labels from prometheus metrics have disappeared, so in our other clusters I can still see those labels and in the one I upgraded we don't have them. Here is the example: 
**Before**:
```
otelcol_process_cpu_seconds{attributes="k8s.namespace.name=\"xyz\", k8s.deployment.name=\"xyz\", k8s.container.name=\"otel-collector\", k8s.pod.name=\"xyz\", k8s.node.name=\"xyz\"",service_instance_id="xyz",service_name="otelcol-contrib",service_version="0.121.0"} 97.13
```
**After**:
```
otelcol_process_cpu_seconds_total{otel_scope_name="go.opentelemetry.io/collector/service",otel_scope_version=""} 3.26
```

The labels are different, but config remains the same (Helm values for the chart), how can we find the upgrade guide, for required config changes? I scrolled a few latest commits in the repo but cannot see other changes except the `telemetry.metrics.address. -> .readers` one.  Also nothing in the [upgrade guide](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/UPGRADING.md) about that. 

The version is: 
```
info	service@v0.123.0/service.go:264	Starting otelcol-contrib...	{"Version": "0.123.0", "NumCPU": 4}
```

dima-sh-papaya:
The related config part looks like this

```
  service:
    telemetry:
      metrics:
        readers:
        - pull:
            exporter:
              prometheus:
                host: $${env:MY_POD_IP}
                port: 8888
      resource:
        attributes: >-
          k8s.namespace.name="xyz",
          k8s.deployment.name="xyz",
          k8s.container.name="otel-collector",
          k8s.pod.ip="$${env:MY_POD_IP}",
          k8s.node.name="$${env:K8S_NODE_NAME}"
```

dima-sh-papaya:
updated the image to a new repo as shown in changelog readme file, now have a single metric that has the attributes:

```
[...]

PR #1645, opened 2025-04-24
[opentelemetry-collector] Fix bug converting from metrics.address to otel sdk prometheus exporter 

TylerHelmuth:
While investigating https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1642 I found a bug with our metrics address conversion.

Closes https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1642

### DIFF
--- a/charts/opentelemetry-collector/examples/old-metrics-url-values/rendered/configmap.yaml
+++ b/charts/opentelemetry-collector/examples/old-metrics-url-values/rendered/configmap.yaml
@@ -6,7 +6,7 @@ metadata:
   name: example-opentelemetry-collector
   namespace: default
   labels:
-    helm.sh/chart: opentelemetry-collector-0.122.2
+    helm.sh/chart: opentelemetry-collector-0.122.3
     app.kubernetes.io/name: opentelemetry-collector
     app.kubernetes.io/instance: example
     app.kubernetes.io/version: "0.123.1"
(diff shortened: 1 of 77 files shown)


charts/opentelemetry-collector/examples/old-metrics-url-values/rendered/configmap.yaml
```diff
@@ -87,5 +87,5 @@ data:
           - pull:
               exporter:
                 prometheus:
-                  host: ${env:MY_POD_IP}
+                  host: example.com
```
TylerHelmuth, line 90: This was the bug. Users that had set a specific value for `metrics.address` were not getting it converted properly.

charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -32,8 +32,15 @@ metrics:
...
-            host: {{ ._0 | replace "{env!" "{env:" }}
-            port: {{ ._1 }}
+            host: {{ .address._0 | replace "{env!" "{env:" }}
+            port: {{ .address._1 }}
+            {{- if .Values.config.service.telemetry.resource }}
+            with_resource_constant_labels:
+              included:
+              {{- range (keys .Values.config.service.telemetry.resource) }}
+              - {{ println . }}
+              {{- end }}
```
TylerHelmuth, line 42: This handles a breaking change that happened with the otel sdk's prometheus exporter where resource attributes were not included by default

jaronoff97 (approved):

Issue #1535, opened 2025-02-20
[otel-kube-stack] Parameterize installing only OTEL CRDs


Bemesko:
My current use case for the otel-kube-stack is installing it alongside an environment that already has kube-prometheus stack. I'm doing this for backwards-compatibility reasons but intend to start like this to fully migrate to otel, eventually.

The otel-kube-stack comes with otel's own CRDs but it also bundles some of the same CRDs kube-prometheus-stack managed. This creates some conflicts in my setup because I now have 2 charts that manage the same CRDs and if I wanted both I would either need to keep the versions of both charts in sync or enable only the CRDs for opentelemetry by setting the `crds.create` parameter to `false` and also installing the `opentelemetry-operator` chart with everything except the CRDs enabled. There's also the option of applying the CRDs myself but I'd like to avoid that since I manage my environment using GitOps.

(All of this is my impression as far as I've researched, please correct me if I'm wrong or if there is another way to do what I want!)

I'd like to sugest adding a parameter to disable installing the `monitoring.coreos` CRDs in the `otel-kube-stack` Helm chart. I'm not familiar with how flexible the bundling system for CRDs is in Helm but I know it is weird about them in general, so I'm not entirely sure whether this is possible.

Thanks in advance!

Sunilwali679:
@Bemesko could you please assign this task to me ?

TylerHelmuth:
@jaronoff97 this idea sounds ok to me, do you agree?

jaronoff97:
Yep definitely reasonable and acceptable! Thanks for opening this.

Aransh:
Hey, have the same exact issue, I see https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1583 by @yuriolisa was closed, is this still planned?

jaronoff97:
@Aransh it seems the PR got auto-closed after being stale, would you want to contribute the fix? I'd be happy to review and merge it.

Aransh:
> @Aransh it seems the PR got auto-closed after being stale, would you want to contribute the fix? I'd be happy to review and merge it.

Sure, I'll handle this tomorrow, thanks for the quick response!

jaronoff97:
@Aransh thanks for your contribution and work here! Much appreciated. :bow:

PR #1667, opened 2025-05-13
[opentelemetry-kube-stack] CRD options

Aransh:
This PR allows selectively disabling CRDs (openTelemetry or monitoringCoreOS).
This resolves conflicts for anyone already installing the monitoringCoreOS CRDs separately.

Due to helm restrictions (CRDs directory is not templated, dependency chart condition can not use logic such as "and"), the crds sub chart and the "crds.install" value have been split into 2.

Resolves https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1535


charts/opentelemetry-kube-stack/README.md
```diff
@@ -148,3 +148,10 @@ $ helm show values open-telemetry/opentelemetry-kube-stack
 ```
 
 When using this chart as a subchart, you may want to unset certain default values. Since Helm v3.13 values handling is improved and null can now consistently be used to remove values (e.g. to remove the default CPU limits).
+
+
+# Upgrading
+
+## From 0.5.x to 0.6.x
+
+Value "crds.install" has been removed in favor of "crds.openTelemetry" and "crds.monitoringCoreOS" to allow selectively disabling them.
```
jaronoff97, line 157: i think we could deprecate the crds.install field and have it take precedence to the new fields in order to not break existing installations. I think this would work where we have [condition](https://helm.sh/docs/topics/charts/#tags-and-condition-fields-in-dependencies): crds.install,crds.openTelemetry and then you can just remove the old field, this will make it so that if a user hasn't upgraded their values and is setting crds.install explicitly, we will use whatever they set, otherwise we'll use the new defaults. Please test this out, but i think this should work and prevent a breaking change.

charts/opentelemetry-kube-stack/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-kube-stack
-version: 0.5.2
+version: 0.6.0
```
jaronoff97, line 3: please run `make generate-examples CHARTS=opentelemetry-kube-stack`

charts/opentelemetry-kube-stack/Chart.yaml
```diff
@@ -18,9 +18,12 @@ icon: https://raw.githubusercontent.com/cncf/artwork/a718fa97fffec1b9fd14147682e
 # release, the collector's latest patch will be manually overridden.
 appVersion: 0.120.0
 dependencies:
-  - name: crds
+  - name: openTelemetry-crds
     version: "0.0.0"
-    condition: crds.install
+    condition: crds.openTelemetry
+  - name: monitoringCoreOS-crds
```
jaronoff97, line 24: i don't love the names here, usually you don't see charts with capitalization. we can name them otel-crds and prometheus-crds, as they're not actually exposed externally to this chart.

charts/opentelemetry-kube-stack/README.md
Aransh: Oh I see, so helm uses the first field *That it finds*, nice find, this is exactly what I was trying to achieve, applied

charts/opentelemetry-kube-stack/Chart.yaml
```diff
@@ -18,9 +18,12 @@ icon: https://raw.githubusercontent.com/cncf/artwork/a718fa97fffec1b9fd14147682e
 # release, the collector's latest patch will be manually overridden.
 appVersion: 0.120.0
 dependencies:
-  - name: crds
+  - name: otel-crds
     version: "0.0.0"
-    condition: crds.install
+    condition: crds.install,crds.openTelemetry
```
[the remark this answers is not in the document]
Aransh: Yeah, kind of thought of that too but guess I was basing too much on the original PR this was based on, anyway, changed.

And yes, I was able to test this using:
```
helm template test charts/opentelemetry-kube-stack --include-crds
helm template test charts/opentelemetry-kube-stack --include-crds --set crds.install=false
helm template test charts/opentelemetry-kube-stack --include-crds --set crds.installOtel=false
helm template test charts/opentelemetry-kube-stack --include-crds --set crds.installPrometheus=false
```

They all returned the expected result

jaronoff97 (approved):

jaronoff97:
this looks good to me, thank you for your contribution! Going to wait on more review and then merge :D

Aransh:
Thanks @jaronoff97 :)
Fixed the chart.lock so lint passes, please let me know if anything else is needed from my side

jaronoff97 (approved):

Aransh:
Rebased to align with 0.5.3 changes.
@TylerHelmuth would appreciate your input here 🙏

jaronoff97 (approved):

charts/opentelemetry-kube-stack/values.yaml
```diff
@@ -26,8 +26,10 @@ cleanupJob:
 
 # Should the CRDs be installed by this chart.
 crds:
-  # Should the CRDs be installed
-  install: true
+  # Control whether the opentelemetry.io CRDS should be installed.
+  installOtel: true
+  # Control whether the monitoring.coreos CRDS should be installed.
+  installPrometheus: true
```
TylerHelmuth, line 32: This will break anyone who is setting `crds.install: false`. 

Can we do this in several steps where we keep support for `crds.install` and if that is set to false both crd charts are not installed? We can include a warning in the NOTES.txt that this field will be dropped later. Also something in the UPGRADING.md file.

jaronoff97: @TylerHelmuth this actually shouldn't break existing users, because of the way conditions work and how we've set up Chart.yaml, the initial condition check will be false and it will remain disabled

Aransh: Yup, as @jaronoff97 stated.
Tested manually with "helm template" this continues to work as intended, whether you use the old or new values.

Aransh: @TylerHelmuth I originally did add upgrading instructions, but following @jaronoff97's notes moved to the multi-condition approach that keeps backwards compatibility. See https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1667#pullrequestreview-2836621144

TylerHelmuth (approved):

PR #1657, opened 2025-05-01
[charts/target-allocator] new helm chart for the target allocator deployment

atoulme:
Fixes #1656

PR #1675, opened 2025-05-14
Example ta collector deployments

atoulme:
Built on top of #1657 - this adds a bit more information on how you can realistically use the helm chart with a set of prometheus targets and collectors.

### DIFF
--- a/charts/opentelemetry-target-allocator/Chart.yaml
+++ b/charts/opentelemetry-target-allocator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-target-allocator
-version: 0.126.0
+version: 0.126.1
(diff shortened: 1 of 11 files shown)


atoulme:
I suppose #1694 is required first

jaronoff97:
yep, needs a rebase

TylerHelmuth:
@atoulme all commits to main that change a chart are automatically released. Since you changes a chart, it needs a version bump.

TylerHelmuth:
https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/CONTRIBUTING.md#versioning

atoulme:
> @atoulme all commits to main that change a chart are automatically released. Since you changes a chart, it needs a version bump.

but I'm just changing an example?

atoulme:
ok bumping chartVersion

PR #1678, opened 2025-05-16
Fix #1582

### LINKED ISSUE #1582: [opentelemetry-kube-stack] autoscaler not working in this chart
In `opentelemetry-kube-stack` 
```
  {{- if (not (eq $collector.mode "daemonset" )) }}
  replicas: {{ $collector.replicas }}
  {{- end }}
```
Removing replicas in collector definition in chart still not helping because default is 1 and we will have always 1.
Even when set `autoscaler` minReplicas all try to sync but replicas from collector is set instead and HPA cannot change that.
This means actually in this chart HPA newer work.

in `opentelemetry-collector` chart I see proper handling of this
```
  replicas: {{ .Values.replicaCount | default 1 }}
{{- end }}
```

szibis:
Fix HPA to work in kube-stack Operator based setup because now it is blocked by static replicacount setting.

https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1582

### DIFF
--- a/charts/opentelemetry-kube-stack/Chart.yaml
+++ b/charts/opentelemetry-kube-stack/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-kube-stack
-version: 0.5.2
+version: 0.5.3
 description: |
   OpenTelemetry Quickstart chart for Kubernetes.
   Installs an operator and collector for an easy way to get started with Kubernetes observability.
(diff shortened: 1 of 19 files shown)


jaronoff97 (commented):
change is fine, can you bump the patch version of the chart and re-run `make generate-examples CHARTS=opentelemetry-kube-stack`

jaronoff97 (approved):

jaronoff97:
thank you for your contribution! :bow:

PR #1682, opened 2025-05-19
Add k8sLeaderElector to clusterMetrics preset

zparnold:
This pull request introduces enhancements to the OpenTelemetry Collector Helm chart to support high availability for cluster metrics collection using the `k8sleaderelector` extension. The changes include documentation updates, configuration examples, validation improvements, and support for leader election in multi-replica setups.

### High Availability and Leader Election Support:
* Added documentation in `charts/opentelemetry-collector/README.md` to explain how to configure the `k8sleaderelector` extension for high availability when running multiple replicas. This ensures only one collector instance collects cluster metrics at a time.
* Introduced a new example configuration in `charts/opentelemetry-collector/examples/clusteMetricsWithLeaderElector/values.yaml` to demonstrate the use of `k8sleaderelector` with deployment mode and multiple replicas.

### Validation and Configuration Updates:
* Updated `charts/opentelemetry-collector/templates/NOTES.txt` to include validation for the `k8sLeaderElector` extension and warnings about potential duplicate metrics when running multiple replicas without leader election.
* Modified `charts/opentelemetry-collector/templates/_config.tpl` to conditionally add the `k8s_leader_elector` configuration to the Kubernetes Cluster Receiver based on the `k8sLeaderElector` value.
* Updated `charts/opentelemetry-collector/values.yaml` to include a new `k8sLeaderElector` field under `clusterMetrics` for specifying the leader election extension ID. Adjusted comments to reflect support for deployment, statefulset, and daemonset modes.

### DIFF
--- a/charts/opentelemetry-collector/templates/_config.tpl
+++ b/charts/opentelemetry-collector/templates/_config.tpl
@@ -204,6 +204,9 @@ receivers:
 receivers:
   k8s_cluster:
     collection_interval: 10s
+    {{- if and .Values.presets.clusterMetrics.k8sLeaderElector (regexMatch "^[^[:space:]]+$" .Values.presets.clusterMetrics.k8sLeaderElector) }}
+    k8s_leader_elector: {{ .Values.presets.clusterMetrics.k8sLeaderElector }}
+    {{- end }}
 {{- end }}
 
 {{- define "opentelemetry-collector.applyKubeletMetricsConfig" -}}
--- a/charts/opentelemetry-collector/templates/_config.tpl
+++ b/charts/opentelemetry-collector/templates/_config.tpl
@@ -195,7 +195,19 @@ receivers:
 {{- end }}
 
 {{- define "opentelemetry-collector.applyClusterMetricsConfig" -}}
-{{- $config := mustMergeOverwrite (dict "service" (dict "pipelines" (dict "metrics" (dict "receivers" list)))) (include "opentelemetry-collector.clusterMetricsConfig" .Values | fromYaml) .config }}
+{{- $config := .config }}
+
+{{- /* Add leader election extension if enabled */ -}}
+{{- if .Values.presets.clusterMetrics.enableLeaderElection }}
+{{- $config = mustMergeOverwrite (dict "extensions" (dict "k8s_leader_elector" (dict "lease_name" "otel-collector-leader" "lease_namespace" (include "opentelemetry-collector.namespace" .)))) $config }}
+{{- $config = mustMergeOverwrite (dict "service" (dict "extensions" list)) $config }}
+{{- if not (has "k8s_leader_elector" $config.service.extensions) }}
+{{- $_ := set $config.service "extensions" (append $config.service.extensions "k8s_leader_elector" | uniq) }}
+{{- end }}
+{{- end }}
+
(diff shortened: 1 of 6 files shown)


charts/opentelemetry-collector/values.yaml
```diff
@@ -72,10 +72,16 @@ presets:
...
+  # Can be used with mode = deployment, statefulset, or daemonset.
+  # When running multiple replicas, the k8sleaderelector extension can be configured
+  # to ensure only one collector instance is collecting metrics at a time.
   # See https://opentelemetry.io/docs/kubernetes/collector/components/#kubernetes-cluster-receiver for details on the receiver.
   clusterMetrics:
     enabled: false
+    # Specify the k8sleaderelector extension ID to enable leader election when multiple replicas are used.
+    # This requires configuring and enabling the k8sleaderelector extension separately.
+    # Example: k8sleaderelector
+    k8sLeaderElector: ""
```
TylerHelmuth, line 84: We have, as a general rule, that presets should not require any user input to configure. Instead of a the preset accepting an extension id, could we do something like `enabledLeaderElection: false` and when enabled all the necessary configuration/components are automatically injected?

zparnold: Great suggestion! Implemented. The only thing I think is potentially a problem is that its now a "warning" when you install it without configuring the leader elector. If you want me to switch it to a "fail" I can.

charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -195,7 +195,19 @@ receivers:
 {{- end }}
 
 {{- define "opentelemetry-collector.applyClusterMetricsConfig" -}}
-{{- $config := mustMergeOverwrite (dict "service" (dict "pipelines" (dict "metrics" (dict "receivers" list)))) (include "opentelemetry-collector.clusterMetricsConfig" .Values | fromYaml) .config }}
+{{- $config := .config }}
+
+{{- /* Add leader election extension if enabled */ -}}
+{{- if .Values.presets.clusterMetrics.enableLeaderElection }}
+{{- $config = mustMergeOverwrite (dict "extensions" (dict "k8s_leader_elector" (dict "lease_name" "otel-collector-leader" "lease_namespace" (include "opentelemetry-collector.namespace" .)))) $config }}
```
TylerHelmuth, line 202: Please break out the k8s leader election config into its own helper template like `opentelemetry-collector.clusterMetricsConfig` does.

charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -195,7 +195,19 @@ receivers:
 {{- end }}
 
 {{- define "opentelemetry-collector.applyClusterMetricsConfig" -}}
-{{- $config := mustMergeOverwrite (dict "service" (dict "pipelines" (dict "metrics" (dict "receivers" list)))) (include "opentelemetry-collector.clusterMetricsConfig" .Values | fromYaml) .config }}
+{{- $config := .config }}
+
+{{- /* Add leader election extension if enabled */ -}}
```
TylerHelmuth, line 200: Lets move the logic for injecting the extension into its own helper. It can still be invoked within `opentelemetry-collector.applyClusterMetricsConfig`.

TylerHelmuth (dismissed):

PR #1685, opened 2025-05-25
fix(opentelemetry-operator): allow rbac subjectaccessreviews

Intuinewin:
Hi,

The operator tries to create multiple SubjectAccessReview at startup in order to test rights and enable/disable features dynamically such as RBAC creation (cf [this](https://github.com/open-telemetry/opentelemetry-operator/blob/a419080834a9f5e382652e795e499cd3689b0ff5/internal/config/config.go#L143)). Without this right all these features remain disabled.

Some error logs I was having:
```
{"level":"LEVEL(-2)","timestamp":"2025-05-25T19:34:46Z","logger":"config","message":"the rbac permissions are not set for the operator","reason":"unable to check rbac rules: subjectaccessreviews.authorization.k8s.io is forbidden: User \"system:serviceaccount:opentelemetry-operator:opentelemetry-operator\" cannot create resource \"subjectaccessreviews\" in API group \"authorization.k8s.io\" at the cluster scope"}
```

### DIFF
--- a/charts/opentelemetry-operator/Chart.yaml
+++ b/charts/opentelemetry-operator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.90.2
+version: 0.90.3
 description: OpenTelemetry Operator Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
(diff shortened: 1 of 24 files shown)


jaronoff97:
I do believe we want this still to only be allowed if kube rbac proxy is enabled ([docs](https://github.com/brancz/kube-rbac-proxy)). is there a reason you have disabled the kube rbac proxy?

Intuinewin:
I don't understand what this has to do with the rbac proxy. I haven't enabled the rbac proxy because I don't need it. As far as I know, the proxy is used only to protect the operator's /metrics.

I agree that enabling the proxy in the chart would solve my problem but IMO the operator should be able to run with all its features enabled without requiring the rbac proxy.

Let me know if I'm missing something

jaronoff97 (approved):

jaronoff97 (approved):

jaronoff97:
waiting for tests to pass and then will merge, thank you for your patience :bow:

Allex1 (approved):

PR #1688, opened 2025-05-27
[opentelemetry-kube-stack] Support extra kubernetes manifests

JorTurFer:
Support for extra kubernetes manifest without coupling the chart to other CRDs.

For example, if secrets need to be recovered from Secrets CSI Driver, it can be done like

```yaml
extraObjects: 
- apiVersion: secrets-store.csi.x-k8s.io/v1
  kind: SecretProviderClass
  metadata:
    name: demo
    namespace: test-app
  spec:
    parameters:
      objects: >
        - secretPath: xxxxxxx/data/registry-credentials
          objectName: "imagePullSecret"
          secretKey: registry-credentials
      vaultAddress: xxxxxxx
    provider: vault
    secretObjects:
      - data:
          - key: .dockerconfigjson
            objectName: imagePullSecret
        secretName: demo-image-pull-secrets
        type: kubernetes.io/dockerconfigjson
```

### DIFF
--- a/charts/opentelemetry-kube-stack/templates/extramanifests.yaml
+++ b/charts/opentelemetry-kube-stack/templates/extramanifests.yaml
@@ -1,0 +1,4 @@
+{{ range .Values.extraObjects }}
+---
+{{ tpl (toYaml .) $ }}
+{{ end }}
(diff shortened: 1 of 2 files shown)


jaronoff97 (commented):
not opposed to this, i've seen it done in some other charts. Not my personal favorite way of deploying this extra things, but i certainly understand why people do prefer this. Can you:
* bump the patch version of the chart
* update an example to include an extra object
* re-run `make generate-examples CHARTS=opentelemetry-kube-stack`

PR #1700, opened 2025-06-02
[chore] Update Kubernetes to 1.32.5 and Helm to 3.18.1; add version docs; improve code uniformity

jvoravong:
- Kubernetes and Helm Dependency Update:
  - Updated Kubernetes (kind) version from 1.24 (maintenance support ended on July 28th, 2023) to 1.32.5 (latest)
  - Updated Helm and related dependencies in GitHub workflows from version 3.X to 3.18.1.
  - This update ensures quality of Helm-rendered examples, in other projects I noticed example render issues caused by using outdated Helm dependencies. 
  - No breaking changes were identified in this project when using Kubernetes 1.32 and Helm 3.18.1.
    - But a breaking change was identified specifically for Helm 3.18.0 which causes a chart lint bug and maybe a Helm install/upgrade bug in the opentelemetry-kube-stack chart (CC: @jaronoff97).
    - [Github workflow run example](https://github.com/open-telemetry/opentelemetry-helm-charts/actions/runs/15393266228/job/43307474948?pr=1698#step:5:821)  
    - _"error converting YAML to JSON: yaml: line X: mapping values are not allowed in this context" in charts/opentelemetry-kube-stack/examples/prometheus-otel/values.yaml._ 
    - The fix is already out, just use Helm 3.18.1 which is currently latest.
- Documentation Enhancement:  
  - Documented versions Kubernetes and Helm versions used to maintain this project in README.md. This is important for customers and really should be a matrix like document that grows over time, keeping it simple for now and just adding a couple static doc lines.
- Code Cleanup:  
  - Improved uniformity across related areas for better consistency and maintainability.

### DIFF
--- a/.github/actions/setup/action.yaml
+++ b/.github/actions/setup/action.yaml
@@ -12,23 +12,23 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - name: Set up Helm
-      uses: azure/setup-helm@v4.2.0
+    - name: Install Helm
+      uses: azure/setup-helm@v4
       with:
-        version: ${{ inputs.helm-version }}
+        version: v3.18.1
 
     - uses: actions/setup-python@v4
       with:
         python-version: 3.8
 
     - name: Set up chart-testing
-      uses: helm/chart-testing-action@v2.6.0
+      uses: helm/chart-testing-action@v2.7.0
 
     - name: Create kind cluster
-      uses: helm/kind-action@v1.8.0
+      uses: helm/kind-action@v1.12.0
       if: ${{ inputs.create-kind-cluster == 'true' }}
       with:
-        config: ./.github/kind-1.24.yaml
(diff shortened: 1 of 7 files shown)


TylerHelmuth:
I like that before we had confidence when telling customers a helm minimum version for each chart, since not everyone is using the latest helm version. A couple thoughts:

1. I like rendering the examples with the latest helm
2. I like testing with our minimum supported version of k8s and helm. We can also add testing for the latest versions. [This is what the operator repo does.](https://github.com/open-telemetry/opentelemetry-operator/blob/e8f834acb8643c3e68fa6b268819c931ffbceace/.github/workflows/e2e.yaml#L61-L63)

jvoravong:
@TylerHelmuth great thoughts. I can update testing workflows to use a matrix input to test both oldest and latest supported versions for Kubernetes and Helm. I'll add that shortly.

PR #1701, opened 2025-06-02
chore(Telemetry): set service.name from  in telemetry release.name

Syndlex:
set service.name in telemetry from release.name

the default value is "otel-contrib" and this is irritating in telemetry values

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.126.0
+version: 0.126.1
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-collector/templates/_config.tpl
+++ b/charts/opentelemetry-collector/templates/_config.tpl
@@ -43,6 +43,11 @@ metrics:
             {{- end }}
 {{- end }}
 
+{{- define "opentelemetry-collector.telemetry.resource"  -}}
+resource:
+  service.name: "{{ .Release.Name }}"
+{{- end }}
+
 {{- define "opentelemetry-collector.otelsdkotlp.logs" -}}
 logs:
   processors:
@@ -62,6 +67,7 @@ logs:
 {{- .Values.alternateConfig | toYaml }}
 {{- else}}
 {{- $config := deepCopy .Values.config }}
+{{- $_ := set $config.service "telemetry" (mustMerge $config.service.telemetry (include "opentelemetry-collector.telemetry.resource" . | fromYaml)) }}
 {{- if .Values.internalTelemetryViaOTLP.traces.enabled }}
 {{- $_ := set $config.service "telemetry" (mustMerge $config.service.telemetry (include "opentelemetry-collector.otelsdkotlp.traces" . | fromYaml)) }}
 {{- end }}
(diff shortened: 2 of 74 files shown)


charts/opentelemetry-collector/Chart.yaml
```diff
@@ -12,4 +12,4 @@ maintainers:
   - name: jaronoff97
   - name: TylerHelmuth
 icon: https://opentelemetry.io/img/logos/opentelemetry-logo-nav.png
-appVersion: 0.127.0
+appVersion: 0.127.1
```
jaronoff97: you need to change the chart version not the app version here

Syndlex: wow sorry.

Never contribute after 9 hours of work.

charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -43,6 +43,11 @@ metrics:
             {{- end }}
 {{- end }}
 
+{{- define "opentelemetry-collector.telemetry.resource"  -}}
+resource:
+  service.name: "{{ .Release.Name }}"
+{{- end }}
```
TylerHelmuth, line 49: We should make this configurable in the preset

Syndlex: what do you mean with preset?

`.Values.presets`?

TylerHelmuth: Sorry Github's diff view made it look like this was being added to `opentelemetry-collector.otelsdkotlp.logs`

charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -43,6 +43,11 @@ metrics:
             {{- end }}
 {{- end }}
 
+{{- define "opentelemetry-collector.telemetry.resource"  -}}
+resource:
+  service.name: "{{ .Release.Name }}"
```
TylerHelmuth, line 48: I'm not actually sure whats typical for OTel's `service.name` semantic and a helm install. Should it be the release name or the fullname?

Syndlex: that's a create question.
I like name more. Because in full name I always have Somethink like "PROD-"
Everything is better than "otel-contrib"

jvoravong: If the default service.name was functional in previous releases, this could be a breaking change and should be documented in release notes. Users may have alerts tied to the current service.name value, especially for collector internal metrics.

I’d suggest using `fullname` instead, as it matches the [collector’s k8s service name](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/97ea912f3e4e1ec9d43e730fb47602ec1abfd0b4/charts/opentelemetry-collector/templates/service.yaml#L5) and better supports multiple chart instances in the same namespace or as subcharts. Open to considering all options though.

TylerHelmuth: Oh good point, I forgot the collector had a default service name.

@Syndlex we need to make this optional.

Syndlex: What was the default service name? 
When I do not set anything it is "otel-contrib" for me.
In my eyes this is would be wrong.

Optional would just mean don't setting it. Because it's setable from the current chart.

TylerHelmuth: > When I do not set anything it is "otel-contrib" for me.

 ya this is the default for the otel-contrib distro. For the k8s distro it would be otel-k8s.
 
 We can't automatically set a new value for users since it will change their internal telemetry. They might have triggers, SLOs, boards, or other things that are expecting `otel-contrib`.
 
 I agree it is already settable in `.Values.config`, but a lot of people don't realize you can do that. We could offer an option in the chart like `serviceName: ""` that when unset does nothing, but when set automatically configures the resource.

Issue #1111, opened 2024-04-02
add automountServiceAccountToken to deployment


ethanwllms:
Adding this option would help for those using Kyverno-Scan in their pipeline.

I have proposed the change [here](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1110)

jaronoff97:
@ethanwllms does kyverno-scan not automount the service account token by default right now?

amikoliunas:
This change would also help to be compliant with Azure AKS policy "Kubernetes clusters should disable automounting API credentials". Setting `automountServiceAccountToken: false` is required for both DaemonSet and ReplicaSet.

marcschaeferger:
Can be closed. Implemented by #1707. The chart now exposes `automountServiceAccountToken` and renders it on the workload and ServiceAccount

PR #1707, opened 2025-06-06
feat(opentelemetry-collector): add automountServiceAccountToken param for sa

ppatel1604:
The PR implements changes to fix the issue #1111. Changes are as below
- Support for `automountServiceAccountToken` with a check to avoid breaking changes
- Accepts the value for `automountServiceAccountToken` through helm values
- Default value for `automountServiceAccountToken` as part of the `values.yaml`
- opentelemetry-collector schema update
- Example values files with `true` and `false`
- Version bump

### DIFF
--- a/charts/opentelemetry-collector/ci/serviceaccount-automount-false-values.yaml
+++ b/charts/opentelemetry-collector/ci/serviceaccount-automount-false-values.yaml
@@ -1,0 +1,37 @@
+# Test configuration for disabling automountServiceAccountToken
+mode: deployment
+
+image:
+  repository: otel/opentelemetry-collector-contrib
+
+serviceAccount:
+  create: true
+  automountServiceAccountToken: false
+
+config:
+  receivers:
+    otlp:
+      protocols:
+        grpc:
+          endpoint: 0.0.0.0:4317
+        http:
+          endpoint: 0.0.0.0:4318
+  processors:
+    batch:
+  exporters:
+    debug:
+      verbosity: detailed
+  service:
+    pipelines:
+      traces:
+        receivers: [otlp]
+        processors: [batch]
+        exporters: [debug]
+      metrics:
+        receivers: [otlp]
+        processors: [batch]
+        exporters: [debug]
+      logs:
+        receivers: [otlp]
+        processors: [batch]
+        exporters: [debug]
--- a/charts/opentelemetry-collector/ci/serviceaccount-automount-true-values.yaml
+++ b/charts/opentelemetry-collector/ci/serviceaccount-automount-true-values.yaml
@@ -1,0 +1,37 @@
+# Test configuration for explicitly enabling automountServiceAccountToken
+mode: daemonset
+
+image:
+  repository: otel/opentelemetry-collector-contrib
+
+serviceAccount:
+  create: true
+  automountServiceAccountToken: true
+
+config:
+  receivers:
+    otlp:
+      protocols:
+        grpc:
+          endpoint: 0.0.0.0:4317
+        http:
+          endpoint: 0.0.0.0:4318
+  processors:
+    batch:
+  exporters:
+    debug:
+      verbosity: detailed
+  service:
+    pipelines:
+      traces:
+        receivers: [otlp]
+        processors: [batch]
+        exporters: [debug]
+      metrics:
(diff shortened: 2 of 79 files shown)


charts/opentelemetry-collector/templates/_pod.tpl
```diff
@@ -4,6 +4,9 @@ imagePullSecrets:
   {{- toYaml . | nindent 2 }}
 {{- end }}
 serviceAccountName: {{ include "opentelemetry-collector.serviceAccountName" . }}
+{{- if hasKey .Values.serviceAccount "automountServiceAccountToken" }}
```
TylerHelmuth: This check will always pass since you added `.Values.serviceAccount.automountServiceAccountToken` to the default values.yaml. Setting the value to false will not make this check fail.

charts/opentelemetry-collector/templates/_pod.tpl
```diff
@@ -4,6 +4,9 @@ imagePullSecrets:
...
+automountServiceAccountToken: {{ .Values.serviceAccount.automountServiceAccountToken }}
```
TylerHelmuth, line 7: Is it safe to put the token in the object definition?

charts/opentelemetry-collector/templates/_pod.tpl
ppatel1604: You are correct. I was trying to avoid breaking change, so I was trying to decide between conditional rendering and default values. Forgot to remove it. It's fixed now. Thank you for the review 👍

charts/opentelemetry-collector/templates/_pod.tpl
ppatel1604: This is not actually a token. This is just a boolean value. As per my understanding, the actual token supposed to be managed by Kubernetes. Maybe if I should rename it to something like `allowAutomountServiceAccountToken`?

Apologies if I did not understand the question correctly. Perhaps if I can have an example that I can follow, I will more than happy to change the implementation.

TylerHelmuth: Oh I see now. Matching the k8s field name is correct, I just misread it.

charts/opentelemetry-collector/ci/serviceaccount-automount-true-values.yaml
TylerHelmuth, line 1: Since true is the default k8s behavior you can remove this test.

charts/opentelemetry-collector/ci/serviceaccount-automount-false-values.yaml
```diff
@@ -0,0 +1,37 @@
+# Test configuration for disabling automountServiceAccountToken
+mode: deployment
+
+image:
+  repository: otel/opentelemetry-collector-contrib
+
+serviceAccount:
+  create: true
+  automountServiceAccountToken: false
```
TylerHelmuth: Since there is no interesting logic in the chart utilizing this field we can omit this CI test.

charts/opentelemetry-collector/ci/serviceaccount-automount-true-values.yaml
ppatel1604: Test removed. Thank you :)

charts/opentelemetry-collector/ci/serviceaccount-automount-false-values.yaml
ppatel1604: Test removed. Thank you :)

TylerHelmuth (approved):

TylerHelmuth:
@ppatel1604 are you by any change generating the examples with an old version of helm? They are still out of sync. You can also simplify the generation with `make generate-examples CHARTS=opentelemetry-collector`

TylerHelmuth:
When I checkout your branch and run `make generate-examples CHARTS=opentelemetry-collector` locally I do get changes.

ppatel1604:
> When I checkout your branch and run `make generate-examples CHARTS=opentelemetry-collector` locally I do get changes.

Thank you @TylerHelmuth. I have pulled latest changes and updated the examples. Can you please re-review this PR?

TylerHelmuth (approved):

Issue #1585, opened 2025-03-20
[opentelemetry-operator] set auto instrumentation images without specifing the tag


davgia:
# Please describe the problem

My goal is to pull all images from a k8s cluster using a pull through cache strategy. To do that I need to change all the deployed helm charts.
Relative to this chart, I would like to define default auto-instrumentation images repository **without** specifying the tag. I would like the operator to take care of the tag (so it is still updated when necessary). Those are my values for the `opentelemetry-operator` chart (version: `0.83.0`):

``` yaml
# NB: I am in terraform, so ${image_registry} is substituted with my private oci registry url
manager:
  image:
    repository: "${image_registry}/ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator"
  collectorImage:
    repository: "${image_registry}/docker.io/otel/opentelemetry-collector-contrib"
  autoInstrumentationImage:
    dotnet:
      repository: "${image_registry}/ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-dotnet"
      tag: "1.2.0" # it should not be required (or should be defined in the helm chart default values), the operator should take care of this
    java:
      repository: "${image_registry}/ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-java"
      tag: "1.33.6" # it should not be required (or should be defined in the helm chart default values), the operator should take care of this
    python:
      repository: "${image_registry}/ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-python"
      tag: "0.51b0" # it should not be required (or should be defined in the helm chart default values), the operator should take care of this
```

### Describe the solution you'd like

The solution for me would be to expose a single helm chart value to override the default registry of all instrumentation images and to specify the tag in the default helm chart values for each instrumentation (just like the collector tag).

### Describe alternatives you've considered

The alternative is to hardcode the tag, but I don't like this approach because I may forget to update it (and it may be necessary in order to function properly). Also, I would avoid using latest tag just because it may force the operator to use incompatible images (I see that new major are available but are not yet used by default).

jvoravong:
I don’t have a strong opinion on overriding only the image registry via a global variable, but I would find it useful if we had a more flexible approach to defining Docker images used across Helm values, the manager’s CRD spec, and the Instrumentation CRD spec.  

When using the operator as a subchart, I often desire or need maximum flexibility because:  
- Images may come from multiple or a single registry.  
[...]

PR #1711, opened 2025-06-13
Pin image tags and repositories on values.yaml file

YGwen:
Hi Team, 

When using this chart in an air gape environment, we should not have to specify tags but only the new repository.
This problem was already mentioned in this [issue](https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1585).


YGwen:
@Allex1 @jvoravong, any chance we could discuss this ?
This would drastically improve usage of this chart in air gaped environments.

PR #1715, opened 2025-06-13
feat(opentelemetry-operator): add imagePullPolicy for the operator image

KyriosGN0:
Signed-off-by: AvivGuiser <avivguiser@gmail.com>

### DIFF
--- a/charts/opentelemetry-operator/values.yaml
+++ b/charts/opentelemetry-operator/values.yaml
@@ -44,6 +44,7 @@ manager:
   image:
     repository: ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator
     tag: ""
+    imagePullPolicy: IfNotPresent
   collectorImage:
     repository: ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector-k8s
     tag: 0.126.0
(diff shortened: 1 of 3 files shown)


charts/opentelemetry-operator/values.yaml
```diff
@@ -104,6 +105,8 @@ manager:
   #     memory: 64Mi
   #     ephemeral-storage: 50Mi
 
+  # Image pull policy for the manager container.
+  imagePullPolicy: IfNotPresent
```
Allex1: I don't think this is needed here

KyriosGN0: Would you like me to comment it and use a default in the template itself?

TylerHelmuth: We should use the new field you added under `manager.image` instead.

TylerHelmuth (commented):
you'll also need to run `make generate-examples CHARTS=opentelemetry-operator`

charts/opentelemetry-operator/values.yaml
KyriosGN0: ah ok, i will remove it

KyriosGN0:
@TylerHelmuth it looks like i need to regenerate the values.schema.json, is there some preferred way to do it ?

PR #1716, opened 2025-06-16
fix: convertMemToBytes helper

darkweaver87:
This PR fixes `convertMemToBytes` helper.

This function aims to handle SI, IEC prefixes or no prefixes (cf. https://github.com/kubeflow/crd-validation/blob/master/vendor/k8s.io/apimachinery/pkg/api/resource/quantity.go#L44).
SI prefixes use power of 10 (e.g. 1e18 = 1 x 10^18) (m | "" | k | M | G | T | P | E).
IEC prefixes use power of 2 (e.g. 0x1p60 = 2^60) (Ki | Mi | Gi | Ti | Pi | Ei).

The values should be the same:

- 128974848 ✅
- 129e6 ✅
- 129M ✅
- 128974848000m ❌
- 123Mi ✅

Test:

```bash
$ helm template test ./charts/opentelemetry-collector/ -f charts/opentelemetry-collector/examples/using-GOMEMLIMIT/values.yaml --set resources.limits.memory=128974848000m | grep GOMEMLIMIT -A 1
            - name: GOMEMLIMIT
              value: "98400000000MiB"
$ git checkout fix/helper 
$ helm template test ./charts/opentelemetry-collector/ -f charts/opentelemetry-collector/examples/using-GOMEMLIMIT/values.yaml --set resources.limits.memory=128974848000m | grep GOMEMLIMIT -A 1
            - name: GOMEMLIMIT
              value: "98MiB"
```


charts/opentelemetry-collector/templates/_helpers.tpl
```diff
@@ -190,7 +190,9 @@ Allow the release namespace to be overridden
     {{- $mem = mulf (trimSuffix "g" $mem | float64) 1e9 -}}
   {{- else if hasSuffix "gi" $mem -}}
     {{- $mem = mulf (trimSuffix "gi" $mem | float64) 0x1p30 -}}
-  {{- else if hasSuffix "m" $mem -}}
+  {{- else if hasSuffix "m" . -}}
+    {{- $mem = divf (trimSuffix "m" $mem | float64) 1e3 -}}
+  {{- else if hasSuffix "M" . -}}
     {{- $mem = mulf (trimSuffix "m" $mem | float64) 1e6 -}}
```
TylerHelmuth, line 196: The trim command needs updated to be M

darkweaver87: The `$mem` is lowered at the top of the function.

TylerHelmuth:
@darkweaver87 please bump the chart's patch version and run `make generate-examples CHARTS=opentelemetry-collector`

PR #1725, opened 2025-06-22
Add custom metrics support in opentelemetry-collector

boqu:
We should be able to scale pod with custom metrics in HPA. e.g. metrics from OTEL.


charts/opentelemetry-collector/templates/hpa.yaml
```diff
@@ -34,4 +34,66 @@ spec:
           type: Utilization
           averageUtilization: {{ .Values.autoscaling.targetCPUUtilizationPercentage }}
   {{- end }}
+  {{- if .Values.autoscaling.podsMetrics }}
+  {{- range .Values.autoscaling.podsMetrics }}
```
TylerHelmuth: We either need to require the user provide all the yaml structure in the 3 new fields or we need to tell the user what they need to set in the field list. At the moment it looks like the new fields are an array of complex objects. 

I'd prefer we solve this by adding a new `additionalMetrics` field and then the user supplies all the yaml they want. This lets them add pod, object, or external metrics without us having to define an interface.

boqu: @TylerHelmuth Thank you for your feedback. I updated the PR to use `additionalMetrics` per as your suggestion.

PR #1728, opened 2025-06-25
[opentelemetry-operator]: bump operator to 0.127.0

KyriosGN0:


KyriosGN0:
the test failures are weird, the [release notes](https://github.com/open-telemetry/opentelemetry-operator/releases/tag/v0.127.0) don't say that something has changed in regards to the metrics service

TylerHelmuth:
@open-telemetry/operator-approvers @jaronoff97 do you know if anything specific was needed for the new `operator-metrics` test?

@KyriosGN0 you can skip the test for now.

swiatekm:
> @open-telemetry/operator-approvers @jaronoff97 do you know if anything specific was needed for the new `operator-metrics` test?
> 
> @KyriosGN0 you can skip the test for now.

That test finds the operator pod using the following labels:

```yaml
control-plane: controller-manager
```

My guess would be that the Helm Chart either doesn't set them or uses different values. The test can be skipped for now and we can fix the test in the operator repo to make it less dependent on the deployment method.

KyriosGN0:
@swiatekm @Allex1 do you want me to remove the test from the exclude list ?

swiatekm (approved):

Issue #1726, opened 2025-06-23
[opentelemetry-kube-stack] Unify both collectors in a single one


JorTurFer:
Currently, the chart deploys 2 different collectors, one to scrape the cluster resources as daemonset and other to retrieve k8s api information as deployment with a single replica. 

In latest versions, both receivers (`k8s_cluster` and `k8sobjects`) support a parameter to enable leader election `k8s_leader_elector`, which improves the reliability of the receiver adding HA capabilities.

My proposal is unifying both collectors in a single one, running as daemonset with `k8s_leader_elector` to keep a single instance pulling that info from k8s api.

I'm willing to open a PR with the needed changes, but I'm not fully sure if this will be accepted as removing the second collector could be a breaking change. If you think that this is worth, I'll open a PR later on

In case of not agreeing with unifying them, I'd like to as to at least, support multiple instances of cluster collector to have it with HA config just adding configurable instances to that collector + leader election

swiatekm:
Have you tested this idea in a large production cluster? I'm worried that we're going to end up with a single DaemonSet Pod doing a lot more work, which Kubernetes makes difficult to set resources for. At the very least, if we decide to ship this by default, we should document the downsides.

PR #1735, opened 2025-06-27
[opentelemetry-kube-stack]: Merge both collectors in a single one with leader election

JorTurFer:
This PR bumps the operator dependency and merges both collectors in a single one running as daemonset with support to leader election for k8s api receivers, getting rid of the collector deployed as single instance deployment

Fixes https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1726

### DIFF
--- a/charts/opentelemetry-kube-stack/templates/_config.tpl
+++ b/charts/opentelemetry-kube-stack/templates/_config.tpl
@@ -34,11 +34,11 @@ the config is written as YAML.
 {{- $_ := set $collector "config" $config }}
 {{- end }}
 {{- if .collector.presets.kubernetesEvents.enabled }}
-{{- $config = (include "opentelemetry-kube-stack.collector.applyKubernetesEventsConfig" (dict "collector" $collector) | fromYaml) -}}
+{{- $config = (include "opentelemetry-kube-stack.collector.applyKubernetesEventsConfig" (dict "collector" $collector "namespace" .namespace) | fromYaml) -}}
 {{- $_ := set $collector "config" $config }}
 {{- end }}
 {{- if .collector.presets.clusterMetrics.enabled }}
-{{- $config = (include "opentelemetry-kube-stack.collector.applyClusterMetricsConfig" (dict "collector" $collector) | fromYaml) -}}
+{{- $config = (include "opentelemetry-kube-stack.collector.applyClusterMetricsConfig" (dict "collector" $collector "namespace" .namespace) | fromYaml) -}}
 {{- $_ := set $collector "config" $config }}
 {{- end }}
 {{- toYaml $collector.config | nindent 4 }}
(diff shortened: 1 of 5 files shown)


JorTurFer:
I think that waiting for https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1728 and bumping the operator to the new version can be worth

charts/opentelemetry-kube-stack/templates/_config.tpl
```diff
@@ -227,16 +227,23 @@ receivers:
...
 {{- end }}
 {{- $config | toYaml }}
 {{- end }}
 
 {{- define "opentelemetry-kube-stack.collector.clusterMetricsConfig" -}}
+extensions:
+  k8s_leader_elector/k8s_cluster:
+    auth_type: serviceAccount
+    lease_name: k8s.cluster.receiver.opentelemetry.io
+    lease_namespace: {{ .namespace }}
```
TylerHelmuth: I know it will merge cleanly, but if the leader elector extension ever needs changed we'll have to make sure to update all the instances. Can you move this to its own template and reuse it in the other templates?

JorTurFer: yeah, nice idea! Tomorrow I'll do it and I'll rebase my branch too

JorTurFer: PTAL @TylerHelmuth , I've extracted the generation to another block and reused it for both cases

charts/opentelemetry-kube-stack/values.yaml
```diff
@@ -527,48 +531,6 @@ collectors:
               - batch
             exporters:
               - debug
-  cluster:
```
jaronoff97, line 530: do you think users would want this to be configurable? i.e. we give the user the ability to choose if they want the standalone or use leader election? I think there's value in having the option, but this is meant to be an opinionated chart... my only hesitation is that using leader election can be a bit frustrating for some users architectures.

jaronoff97 (commented):
One thought as to whether we want to opt users in. Also, helm deletes can be a bit odd, so i think it would be worth writing some type of changelog entry in the README that encourages Gitops users to manually prune the resources

charts/opentelemetry-kube-stack/values.yaml
JorTurFer: Considering that this is opinionated, I'd get rid of it because the current `cluster` collector doesn't work in HA, so as default it's not "safe to use". I thought about keeping both collectors as they were, but adding leader election to `cluster` one. The problem with that approach was the more resources than needed were used when the daemonset can do that taks quite better.

Maybe, we can create an example with the previous configuration with both collectors and clarify in the readme that for those cases where leader election isn't an option, they can just check the example to use this approach of 2 collectors without leader election, but emphasizing the fact of not being in HA the `cluster` one. 

WDYT?

rogercoll: +1 on getting rid of the cluster collector but still making it configurable (probably some docs for the following):

- Override daemonset collector to remove the `k8s_leader_elector` extension from the configuration -> some Otel collector distributions might not include this extension
- Being able to manually define back the `cluster` collector.

That will allow current configurations that rely on the `cluster` collector to use future Chart versions with minimum changes.

JorTurFer: I've added an example about how to define back the previous approach as well as some notes in readme.md to explain how to upgrade and to announce the "new requirement" of leader election extension and linking the example that defines back the previous approach.

To disable the leader election I have had to add a new field to the preset because I haven't found any other good way to disable it without adding a lot of code (but if someone has a good idea that I've missed just say them to me, I'm quite newbie with advanced helm templating 😄 )

PTAL

jaronoff97: @JorTurFer your implementation is fantastic in my eyes :) thanks for the work here! I like how this ended up.

jaronoff97 (approved):

rogercoll (approved):

jaronoff97:
@JorTurFer thank you for the work here!! this is a huge add :D

JorTurFer:
Happy to help, this stack is awesome and we are movimg our clusters to this from prometheus stack as this is quite lighter.
Huge thanks to you for it!

JorTurFer:
BTW, we have already rolled out this version in some of our clusters and it works as expected (at least for the moment 🤣 )

PR #1742, opened 2025-07-01
[target-allocator] container resources

nlamirault:


charts/opentelemetry-target-allocator/values.yaml
```diff
@@ -47,3 +47,11 @@ targetAllocator:
     filter_strategy: relabel-config
     config:
       scrape_configs: []
+
+  resources: {}
```
TylerHelmuth, line 51: Lets add a warning in NOTES.txt when this is not set. See the collector chart for an example.

nlamirault: done.

kayahk: When will this be merged? Defining resource allocations is mandatory for us if we want to have any helm chart deployed.
Thanks!

PR #1748, opened 2025-07-03
fix(opentelemetry-collector): always sort prometheus resource attrs

snuggie12:
Need to sort these keys in order to keep the manifests gitops friendly.


Allex1 (approved):

snuggie12:
apologies for the noise as I clicked the wrong button in the UI for syncing the fork and didn't notice until it was too late.

I think I'm properly rebased, chart was bumped, and ran the above `make` command

PR #1752, opened 2025-07-07
fix: use part-of kubernetes label instead of instance in demo chart

Machyne:
The [Kubernetes Recommended Labels docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/) say that instance should be a unique identifier for an instance of an application. Here, it's set to the helm installation name everywhere. I think `part-of` is more appropriate in this case.

Using `instance` creates issues when scraping service labels per the [semantic conventions](https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/#how-servicename-should-be-calculated): since no `resource.opentelemetry.io/service.name` annotation is provided, this well known k8s label `app.kubernetes.io/instance` is checked next. The result is that all of the data from the workload pods sent through `k8sattributes processor` using the `service.name` metadata will come out with the same `service.name` (ex: "my-otel-demo" when using [the example install command](https://opentelemetry.io/docs/platforms/kubernetes/helm/demo/#installation)).

If you think this is an issue with the semantic conventions, please let me know and I will take this discussion to that repo instead. Thanks in advance!

(note: this is a one-line diff on `charts/opentelemetry-demo/templates/_helpers.tpl` plus the result of `make generate-examples`)

### DIFF
--- a/charts/opentelemetry-demo/Chart.yaml
+++ b/charts/opentelemetry-demo/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: opentelemetry-demo
-version: 0.37.1
+version: 0.37.2
 description: opentelemetry demo helm chart
 home: https://opentelemetry.io/
 sources:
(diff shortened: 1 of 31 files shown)


Machyne:
> @Machyne can you bump the patch version of the chart please and then re-run `make generate-examples CHARTS=opentelemetry-demo`

Done! Thank you everyone for the reviews!

PR #1753, opened 2025-07-07
[opentelemetry-kube-stack] Update opentelemetry-operator dependency to v0.90.4

rakeshadk7:
## Description
Updates the opentelemetry-operator chart dependency from v0.84.2 to v0.90.4, which brings the operator version from 0.120.0 to 0.126.0.

## Problem
The current kube-stack chart uses an older operator version (0.120.0) that has telemetry configuration compatibility issues. Users experience collector startup failures with the error: 

```
'service.telemetry.metrics' decoding failed due to the following error(s): '' has invalid keys: address
```

## Solution
- Bump chart version: 0.6.2 → 0.6.3
- Bump appVersion: 0.120.0 → 0.126.0  
- Update opentelemetry-operator dependency: 0.84.2 → 0.90.4

## Testing
- [ ] Helm chart linting passes
- [ ] Chart installs successfully with new operator version
- [ ] Resolves telemetry configuration conflicts

## Related Issues
Fixes compatibility issues with deprecated telemetry address field injection.

### DIFF
--- a/charts/opentelemetry-kube-stack/Chart.yaml
+++ b/charts/opentelemetry-kube-stack/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-kube-stack
-version: 0.6.2
+version: 0.6.3
 description: |
   OpenTelemetry Quickstart chart for Kubernetes.
   Installs an operator and collector for an easy way to get started with Kubernetes observability.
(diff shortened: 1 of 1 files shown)


rakeshadk7:
The operator auto inject the `telemetry` section and there doesnt seem to be way to override it. 

My current workaround is override the version in the values.yaml

```
# Override the operator version to use the latest available
opentelemetry-operator:
  manager:
    image:
      repository: ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator
      tag: "0.126.0"  # Latest version that should fix telemetry config issues
```

charts/opentelemetry-kube-stack/Chart.yaml
```diff
@@ -26,7 +26,7 @@ dependencies:
     condition: crds.install,crds.installPrometheus
   - name: opentelemetry-operator
     repository: https://open-telemetry.github.io/opentelemetry-helm-charts
-    version: 0.84.2
+    version: 0.90.4
```
TylerHelmuth, line 29: We just released an new operator chart version and you bump this version again?

jaronoff97:
Looks like Roger upgraded to latest already, so i think we're good to close this. Thank you!

PR #1758, opened 2025-07-10
[opentelemetry-operator] allow for overriding namespace in values file

CharlieTLe:


CharlieTLe:
Hi @TylerHelmuth, could you help review?

charts/opentelemetry-operator/templates/_helpers.tpl
```diff
@@ -23,6 +23,17 @@ If release name contains chart name it will be used as a full name.
...
+{{/*
+Allow the release namespace to be overridden
+*/}}
+{{- define "opentelemetry-operator.namespace" -}}
+  {{- if .Values.namespaceOverride -}}
+    {{- .Values.namespaceOverride -}}
+  {{- else -}}
+    {{- .Release.Namespace -}}
+  {{- end -}}
+{{- end -}}
```
jvoravong, line 31: Namespace names have a [char size limit of 63](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names) I've seen users hit before, I'd suggest truncating the value based on the limit with something like this.
`{{- default .Release.Namespace .Values.namespaceOverride | trunc 63 | trimSuffix "-" -}}
`

jvoravong:
> @CharlieTLe the linter is failing because you need to bump the patch version of the chart, otherwise i think this looks fine to me.

Yeah just bump the patch version in the Chart.yaml and run `make generate-examples`, otherwise LGTM.

jvoravong (approved):

PR #1761, opened 2025-07-14
[kube-stack] bump operator to 0.91.0 and add target allocator crd

rogercoll:
Current deployments of the Chart for collectors that set horizontal autoscaling have the following failure:

```
status:
  conditions:
  - lastTransitionTime: "2025-07-14T10:37:48Z"
    message: 'the HPA controller was unable to get the target''s current scale: Internal
      error occurred: the spec replicas field ".spec.replicas" does not exist'
    reason: FailedGetScale
    status: "False"
    type: AbleToScale
  currentMetrics: null
  desiredReplicas: 0
  ```
  
 Fixed by: https://github.com/open-telemetry/opentelemetry-operator/pull/4047
 
 This PR updates the operator dependency to latest + adds the new TargetAllocator CR (operator fails at startup if not available)

### DIFF
--- a/charts/opentelemetry-kube-stack/Chart.lock
+++ b/charts/opentelemetry-kube-stack/Chart.lock
@@ -7,12 +7,12 @@ dependencies:
   version: 0.0.0
 - name: opentelemetry-operator
   repository: https://open-telemetry.github.io/opentelemetry-helm-charts
-  version: 0.84.2
+  version: 0.91.0
 - name: kube-state-metrics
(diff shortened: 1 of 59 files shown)


rogercoll:
> @rogercoll were you able to test that these bumps don't break the functionality manually?

Yes, with these changes I was able to run a configuration previously pinned at `0.3.9` without problem. Manual deployment of the target allocator seems to be fine too.

PR #1762, opened 2025-07-15
[opentelemetry-collector]: bump to 0.129.0

KyriosGN0:


TylerHelmuth:
@KyriosGN0 we need to do the v0.129.0 bump first. We typically dont skip versions so that users always have a chart that corresponds with a released version.

charts/opentelemetry-collector/Chart.yaml
```diff
@@ -12,4 +12,4 @@ maintainers:
   - name: jaronoff97
   - name: TylerHelmuth
 icon: https://opentelemetry.io/img/logos/opentelemetry-logo-nav.png
-appVersion: 0.128.0
+appVersion: 0.129.0
```
TylerHelmuth:
```suggestion
appVersion: 0.129.1
```

TylerHelmuth:
@KyriosGN0 the v0.129.0 release went wonky and I think all the images were released as v0.129.1.  You'll need to regenerate the examples

KyriosGN0:
> @KyriosGN0 the v0.129.0 release went wonky and I think all the images were released as v0.129.1.  You'll need to regenerate the examples

Sorry, Done!

TylerHelmuth (approved):

PR #1780, opened 2025-08-01
Add revisionHistoryLimit to statefulset and daemonset

JBodkin-Amphora:


charts/opentelemetry-collector/templates/statefulset.yaml
```diff
@@ -23,6 +23,7 @@ spec:
     whenDeleted: {{ .Values.statefulset.persistentVolumeClaimRetentionPolicy.whenDeleted }}
     whenScaled: {{ .Values.statefulset.persistentVolumeClaimRetentionPolicy.whenScaled }}
 {{- end }}
+  revisionHistoryLimit: {{ .Values.revisionHistoryLimit }}
```
TylerHelmuth, line 26: I suspect the values schema needs updated. Can you also add a field to the values.yaml with a comment how to use this field?

JBodkin-Amphora: The value already exists in the values.yaml, as it works for deployment. I've removed the comment from the values.yaml, as now it works for deployment, statefulset and daemonset

TylerHelmuth (approved):

TylerHelmuth:
@JBodkin-Amphora looks like you need another `make generate-examples CHARTS=opentelemetry-collector`

PR #1782, opened 2025-08-01
Update values.yaml

### LINKED ISSUE #1781: operator.autoinstrumentation.go - no such feature gate
In `values.yaml` there are instructions on how to enable go auto telemetry by specifying ` operator.autoinstrumentation.go` feature gate:

https://github.com/open-telemetry/opentelemetry-helm-charts/blob/3d45e2bd181adac69b2e205724d7168889c92d20/charts/opentelemetry-operator/values.yaml#L76C1-L77C74

However, when specified, the operator will panic:

```
panic: invalid argument "operator.autoinstrumentation.go" for "--feature-gates" flag: no such feature gate "operator.autoinstrumentation.go". valid gates: [operator.collector.default.config operator.collector.targetallocatorcr operator.golang.flags operator.sidecarcontainers.native operator.targetallocator.fallbackstrategy operator.targetallocator.mtls]
goroutine 1 [running]:
main.main()
	/home/runner/work/opentelemetry-operator/opentelemetry-operator/main.go:92 +0x2c40
```

Chart version `0.92.1`

David-VTUK:
Fixes #1781

### DIFF
--- a/charts/opentelemetry-operator/values.yaml
+++ b/charts/opentelemetry-operator/values.yaml
@@ -74,7 +74,7 @@ manager:
       repository: ""
       tag: ""
     # The Go instrumentation support in the operator is disabled by default.
-    # To enable it, use the operator.autoinstrumentation.go feature gate.
+    # To enable it, add --enable-go-instrumentation to extraArgs (source: https://github.com/open-telemetry/opentelemetry-operator/blob/5ddc44026eede78bd8cdf501e3689f9a143d9860/CHANGELOG.md?plain=1#L1153)
     go:
       repository: ""
       tag: ""


charts/opentelemetry-operator/values.yaml
```diff
@@ -74,7 +74,7 @@ manager:
       repository: ""
       tag: ""
     # The Go instrumentation support in the operator is disabled by default.
-    # To enable it, use the operator.autoinstrumentation.go feature gate.
+    # To enable it, add --enable-go-instrumentation to extraArgs (source: https://github.com/open-telemetry/opentelemetry-operator/blob/5ddc44026eede78bd8cdf501e3689f9a143d9860/CHANGELOG.md?plain=1#L1153)
```
TylerHelmuth, line 77: We should update the chart to help set this feature.

David-VTUK: Good point @TylerHelmuth - do you suggest we have something specific in the `values.yaml` that, when enabled, adds the respective argument to `extraArgs`. Something like:

`go.autoinstrumentation.enabled = true` or reuse `operator.autoinstrumentation.go` but change the outcome to add `--enable-go-instrumentation` to `extraArgs`

I'll then update accordingly.

TylerHelmuth: Lets add a section in `manager` that controls [these flags](https://github.com/open-telemetry/opentelemetry-operator/blob/main/README.md#controlling-instrumentation-capabilities). For now we'll only add the go flag, but we can add the other ones later.

Something like

```yaml
manager:
  autoInstrumentation:
    go:
      enabled: false
```

PR #1783, opened 2025-08-01
replace bitnami/kubectl image references with rancher/kubectl fixes#1768

lakkawardhananjay:
This PR replaces deprecated Bitnami container image reference with Rancher-maintained equivalent to avoid future image pull failures, in line with [bitnami/containers#83267](https://github.com/bitnami/containers/issues/83267).

🔧 Changes Made
Replaced `bitnami/kubectl` with `rancher/kubectl` in charts/opentelemetry-kube-stack/values.yaml

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.129.0
+version: 0.130.0
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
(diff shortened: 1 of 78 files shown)


charts/opentelemetry-kube-stack/values.yaml
```diff
@@ -17,7 +17,7 @@ cleanupJob:
   enabled: true
   # Image details for the kubectl
   image:
-    repository: bitnami/kubectl
+    repository: rancher/kubectl
```
jaronoff97: please bump the patch version of the chart and then run `make generate-examples CHARTS=opentelemetry-kube-stack`

lakkawardhananjay: ok on it!

PR #1791, opened 2025-08-07
Bump OTel Demo sub charts

cyrille-leclerc:
Just bump the following sub charts: opentelemetry collector, jaeger, Prometheus, and Grafana.
I didn't bump opensearch as I am not knowledgeable enough on this database.

I just bumped the sub chart versions in `Chart.yaml` then run `make`

PR #1792, opened 2025-08-08
[otel-demo] Fix and update prometheus config

cyrille-leclerc:
Depend on:
- https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1791

Fix Prometheus config: the `otlp` block was in the wrong place and ignored.
Update Prometheus OTLP resource attribute promotion to cover more use cases and get parity with the otel-demo Docker compose.

### DIFF
--- a/charts/opentelemetry-demo/values.yaml
+++ b/charts/opentelemetry-demo/values.yaml
@@ -832,34 +832,43 @@ prometheus:
     extraFlags:
       - "enable-feature=exemplar-storage"
       - "web.enable-otlp-receiver"
-    global:
-      scrape_interval: 5s
-      scrape_timeout: 3s
-      evaluation_interval: 30s
     tsdb:
       out_of_order_time_window: 30m
-    prometheus.yml:
-      otlp:
-        keep_identifying_resource_attributes: true
-        # Recommended attributes to be promoted to labels.
-        promote_resource_attributes:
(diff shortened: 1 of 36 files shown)


charts/opentelemetry-demo/values.yaml
```diff
@@ -762,14 +762,14 @@ opentelemetry-collector:
       pipelines:
         traces:
           processors: [memory_limiter, resource, transform, batch]
-          exporters: [otlp, debug, spanmetrics]
+          exporters: [otlp, spanmetrics]
```
puckpuck: We should keep the debug exporter for traces, as well as for metrics and logs. With it the user can see that data is being exported successfully from the collector.

puckpuck (changes requested):

charts/opentelemetry-demo/values.yaml
cyrille-leclerc:
```suggestion
          exporters: [otlp, debug, spanmetrics]
```

cyrille-leclerc: ack, I'll restore de debug exporter.

cyrille-leclerc: Debug exporter restored

puckpuck (approved):

PR #1801, opened 2025-08-13
[opentelemetry-kube-stack] Use daemonSetUpdateStrategy instead of updateStrategy

### LINKED ISSUE #1574: [opentelemetry-kube-stack] the `updateStrategy` is not working
Currently, the template makes use of `updateStrategy` in `collector.yaml`. However, the Custom Resource Definition (CRD) only contains `daemonSetUpdateStrategy` and `deploymentUpdateStrategy`. As a result, it needs to switch to `daemonSetUpdateStrategy`. 

Refer to: 
https://github.com/open-telemetry/opentelemetry-helm-charts/blob/ac8c62cb0c4176b13887b19ede085a4b0be1957f/charts/opentelemetry-kube-stack/templates/collector.yaml#L107-L110

byehn:
Fixes #1574 

OpenTelemetryCollector v1beta1 API uses `daemonSetUpdateStrategy` instead of `updateStrategy`([link](https://github.com/open-telemetry/opentelemetry-operator/blob/v0.129.1/apis/v1beta1/opentelemetrycollector_types.go#L132)). This PR updates the collector template and values to reflect that change.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 TMP_DIRECTORY = ./tmp
-CHARTS ?= opentelemetry-collector opentelemetry-operator opentelemetry-demo opentelemetry-ebpf opentelemetry-kube-stack
+CHARTS ?= opentelemetry-collector opentelemetry-operator opentelemetry-demo opentelemetry-ebpf opentelemetry-kube-stack opentelemetry-target-allocator
 OPERATOR_APP_VERSION ?= "$(shell cat ./charts/opentelemetry-operator/Chart.yaml | sed -nr 's/appVersion: ([0-9]+\.[0-9]+\.[0-9]+)/\1/p')"
 
 .PHONY: generate-examples
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.134.0
+version: 0.134.1
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-collector/examples/alternate-config/rendered/clusterrole.yaml
+++ b/charts/opentelemetry-collector/examples/alternate-config/rendered/clusterrole.yaml
@@ -5,7 +5,7 @@ kind: ClusterRole
 metadata:
   name: example-opentelemetry-collector
   labels:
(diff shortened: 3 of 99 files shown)


charts/opentelemetry-kube-stack/values.yaml
```diff
@@ -307,8 +307,12 @@ defaultCRConfig:
   #   metrics:
   #     enableMetrics: true
 
-  # Update strategy for the collector
-  updateStrategy: {}
```
TylerHelmuth, line 311: @jaronoff97 this is a breaking change, are you ok with that? In the collector chart we'd handle this by supporting both for a bit. Do you want to do that here?

jaronoff97: @TylerHelmuth @byehn, I do think we should have a migration path here actually. Let's add in the two new strategies, and put a note in the NOTES to inform users of the deprecation. Sorry for the delayed response.

byehn:
@TylerHelmuth @jaronoff97, thanks for the feedback. I kept the existing updateStrategy for backward compatibility, but added a note so that it can also be switched to the new value.

charts/opentelemetry-kube-stack/templates/collector.yaml
```diff
@@ -104,14 +104,18 @@ spec:
   observability:
   {{- toYaml . | nindent 4}}
   {{- end }}
-  {{- with $collector.updateStrategy }}
-  updateStrategy:
+  {{- if eq $collector.mode "daemonset" }}
+  {{- with (or $collector.daemonSetUpdateStrategy $collector.updateStrategy) }}
```
TylerHelmuth: Instead of setting daemonSetUpdateStrategy, `$collector.updateStrategy` should continue to set `updateStrategy` when set. In `NOTES.txt` we can add validation that enforces `updateStrategy` cannot be set when `deploymentUpdateStrategy` or  `daemonSetUpdateStrategy` are set.

TylerHelmuth: Unless in the operate `daemonSetUpdateStrategy` and `updateStrategy` are treated exactly the same.

byehn: @TylerHelmuth, got it. I reverted it back so that it’s set directly with updateStrategy.
Since the current chart already uses the v1beta1 collector, I think just adding a note about the deprecation (as it is now) should be enough. What do you think?

PR #1803, opened 2025-08-14
Fix(opentelemetry-operator): update featureGates schema to array

Kescefko:
Starting from Helm v3.18.5, rendering the Opentelemetry Operator chart can fail due to a schema validation error:

**opentelemetry-collector
opentelemetry-operator:
"file:///values.schema.json#" is not valid against metaschema: jsonschema validation failed
at '/properties/manager/properties/featureGates/examples': got string, want array**

This PR fixes the issue by updating the featureGates schema from a single string to an array, ensuring:

- Helm values now correctly accept multiple feature gates.
- Compatibility with Helm v3.18.5+ and operator auto-generated CRDs.
- Backward compatibility with existing configurations.
- No functional changes to operator behavior; this is purely a schema fix.

### DIFF
--- a/charts/opentelemetry-operator/values.schema.json
+++ b/charts/opentelemetry-operator/values.schema.json
@@ -172,7 +172,8 @@
         "leaderElection",
         "verticalPodAutoscaler",
         "rolling",
-        "securityContext"
+        "securityContext",
+        "ignoreMissingCollectorCRDs"
       ],
       "additionalProperties": false,
       "properties": {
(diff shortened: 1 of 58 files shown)


charts/opentelemetry-operator/values.schema.json
```diff
@@ -573,7 +573,9 @@
           "type": "string",
           "default": "",
           "title": "The featureGates to enable",
-          "examples": ""
+          "examples": [
+            ""
+          ]
```
dmitryax, line 578: This isn't a useful example. I would just remove this line instead

dmitryax:
Fixed in https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1805

PR #1804, opened 2025-08-14
Fix example block for featureGates

tillig:
Helm v3.18.5 is more strict about validating the `values.schema.json` in a Helm chart. When providing examples (plural), the data must be in an array. This updates the `examples` for the `featureGates` property to correctly be an array, matching the other properties that also use array for examples.


dmitryax:
Fixed in https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1805

PR #1808, opened 2025-08-19
Add additionalVolumeMounts to the opentelemetry-demo chart

TimBosman:
## Summary

This PR adds support for specifying extra volume mounts for containers in the Helm chart. Currently, the extra volumes can be mounted only on initContainers and sidecarContainers, not on the regular container.

## Motivation

Some use cases require mounting additional volumes (e.g., configuration files, secrets, or shared storage) for containers. Currently, the chart does not provide a way to specify extra volume mounts for these containers.

## Changes

Adds extraVolumeMounts values to values.yaml and values.schema.json.

## Usage

Example configuration in values.yaml:

```yaml
components:
  image-provider:
    additionalVolumes:
      - name: my-volume
        hostPath:
        path: /home/user/test
        type: ""
    extraVolumeMounts:
      - name: my-volume
        mountPath: /place/to/mount
```


charts/opentelemetry-demo/values.schema.json
```diff
@@ -174,6 +174,12 @@
           "type": "array",
           "items": {}
         },
+        "extraVolumeMounts": {
```
TylerHelmuth: Also update the values.yaml file with a default value for extraVolumeMounts or at least a commented-out entry so that users know it is available.

TimBosman:
Thank you very much for you help. I have adjusted the things you raised. Let me know if I'm missing anything else.

charts/opentelemetry-demo/values.yaml
```diff
@@ -425,6 +425,14 @@ components:
...
       limits:
         memory: 50Mi
+    # additionalVolumes:
+    #   - name: nginx-logs
+    #     hostPath:
+    #     path: /var/log/nginx
+    #     type: ""
+    # extraVolumeMounts:
+    #   - name: nginx-logs
+    #     mountPath: /var/log/nginx
```
TylerHelmuth: Please update the comment at the `components:` field that lists all the available fields

charts/opentelemetry-demo/templates/_objects.tpl
```diff
@@ -72,6 +72,13 @@ spec:
             {{- .livenessProbe | toYaml | nindent 12 }}
           {{- end }}
           volumeMounts:
+          {{- range .extraVolumeMounts }}
+            - name: {{ .name | lower }}
+              mountPath: {{ .mountPath }}
+              {{- if .subPath }}
+              subPath: {{ .subPath }}
+              {{- end }}
+          {{- end }}
```
TylerHelmuth: Lets not define any fields on the extra-volume mount and take the yaml as the user gives it. This is what we do in other charts, like the opentelemetry-collector chart. The json schema should be 

```

    "extraVolumeMounts": {
      "type": "array",
      "items": {
        "type": "object"
      }
    },
```

charts/opentelemetry-demo/templates/_objects.tpl
```diff
@@ -72,6 +72,9 @@ spec:
             {{- .livenessProbe | toYaml | nindent 12 }}
           {{- end }}
           volumeMounts:
+          {{- if .Values.extraVolumeMounts }}
```
TylerHelmuth: The chart uses `additionalVolumes` so lets match that naming and use `additionalVolumeMounts`

TylerHelmuth:
@TimBosman please bump the chart's patch version and run `make generate-examples CHARTS=opentelemetry-demo`

TylerHelmuth (approved):

PR #1822, opened 2025-09-02
[operator] bump operator version to v0.132.0

frzifus:


frzifus:
@TylerHelmuth yes, slightly. We flipped native-sidecars to be always on. Whenever we run `chainsaw/e2e-prometheuscr` or `chainsaw/e2e-sidecar` we need to provide a selector `--selector sidecar=legacy` or `--selector sidecar=natiive` to match the expected result.

Here we run on k8s 1.24, thats why we want to use `--selector sidecar=legacy`.

pavolloffay:
@frzifus I think this PR should resolve https://github.com/open-telemetry/opentelemetry-operator/issues/4331

frzifus:
> @frzifus let me know if you want me to take over this bump.

oh that would be rly great @TylerHelmuth!

PR #1835, opened 2025-09-08
fix(opentelemetry-kube-stack): correct container name for collector logs exclusion

### LINKED ISSUE #1697: [opentelemetry-kube-stack] Cannot exclude collector logs with logsCollection preset
Disabling any [.presets.logsCollection.includeCollectorLogs](https://artifacthub.io/packages/helm/opentelemetry-helm/opentelemetry-kube-stack/0.6.1?modal=values&path=defaultCRConfig.presets.logsCollection.includeCollectorLogs) does not work for [OpenTelemetry Kube Stack Helm Chart v0.6.1](https://artifacthub.io/packages/helm/opentelemetry-helm/opentelemetry-kube-stack/0.6.1) with any collector mode.

I believe I narrowed this down to the container name for the exclude rule being set to the Chart name (defaults to `opentelemetry-kube-stack`) instead of `otc-container`, which is said to be the container name in the [Operator API Docs](https://github.com/open-telemetry/opentelemetry-operator/blob/1c2913a/docs/api/opentelemetrycollectors.md#opentelemetrycollectorspec).

My daemonset OpenTelemetryCollector CRD YAML snippet:
```
    receivers:
      filelog:
        exclude: /var/log/pods/observability_opentelemetry-kube-stack-daemon*_*/opentelemetry-kube-stack/*.log
        include:
        - /var/log/pods/*/*/*.log
```

My daemonset YAML snippet:
```
      containers:
      - args:
        - --config=/conf/collector.yaml
        image: otel/opentelemetry-collector-k8s:0.120.0
        name: otc-container
```

One of my log filepaths:
```
log.file.path:
    /var/log/pods/observability_opentelemetry-kube-stack-daemon-collector-wzjz6_51637406-3496-4291-a188-39e9fa8523d4/otc-container/0.log
```

From [OpenTelemetryCollector (v1beta1) API](https://github.com/open-telemetry/opentelemetry-operator/blob/1c2913a/docs/api/opentelemetrycollectors.md#opentelemetrycollectorspec-1):

> Container names managed by the operator:
>
> - `otc-container`
>

From [OpenTelemetry Kube Stack Helm Chart _config.tpl](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/4c6e67e/charts/opentelemetry-kube-stack/templates/_config.tpl#L20):

> ```
> {{- if .collector.presets.logsCollection.enabled }}
> {{- $_ := set $collector "exclude" (printf "/var/log/pods/%s_%s*_*/%s/*.log" .namespace (include "opentelemetry-kube-stack.collectorFullname" .) (.Chart.Name | lower)) }}
> ```

t3mi:
Fixes #1697

### DIFF
--- a/charts/opentelemetry-kube-stack/templates/_config.tpl
+++ b/charts/opentelemetry-kube-stack/templates/_config.tpl
@@ -17,7 +17,7 @@ the config is written as YAML.
 {{- $_ := set $collector "config" $config }}
 {{- end }}
 {{- if .collector.presets.logsCollection.enabled }}
-{{- $_ := set $collector "exclude" (printf "/var/log/pods/%s_%s*_*/%s/*.log" .namespace (include "opentelemetry-kube-stack.collectorFullname" .) (.Chart.Name | lower)) }}
+{{- $_ := set $collector "exclude" (list (printf "/var/log/pods/%s_%s*_*/otc-container/*.log" .namespace (include "opentelemetry-kube-stack.collectorFullname" .))) }}
 {{- $config = (include "opentelemetry-kube-stack.collector.applyLogsCollectionConfig" (dict "collector" $collector) | fromYaml) -}}
 {{- $_ := set $collector "config" $config }}
 {{- end }}
@@ -317,7 +317,8 @@ receivers:
     exclude: []
     {{- else }}
     # Exclude collector container's logs. The file format is /var/log/pods/<namespace_name>_<pod_name>_<pod_uid>/<container_name>/<run_id>.log
-    exclude: {{ .exclude }}
+    exclude:
+    {{- toYaml .exclude | nindent 4 }}
     {{- end }}
     start_at: end
     retry_on_failure:
--- a/charts/opentelemetry-kube-stack/Chart.yaml
+++ b/charts/opentelemetry-kube-stack/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-kube-stack
-version: 0.10.1
+version: 0.10.2
 description: |
   OpenTelemetry Quickstart chart for Kubernetes.
   Installs an operator and collector for an easy way to get started with Kubernetes observability.
--- a/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/bridge.yaml
+++ b/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/bridge.yaml
@@ -5,7 +5,7 @@ kind: OpAMPBridge
 metadata:
   name: example
   labels:
-    helm.sh/chart: opentelemetry-kube-stack-0.10.1
+    helm.sh/chart: opentelemetry-kube-stack-0.10.2
     app.kubernetes.io/version: "0.129.1"
     app.kubernetes.io/managed-by: Helm
     release: "example"    
--- a/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/collector.yaml
+++ b/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/collector.yaml
@@ -6,7 +6,7 @@ metadata:
   name: example-daemon
   namespace: default
   labels:
-    helm.sh/chart: opentelemetry-kube-stack-0.10.1
+    helm.sh/chart: opentelemetry-kube-stack-0.10.2
     app.kubernetes.io/version: "0.129.1"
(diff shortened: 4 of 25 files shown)


charts/opentelemetry-kube-stack/templates/_config.tpl
```diff
@@ -17,7 +17,7 @@ the config is written as YAML.
 {{- $_ := set $collector "config" $config }}
 {{- end }}
 {{- if .collector.presets.logsCollection.enabled }}
-{{- $_ := set $collector "exclude" (printf "/var/log/pods/%s_%s*_*/%s/*.log" .namespace (include "opentelemetry-kube-stack.collectorFullname" .) (.Chart.Name | lower)) }}
+{{- $_ := set $collector "exclude" (printf "/var/log/pods/%s_%s*_*/otc-container/*.log" .namespace (include "opentelemetry-kube-stack.collectorFullname" .)) }}
```
jaronoff97: i think this will only exclude default images, but not necessarily custom images. can we provide a way to override?

t3mi: Added separate input for container names with default set to `otc-container`.

charts/opentelemetry-kube-stack/templates/_config.tpl
```diff
@@ -317,7 +321,8 @@ receivers:
     exclude: []
     {{- else }}
     # Exclude collector container's logs. The file format is /var/log/pods/<namespace_name>_<pod_name>_<pod_uid>/<container_name>/<run_id>.log
-    exclude: {{ .exclude }}
+    exclude:
```
jaronoff97, line 320: it seems like we don't have a test for this preset yet, or if we do nothing is changing today which is unexpected. can you update an example to verify this works?

t3mi: I've updated existing prometheus-otel example to include logs collection. If that's not desired I could add new example to cover that.

charts/opentelemetry-kube-stack/examples/prometheus-otel/values.yaml
```diff
@@ -35,7 +35,8 @@ collectors:
             name: otel-collector-secret
     presets:
       logsCollection:
-        enabled: false
+        enabled: true
+        includeCollectorLogs: false
```
jaronoff97: lets change an example that's already using logs collection, this example serves as a replacement for the kube-prom-stack chart.

jaronoff97: should we also test setting `collectorContainerNames`?

t3mi: added new example as there were none with logs enabled. Multiple containers were also specified for testing.

charts/opentelemetry-kube-stack/values.yaml
```diff
@@ -397,6 +397,12 @@ defaultCRConfig:
     logsCollection:
       enabled: false
       includeCollectorLogs: true
+      # If includeCollectorLogs is set to false, below list of container names
+      # used in collector would be used to generate exclusion rules
+      collectorContainerNames:
+        - otc-container
+        # - init-nginx
+        # - additional-container
```
TylerHelmuth: I dont like the precedence this sets. We have historically not allowed the preset logic to be configurable. I'd prefer the preset make an assumption for the collector container name based on the chart's default. @jaronoff97 what do you think?

jaronoff97: yeah i think that's a fair assertion... @t3mi after thinking about it, i think Tyler is right here. Let's just leave it at `otc-container`. If someone wants to override they can disable the preset and copy the config manually. Apologies for the back and forth 🙇

t3mi: No worries, reverted to initial fix. Thanks for reviews!
[Here](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1835/files#diff-32591affb5b32287ce2f9c5bfedee3d36bf33564e8df00bf6e85c168d1e228feR98) is an updated example.

jaronoff97 (approved):

TylerHelmuth:
@t3mi thats for this fix. Please handle the merge conflicts and then I'll try to make this the next merge for kube-stack

Issue #1826, opened 2025-09-04
[opentelemetry-collector] Cannot set zero replicas for deployment


yurytaranau:
We are planning to deploy specific collectors to the standby region and keep them scaled down to zero replicas. However, it seems that setting zero replicas for a deployment is currently not possible, as Helm interprets `0` as an unset value and defaults to `1`: https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/templates/deployment.yaml#L16-L18

For further details, you can refer to the related Helm issue: [Helm Issue #3164](https://github.com/helm/helm/issues/3164).

May I kindly ask if there is a specific reason for using a default value in this case? Or could the changes introduced in this [PR](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1504) potentially be reverted?

I would appreciate your insights on this. Thank you!

PR #1839, opened 2025-09-11
[chore] [collector] Revert replica count in case of Deployment mode

yurytaranau:
Fixes: #1826 
Revers: https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1504

Setting the replica count to zero in a deployment is prevented because Helm interprets a value of `0` as `unset` and defaults it to `1`. This update removes the `default 1` logic, enabling explicit control over the replica count, including support for setting it to `zero`. If the replicaCount value is not specified, the default value of `1` from values.yaml will be applied.

For more details, refer to the related Helm issue: [Helm Issue #3164](https://github.com/helm/helm/issues/3164).


charts/opentelemetry-collector/templates/deployment.yaml
```diff
@@ -14,7 +14,7 @@ metadata:
   {{- end }}
 spec:
 {{- if not .Values.autoscaling.enabled }}
-  replicas: {{ .Values.replicaCount | default 1 }}
+  replicas: {{ .Values.replicaCount }}
```
jaronoff97, line 17: this should be fine given we set replicaCount to 1 [here ](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/595204cdafda21c73099f12686b6f1eac8822d33/charts/opentelemetry-collector/values.yaml#L406)

PR #1849, opened 2025-09-14
feat(target-allocator): add scheduling configuration support

chenlujjj:
Add affinity, tolerations, nodeSelector, and topologySpreadConstraints support to target-allocator chart:

- Add scheduling fields to values.yaml with default nodeSelector
- Update deployment template to support all scheduling configurations
- Bump chart version from 0.126.4 to 0.126.5
- Regenerate examples with updated templates


charts/opentelemetry-target-allocator/values.yaml
```diff
@@ -55,3 +55,9 @@ targetAllocator:
     # requests:
     #   cpu: 100m
     #   memory: 64Mi
+
+affinity: {}
+tolerations: []
+nodeSelector:
+  kubernetes.io/os: linux
```
TylerHelmuth: Does there need to be a default?

chenlujjj: Copied from opentelemetry-operator values, I think we can leave it empty. 
Updated

chenlujjj:
Bumped chart version to 0.126.8 and regenerated the examples.

PR #1850, opened 2025-09-14
feat(target-allocator): add ServiceMonitor support

chenlujjj:
Add ServiceMonitor resource support to target-allocator chart.

- Add serviceMonitor configuration section to values.yaml
- Create servicemonitor.yaml template with full feature support
- Support extraLabels, annotations, relabelings, and metricRelabelings

### DIFF
--- a/charts/opentelemetry-target-allocator/values.yaml
+++ b/charts/opentelemetry-target-allocator/values.yaml
@@ -55,3 +55,17 @@ targetAllocator:
     # requests:
     #   cpu: 100m
     #   memory: 64Mi
+
+  ## Enable ServiceMonitor for Prometheus metrics scrape
+  serviceMonitor:
+    enabled: false
+    # additional labels on the ServiceMonitor
+    extraLabels: {}
+    # add annotations on the ServiceMonitor
+    annotations: {}
+    metricsEndpoints:
+      - port: http-port
+    # Used to set relabeling and metricRelabeling configs on the ServiceMonitor
+    # https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config
+    relabelings: []
+    metricRelabelings: []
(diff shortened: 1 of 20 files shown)


charts/opentelemetry-target-allocator/values.yaml
TylerHelmuth, line 1: @jaronoff97 does the target allocator support configuring OTel metrics via the otel sdk configuration file? If so I'd prefer we not add dependencies on the prometheus operator

atoulme: See https://github.com/open-telemetry/opentelemetry-operator/pull/4062 - we moved to the otel sdk. We didn't change how we expose metrics though.

TylerHelmuth: My opinion would be to not add an indirect dependency on the prometheus operator for this chart like we did for the collector chart. Back then we didn't have any solution but prometheus to export the metrics. Since we do now, I'd prefer not to add features that depend on a resource being installed that isn't in this chart.

We can add an `extraObjects` instead for people who what the install to also install a service monitor. The kube-stack chart already allows this.

chenlujjj: Agreed that `extraObjects` is needed to provide a more flexible way for users to add any k8s objects they want, I'd like to create another PR for that.

Regarding ServiceMonitor, since we disable it by default, this chart wouldn't depend on the ServiceMonitor CRD. Actually I guess in lots of use cases of target allocator, there should be ServiceMonitor CRD in the same k8s cluster, so it should not be a big problem. By adding this template, users can enable ServiceMonitor in a simpler way.

TylerHelmuth: The concern I have (and why in general I dont like adding dependencies on stuff outside the chart) is that we have to support it and that support burden is high. If someone opens an issue saying our ServiceMonitor doesn't work, I dont have a good way to investigate that.

chenlujjj: Understood your concern. I'll close this PR and make another PR adding `extraObjects`

PR #1851, opened 2025-09-14
feat(target-allocator): add health probes and replica count support

chenlujjj:
Add liveness probe, readiness probe, and configurable replica count support to target-allocator chart.

- Add replicaCount configuration with default value of 1
- Add livenessProbe configuration with default /livez endpoint
- Add readinessProbe configuration with default /readyz endpoint
- Update deployment template to support all probe configurations

All probes are enabled by default with sensible defaults, and can be fully customized including timeouts, thresholds, and intervals.

### DIFF
--- a/charts/opentelemetry-target-allocator/templates/deployment.yaml
+++ b/charts/opentelemetry-target-allocator/templates/deployment.yaml
@@ -6,7 +6,7 @@ metadata:
   labels:
     {{- include "helper.commonLabels" . | nindent 4 }}
 spec:
-  replicas: 1
+  replicas: {{ .Values.replicaCount }}
   selector:
     matchLabels:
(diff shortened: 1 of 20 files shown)


charts/opentelemetry-target-allocator/templates/deployment.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
   labels:
     {{- include "helper.commonLabels" . | nindent 4 }}
 spec:
-  replicas: 1
+  replicas: {{ .Values.replicaCount }}
```
TylerHelmuth, line 9: does the targetallocator support being deployed with multiple replicas?

jaronoff97: yes, in its default consistent hashing configuration, it's horizontally scalable.

PR #1853, opened 2025-09-16
feat(opentelemetry-operator): opt-in to use user namespace

t3mi:
With Kubernetes 1.33, [user namespaces](https://kubernetes.io/docs/concepts/workloads/pods/user-namespaces/) are supported to enhance pod security by isolating pod processes in a separate user namespace, reducing the risk of privilege escalation. This PR adds an opt-in configuration to enable user namespaces for OpenTelemetry Operator pod only.

### DIFF
--- a/charts/opentelemetry-operator/templates/deployment.yaml
+++ b/charts/opentelemetry-operator/templates/deployment.yaml
@@ -37,7 +37,7 @@ spec:
     spec:
       automountServiceAccountToken: {{ .Values.automountServiceAccountToken }}
       hostNetwork: {{ .Values.hostNetwork }}
-      {{- if and (semverCompare ">=1.33-0" .Capabilities.KubeVersion.Version) (kindIs "bool" .Values.hostUsers) }}
+      {{- if and (semverCompare ">=1.30-0" .Capabilities.KubeVersion.Version) (kindIs "bool" .Values.hostUsers) }}
       hostUsers: {{ .Values.hostUsers }}
       {{- end }}
       {{- if .Values.priorityClassName }}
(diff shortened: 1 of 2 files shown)


charts/opentelemetry-operator/templates/deployment.yaml
```diff
@@ -37,6 +37,9 @@ spec:
     spec:
       automountServiceAccountToken: {{ .Values.automountServiceAccountToken }}
       hostNetwork: {{ .Values.hostNetwork }}
+      {{- if and (semverCompare ">=1.33-0" .Capabilities.KubeVersion.Version) (kindIs "bool" .Values.hostUsers) }}
```
TylerHelmuth: Using `.Values.hostUsers` is acceptable.


```suggestion
      {{- if and (semverCompare ">=1.33-0" .Capabilities.KubeVersion.Version) .Values.hostUsers }}
```

charts/opentelemetry-operator/values.yaml
```diff
@@ -360,6 +360,9 @@ nodeSelector:
 topologySpreadConstraints: []
 hostNetwork: false
 
+# Whether to use user namespace or not
```
TylerHelmuth, line 370: Please add a comment explaining this is only available on k8s 1.33 and above.

t3mi: added

charts/opentelemetry-operator/templates/deployment.yaml
t3mi: If there will be no check for kind, it would not be possible actually to enable user namespaces as condition would not be satisfied (e.g. `hostUsers: false`). It's either check only for kubernetes version and set default value to `true` or have it like this to omit the property in the rendered manifests unless user explicitly decide to set the value.

TylerHelmuth: @t3mi based on the docs you linked this has been beta since 1.30. Can we reduce this semver check?

t3mi: Changed to 1.30

PR #1830, opened 2025-09-07
fix(opentelemetry-kube-stack): support tpl for affinity/volumes/volumeMounts

t3mi:
This will allow dynamic references to collector name or other template information like below:

```yaml
defaultCRConfig:
  enabled: true
  affinity:
    podAntiAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        - labelSelector:
            matchLabels:
              app.kubernetes.io/component: opentelemetry-collector
              app.kubernetes.io/instance: "{{ .namespace }}.{{ include \"opentelemetry-kube-stack.collectorFullname\" . }}"
          topologyKey: kubernetes.io/hostname
  volumeMounts:
    - mountPath: /var/lib/otelcol/{{ include "opentelemetry-kube-stack.collectorFullname" . }}
      name: varlibotelcol
  volumes:
    - hostPath:
        path: /var/lib/otelcol/{{ include "opentelemetry-kube-stack.collectorFullname" . }}
        type: DirectoryOrCreate
      name: varlibotelcol
```

PR #1855, opened 2025-09-16
fix(opentelemetry-kube-stack): support tpl for collector configuration

t3mi:
Expands #1830 with template support for collector configuration:
- adds new example for isolated multi-collector deployment sharing same node storage for the queues
- adds clusterName into the collector context

Actual changes and impact are part of second commit.


charts/opentelemetry-kube-stack/templates/_config.tpl
```diff
@@ -41,7 +41,7 @@ the config is written as YAML.
 {{- $config = (include "opentelemetry-kube-stack.collector.applyClusterMetricsConfig" (dict "collector" $collector "namespace" .namespace) | fromYaml) -}}
 {{- $_ := set $collector "config" $config }}
 {{- end }}
-{{- toYaml $collector.config | nindent 4 }}
+{{- tpl (toYaml $collector.config) . | nindent 4 }}
```
jaronoff97, line 44: the change here makes sense, but it would be useful to see this working with a modified example. It would be easy enough to template in something like an annotation to check

t3mi: Values for newly added example cover such functionality as replacing namespace, clusterName and collector name, specifically the diff for the second commit shows that. Do you want one more example to be modified?

jaronoff97: oh i must have missed that! nope youre all good. The examples tend to be blend together during review. Thank you :bow:

jaronoff97:
Thank you for your contribution!

Issue #1743, opened 2025-07-01
[target-allocator] Additional property global is not allowed


nlamirault:
I would like to use the new chart Target Allocator, and I've got an error when I try to template:

```yaml
---
apiVersion: v2
type: application
name: opentelemetry-stack
version: 1.0.0
appVersion: 1.0.0
dependencies:
- name: opentelemetry-target-allocator
  repository: https://open-telemetry.github.io/opentelemetry-helm-charts
  version: 0.126.1
```

and the error:

```shell
$ helm template otel . -n opentelemetry
Error: values don't meet the specifications of the schema(s) in the following chart(s):
opentelemetry-target-allocator:
- (root): Additional property global is not allowed
```

This version works fine (using the Collector Helm chart):

```yaml
apiVersion: v2
type: application
name: opentelemetry-stack
version: 1.0.0
appVersion: 1.0.0
dependencies:
# - name: opentelemetry-target-allocator
#   repository: https://open-telemetry.github.io/opentelemetry-helm-charts
#   version: 0.126.1
- name: opentelemetry-collector
  repository: https://open-telemetry.github.io/opentelemetry-helm-charts
  version: 0.122.5
```

Any idea how to fix that ?

JaredTan95:
You need to check your `values.yaml` content and if there is `global` field under `opentelemetry-target-allocator` section.

PR #1856, opened 2025-09-19
[opentelemetry-target-allocator] support for `global` and `enabled` in `values.schema.json`

JaredTan95:
Fixes https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1743

### DIFF
--- a/charts/opentelemetry-target-allocator/values.schema.json
+++ b/charts/opentelemetry-target-allocator/values.schema.json
@@ -2,6 +2,13 @@
   "$schema": "https://json-schema.org/draft/2020-12/schema",
   "type": "object",
   "properties": {
+    "enabled": {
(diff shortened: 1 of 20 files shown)


charts/opentelemetry-target-allocator/values.schema.json
```diff
@@ -2,6 +2,13 @@
   "$schema": "https://json-schema.org/draft/2020-12/schema",
   "type": "object",
   "properties": {
+    "enabled": {
```
jaronoff97, line 5: @JaredTan95 can you please bump the patch version for this one?

JaredTan95: Oh, it was overwritten after merging the main branch

PR #1859, opened 2025-09-21
fix: namespace compromise via hostpid

### LINKED ISSUE #1858: Namespace Compromise via hostPID
We identified the usage of hostPID in your Kubernetes manifest enables unauthorized namespace access and container introspection. Processes inside the container can observe or interfere with other workloads running on the node, violating container isolation.

We provide supporting evidence from Kubernetes Pod Security Standards (Baseline & Restricted) and https://github.com/BishopFox/badPods/tree/main/manifests/hostpid which explicitly discourage hostPID: true unless absolutely required, due to risk of privilege escalation and information disclosure.

https://github.com/open-telemetry/opentelemetry-helm-charts/blob/ae5fb8291dde7e5984469f049ce7f037475709b2/charts/opentelemetry-ebpf/templates/kernel-collector-daemonset.yaml#L153

Expected Behavior: 

do not set hostPID: true unless absolutely required.

zyue110026:
fix #1858 by setting hostPID to false

### DIFF
--- a/charts/opentelemetry-ebpf/templates/kernel-collector-daemonset.yaml
+++ b/charts/opentelemetry-ebpf/templates/kernel-collector-daemonset.yaml
@@ -150,7 +150,7 @@ spec:
 {{- end }}
       dnsPolicy: ClusterFirstWithHostNet
       hostNetwork: true
-      hostPID: true
+      hostPID: false
       serviceAccountName: {{ include "opentelemetry-collector-kernel-collector.serviceAccountName" . }}
       volumes:
       - name: {{ include "opentelemetry-ebpf.fullname" . }}-config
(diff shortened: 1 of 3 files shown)


charts/opentelemetry-ebpf/templates/kernel-collector-daemonset.yaml
```diff
@@ -150,7 +150,7 @@ spec:
 {{- end }}
       dnsPolicy: ClusterFirstWithHostNet
       hostNetwork: true
-      hostPID: true
+      hostPID: false
```
JaredTan95, line 153: I think this is a crucial parameter for ebpf programs.

PR #1872, opened 2025-09-25
feat(demo): add ipv6 environment vars

gnmeyer:
[Link to otel-demo feature](https://github.com/open-telemetry/opentelemetry-demo/pull/2594)

### DIFF
--- a/charts/opentelemetry-demo/examples/bring-your-own-observability/rendered/opensearch/poddisruptionbudget.yaml
+++ b/charts/opentelemetry-demo/examples/bring-your-own-observability/rendered/opensearch/poddisruptionbudget.yaml
@@ -1,6 +1,6 @@
 ---
 # Source: opentelemetry-demo/charts/opensearch/templates/poddisruptionbudget.yaml
-apiVersion: policy/v1beta1
+apiVersion: policy/v1
 kind: PodDisruptionBudget
 metadata:
   name: "opensearch-pdb"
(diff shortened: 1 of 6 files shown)


julianocosta89:
> The rendered files were made from the `make generate-examples` command as a target from the Makefile. Looking at the other examples, they also have the rendered out files. Unless one of the workflows runs the `make generate-examples` command, all of these files should be fine to be merged in.

Yes, but that's because we are adding a "new" example.
My question is, do we need that new example?

I'm asking because every single example we add, we have hundreds of extra lines of code generated, that need to be maintained.

julianocosta89:
@gnmeyer / @iturgut  can we get the lint fixed?
This one should be good to go!

charts/opentelemetry-demo/examples/bring-your-own-observability/rendered/opensearch/poddisruptionbudget.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 # Source: opentelemetry-demo/charts/opensearch/templates/poddisruptionbudget.yaml
-apiVersion: policy/v1
+apiVersion: policy/v1beta1
```
TylerHelmuth: @gnmeyer lint is failing bc of all these poddistruptionbudget.yaml changes. Maybe a holdover from a bad merge?

gnmeyer: @TylerHelmuth fixed!

Issue #1879, opened 2025-09-30
OpenTelemetry demo chart default config not working correctly


phgeraldeli:
# Bug Report

I`m using **ChartVersion 0.38.0** and **app version 2.1.3** with the following values:

```
  default:
    image:
      tag: "2.1.3"
  components:
    frontend-proxy:
      service:
        type: LoadBalancer
```

## Symptom

Several components are not working correctly.
For example:

- Prometheus shows many targets as down
- Jaeger UI shows nothing in the monitor tab

Prometheus errors:

- Error scraping target: Get "http://10.0.44.23:4317/metrics": net/http: HTTP/1.x transport connection broken: malformed HTTP response "\x00\x00\x06\x04\x00\x00\x00\x00\x00\x00\x05\x00\x00@\x00"
- Error scraping target: server returned HTTP status 404 Not Found
- Error scraping target: server returned HTTP status 400 Bad Request
- Error scraping target: Get "http://10.0.44.23:6831/metrics": dial tcp 10.0.44.23:6831: connect: connection refused
- Error scraping target: Get "http://10.0.44.23:14250/metrics": net/http: HTTP/1.x transport connection broken: malformed HTTP response "\x00\x00\x06\x04\x00\x00\x00\x00\x00\x00\x05\x00\x00@\x00"
- Error scraping target: server returned HTTP status 404 Not Found
- Error scraping target: Get "http://10.0.44.23:8888/metrics": dial tcp 10.0.44.23:8888: connect: connection refused

See the following images:

<img width="1067" height="769" alt="Image" src="https://github.com/user-attachments/assets/4cff1aa3-eec0-452b-8908-4c1c2ed7be15" />

<img width="1397" height="687" alt="Image" src="https://github.com/user-attachments/assets/8bd33170-d7da-4824-87bf-3c4f14e085fe" />

## Reproduce

1. Create brand new cluster EKS

2. Create Chart.yaml

```
apiVersion: v2
name: example-chart
description: A Helm chart to install all example demo applications
version: 1.0.0
dependencies:
  - name: opentelemetry-demo
    version: "0.38.0"
    repository: "https://open-telemetry.github.io/opentelemetry-helm-charts"
```

3. Create values.yaml

```
opentelemetry-demo:
  default:
    image:
      tag: "2.1.3"
  components:
    frontend-proxy:
      service:
        type: LoadBalancer
```

4. Install Helm chart

# Additional Question

Someone has a config that already have everything working?

[...]

PR #1880, opened 2025-10-01
[demo] add missing params and fix dashboards

julianocosta89:
Working on: https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1879

This PR fixes the Shipping endpoint that now must have the protocol as part of the ADDR.
It also syncs the Grafana dashboards we have on the Demo.

This doesn't fix the Jaeger UI not showing span metrics, since we use the Jaeger Helm chart and the last update from it was on 20 Mar, 2025 (v3.4.1), which shipped Jaeger v `1.53.0`.

The fix to the issue was merged on Sep 22, 2024, but it is not part of the Jaeger `1.53.0` release. (https://github.com/jaegertracing/jaeger/pull/6007)

Whenever the Jaeger Helm chart v2 is available we will migrate to it. Till there, Jaeger in k8s will only work for traces.


julianocosta89:
> Whenever the Jaeger Helm chart v2 is available we will migrate to it. Till there, Jaeger in k8s will only work for traces.

@TylerHelmuth done! ✅ 

I also asked the Jaeger folks on slack about that: https://cloud-native.slack.com/archives/CGG7NFUJ3/p1759389861625699

PR #1881, opened 2025-10-01
fix: wrong values schema definition for operator

gamerslouis:


TylerHelmuth (commented):
@gamerslouis thanks for fixing this. Please update the chart's patch version and then run `make generate-examples CHARTS=opentelemetry-operator`

PR #1882, opened 2025-10-01
Update grafana resources to new opensearch database name

grzesuav:
Similar to https://github.com/open-telemetry/opentelemetry-demo/pull/2363/


grzesuav:
Ok, will close the . Is there any previous version where both gradana and Jaeger were working ?

PR #1893, opened 2025-10-07
[otel-demo] Bump Grafana to v12.2.0

cyrille-leclerc:
Follow up on OTel Demo's Docker Compose change:
* https://github.com/open-telemetry/opentelemetry-demo/pull/2615

Fixes:
* https://github.com/open-telemetry/opentelemetry-demo/issues/2589

### DIFF
--- a/charts/opentelemetry-demo/values.yaml
+++ b/charts/opentelemetry-demo/values.yaml
@@ -982,6 +982,8 @@ prometheus:
 grafana:
   enabled: true
   fullnameOverride: grafana
+  image:
+    tag: "12.2.0"
   testFramework:
     enabled: false
   grafana.ini:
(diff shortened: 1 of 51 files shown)


charts/opentelemetry-demo/values.yaml
```diff
@@ -982,6 +982,8 @@ prometheus:
 grafana:
   enabled: true
   fullnameOverride: grafana
+  image:
```
JaredTan95, line 985: Why not upgrade together through `grafana` chart in chart.yaml ?

julianocosta89: @cyrille-leclerc, @JaredTan95 is correct.
The dependent chart version should be the one bumped: https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-demo/Chart.yaml#L29

Issue #1895, opened 2025-10-07
[otel-kube-stack] host metrics lack `host.name` or `k8s.node.name` attribute


cyrille-leclerc:
When using the OTel Collector `presets/hostmetrics` ([values.yaml](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/f4a8cba72f14c8d7c036cd3914b957628f5a5997/charts/opentelemetry-kube-stack/values.yaml#L427)), the system metrics like `system.cpu.time` don’t get enriched with the `host.name` or `k8s.node.name` attributes, just getting enriched with `k8s.cluster.name`, making the metric unusable.

Example:

```
system_cpu_time_seconds_total{cpu="cpu2", k8s_cluster_name="otel-demo-stack", state="steal"}7.96
```

cyrille-leclerc:
No worries Dylan. I happen to be a PM working at simplifying the Mimir OTel experience, I understand how this config is error prone.
Cc @aknuds1 and @ywwg with whom I work on Mimir resource attributes promotion.

PR #1896, opened 2025-10-07
[otel-kube-stack] Add `host.name` and `k8s.node.name` to host metrics

cyrille-leclerc:
Fixes:
* https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1895

Introduce in the OTel Collector telemetry pipeline:
* The `k8snode`  detector ([docs](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor#k8s-node-metadata)) of the `resourcedetection` processor. Note that the K8s permissions are already good.
* Recopy of `k8s.node.name` as `host.name` so that K8s agnostic dashboards and alerts can be used.

### DIFF
--- a/charts/opentelemetry-kube-stack/values.yaml
+++ b/charts/opentelemetry-kube-stack/values.yaml
@@ -551,7 +551,7 @@ collectors:
               - otlp
             processors:
               - resourcedetection/env
-              - transform/set-hostname
+              - resource/hostname
               - batch
             exporters:
               - debug
--- a/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/collector.yaml
+++ b/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/collector.yaml
@@ -412,7 +412,7 @@ spec:
           processors:
           - k8sattributes
           - resourcedetection/env
-          - transform/set-hostname
+          - resource/hostname
           - batch
           receivers:
           - otlp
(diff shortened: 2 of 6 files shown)


charts/opentelemetry-kube-stack/values.yaml
```diff
@@ -510,14 +510,22 @@ collectors:
...
+          timeout: 2s
+          override: false
+        transform/set-hostname:
+          error_mode: ignore
+          trace_statements:
+            - set(resource.attributes["host.name"], resource.attributes["k8s.node.name"]) where resource.attributes["host.name"] == nil
+          metric_statements:
+            - set(resource.attributes["host.name"], resource.attributes["k8s.node.name"]) where resource.attributes["host.name"] == nil
+          log_statements:
+            - set(resource.attributes["host.name"], resource.attributes["k8s.node.name"]) where resource.attributes["host.name"] == nil
```
rogercoll: What do you think if we use the [resource](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourceprocessor) processor instead? That will prevent having to check the existence of the `host.name` and `k8s.node.name` resource attributes (and the configuration will be the same for any signal).

```suggestion
        resource/hostname:
          attributes:
            - key: host.name
              from_attribute: k8s.node.name
              action: insert
```

cyrille-leclerc: Thanks for the suggestion, please review updated code.

rogercoll (approved):

Issue #1713, opened 2025-06-13
Add  helm.sh/resource-policy: keep in CRDs


younusmohmmad:
Hi,

I noticed that the CRD manifest files in the OpenTelemetry Operator Helm chart repository ([link](https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-operator/conf/crds)) do not include the label:

helm.sh/resource-policy: keep

As a result, when the Helm chart is uninstalled, the CRDs are also deleted. Since CRDs are cluster-scoped resources, it is common to install them once and have multiple namespaces use the corresponding Custom Resources.

We have a use case where we want to install the CRDs in a separate namespace and ensure that they persist even if the operator Helm chart is deleted. This allows us to manage CRs in different namespaces without affecting the CRDs.

Would it be possible to update the Helm chart to add the helm.sh/resource-policy: keep annotation to the CRD manifests? This would prevent Helm from deleting the CRDs during uninstallation and better support this multi-namespace use case.

Thank you!

PR #1901, opened 2025-10-14
feat(opentelemetry-operator): add annotations support to CRDS

DexterDixon:
fix: #1713

This is also my first contribution.


TylerHelmuth:
@DexterDixon thanks for contributing, you picked a very tricky topic to start lol we've had a lot of pain in the past with the CRDs because of helm uninstall process not removing them. We have a lot of logic in the repo to keep these CRDs up-to-date with the operator version on upgrades as well (all the logic is in the Makefile). We'll need to be very careful how we add this feature.

@jaronoff97 is this safe to do (assuming our automated upgrade script is updated)? How would this affect a `helm upgrade`?

PR #1912, opened 2025-10-23
fix(opentelemetry-demo): Add missing ENVOY_ADDR env variable to frontend-proxy

girodav:
[This PR](https://github.com/open-telemetry/opentelemetry-demo/pull/2594/files) introduced a new environment variable, ENVOY_ADDR. It was added to the .env file used for Docker deployments, but it’s missing from the Helm chart.

### DIFF
--- a/charts/opentelemetry-demo/Chart.yaml
+++ b/charts/opentelemetry-demo/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: opentelemetry-demo
-version: 0.38.3
+version: 0.38.4
 description: opentelemetry demo helm chart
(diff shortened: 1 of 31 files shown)


julianocosta89 (commented):
hello @girodav thx for submitting this PR.
The new env var must be added just when we get a new release.

The Helm chart has the container version set to the latest release:
`image: 'ghcr.io/open-telemetry/demo:2.1.3-frontend-proxy'`

frontend-proxy v2.1.3 doesn't need this env var.

TylerHelmuth:
Closing for now since we can't take this until the otel deom does a release and the helm chart is updated to use that release.

PR #1738, opened 2025-06-28
Use k8sattributes processor's native support for service attributes

priyalakshumi:
This PR updates the Helm chart to leverage native support for extracting service-related attributes
1.` service.version` is now resolved using native support via the k8sattributes processor.
2. `service.name` follows the below hierarchy defined by the OpenTelemetry semantic conventions. To ensure the correct value is selected, a pod annotation has been added
  ```
  1. pod.annotation[resource.opentelemetry.io/service.name]
  2. pod.label[app.kubernetes.io/instance] 
  3. pod.label[app.kubernetes.io/name]
```
3. The `k8s-app` label appears to be a custom, so it has been left unchanged.
Fixes: #1723

PR #1918, opened 2025-10-28
[opentelemetry-collector] Align `presets/kubernetesAttributes` on the opentelemetry-kube-stack presets

cyrille-leclerc:
(opentelemetry-collector) Align presets/kubernetesAttributes on the opentelemetry-kube-stack preset and benefit of:
* https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1738

### DIFF
--- a/charts/opentelemetry-operator/Chart.yaml
+++ b/charts/opentelemetry-operator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.99.0
+version: 0.99.1
 description: OpenTelemetry Operator Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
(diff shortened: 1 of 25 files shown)


charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -344,17 +344,31 @@ processors:
...
-        - "k8s.deployment.name"
-        - "k8s.statefulset.name"
-        - "k8s.daemonset.name"
-        - "k8s.cronjob.name"
-        - "k8s.job.name"
-        - "k8s.node.name"
-        - "k8s.pod.name"
-        - "k8s.pod.uid"
-        - "k8s.pod.start_time"
+        - k8s.namespace.name
```
cyrille-leclerc, line 349: FYI I changed the order of the attributes to strictly align with the preset of the otel-kube-stack, just copy/pasting the code.

TylerHelmuth, line 349: @open-telemetry/helm-approvers can you think of any reason adding addition resource attributes would be a breaking change?

PR #1922, opened 2025-10-31
Fix jaeger-query endpoint in otel-demo

lmolkova:
Trying to run k8s version of otel demo following https://opentelemetry.io/docs/demo/kubernetes-deployment/.

Everything works, but when I go to grafana dashboard and try to get traces from jaeger, I see error `dial tcp: lookup jaeger on 10.96.0.10:53: server misbehaving`.

The datasource is configured to `jaeger:16686`, but there is no `jaeger` endpoint, there are `jaeger-agent`, `jaeger-collector` and `jaeger-query`. 

Changing Grafana datasource to `jaeger-query` fixes the problem.

I tested it by installing chart locally - otel-demo including grafana works fine.

/cc @cyrille-leclerc

### DIFF
--- a/charts/opentelemetry-collector/examples/alternate-config/rendered/deployment.yaml
+++ b/charts/opentelemetry-collector/examples/alternate-config/rendered/deployment.yaml
@@ -26,7 +26,7 @@ spec:
   template:
     metadata:
       annotations:
-        checksum/config: 49bc4f1aa532d0e6c85ef87af41dbd261a17c021f4a3aebf386465e5c4cd5786
+        checksum/config: 3fdb7e9482f11f60ac4a82c488d7b1f70d9ebecd0a13c287d74c5beb2192f61f
         
       labels:
         app.kubernetes.io/name: opentelemetry-collector
--- a/charts/opentelemetry-demo/examples/bring-your-own-observability/rendered/flagd-config.yaml
+++ b/charts/opentelemetry-demo/examples/bring-your-own-observability/rendered/flagd-config.yaml
@@ -14,41 +14,114 @@ metadata:
     app.kubernetes.io/managed-by: Helm
 data:
   
-  demo.flagd.json: "{\r\n  \"$schema\": \"https://flagd.dev/schema/v0/flags.json\",\r\n
-    \ \"flags\": {\r\n    \"productCatalogFailure\": {\r\n      \"description\": \"Fail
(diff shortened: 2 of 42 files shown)


TylerHelmuth (approved):

julianocosta89 (approved):

charts/opentelemetry-collector/examples/alternate-config/rendered/deployment.yaml
```diff
@@ -26,7 +26,7 @@ spec:
   template:
     metadata:
       annotations:
-        checksum/config: 3fdb7e9482f11f60ac4a82c488d7b1f70d9ebecd0a13c287d74c5beb2192f61f
+        checksum/config: 49bc4f1aa532d0e6c85ef87af41dbd261a17c021f4a3aebf386465e5c4cd5786
```
lmolkova: I'm not sure why this and similar need to be changed. 

I ran `make generate-examples CHARTS=opentelemetry-demo`, but then CI has still failed and told me to generate all examples. I did and here we are. 

I'm investigating, but  please let me know if it's a normal part of the process.

charts/opentelemetry-demo/examples/bring-your-own-observability/rendered/flagd-config.yaml
```diff
@@ -6,122 +6,49 @@ metadata:
...
-          "variants": {
-            "10sec": 10000,
-            "5sec": 5000,
-            "off": 0
-          },
-          "defaultVariant": "off"
-        }
-      }
-    }
+  demo.flagd.json: "{\r\n  \"$schema\": \"https://flagd.dev/schema/v0/flags.json\",\r\n
```
lmolkova: will fix this

charts/opentelemetry-collector/examples/alternate-config/rendered/deployment.yaml
TylerHelmuth: It is not expected that your changes in the demo charts would result in other chart's examples changing. It is possible you ran `make generate-examples` without a `CHARTS` filter and that would generate all the examples, which can fail if you're missing chart dependencies or the latest from main.

To proceed, undo all the changes made to other charts. Then run `make generate-examples CHARTS=opentelemetry-demo`.

lmolkova: not sure what was the problem, but generating on windows + wsl results in such extra changes. generating on mac works fine.

PR #1923, opened 2025-10-31
fix(opentelemetry-demo): update flagd config with email memory link flag

Red-GV:
This PR updates the flagd config after the last version update for the demo. This added the memory link for the email service, but the configuration file was not updated. This corrects it so that the fault can be used.


julianocosta89:
@Red-GV thx for that!

Would you be able to run `make generate-examples CHARTS=opentelemetry-demo` and push the changes?

Red-GV:
Np, @julianocosta89 . Though I thought I already did that in the PR? Do I need to generate examples for the whole repo?

PR #1926, opened 2025-11-03
[opentelemetry-ebpf-instrumentation] Fix image repository to match ghcr and fix make generate-examples

### LINKED ISSUE #1925: Error running make generate-examples
I'm getting the following error when running `make generate-examples` with the latest chart: 

````
Error: could not find charts/opentelemetry-ebpf-instrumentationopentelemetry-ebpf-instrumentation: stat charts/opentelemetry-ebpf-instrumentationopentelemetry-ebpf-instrumentation: no such file or directory
find: charts/opentelemetry-ebpf-instrumentationopentelemetry-ebpf-instrumentation/examples: No such file or directory
````

I believe the issue is due to the following line of the Makefile: 

````
CHARTS ?= opentelemetry-collector opentelemetry-operator opentelemetry-demo opentelemetry-ebpf opentelemetry-kube-stack opentelemetry-target-allocator opentelemetry-ebpf-instrumentationopentelemetry-ebpf-instrumentation
````

Which was changed as part of https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1704.   

Updating it as follows worked for me locally: 

````
CHARTS ?= opentelemetry-collector opentelemetry-operator opentelemetry-demo opentelemetry-ebpf opentelemetry-kube-stack opentelemetry-target-allocator opentelemetry-ebpf-instrumentation
````

### PR DESCRIPTION
**Pull request by NimrodAvni78**
- fixes #1925 
- fixes image repository to match ghcr
- change k8s-cache to stable version v0.1.0

### DIFF
--- a/charts/opentelemetry-ebpf-instrumentation/Chart.yaml
+++ b/charts/opentelemetry-ebpf-instrumentation/Chart.yaml
@@ -14,4 +14,4 @@ maintainers:
   - name: jaronoff97
   - name: TylerHelmuth
   - name: nimrodavni78
-appVersion: 0.1.1
+appVersion: 0.1.0
--- a/charts/opentelemetry-ebpf-instrumentation/examples/default/rendered/clusterrole.yaml
+++ b/charts/opentelemetry-ebpf-instrumentation/examples/default/rendered/clusterrole.yaml
@@ -8,7 +8,7 @@ metadata:
     helm.sh/chart: opentelemetry-ebpf-instrumentation-0.1.1
     app.kubernetes.io/name: opentelemetry-ebpf-instrumentation
     app.kubernetes.io/instance: example
-    app.kubernetes.io/version: "0.1.1"
+    app.kubernetes.io/version: "0.1.0"
     app.kubernetes.io/managed-by: Helm
     app.kubernetes.io/part-of: obi
     app.kubernetes.io/component: rbac
--- a/charts/opentelemetry-ebpf-instrumentation/examples/default/rendered/clusterrolebinding.yaml
+++ b/charts/opentelemetry-ebpf-instrumentation/examples/default/rendered/clusterrolebinding.yaml
@@ -8,7 +8,7 @@ metadata:
     helm.sh/chart: opentelemetry-ebpf-instrumentation-0.1.1
     app.kubernetes.io/name: opentelemetry-ebpf-instrumentation
     app.kubernetes.io/instance: example
-    app.kubernetes.io/version: "0.1.1"
+    app.kubernetes.io/version: "0.1.0"
     app.kubernetes.io/managed-by: Helm
     app.kubernetes.io/part-of: obi
     app.kubernetes.io/component: rbac
--- a/charts/opentelemetry-ebpf-instrumentation/examples/default/rendered/configmap.yaml
+++ b/charts/opentelemetry-ebpf-instrumentation/examples/default/rendered/configmap.yaml
@@ -9,7 +9,7 @@ metadata:
     helm.sh/chart: opentelemetry-ebpf-instrumentation-0.1.1
     app.kubernetes.io/name: opentelemetry-ebpf-instrumentation
     app.kubernetes.io/instance: example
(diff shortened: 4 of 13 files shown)

### CODE REVIEW

File: charts/opentelemetry-ebpf-instrumentation/Chart.yaml [hunk 1]
```diff
@@ -14,4 +14,4 @@ maintainers:
   - name: jaronoff97
   - name: TylerHelmuth
   - name: nimrodavni78
-appVersion: 0.1.0
+appVersion: 0.1.1
```
skl: Should appVersion remain as 0.1.0? As that is the version of the image tag being used.

NimrodAvni78: usually i keep the helm chart version and appVersion the same, but i can change it
from helm docs i think you are right

> appVersion: The version of the app that this contains (optional). Needn't be SemVer. Quotes recommended.

skl: Sometimes the `appVersion` is used as the target image tag, that doesn't seem to be the case here so I don't think this value actually has any impact. But I thought it was worth pointing out in case we decide to use it like this in future?

NimrodAvni78: in the meanwhile i changed it to v0.1.0 but i can change back, whatever you think
i like it that it shows 2 different values and not just a duplication, that one repersents the obi image version and one shows to chart version

skl: The other advantage of having appVersion same as image version, is that we could just use appVersion in the templates, so that when we come to release new versions in future - we just change the tag in one place.

So yeah, I also agree with having appVersion match the target image tag.

charts/opentelemetry-ebpf-instrumentation/values.yaml
```diff
@@ -293,9 +293,9 @@ k8sCache:
     # -- K8s Cache image registry (defaults to ghcr.io)
     registry: "ghcr.io"
     # -- K8s Cache image repository.
-    repository: otel/opentelemetry-ebpf-k8s-cache
+    repository: open-telemetry/opentelemetry-ebpf-instrumentation/opentelemetry-ebpf-k8s-cache
     # -- (string) K8s Cache image tag. When empty, the Chart's appVersion is used.
-    tag: main
+    tag: v0.1.0
```
jaronoff97, line 298: do we not want to be using the Chary.yaml app version here per the comment?

NimrodAvni78: https://github.com/open-telemetry/opentelemetry-helm-charts/blob/9bc3f27c5f65dbf57a856741b94614fae03e7ed2/charts/opentelemetry-collector/templates/_pod.tpl#L40
i see its done like that in the collector chart, i can also do something like this, maybe in a different PR tho? since this fixes the chart to pull the correct images

JaredTan95: Are you sure the registry address `open-telemetry/opentelemetry-ebpf-instrumentation/` here is correct? I think the original `otel/` is correct.

following https://hub.docker.com/r/otel/opentelemetry-ebpf-k8s-cache

NimrodAvni78: @JaredTan95  yeah, its a different image repo and name when downloading from ghcr.io instead of docker.io

JaredTan95: Oh, I didn't notice the `registry: "ghcr.io"`  above

JaredTan95: > i see its done like that in the collector chart, i can also do something like this, maybe in a different PR tho? since this fixes the chart to pull the correct images

I created another issue to track it. see https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1931

JaredTan95 (approved):

PR #1928, opened 2025-11-03
[kube-stack] Add popular metrics to the `presets/hostMetrics`

cyrille-leclerc:
Add popular Linux monitoring metrics to the `presets/hostMetrics`:

* `system.cpu.logical.count` so that
   * the CPU usage metrics like `system.cpu.time` & `system.cpu.utilization` can be compared to the max available CPU.
   * the Linux monitoring dashboard can easily report on the CPU capacity of the server
* `system.memory.limit` so that
   * the Memory usage metrics like `system.memory.usage` can 
   * the Linux monitoring dashboard can easily report on the RAM capacity of the server
* `system.paging.usage` because we always want to verify the paging activity when assessing memory usage
* `system.uptime` because we are alway interested in understanding when a server started

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.139.0
+version: 0.139.1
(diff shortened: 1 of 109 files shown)


jaronoff97 (approved):
all fine, please rebase when you get a chance

PR #1933, opened 2025-11-05
bump obi to v0.2.0 & take image tag from appVersion by default

### LINKED ISSUE #1931: [ebpf-instrumentation] make appVersion as default image tag
https://github.com/open-telemetry/opentelemetry-helm-charts/blob/9bc3f27c5f65dbf57a856741b94614fae03e7ed2/charts/opentelemetry-collector/templates/_pod.tpl#L40
i see its done like that in the collector chart, i can also do something like this, maybe in a different PR tho? since this fixes the chart to pull the correct images

_Originally posted by @NimrodAvni78 in https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1926#discussion_r2487028253_

NimrodAvni78:
resolves #1931

### DIFF
--- a/charts/opentelemetry-ebpf-instrumentation/Chart.yaml
+++ b/charts/opentelemetry-ebpf-instrumentation/Chart.yaml
@@ -1,7 +1,7 @@
 ---
 apiVersion: v2
 name: opentelemetry-ebpf-instrumentation
-version: 0.1.1
+version: 0.2.0
 description: OpenTelemetry eBPF instrumentation Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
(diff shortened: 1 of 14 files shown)


charts/opentelemetry-ebpf-instrumentation/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 apiVersion: v2
 name: opentelemetry-ebpf-instrumentation
-version: 0.1.1
```
JaredTan95, line 4: maybe 0.1.2 is fine?

NimrodAvni78: What's the convention when doing a minor update to the application? Is it also a minor to the helm chart or a patch?

PR #1937, opened 2025-11-10
[otel-kube-stack] Use `defaultCRConfig` enabled

### LINKED ISSUE #1936: [kube-stack] defaultCRConfig.enabled field is not respected when merging collector configuration
## Description

The `defaultCRConfig.enabled` field in `values.yaml` is not being checked before merging the default configuration into individual collectors. Currently, even when `defaultCRConfig.enabled: false`, the default configuration is still applied to all collectors.

## Current Behavior

In `charts/opentelemetry-kube-stack/templates/collector.yaml:2-3`:
```yaml
{{ range $_, $collector := $.Values.collectors -}}
{{- if $collector.enabled -}}
{{- $collector := (mergeOverwrite (deepCopy $.Values.defaultCRConfig) $collector) }}
```
The template only checks if the individual `$collector.enabled` is true, but does not verify whether `defaultCRConfig.enabled` is true before merging the default configuration.

## Expected Behavior

The defaultCRConfig should only be merged into collectors when `defaultCRConfig.enabled: true`. When `defaultCRConfig.enabled: false`, collectors should use only their own configuration without inheriting from defaultCRConfig.

## Suggested Fix

Add a check for defaultCRConfig.enabled before performing the merge operation in collector.yaml:

```yaml
{{ range $_, $collector := $.Values.collectors -}}
{{- if $collector.enabled -}}
{{- if $.Values.defaultCRConfig.enabled }}
{{- $collector := (mergeOverwrite (deepCopy $.Values.defaultCRConfig) $collector) }}
{{- end }}
```

gimsesu:
fixes #1936

### DIFF
--- a/charts/opentelemetry-kube-stack/Chart.yaml
+++ b/charts/opentelemetry-kube-stack/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-kube-stack
-version: 0.12.4
+version: 0.12.5
 description: |
   OpenTelemetry Quickstart chart for Kubernetes.
   Installs an operator and collector for an easy way to get started with Kubernetes observability.


charts/opentelemetry-kube-stack/values.yaml
```diff
@@ -56,7 +56,7 @@ opentelemetry-operator:
 # This is the default configuration for all collectors generated by the chart.
 # Any collectors in the `collectors` are overlayed on top of this configuration.
 defaultCRConfig:
-  enabled: false
+  enabled: true
```
TylerHelmuth, line 59: @gimsesu based on the issue you filed the claim is that bc of this bug the chart has been behaving as if this value was true?

gimsesu: @TylerHelmuth Yes, exactly. Before this fix, the chart was behaving as if `defaultCRConfig.enabled` was always `true`, regardless of the actual value set by users.

To maintain backward compatibility, the default value has been changed to true.

gimsesu:
@puckpuck @julianocosta89 
Hello, I'd appreciate your review.

@jaronoff97 
Do I have to keep waiting until the assigned reviewer reviews it?

jaronoff97 (approved):

gimsesu:
@jaronoff97 
I forgot to update the example. My apologies. I'm still getting used to it.

Issue #1944, opened 2025-11-14
When disableLeaderElection=true, k8s_leader_elector extensions are still loaded


hvindin:
# Description

When I configured my values file with the following:

```yaml
collector:
  presets:
    kubernetesEvents:
      enabled: true
      disableLeaderElection: true
```

The `OpenTelemetryCollector` object generated for the daemon becomes invalid.

The k8s_leader_election extensions are not declared under `config.extensions` (correct behaviour) however they are loaded under `service.extensions` like so:

```yaml
    service:
      extensions:
      - k8s_leader_elector/k8s_objects
      - k8s_leader_elector/k8s_cluster
```

This results in the daemonset crashing with the following error:

```
invalid configuration: service::extensions: references extension "k8s_leader_elector/k8s_objects" which is not configured
```

# Current Behavior

In `charts/opentelemetry-kube-stack/templates/_config.tpl` the following line:

```
{{- $_ := set $config.service "extensions" (append ($config.service.extensions | default list) (printf "k8s_leader_elector/%s" $electorName) | uniq)  }}
```

Always renders the addition of the extensions under the service.extensions.

# Expected Behaviour

When `disableLeaderElection` is set to `true` the extensions should not be rendered.

# Suggested fix

Add a check for `disableLeaderElection` before rendering the extensions, this same pattern exists plenty of times in the `_config.tpl` file already.

I am pretty sure this is a very straightforward change so I will attempt to raise a pull request in a moment.

PR #1945, opened 2025-11-14
Add a check to prevent loading `k8s_leader_elector` extensions when `disableLeaderElection` is true.

hvindin:
This fixes issue #1944 - https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1944


JaredTan95 (changes requested):
The flag bits need to be added accordingly in `values.yaml`.

BTW, pls following the contributing docs and bump up the version in `Chart.yaml` ,then run `CHARTS=opentelemetry-operator make generate-examples`

charts/opentelemetry-kube-stack/templates/_config.tpl
```diff
@@ -227,8 +227,11 @@ receivers:
 {{- $config := mustMergeOverwrite (include "opentelemetry-kube-stack.collector.clusterMetricsConfig" (dict "collector" .collector "namespace" .namespace "electorName" $electorName) | fromYaml) .collector.config }}
 {{- if and (dig "service" "pipelines" "metrics" false $config) (not (has "k8s_cluster" (dig "service" "pipelines" "metrics" "receivers" list $config))) }}
 {{- $_ := set $config.service.pipelines.metrics "receivers" (append ($config.service.pipelines.metrics.receivers | default list) "k8s_cluster" | uniq)  }}
+{{- $disableLeaderElection := .collector.presets.kubernetesEvents.disableLeaderElection }}
```
TylerHelmuth: Can you update this template so that it is using `.collector.presets.clusterMetrics.disableLeaderElection` instead of the kuberenetes events setting?

TylerHelmuth: I think there are other places in this chart that the kubernetes events leader election is used for the cluster metrics, can you fix those too?

JaredTan95 (approved):

PR #1946, opened 2025-11-14
[opentelemetry-collector] Add namespace exclusion for logs collection

YanivZl:
## Description

This PR adds the ability to exclude logs from specific Kubernetes namespaces when using the `logsCollection` preset. This feature provides a more convenient way to filter out logs from unwanted namespaces.

## Motivation

Currently, users **cannot** easily exclude logs from certain namespaces (like `kube-system`, `kube-public`, etc.). This makes it difficult to filter out logs that are not needed for observability.

Many observability tools charge based on the volume of logs ingested. Without an easy way to exclude logs from system namespaces or other non-critical sources, users may face unexpected and unwanted charges for collecting logs they don't actually need. This PR introduces a simple `excludeNamespaces` list that makes it easy to control which logs are collected, helping users **Reduce costs** by excluding logs from namespaces that don't provide business value

## Usage Example

```yaml
mode: daemonset

presets:
  logsCollection:
    enabled: true
    excludeNamespaces:
      - kube-system
      - kube-public
      - kube-node-lease
```

This generates the following exclude patterns in the filelog receiver:
```yaml
exclude:
  - /var/log/pods/kube-system_*/*/*.log
  - /var/log/pods/kube-public_*/*/*.log
  - /var/log/pods/kube-node-lease_*/*/*.log
  - /var/log/pods/default_example-opentelemetry-collector*_*/opentelemetry-collector/*.log
```

### DIFF
--- a/charts/opentelemetry-collector/README.md
+++ b/charts/opentelemetry-collector/README.md
@@ -129,6 +129,20 @@ config:
         - /var/log/pods/example-namespace_*/*/*.log
 ```
 
+Alternatively, you can exclude logs from specific namespaces using the `excludeNamespaces` property:
+
+```yaml
+mode: daemonset
+
+presets:
+  logsCollection:
+    enabled: true
+    excludeNamespaces:
+      - kube-system
+      - kube-public
+      - kube-node-lease
+```
+
 The container logs pipeline uses the `debug` exporter by default.
 Paired with the default `filelog` receiver that receives all containers' console output,
 it is easy to accidentally feed the exported logs back into the receiver.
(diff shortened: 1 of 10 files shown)


TylerHelmuth (changes requested):
@YanivZl we have a rule not to make presets customizable using non-boolean flags. This allows us to keep them very simple and very opinionated.

For your situation, I suggest enabling the logsCollection preset and then including the exclude configuration in your config. The chart will handle merging the preset config and your custom exlcude:

```yaml
mode: daemonset

image:
  repository: "ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector-k8s"

command:
  name: "otelcol-k8s"

presets:
  logsCollection:
    enabled: true

config:
  receivers:
    filelog:
      exclude:
        - kube-system
        - kube-public
        - kube-node-lease
        - /var/log/pods/default_example-opentelemetry-collector*_*/opentelemetry-collector/*.log
```

PR #1947, opened 2025-11-14
fix(opentelemetry-kube-stack): add jobLabel on node-exporter service

afarbos:
See the doc https://prometheus-operator.dev/docs/api-reference/api/#monitoring.coreos.com/v1.ServiceMonitorSpec
```
If the value of this field is empty or if the label doesn’t exist for the given Service, the job label of the metrics defaults to the name of the associated Kubernetes Service.
```
This mean the jobLabel is today set to `job="opentelemetry-kube-stack-prometheus-node-exporter"` for example.


JaredTan95:
pls following the contributing docs and bump up the version in `Chart.yaml` ,then run `CHARTS=opentelemetry-kube-stack make generate-examples`

TylerHelmuth (approved):

PR #1950, opened 2025-11-18
[collector] fix example

TylerHelmuth:


charts/opentelemetry-collector/examples/kubernetesAttributes/values.yaml
```diff
@@ -9,13 +9,4 @@ command:
...
     enabled: true
-
-config:
-  service:
-    pipelines:      
-      traces:
-        processors:
-          - resource
-          - k8sattributes
-          - batch
```
TylerHelmuth, line 20: This config is unnecessary since we added logic to inject k8sattributes at the beginning

TylerHelmuth:
Replaced by https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1951

PR #1951, opened 2025-11-18
[maint]: remove check-examples

TylerHelmuth:
Recently we move to slow on reviews to keep up with submitters handling the constant chart bump/example regeneration needed to merge PRs. This PR removes the `check-examples` target and deletes the rendered examples from kube-stack as an example for what this will mean going forward.

I propose we switch to having examples values.yamls, but not checking in their rendered contents. During PRs if we want to see how a PR affects the chart we can checkout the PR and run `make generate-examples` locally. By not checking examples in CI we'll be able to handle merge conflicts ourselves by bumping chart versions.

If we agree with this switch I'll submit other PRs for the rest of the charts to remove their examples.

### DIFF
--- a/.github/workflows/ossf-scorecard.yml
+++ b/.github/workflows/ossf-scorecard.yml
@@ -42,6 +42,6 @@ jobs:
       # Upload the results to GitHub's code scanning dashboard (optional).
       # Commenting out will disable upload of results to your repo's Code Scanning dashboard
       - name: "Upload to code-scanning"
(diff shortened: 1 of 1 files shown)


JaredTan95 (approved):
make sense

jaronoff97 (approved):
i'd definitely prefer it if we could just automate the example generation, i derive value from being able to see the changes, but i understand that this is a burden and i don't wanna hold it up.

PR #1952, opened 2025-11-18
feat(opentelemetry-demo): add KEDA-based horizontal scaling support

yonch:
this adds KEDA scale-up support to the opentelemetry-demo, so it can stress a large system.

The Unvariance collector uses this to run its benchmark on a 96-core m7i.metal-24xl and reach high transactions per second count (multiple hundreds) and high CPU utilization, e.g., https://unvariance.github.io/collector/benchmarks/workload_performance.png


julianocosta89:
@puckpuck, @rogercoll is that something we would like to add and maintain in the Demo?

PR #1972, opened 2025-12-03
Allow setting custom collector image repo without custom tag

Aransh:
Current chart logic requires setting collector.image.tag alongside collector.image.repository in order to use a custom image on a collector.
For a use-case of using a custom image repo but the same tag, which follows the same versions (for example using "otel/opentelemetry-collector-contrib", or maybe a mirror), a user may want to set the image repo but still allow the chart maintainers to set the image tag that the chart is intended for.

This is implemented by keeping the default behavior, but using Chart.AppVersion as the image tag in case collector.image.repository is provided  but collector.image.tag isn't

### DIFF
--- a/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/collector.yaml
+++ b/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/collector.yaml
@@ -456,7 +456,7 @@ spec:
           - batch
           receivers:
           - otlp
-  image: "otel/opentelemetry-collector-k8s:0.138.0"
+  image: "otel/opentelemetry-collector-k8s:0.141.0"
   imagePullPolicy: IfNotPresent
   upgradeStrategy: automatic
   terminationGracePeriodSeconds: 30
--- a/charts/opentelemetry-kube-stack/examples/daemonset-annotation-discovery-logs/rendered/collector.yaml
+++ b/charts/opentelemetry-kube-stack/examples/daemonset-annotation-discovery-logs/rendered/collector.yaml
@@ -82,7 +82,7 @@ spec:
           - batch
           receivers:
           - otlp
-  image: "otel/opentelemetry-collector-k8s:0.138.0"
+  image: "otel/opentelemetry-collector-k8s:0.141.0"
   imagePullPolicy: IfNotPresent
   upgradeStrategy: automatic
   terminationGracePeriodSeconds: 30
--- a/charts/opentelemetry-kube-stack/examples/daemonset-annotation-discovery-metrics/rendered/collector.yaml
+++ b/charts/opentelemetry-kube-stack/examples/daemonset-annotation-discovery-metrics/rendered/collector.yaml
@@ -80,7 +80,7 @@ spec:
           - batch
           receivers:
           - otlp
-  image: "otel/opentelemetry-collector-k8s:0.138.0"
+  image: "otel/opentelemetry-collector-k8s:0.141.0"
   imagePullPolicy: IfNotPresent
   upgradeStrategy: automatic
   terminationGracePeriodSeconds: 30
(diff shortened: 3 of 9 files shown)


charts/opentelemetry-kube-stack/templates/collector.yaml
```diff
@@ -32,6 +32,8 @@ spec:
   image: "{{ $collector.image.repository }}@{{ $collector.image.digest }}"
   {{- else if $collector.image.tag }}
   image: "{{ $collector.image.repository }}:{{ $collector.image.tag }}"
+  {{- else if $collector.image.repository }}
+  image: "{{ $collector.image.repository }}:{{ $.Chart.AppVersion }}"
```
jaronoff97: I think we could improve this by doing what the collector chart does `{{ .Values.image.tag | default .Chart.AppVersion }}`

Aransh: @jaronoff97 like this better yes, thanks, done
Also, probably a matter for separate PR/issue, but seems `defaultCRConfig.enabled` doesn't actually do anything 🤔

TylerHelmuth:
@Aransh please split your PR into a PR per chart.

Aransh:
> @Aransh please split your PR into a PR per chart.

@TylerHelmuth The ebpf chart is only included here since the "generate examples" command changed files in it (which later failed the CI as it required bumping version for the changes too).
If I'll split it to another PR, the kube-stack PR will fail unless I bump the version in it again...

Optimally I would make no changes or bump a version for the ebpf chart, that's just what the PR requirements forced me to do 😅

charts/opentelemetry-kube-stack/templates/collector.yaml
```diff
@@ -32,8 +32,8 @@ spec:
   {{- end }}
   {{- if $collector.image.digest }}
   image: "{{ $collector.image.repository }}@{{ $collector.image.digest }}"
-  {{- else if $collector.image.tag }}
-  image: "{{ $collector.image.repository }}:{{ $collector.image.tag }}"
+  {{- else if or $collector.image.tag $collector.image.repository }}
+  image: "{{ $collector.image.repository }}:{{ $collector.image.tag | default $.Chart.AppVersion }}"
```
TylerHelmuth, line 36: @jaronoff97 correct me if I'm wrong, but I thought this chart's `appVersion` tracked the operator version not the collector version. I know they are supposed to be the same, but often enough the collector has a patch release that isn't reflected in the operator's version.

Aransh:
Not stale 🥲

charts/opentelemetry-kube-stack/templates/collector.yaml
TylerHelmuth: @Aransh I can definitively say the `appVersion` is not the correct value to use for the collector image. `appVersion` tracks on the operator version, which may or may not match the desired Collector Version for that particular Operator version.

Aransh: @TylerHelmuth so do you think this PR is obsolete then? Or is there something else to be done?
Or is this only something that can be handled on the operator's code side?

Feels like kind of a miss being forced to set a tag manually when there are *official* alternative images...

TylerHelmuth: Each operator version already has a default collector image version baked in. The reason our helm chart is stuck setting this value in the command flags is because the operator's default image is Core distro and we wanted to use the k8s distro. If the operator is defaulting to the k8s distro now, we could simply remove the values for `collectorImage` the values.yaml file.

TylerHelmuth: I would agree tho that this PR's request to use the appVersion for the collector image unfortunately isn't safe enough.

Aransh: @TylerHelmuth I see... thanks for your response!
So what should I do then? open an issue requesting this change? If so, on which repo would it be most relevant?

TylerHelmuth: Ya the [opentelemetry-operator repo](https://github.com/open-telemetry/opentelemetry-operator). For now we can close this pr

PR #1981, opened 2025-12-10
[otel-kube-stack] Add `presets/resourcedetection` to  activate `aks`, `eks`, and `gcp` detectors ( `cloud.*`, `k8s.cluster.name`, etc attributes)

cyrille-leclerc:
Enriching telemetry with the Kubernetes provider context is very valuable but complicated to setup due to the error prone config of the resource detection processor detectors that require proper ordering with the `k8snode` and `env` detectors.

This PR simplifies this setup adding `resourcedetection` presets to detect cloud-provider / kubernetes-provider metadata and `k8s.cluster.name` hooking a [Resource Detection Processor detector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/resourcedetectionprocessor/README.md#supported-detectors) on `eks`, `aks`, and `gcp`.

These new presets prevent confusions making it clear that it's a predefined setup for frequently used resourcedetection processor detectors.

Other detectors or advanced config options for the detectors supported in this PR can be added later.

```yaml
    presets:
      ...
      resourceDetection:
        aks:
          enabled: false
        eks:
          enabled: false
        gcp:
          enabled: false
```

### No activation of any `presets/resourcedetection` `eks`, `aks`, or `gcp

```yaml
collectors:
  daemon:
    presets:
```

Will keep the default `resourcedetection/env` config fragment in the OTel Collector config:

```yaml
resourcedetection/env:
  detectors:
  - env
  - k8snode
  override: false
  timeout: 2s
```

### Amazon EKS

Enabling the `eks` resource detector:

```yaml
collectors:
  daemon:
    presets:
      resourceDetection:
        eks:
           enabled: true
```

Will enrich the `resourcedetection/env` config fragment in the OTel Collector config:

```yaml
resourcedetection/env:
  detectors:
  - env
  - k8snode
  - eks
  eks:
    resource_attributes:
      k8s.cluster.name:
        enabled: true
  override: false
  timeout: 15s
```

### Azure AKS

Enabling the `aks` resource detector:

```yaml
collectors:
  daemon:
    presets:
      resourceDetection:
        aks:
          enabled: true
```

Will enrich the `resourcedetection/env` config fragment in the OTel Collector config:

```yaml
resourcedetection/env:
  detectors:
  - env
  - k8snode
  - aks
  aks:
    resource_attributes:
      k8s.cluster.name:
        enabled: true
  override: false
  timeout: 2s
```

### GCP

Enabling the `gcp` resource detector:

```yaml
collectors:
  daemon:
    presets:
      resourceDetection:
        gcp:
          enabled: true
```

Will enrich the `resourcedetection/env` config fragment in the OTel Collector config:

```yaml
resourcedetection/env:
  detectors:
  - env
  - k8snode
  - gcp
  timeout: 2s
```

### DIFF
--- a/charts/opentelemetry-kube-stack/Chart.yaml
+++ b/charts/opentelemetry-kube-stack/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-kube-stack
-version: 0.14.9
+version: 0.14.10
 description: |
   OpenTelemetry Quickstart chart for Kubernetes.
   Installs an operator and collector for an easy way to get started with Kubernetes observability.
--- a/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/bridge.yaml
+++ b/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/bridge.yaml
@@ -5,7 +5,7 @@ kind: OpAMPBridge
 metadata:
   name: example
   labels:
-    helm.sh/chart: opentelemetry-kube-stack-0.14.9
+    helm.sh/chart: opentelemetry-kube-stack-0.14.10
     app.kubernetes.io/version: "0.144.0"
     app.kubernetes.io/managed-by: Helm
     release: "example"    
--- a/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/collector.yaml
+++ b/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/collector.yaml
@@ -6,7 +6,7 @@ metadata:
   name: example-daemon
   namespace: default
   labels:
-    helm.sh/chart: opentelemetry-kube-stack-0.14.9
+    helm.sh/chart: opentelemetry-kube-stack-0.14.10
     app.kubernetes.io/version: "0.144.0"
     app.kubernetes.io/managed-by: Helm
     release: "example"        
--- a/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/hooks.yaml
+++ b/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/hooks.yaml
@@ -61,4 +61,4 @@ spec:
           - "delete"
           - "instrumentations,opampbridges,opentelemetrycollectors"
           - "-l"
-          - "helm.sh/chart=opentelemetry-kube-stack-0.14.9"
+          - "helm.sh/chart=opentelemetry-kube-stack-0.14.10"
--- a/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/instrumentation.yaml
+++ b/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/instrumentation.yaml
@@ -5,7 +5,7 @@ kind: Instrumentation
 metadata:
   name: example
   labels:
-    helm.sh/chart: opentelemetry-kube-stack-0.14.9
+    helm.sh/chart: opentelemetry-kube-stack-0.14.10
     app.kubernetes.io/version: "0.144.0"
     app.kubernetes.io/managed-by: Helm
     release: "example"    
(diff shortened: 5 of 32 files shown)


charts/opentelemetry-kube-stack/values.yaml
```diff
@@ -473,7 +473,9 @@ defaultCRConfig:
         enabled: false
       metrics:
         enabled: false
-
+    resourceDetection:
+      enabled: false
+#        kubernetesDetector: ""
```
rogercoll: Could this be an array instead? One might be interested of setting more than one k8s/cloud detector at once. Also, what do you think of renaming it to  ~~kubernetes~~detector? (For example, the `gcp` detector will populate non-k8s attributes like `cloud.account.id`)

cyrille-leclerc: Please review change `kubernetesDetector` -> `detectors`

jaronoff97 (commented):
I think @rogercoll is correct though and we should have some guardrails here, otherwise it feels like the preset becomes less useful, in which case i would say this isn't necessary as a change.

charts/opentelemetry-kube-stack/values.yaml
```diff
@@ -473,7 +473,10 @@ defaultCRConfig:
         enabled: false
       metrics:
         enabled: false
-
+    resourceDetection:
+      enabled: false
+      # Detectors to add to the preconfigured 'env' and 'k8snode' ones. Typically used to resolve the 'k8s.cluster.name' resource attribute like 'eks', 'aks', or 'gcp'.
+      # detectors: []
```
TylerHelmuth: We have a strict stance in our charts' presets to not allow complex types. Optimally we'd have enabling this preset configure a very specific resourcedetection processor with very specific detectors. Another option is to have enable flags for specific detectors. 

We want to avoid having user-supplied strings in a preset as it requires the user to understand the component to configure the preset, which is in contrast with the goal of the preset.

charts/opentelemetry-kube-stack/templates/collector.yaml
```diff
@@ -164,7 +164,7 @@ spec:
   {{- toYaml . | nindent 4 }}
   {{- end }}
   env:
-  - name: K8S_NODE_NAME
+  - name: K8S_NODE_NAME{{/* used by detectors such as the EKS resource detector */}}
```
rogercoll:
```suggestion
  - name: K8S_NODE_NAME {{/* used by detectors such as the EKS resource detector */}}
```

charts/opentelemetry-kube-stack/values.yaml
```diff
@@ -493,7 +493,15 @@ defaultCRConfig:
         enabled: false
       metrics:
         enabled: false
-
+    resourceDetection:
```
rogercoll, line 496: I’m not sure about adding these changes, at least not with the current naming. Wouldn’t this create confusion for users and mix terminology with the collector’s resource detection processor? That processor has many more capabilities beyond just the EKS, AKS, and GCP detectors.

Although I’m not fully convinced, I’ll leave the final decision to the approvers if they want to include this opinionated preset. The code itself LGTM.

cyrille-leclerc: Thanks Roger.
I have updated the beginning of the description of this PR to clarify the points you raised.

> Wouldn’t this create confusion for users and mix terminology with the collector’s resource detection processor?

I'm not sure it would create confusion as this `preset`:
* Reuses exactly the names of the detectors of the resourcedetection processor.
* Doesn't introduce any new configuration vocabulary / concept

> That processor has many more capabilities beyond just the EKS, AKS, and GCP detectors.

Agreed, I feel it's common for Helm Chart `presets` to be an opinionated subset of the OTel Collector components for popular use cases.

Moreover, the config of this `resourcdetection` preset should be extensible enough to handle additional use cases, either adding support for new deector or offering more config options on the supported detectors.

The first extension I can think of is adding support for more Kubernetes providers like `openshift`...

WDYT?

rogercoll: > Moreover, the config of this resourcdetection preset should be extensible enough to handle additional use cases, either adding support for new deector or offering more config options on the supported detectors.

Sounds good, thanks for the clarification.

PR #1983, opened 2025-12-11
OBI: fix zero-valued ports

mariomac:
People trying to disable prometheus port by setting it to 0 caused an error:

```
Error: INSTALLATION FAILED: DaemonSet.apps "obi" is invalid: spec.template.spec.containers[0].ports[0].containerPort: Required value
```

### DIFF
--- a/charts/opentelemetry-ebpf-instrumentation/templates/daemonset.yaml
+++ b/charts/opentelemetry-ebpf-instrumentation/templates/daemonset.yaml
@@ -73,22 +73,16 @@ spec:
                 - ALL
           {{- end }}
           ports:
-          {{- if (and
-                    (or (.Values.service.targetPort) (.Values.config.data.prometheus_export))
-                    (ne .Values.service.targetPort 0)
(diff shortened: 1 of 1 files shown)


mariomac:
@TylerHelmuth I haven't found any official reference, neither in the issues, to that breaking chage. But in the last few months I've found some helm charts breaking similarly.

Nevermind, after digging more into it, I think the problem was in our helm chart. I have discarded the previous changes and provided a new, simplified fix.

When prometheus_export.port was set to zero, this template:

```gotemplate
          {{- if or (.Values.service.targetPort) (.Values.config.data.prometheus_export) }}
          - name: {{ .Values.service.portName }}
            containerPort: {{ .Values.service.targetPort | default .Values.config.data.prometheus_export.port }}
            protocol: TCP
          {{- end }}
```
Has been always rendering to:
```yaml
          - name: "whatever"
            containerPort: 0
            protocol: TCP
```

This has been never correct, but for some reason Kubernetes was not complaining before. I have amended the code to avoid rendering this zero-valued port.

TylerHelmuth (approved):

PR #2004, opened 2025-12-23
[opentelemetry-operator] Remove kube-rbac-proxy and use built-in metrics auth

### LINKED ISSUE #1999: Migrate away from kube-rbac-proxy
The OTEL operator 0.142.0 removed depenrency on the kube-rback-proxy. The authentication and authorization is now handled by the native controller runtime implementation.

* kube-rbac-proxy references https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-helm-charts%20kube-rbac-proxy&type=code 
* operator PR https://github.com/open-telemetry/opentelemetry-operator/pull/4576 

Users should not directly depend on the kube-rbac-proxy. The proxy just provides auth for the /metrics operator endpoint.

YASHMAHAKAL:
## Description
Removes kube-rbac-proxy dependency and migrates to controller-runtime's built-in metrics authentication, aligning with upstream operator v0.142.0+ changes.

## Changes
- Removed kube-rbac-proxy sidecar container
- Replaced `kubeRBACProxy` config with `metricsServer` config
- Added support for `--metrics-secure` flag and TLS options
- Updated RBAC to remove proxy-specific permissions
- Simplified service to single metrics port
- Chart version bumped to 0.103.0

## Migration
Default behavior unchanged (metrics on port 8080). For secure metrics, set `metricsServer.secure: true`, as described in UPGRADING.md.

## Testing
Helm lint passes
Templates render correctly
Examples regenerated
Tested on local cluster (Minikube)

Fixes #1999


TylerHelmuth:
Switching to draft until CoPilot is done reviewing.

YASHMAHAKAL:
@TylerHelmuth Copilot review is complete. I've already addressed the port default issue (commit d9931e6). The remaining comments are style suggestions. Should I address them all, or would you like to review first? because i'm not much sure as first two are regarding UPGRADING.md...

TylerHelmuth:
@YASHMAHAKAL close any copilot reviews you're ignoring and then mark the PR ready for review when you want us to review it. I dont want to read through copilot comments.

jaronoff97 (approved):
Should be good @pavolloffay mind taking a look?

pavolloffay (approved):

rogercoll (approved):

rogercoll (commented):
Should we bump `opentelemetry-collector-k8s` image to `0.142.0` as well? https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2004/files#diff-56c52957e9b80d7dcb9b9771e329d7dc4f1a3ee8fb22acdcb3a75c1c92e75c18R53

jaronoff97 (approved):

jaronoff97:
@rogercoll yes i think we should @YASHMAHAKAL apologies one more change

YASHMAHAKAL:
> @rogercoll yes i think we should @YASHMAHAKAL apologies one more change

Thanks @jaronoff97, it's absolutely okay ! i'll be updating the PR with necessary changes : )

rogercoll (approved):

jaronoff97 (approved):

TylerHelmuth:
@YASHMAHAKAL looks like CI is still failing

YASHMAHAKAL:
@TylerHelmuth, Thanks for your patience.. i've made a fix, please review it : )

jaronoff97 (commented):
@YASHMAHAKAL PR looks good, but youll need to bump the Chart version. Given this is technically a breaking change, please bump the minor version instead of the patch version

swiatekm:
@YASHMAHAKAL you'll need to revert the changes from https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2057/changes/6c98bba5e87aa92642fb7125329e10dad061c594 to get the e2e tests to pass.

.github/workflows/operator-test.yaml
```diff
@@ -64,9 +64,7 @@ jobs:
             --set "manager.collectorImage.repository=ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector-k8s" \
             --set "manager.podLabels.control-plane=controller-manager" \
             --set "manager.extraEnvs[0].name=NAMESPACE" \
-            --set "manager.extraEnvs[0].valueFrom.fieldRef.fieldPath=metadata.namespace" \
-            --set "kubeRBACProxy.enabled=false" \
-            --set "manager.ports.metricsPort=8443"
+            --set "manager.extraEnvs[0].valueFrom.fieldRef.fieldPath=metadata.namespace"
```
YASHMAHAKAL, line 67: @swiatekm, i have reverted the changes as per your recommendation.....

charts/opentelemetry-operator/examples/default/rendered/tests/test-service-connection.yaml
```diff
@@ -59,7 +59,7 @@ metadata:
   name: "example-opentelemetry-operator-webhook"
   namespace: default
   labels:
-    helm.sh/chart: opentelemetry-operator-0.105.0
+    helm.sh/chart: opentelemetry-operator-0.106.0
```
YASHMAHAKAL, line 62: @jaronoff97 minor version bumped to 0.106.0 as per your recommendation, Thanks !

charts/opentelemetry-operator/values.yaml
```diff
@@ -233,38 +233,25 @@ manager:
...
-    capabilities:
-      drop:
-        - ALL
-    runAsNonRoot: true
-    seccompProfile:
-      type: RuntimeDefault
+## Metrics server configuration
+## The operator uses controller-runtime built-in auth for the metrics server
+## Authentication is disabled by default for backward compatibility
+metricsServer:
```
swiatekm, line 239: Is there a reason we're putting this in a new section? Intuitively, this should be under a path like `manager.metrics`, since it's really just configuring metrics for the operator binary. @jaronoff97 WDYT?

charts/opentelemetry-operator/templates/deployment.yaml
swiatekm, line 126: This port should be conditional on the settings as well.

charts/opentelemetry-operator/values.yaml
```diff
@@ -233,38 +233,25 @@ manager:
...
+  # Enable authentication and authorization for the metrics server
+  # Corresponds to --metrics-secure flag in the operator
+  secure: true
+
+  # Port for the metrics server
+  # Use 8443 when secure is true, 8080 when false
+  port: 8443
```
swiatekm, line 246: Can we not use `manager.ports.metrics` instead? That would remove a lot of the conditional logic we have around ports and we wouldn't have two sources of truth for them. The HTTPS settings could go into a separate section, like `manager.metrics`.

swiatekm (changes requested):
I'd prefer if we put the new config section under `manager` and only had one way of setting the metrics port. I've left some comments with suggestions.

Notably, if we're enabling HTTPS by default, we also need to deal with https://github.com/open-telemetry/opentelemetry-helm-charts/issues/2063 by configuring the ServiceMonitor appropriately.

charts/opentelemetry-operator/values.yaml
jaronoff97: this makes sense to me.

charts/opentelemetry-operator/templates/deployment.yaml
```diff
@@ -45,7 +45,16 @@ spec:
       {{- end }}
       containers:
         - args:
-            - --metrics-addr=0.0.0.0:{{ .Values.manager.ports.metricsPort }}
+            - --metrics-addr=0.0.0.0:{{ if .Values.metricsServer.secure }}{{ .Values.metricsServer.port }}{{ else }}{{ .Values.manager.ports.metricsPort }}{{ end }}
+            {{- if .Values.metricsServer.secure }}
+            - --metrics-secure=true
```
kravvcu, line 50: @YASHMAHAKAL 

Sorry for butting in but I've stumbled upon this PR as I'm interested in it.

I'm wondering whether this should be passed always as such:

`
- --metrics-secure={{ .Values.metricsServer.secure }}
`

The reason being this seems to be enabled by default in the operator (https://github.com/open-telemetry/opentelemetry-operator/blob/v0.144.0/internal/config/config.go#L179). Though I'm no go developer so I may be mistaken this is the default value and isn't overridden.

swiatekm: Is that a problem? We're always explicitly passing the value we want, without relying on defaults.

kravvcu: Correct me if I'm wrong but if currently the default within operator itself is 'enabled' and the current state of the Helm chart code in *this* PR *conditionally enables* then there's no way to disable secure metrics endpoint with Helm chart with the implementation this PR proposes.

swiatekm: Ah, I misread your comment. You're right, we should set it unconditionally.

eenchev:
Hi @swiatekm @YASHMAHAKAL , do you think anything can be done re this PR? Essentially, the chart has no proper way to configure secure metrics collection on the newer versions of operator. So there is discrepency between the chart configuration and the actual operator until this is merged.

Issue #2013, opened 2026-01-05
[opentelemetry-collector] Helm exception activating `prometheusRule/defaultRules`: `can't evaluate field Values in type interface {}`


cyrille-leclerc:
Helm template exception when activating `prometheusRule/defaultRules` in `values.yml`: `can't evaluate field Values in type interface {}`

To reproduce, define in `values.yaml`:

```yaml
mode: daemonset

image:
  repository: "ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector-k8s"

prometheusRule:
  enabled: true
  defaultRules:
    enabled: true
```

Exception message:

```
Error: opentelemetry-collector/templates/prometheusrule.yaml:125:24
  executing "opentelemetry-collector/templates/prometheusrule.yaml" at <.Values.prometheusRule.defaultRules.additionalRuleLabels>:
    can't evaluate field Values in type interface {}
```

PR #2014, opened 2026-01-05
Fix Helm `prometheusrule.yaml` template syntax error

cyrille-leclerc:
Fix
* https://github.com/open-telemetry/opentelemetry-helm-charts/issues/2013

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.144.0
+version: 0.144.1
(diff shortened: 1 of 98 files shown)


TylerHelmuth:
@cyrille-leclerc something doesn't seem right anymore. I can see the new example, but dont see any fixes in values.yaml or the template files. Was there a bad merge?

klubi:
@cyrille-leclerc Looks like changes to `templates/prometheusrule.yaml` were reverted during merge.
They can be seen in your [first commit](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2014/changes/820262c54fe04aca690cba1927eda0e26917010c#diff-d5de3c3af5face24dd483c21f28e6992088856dba52ca2f77552500670d17e17), but not anymore.

PR #2005, opened 2025-12-23
Upgrade Jaeger subchart to 4.1.5

YASHMAHAKAL:
Upgrades Jaeger from 3.4.1 to 4.1.5 to fix SPM compatibility.

Fixes #1994

Changes:
- Jaeger: 3.4.1 → 4.1.5
- Chart version: 0.39.0 → 0.40.0
- Examples regenerated

Testing:
- ✅ Helm lint passes
- ✅ Dependencies updated
- ✅ Chart.lock shows 4.1.5

PR #2016, opened 2026-01-07
[demo] Demo release version 2.2.0

puckpuck:
Updates to the Demo 2.2.0 release:
- new product-reviews service and its accompanying LLM service
- updates to postgres deployment and initialization (can use custom configmaps to init data)
- updates to all sub-chart dependencies, including Jaeger (this supersedes #2005 )

Marked as draft until Demo 2.0.0 is released

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Please refer to Helm's [documentation](https://helm.sh/docs/) to get started.
 Once Helm is set up properly, add the repo as follows:
 
 ```console
-$ helm repo add open-telemetry https://open-telemetry.github.io/opentelemetry-helm-charts
+helm repo add open-telemetry https://open-telemetry.github.io/opentelemetry-helm-charts
 ```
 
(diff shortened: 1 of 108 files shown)


charts/opentelemetry-demo/values.yaml
```diff
@@ -893,38 +948,88 @@ opentelemetry-collector:
...
+              normalize_duration: true
+
+    receivers:
+      otlp:
+        protocols:
+          grpc:
+            endpoint: ${env:JAEGER_HOST}:${env:JAEGER_GRPC_PORT}
+
+    processors:
+      batch:
```
julianocosta89, line 1032: Should we use `batch` processor here?

puckpuck: This matches the same config we have in the otel-demo for Jaeger

rogercoll (approved):

PR #2018, opened 2026-01-07
[kube-stack] Add `k8s.*` and `host.name` resource attributes to the OTel Collector internal telemetry

cyrille-leclerc:
Add `k8s.*` and `host.name` resource attributes to the OTel Collector `OTEL_RESOURCE_ATTRIBUTES` so internal telemetry is fully contextualized.

This enables the internal telemetry of the collector, `otelcol.*` metrics and logs, to be enriched with traditional K8s attributes for better dashboarding and alerting.

Example of generated OTelCollector internal telemetry metric:

```
otelcol_process_uptime_seconds_total{
    host_name="k3d-k3s-default-server-0",
    instance="bd81775d-d7d9-40e7-8bab-a6ca852bfc3d",
    job="otelcol-contrib",
    k8s_cluster_name="my_k8s_cluster",
    k8s_namespace_name="opentelemetry-operator-system",
    k8s_node_name="k3d-k3s-default-server-0",
    k8s_pod_name="opentelemetry-stack-daemon-collector-d2zrh",
    service_instance_id="bd81775d-d7d9-40e7-8bab-a6ca852bfc3d",
    service_name="otelcol-contrib",
    service_version="0.138.0"}
```

### DIFF
--- a/.github/workflows/fossa.yml
+++ b/.github/workflows/fossa.yml
@@ -14,7 +14,7 @@ jobs:
     steps:
       - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
 
-      - uses: fossas/fossa-action@3ebcea1862c6ffbd5cf1b4d0bd6b3fe7bd6f2cac # v1.7.0
+      - uses: fossas/fossa-action@c414b9ad82eaad041e47a7cf62a4f02411f427a0 # v1.8.0
         with:
           api-key: ${{secrets.FOSSA_API_KEY}}
           team: OpenTelemetry
--- a/.github/workflows/ossf-scorecard.yml
+++ b/.github/workflows/ossf-scorecard.yml
@@ -42,6 +42,6 @@ jobs:
       # Upload the results to GitHub's code scanning dashboard (optional).
       # Commenting out will disable upload of results to your repo's Code Scanning dashboard
       - name: "Upload to code-scanning"
-        uses: github/codeql-action/upload-sarif@cdefb33c0f6224e58673d9004f47f7cb3e328b89 # v4.31.10
+        uses: github/codeql-action/upload-sarif@45cbd0c69e560cd9e7cd7f8c32362050c9b7ded2 # v4.32.2
         with:
           sarif_file: results.sarif
--- a/charts/opentelemetry-kube-stack/Chart.lock
+++ b/charts/opentelemetry-kube-stack/Chart.lock
@@ -7,12 +7,12 @@ dependencies:
   version: 0.0.0
 - name: opentelemetry-operator
   repository: https://open-telemetry.github.io/opentelemetry-helm-charts
-  version: 0.102.0
+  version: 0.105.0
 - name: kube-state-metrics
   repository: https://prometheus-community.github.io/helm-charts
   version: 6.3.0
 - name: prometheus-node-exporter
   repository: https://prometheus-community.github.io/helm-charts
   version: 4.48.0
-digest: sha256:aaf420e2151c00a05ead99ff465e82ba593fa8bd555f9a5e4f0044b78a37f936
-generated: "2026-01-22T00:20:19.34647+01:00"
+digest: sha256:b1f808f14ab7379a5af3cf40d576796d3976e8c6adc2a29db4fb7237d87443fb
+generated: "2026-01-28T18:26:05.097217+01:00"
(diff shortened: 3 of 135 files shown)


charts/opentelemetry-kube-stack/templates/collector.yaml
```diff
@@ -178,10 +178,8 @@ spec:
       fieldRef:
         apiVersion: v1
         fieldPath: status.podIP
-  {{- if $.Values.clusterName }}
   - name: OTEL_RESOURCE_ATTRIBUTES
-    value: "k8s.cluster.name={{ $.Values.clusterName }}"
-  {{- end }}
+    value: "k8s.pod.name=$(OTEL_K8S_POD_NAME),k8s.namespace.name=$(OTEL_K8S_NAMESPACE),k8s.node.name=$(OTEL_K8S_NODE_NAME),host.name=$(OTEL_K8S_NODE_NAME),k8s.node.ip=$(OTEL_K8S_NODE_IP),k8s.pod.ip=$(OTEL_K8S_POD_IP){{- if $.Values.clusterName }},k8s.cluster.name={{ $.Values.clusterName }}{{- end }}"
```
jaronoff97, line 188: can we set a default for clusterName to be unknown if its unset?

cyrille-leclerc: I'm brainstorming a lot on the `k8s.cluster.name` at the moment, at reconciling the value manually set through `values.yaml` with the value resolved by the ResourceDetection processor and I would be inclined to decouple the topic of `k8s.cluster.name` from this PR.

A challenge I see is that it's not because the `k8s.cluster.name` is not set manually through `values.yaml` that it is unset. 
`k8s.cluster.name` may be resolved using the resource detection processor but we don't have access to the resource detection processor for internal telemetry as we can dot it in the ingestion pipeline.

Does it make sense?

cyrille-leclerc: @rogercoll you have domain expertise on using kubernetes provider detectors like `eks`, `aks`, or `gcp`, to resolve the `k8s.cluster.name` in the kube-stack-helm-chart. Do you think it could be confusing if the collector internal telemetry was using `k8s.cluster.name=unknown` instead of the name resolved by the Kubernetes provider detector?

rogercoll: Although by default the [resourcedetector](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor#configuration) overrides any resolution, some detectors encourage disabling the override. Also, in some cases it is not that the value is `unknown` but the internal telemetry not being processed with the corresponding resource detectors. In that sense, I would opt for not setting this value and discuss this topic in another issue/PR (maybe self process internal telemetry before exporting it?)

TylerHelmuth: Lets take cluster name out for now and handle it in a different PR. These rest of these attributes are a great, easy addition (please do this for the collector chart too).

cyrille-leclerc: > Lets take cluster name out for now and handle it in a different PR. 

👍 

> These rest of these attributes are a great, easy addition (please do this for the collector chart too).

In this PR or in a distinct one?

TylerHelmuth: different PR

jaronoff97: @cyrille-leclerc rather than doing an inline if could we default to cluster 'unknown'? or set the value in the values.yaml for unknown?

charts/opentelemetry-kube-stack/values.yaml
```diff
@@ -3,8 +3,10 @@ fullnameOverride: ""
 # Top level field indicating an override for the namespace
 namespaceOverride: ""
 
-# Top level field specifying the name of the cluster
-clusterName: ""
+# Top level field specifying the name of the Kubernetes cluster.
+# When using a Managed Kubernetes Service, it's recommended to use the cluster name assigned by the service for
+# better integration.
+clusterName: "unknown_k8s_cluster"
```
cyrille-leclerc, line 9: @jaronoff97 please review:
* added default cluster name
* Improved docs

charts/opentelemetry-kube-stack/templates/collector.yaml
cyrille-leclerc: @jaronoff97 I added the default value `unknown_k8s_cluster` in `values.yaml`. Do you want me to also enforce it's always set or shall we allow empty cluster name and thus keep the check here?

PR #2021, opened 2026-01-09
feat(opentelemetry-operator): auto detect presence of cert-manager CRDs

theSuess:
This PR adds a check for the presence of the `cert-manager.io/v1` API in the target cluster. Using this approach, the helm chart can be installed without any customization in clusters without cert manager, simplifying the installation.

### DIFF
--- a/charts/opentelemetry-operator/templates/certmanager.yaml
+++ b/charts/opentelemetry-operator/templates/certmanager.yaml
@@ -1,4 +1,4 @@
-{{- if and .Values.admissionWebhooks.create .Values.admissionWebhooks.certManager.enabled }}
+{{- if and .Values.admissionWebhooks.create .Values.admissionWebhooks.certManager.enabled (.Capabilities.APIVersions.Has "cert-manager.io/v1") }}
 apiVersion: cert-manager.io/v1
 kind: Certificate
 metadata:
--- a/charts/opentelemetry-operator/templates/_helpers.tpl
+++ b/charts/opentelemetry-operator/templates/_helpers.tpl
@@ -119,7 +119,7 @@ a cert is loaded from an existing secret or is provided via `.Values`
 {{- $caCertEnc := "" }}
 {{- $certCrtEnc := "" }}
 {{- $certKeyEnc := "" }}
-{{- if .Values.admissionWebhooks.autoGenerateCert.enabled }}
+{{- if or (eq .Values.admissionWebhooks.certificateSource "helm-generated") (eq .Values.admissionWebhooks.certificateSource "prefer-cert-manager") }}
 {{- $prevSecret := (lookup "v1" "Secret" (include "opentelemetry-operator.namespace" .) (default (printf "%s-controller-manager-service-cert" (include "opentelemetry-operator.fullname" .)) .Values.admissionWebhooks.secretName )) }}
 {{- if and (not .Values.admissionWebhooks.autoGenerateCert.recreate) $prevSecret }}
 {{- $certCrtEnc = index $prevSecret "data" "tls.crt" }}
@@ -140,7 +140,10 @@ a cert is loaded from an existing secret or is provided via `.Values`
 {{- $certKeyEnc = b64enc $cert.Key }}
 {{- $caCertEnc = b64enc $ca.Cert }}
 {{- end }}
-{{- else }}
+{{- else if eq .Values.admissionWebhooks.certificateSource "self-signed" }}
+{{- if or (or (eq .Values.admissionWebhooks.certFile "") (eq .Values.admissionWebhooks.keyFile "")) (eq .Values.admissionWebhooks.caFile "")}}
+{{ fail "certFile, keyFile and caFile must be set when using self-signed as certificateSource" }}
+{{- end }}
 {{- $certCrtEnc = .Files.Get .Values.admissionWebhooks.certFile | b64enc }}
 {{- $certKeyEnc = .Files.Get .Values.admissionWebhooks.keyFile | b64enc }}
 {{- $caCertEnc = .Files.Get .Values.admissionWebhooks.caFile | b64enc }}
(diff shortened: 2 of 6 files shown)


charts/opentelemetry-operator/templates/certmanager.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if and .Values.admissionWebhooks.create .Values.admissionWebhooks.certManager.enabled }}
+{{- if and .Values.admissionWebhooks.create .Values.admissionWebhooks.certManager.enabled (.Capabilities.APIVersions.Has "cert-manager.io/v1") }}
```
petewall, line 1: My concern here is that if the user says they want it:
```yaml
admissionWebhooks:
  create: true
  certManager:
    enabled: true
```

But the CRDs haven't been installed, this will silently proceed with the installation but wont deploy these objects.

Before, it was ugly, but at least it would fail.

Would it be worth adding the `else` for the capabilities check and then use a `fail` block to print a message like "Cert manager requested, but the CRDs are not installed. Please install the crd and try again."?

theSuess: Failing when the CRDs aren't available is a good approach as well, but I think this doesn't solve the original issue of `install` not working for everyone by default.

IMHO the default should be `certManager.enabled=false` since it isn't a hard requirement (or included in k8s) but I don't think we want to change the defaults here.

@cyrille-leclerc might have an opinion here as well

cyrille-leclerc: i was wondering if we would introduce a preset like 
```
certs-mode: use-existing-cert-manager | create-self-signed | auto-detect_if-crd-found-use-cert-manager_else-create-self-signed
```

would that resonate with you?

petewall: This is a good idea, and makes it much more explicit for the user to choose.

```
certificateSource: cert-manager | self-signed | prefer-cert-manager
```

re: `certificateSource` vs `certs-mode`, dashes in yaml keys are allowed, but a real pain in Helm templating context, because you have to use `index`.

I like the `prefer-...` logic because that matches other things in Kubernetes.

cyrille-leclerc: you found much better names than me, I love it

theSuess: I'm happy with this as well, will update this PR and ping here once implemented

cyrille-leclerc:
If I understand correctly, we would have the following developer experience:

**TODAY**

```yaml
# TODAY - SELF SIGNED SETUP, discouraged in production
admissionWebhooks:
  ...
  certManager:
    enabled: false # Don't use cert-manager (default: true)
  autoGenerateCert:
    enabled: true # activate cert auto-generation (default: true)
    recreate: true # prevent self-signed certs trust problems that are frequent in dev  (default: true)
```


```yaml
# TODAY - CERT MANAGER SETUP, recommended in production
admissionWebhooks:
  ...
  certManager:
    enabled: true #  (default: true)
    ... add more config if needed
  autoGenerateCert:
    enabled: false #  (default: true)
```

**PR PROPOSAL**

introduce

```yaml
certificateSource: 'prefer-cert-manager', 'cert-manager', 'helm-generated', or 'self-signed'
```

```yaml
# PR PROPOSAL - CERT MANAGER REQUIRED, recommended in production
admissionWebhooks:
  ...
  certificateSource: cert-manager # Require cert-manager
  certManager:
    ... add more config if needed
```

```yaml
# PR PROPOSAL - CERT MANAGER PREFERED, FALL BACK TO SELF SIGNED, ok in production, great to get started
admissionWebhooks:
  ...
  certificateSource: prefer-cert-manager # use cert-manager if discovered during install, fallback to self signed
  certManager:
    ... add more config if needed
```

```yaml
# PR PROPOSAL - SELF SIGNED, FALL BACK TO SELF SIGNED, discouraged in production
admissionWebhooks:
  ...
  certificateSource: self-signed # use self signed
```

theSuess:
If we can come up with a good priority ordering, I'm fine with deprecating the fields instead, but the templating logic will grow a lot more complex

cyrille-leclerc:
> If I understand you correctly, this would make `certificateSource` optional. This would work and shouldn't be too complicated but in that case the default behavior would be the deprecated one which doesn't feel quite right.
> 
> As an alternative we can set `certificateSource` to `cert-manager` by default until the fields are removed. This effectively gives us the same behavior with stricter validation and better error messages

You have a good point. I'm thinking of an incremental approach:
* First, introduce `certificateSource: null`, recommending it while not deprecating the existing `/certManager/enabled` and `/autoGenerateCert/enabled` and make `certificateSource` overwrite `/certManager/enabled` and `/autoGenerateCert/enabled` when defined
* Later, deprecate `/certManager/enabled` and `/autoGenerateCert/enabled`
* Eventually remove `/certManager/enabled` and `/autoGenerateCert/enabled`, mking he helm deployment fail if one of them is defined, and define the default value for  `certificateSource: cert-manager` as you rightfully suggested to be equivalent to the current setup or consider defaulting to `prefer-cert-manager` to be moe lenient.

marcschaeferger:
@theSuess any status/update? Failing CI?

theSuess:
To be honest, I kind of forgot about this PR :sweat_smile:

I've rebased this now & verified this still works. If I remember correctly the reason why this hasn't been merged yet is because it is unclear how to best roll this out without breaking existing setups. I implemented a `prefer-cert-manager` method which solves this for most use cases but for people relying on `helm template` this causes an issue as it doesn't respect dynamic API lookups


CI fails due to a missing version bump which I'd fix once the path forward is certain to keep the diff easily reviewable

marcschaeferger:
I think the main remaining decision here is the migration/backward-compatibility.

Should `certManager.enabled` and `autoGenerateCert.enabled` be removed immediately, or deprecated first?

A possible approach would be:

* `certificateSource` takes precedence when explicitly set.
* Otherwise, keep honoring the existing `certManager.enabled` / `autoGenerateCert.enabled` values during a deprecation period.

Once we agree on that behavior, the implementation can be finalized accordingly. 

@open-telemetry/helm-maintainers 
cc: @open-telemetry/operator-maintainers

PR #2042, opened 2026-01-21
feat: improve kube-stack NOTES.txt

theSuess:
This PR improves the `NOTES.txt` template of the `opentelemetry-kube-stack` helm chart by providing the user with next steps directly after installation.

### DIFF
--- a/charts/opentelemetry-kube-stack/Chart.yaml
+++ b/charts/opentelemetry-kube-stack/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-kube-stack
-version: 0.13.0
+version: 0.13.1
 description: |
   OpenTelemetry Quickstart chart for Kubernetes.
   Installs an operator and collector for an easy way to get started with Kubernetes observability.
--- a/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/bridge.yaml
+++ b/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/bridge.yaml
@@ -5,7 +5,7 @@ kind: OpAMPBridge
 metadata:
   name: example
   labels:
-    helm.sh/chart: opentelemetry-kube-stack-0.13.0
+    helm.sh/chart: opentelemetry-kube-stack-0.13.1
     app.kubernetes.io/version: "0.141.0"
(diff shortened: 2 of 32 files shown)


charts/opentelemetry-kube-stack/templates/NOTES.txt
```diff
@@ -6,3 +6,25 @@
 [WARNING] Collector '{{ $name }}': the 'updateStrategy' value is no longer supported in OpenTelemetryCollector v1beta1. Please migrate to use the `daemonSetUpdateStrategy` value.
 {{- end }}
 {{- end }}
+OpenTelemetry Kube Stack Helm Chart installation completed successfully!
+
+With the features you've enabled, you can:
```
theSuess, line 11: Not sure if we should make this conditional as well.
Theoretically, users can disable everything, but then there'd be no need to use this helm chart over the operator & collector directly IMHO

charts/opentelemetry-kube-stack/templates/NOTES.txt
```diff
@@ -6,3 +6,25 @@
...
+{{ range $_, $protocol := list "grpc" "http" -}}
+{{ $endpoint := dig "receivers" "otlp" "protocols" $protocol "endpoint" "" $collector.config}}
+{{- if and $collector.enabled (ne $endpoint "") -}}
+{{- $port := regexFind ":[0-9]+$" $endpoint -}}
+* Send OTLP to 'http://{{$fullname}}-collector.{{ include "opentelemetry-kube-stack.namespace" $ }}.svc.cluster.local{{$port}} ({{$protocol}})
+{{ end -}}
+{{- end -}}
+{{- end -}}
+{{- if .Values.instrumentation.enabled -}}
+* Annotate your pods with: 'instrumentation.opentelemetry.io/inject-<<language>>: opentelemetry-operator-system/otel-instrumentation' to auto-instrument them
```
cyrille-leclerc: I think that `opentelemetry-operator-system/otel-instrumentation` needs to be variablized:
*  `opentelemetry-operator-system` the K8s NS where the operator is installed
* `otel-instrumentation` is the name of the instrumentation CRD

theSuess: Ah didn't realize this was the case - I'll fix it

PR #2047, opened 2026-01-23
[otel-demo] Bump dependencies to latest version: otel-collector, prometheus, and grafana

cyrille-leclerc:
Bump dependencies to latest version: 
* otel-collector 0.142.0 -> 0.144.0
* prometheus 28.2.0 -> 28.6.0
* grafana 10.5.8 -> 10.5.12

Don't bump jaeger from 4.3.0 to 4.4.2 as the build fails, it seems to on the chart requires changes


charts/opentelemetry-demo/Chart.lock
```diff
@@ -4,15 +4,15 @@ dependencies:
   version: 0.144.0
```
julianocosta89, line 4: I believe you also need to update the `Chart` file

cyrille-leclerc: I think this change is also committed in the PR, it's [here](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2047/files#diff-e22c0af93aebf2544029e2a11c828ac4a1b15c9bb8ec4fd580dfd9d202efa170L19-R19)

I'm investigating on he cause of the build failure, I suspect it's a CI platform issue.

cyrille-leclerc: Build failure fixed

PR #2048, opened 2026-01-26
feat(opentelemetry-operator): enable configuration of autoinstrumentation of nginx image

### LINKED ISSUE #2041: [Opentelemetry Operator chart] No variable in values for Autoinstrumentation NGINX image
Opentelemetry Operator support setting default image for [NGINX autoinstrumentation image](https://github.com/open-telemetry/opentelemetry-operator/blob/7b3d0ec9c1028205c3ba031f173d6eeae65a5ed7/internal/config/config.go#L73) 

Currently this is not exposed in Helm chart values, leading to using the default one always. The workaround would be setting it up via CLI flags, but since there's an official way to do it, it should be templated, as the rest of the autoinstrumentation images [here](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/values.yaml#L60).

faizan-exe:
This PR adds the following option in values.yaml

- `manager.image.autoInstrumentationImage.nginx`

Resolves #2041

### DIFF
--- a/.github/workflows/fossa.yml
+++ b/.github/workflows/fossa.yml
@@ -14,7 +14,7 @@ jobs:
     steps:
       - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
 
-      - uses: fossas/fossa-action@3ebcea1862c6ffbd5cf1b4d0bd6b3fe7bd6f2cac # v1.7.0
+      - uses: fossas/fossa-action@c414b9ad82eaad041e47a7cf62a4f02411f427a0 # v1.8.0
         with:
           api-key: ${{secrets.FOSSA_API_KEY}}
           team: OpenTelemetry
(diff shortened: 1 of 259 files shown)


charts/opentelemetry-operator/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.102.0
+version: 0.103.0
```
jaronoff97: this should be a bump to the patch version not the minor version.

faizan-exe: [CONTRIBUTING.md](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/CONTRIBUTING.md) recommends bumping the minor version and setting patch to zero - but sure I'll make the change.

jaronoff97: @faizan-exe that doc is if you are changing the version of the operator which you are not.

eenchevlp:
@jaronoff97 could you please take a look into this one an merge when you get a chance. It would be really nice if this is included in the next release.

PR #2050, opened 2026-01-26
[opentelemetry-collector]: custom apiVersion support

medzin:
Adds configurable apiVersion and extra updateStrategy fields for DaemonSet/StatefulSet/Deployment mode. This enables:

- Using custom controllers like OpenKruise Advanced DaemonSet and Advanced StatefulSet
- Partition-based rollouts for safer canary deployments
- In-place pod updates without recreation
- Testing new API versions before they become stable


charts/opentelemetry-collector/values.yaml
```diff
@@ -664,6 +672,13 @@ statefulset:
     whenDeleted: Retain
     whenScaled: Retain
 
+daemonset:
```
TylerHelmuth: Do we need this nested under a `daemonset` section? Would it be applicable to other modes?

medzin: In case of OpenKruise the same trick would work for StatefulSet but not for Deployment so I can add support for StatefulSet under a separate key like `statefulset` or we can move `apiVersion` to a top level variable and just fail when using a Deployment.

TylerHelmuth: Is it ever valid to have `mode="deployment"` and to set an API? If so, a root-level field would be nice.

medzin: Theoretically, if `apps/v2` were to be released someday, it would allow for quick testing to see if the chart would work in its current state. However, at the moment, I don't know of a use case for it.

medzin: @TylerHelmuth so should I just move it to the top level? The use case for `mode="deployment"` is a stretch but the code will be cleaner :)

TylerHelmuth: I think top level `apiVersion` is the most future-proof solution.

medzin: @TylerHelmuth ok moved the variable to top level and added examples for other modes.

charts/opentelemetry-collector/values.yaml
```diff
@@ -8,6 +8,12 @@ fullnameOverride: ""
 # Valid values are "daemonset", "deployment", and "statefulset".
 mode: ""
 
+# Override the default apiVersion for custom controllers (e.g., OpenKruise)
+# or for testing new API versions.
+# If not set, defaults to "apps/v1".
+# Example: apps.kruise.io/v1alpha1
```
TylerHelmuth:
```suggestion
# Override the default apiVersion for custom controllers or for testing new API versions.
```

charts/opentelemetry-collector/values.yaml
```diff
@@ -8,6 +8,12 @@ fullnameOverride: ""
...
+apiVersion: ""
```
TylerHelmuth:
```suggestion
apiVersion: "apps/v1"
```

charts/opentelemetry-collector/values.schema.json
```diff
@@ -40,6 +40,10 @@
         ""
       ]
     },
+    "apiVersion": {
```
TylerHelmuth, line 43: Make this field required and then the 3 templates using it can use it directly instead of having conditional locig.

charts/opentelemetry-collector/values.schema.json
```diff
@@ -1038,6 +1042,7 @@
       "properties": {
         "rollingUpdate": {
           "type": "object",
+          "additionalProperties": true,
```
TylerHelmuth: This change seems unrelated

TylerHelmuth (commented):
Left some feedback on how we can make the implementation simpler. Also please remove the new examples as the new field is straightforward.

medzin:
@TylerHelmuth removed examples and simplified the code.

TylerHelmuth (approved):

PR #2055, opened 2026-01-27
[operator] Bump version to 0.144.0

swiatekm:
Closes https://github.com/open-telemetry/opentelemetry-operator/issues/4654

### DIFF
--- a/charts/opentelemetry-operator/values.yaml
+++ b/charts/opentelemetry-operator/values.yaml
@@ -50,7 +50,7 @@ manager:
     imagePullPolicy: IfNotPresent
   collectorImage:
     repository: ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector-k8s
-    tag: 0.143.0
+    tag: 0.144.0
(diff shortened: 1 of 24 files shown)


charts/opentelemetry-operator/values.yaml
```diff
@@ -50,7 +50,7 @@ manager:
     imagePullPolicy: IfNotPresent
   collectorImage:
     repository: ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector-k8s
-    tag: 0.141.0
+    tag: 0.144.0
```
jaronoff97, line 53: do you think it matters that we're skipping some versions?

swiatekm: Maybe? Technically, you can use a different image if you want, the Chart is typically both forward and backward compatible. But it might be a good idea to release the previous ones, too.

swiatekm: The previous versions are now published as well.

PR #2066, opened 2026-02-02
feat: remove clusterMetrics restriction in NOTES.txt

yp28:
As the k8s_leader_election extension can be used with the deployment mode, we no longer need a restriction on the amount of replicas that can be run with the clusterMetrics preset enabled.


yp28:
This needs more work. I see the leader election extension is automatically enabled and configured when using it with the DaemonSet mode. The same can be done for the Deployment mode with more than 1 replica configured.

yp28:
Closing as I no longer run the Deployment, in favour of the DaemonSet with k8sleaderelection for all cluster scoped metrics.

Issue #2063, opened 2026-02-02
ServiceMonitor not working after updating to 0.105.0


pozniakas:
I was using 0.102.0 version and everything was working just fine, but after updating to 0.105.0 I started to encounter problems in Prometheus:

<img width="1565" height="310" alt="Image" src="https://github.com/user-attachments/assets/0719cade-f2cf-4c14-8980-d38123c31b6d" />

After checking logs of the operator it was pretty clear that something is not right:

<img width="896" height="119" alt="Image" src="https://github.com/user-attachments/assets/46a933a0-113d-4492-a687-a92e95281bf5" />

`http: TLS handshake error from 10.244.61.37:39236: client sent an HTTP request to an HTTPS server`
It seems that it expects that the metric route would be triggered in HTTPS, but I don't understand why.

### Helm Chart Values

```yaml
        admissionWebhooks:
          create: false
          autoGenerateCert:
            enabled: false
          certManager:
            enabled: false
        manager:
          collectorImage:
            repository: otel/opentelemetry-collector-contrib
          env:
            ENABLE_WEBHOOKS: "false"
          prometheusRule:
            enabled: true
            defaultRules:
              enabled: true
          resources:
            limits:
              memory: 512Mi
            requests:
              cpu: 20m
              memory: 150Mi
          serviceMonitor:
            enabled: true
          leaderElection:
            enabled: false # enable if using more than one replica
        crds:
          create: true
```

### Generated ServiceMonitor manfiest

```yaml
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  annotations:
  generation: 1
  labels:
    app.kubernetes.io/component: controller-manager
    app.kubernetes.io/instance: otel-operator
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: opentelemetry-operator
    app.kubernetes.io/part-of: opentelemetry-operator
    app.kubernetes.io/version: 0.144.0
    argocd.argoproj.io/instance: staging-otel-operator
    helm.sh/chart: opentelemetry-operator-0.105.0
  name: otel-operator-opentelemetry-operator
  namespace: tempo
spec:
  endpoints:
  - port: metrics
  namespaceSelector:
    matchNames:
    - tempo
  selector:
    matchLabels:
      app.kubernetes.io/component: controller-manager
      app.kubernetes.io/name: opentelemetry-operator
```

Any idea what could be happening?

[...]

PR #2075, opened 2026-02-13
[operator] Disable secure metrics serving by default

swiatekm:
Fixes https://github.com/open-telemetry/opentelemetry-helm-charts/issues/2063 by disabling secure metrics serving by default. I've added it as an argument in the deployment template, which essentially makes it a different default, and users can still change via `manager.extraArgs`. https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2004 will remove RBAC proxy and add proper configuration for this feature to values.yaml. Right now, we just want to fix the breakage caused by 0.102.0.

### DIFF
--- a/charts/opentelemetry-operator/values.yaml
+++ b/charts/opentelemetry-operator/values.yaml
@@ -184,8 +184,7 @@ manager:
   createRbacPermissions: false
   ## List of additional cli arguments to configure the manager
   ## for example: --labels, etc.
-  extraArgs:
-  - "--metrics-secure=false"
+  extraArgs: []
 
   ## Enable leader election mechanism for protecting against split brain if multiple operator pods/replicas are started.
   ## See more at https://docs.openshift.com/container-platform/4.10/operators/operator_sdk/osdk-leader-election.html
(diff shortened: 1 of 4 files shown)


charts/opentelemetry-operator/values.yaml
```diff
@@ -184,7 +184,8 @@ manager:
   createRbacPermissions: false
   ## List of additional cli arguments to configure the manager
   ## for example: --labels, etc.
-  extraArgs: []
+  extraArgs:
+  - "--metrics-secure=false"
```
TylerHelmuth: This will work for anyone not setting additional args, but if anyone is using this value then this default value will be overridden. I think we can either:

- document this behavior in UPGRADING.md and direct users to add it in their custom list
- add this arg directly to the pod definition so it is always included no matter what.

swiatekm: I thought that would make it impossible to change, but apparently you can specify multiple values for a flag, and the last one gets used. The more you know. I'll modify this the way you suggest.

swiatekm: https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2075/changes/824a9b555890ccf06f6a067cb700ff9075806da6

PR #2076, opened 2026-02-14
feat(opentelemetry-collector): add container resizePolicy schema for in-place resize

boqu:
## Summary
Add JSON schema and template support for the container-level `spec.resizePolicy` field, enabling VPA in-place pod resource resize (Kubernetes >= 1.27) on the collector container. VPA CRDs, controller, and the `VerticalPodAutoscaler` object itself remain user-managed (e.g. via `extraManifests`); the chart only renders `resizePolicy` because it cannot be expressed via `extraManifests`.

### Changes
- `values.schema.json`: new `resizePolicy` array — `resourceName: string`, `restartPolicy: NotRequired | RestartContainer`.
- `values.yaml`: `resizePolicy: []` default with a commented example.
- `templates/_pod.tpl`: renders the container's `resizePolicy:` block when set.

Scoped down from the original "native VPA support" proposal per https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2076#discussion_r3209657136.


charts/opentelemetry-collector/values.yaml
```diff
@@ -633,6 +642,47 @@ autoscaling:
   # Supply an array of custom metrics to be used for autoscaling. It includes externalMetrics, objectMetrics, and podsMetrics.
   additionalMetrics: []
 
+# Vertical Pod Autoscaler - https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler
```
TylerHelmuth: Lets avoid linking directly to kubernetes github repos. Can you link to their public docs site instead?

charts/opentelemetry-collector/values.yaml
```diff
@@ -633,6 +642,47 @@ autoscaling:
...
+verticalPodAutoscaler:
+  enabled: false
+
+  # Recommender responsible for generating recommendation for the object.
+  # List should be empty (then the default recommender will generate the recommendation)
+  # or contain exactly one recommender.
```
TylerHelmuth: This comment is kinda confusing. Is there a doc we can link to instead?

charts/opentelemetry-collector/templates/vpa.yaml
```diff
@@ -0,0 +1,40 @@
+{{- if and (.Capabilities.APIVersions.Has "autoscaling.k8s.io/v1") .Values.verticalPodAutoscaler.enabled }}
```
TylerHelmuth: I really prefer not having to do APIVersion checks as it complicates the logic of the helm chart. The readme says the chart supports 1.24+, but 1.27 has been EOL for 2+ years. I think we can just check for `.Values.verticalPodAutoscaler.enabled`. If someone is running this chart on v1.24 and enables the feature then 🤷

charts/opentelemetry-collector/values.yaml
boqu: I update the link

charts/opentelemetry-collector/values.yaml
boqu: Yeah, it's bit confusing. I updated the comment.

charts/opentelemetry-collector/templates/vpa.yaml
boqu: The unnecessary check is removed

dmitryax:
I would be strongly against bringing non-standard crds to this chart. Is there currently a way to install VPA separately?

charts/opentelemetry-collector/values.yaml
```diff
@@ -633,6 +642,45 @@ autoscaling:
   # Supply an array of custom metrics to be used for autoscaling. It includes externalMetrics, objectMetrics, and podsMetrics.
   additionalMetrics: []
 
+# Vertical Pod Autoscaler - https://kubernetes.io/docs/concepts/workloads/autoscaling/#vertical-pod-autoscaling
+verticalPodAutoscaler:
+  enabled: false
```
TylerHelmuth: I have misunderstood how this feature is implemented in kubernetes. Is this feature not available in a standard +1.27 k8s install? What are the additional CRDs it needs?

boqu: No, the VPA doesn't come with kubernetes in default. It need to be installed separately. See https://kubernetes.io/docs/concepts/workloads/autoscaling/#scaling-workloads-vertically and https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler. The CRD can be found [here](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/deploy/vpa-crd.yaml).

TylerHelmuth: We've typically been against adding direct support for external CRDs, but seeing as how this is a kubernetes-owned CRD I'm ok adding this feature. @open-telemetry/helm-approvers what do you think?

boqu: Can I know if we can move this PR forward or anything outstanding concern?

boqu: I'd appreciate if we can move this PR forward.

boqu:
Any chance we can move this PR forward?

charts/opentelemetry-collector/values.yaml
boqu: @TylerHelmuth Anything I can do to move this PR forward?

povilasv: I would be ok with having native VPA support.

But didnt understand why extraManifess didnt work?

This error seems that the VPA CRDs wasnt installed on that cluster:
>Error: INSTALLATION FAILED: unable to build kubernetes objects from release manifest: resource mapping not found for name: "opentelemetry-collector-o6n8pl90u0" namespace: "opentelemetry-collector-o6n8pl90u0" from "": no matches for kind "VerticalPodAutoscaler" in version "autoscaling.k8s.io/v1"

boqu: My understanding is, before applying anything, the helm renders the whole release and runs every manifest through the client‑side REST mapper to resolve its GVK(Group, Version and Kind) against the API server's discovery. At that moment, the CRD added via extraManifest has not been posted. Thus, helm fails to build VPA object. That also explains why it works if we pre-install VPA CRD in the CI test. See .github/actions/setup/action.yaml.

TylerHelmuth: Ya you can't install a CRD via the manifest. Managing CRDs in the helm chart that also installs something that relies on those CRDs is really hard (see the kube-stack chart). If you install the CRDs separately I'd expect the `extraManifest` to work. We will definitely not be adding the CRDs to this chart, so no matter what you'll need to manage the CRDs elsewhere.

boqu: I guess I probably didn't describe the PR clearly in the first place. The OTEL collector helm chart won't install VPA CRDs. If users use VPA with the helm chart, they need to make sure the VPA CRDs have been installed already. I just updated the comment to make it clear.

I also bump the version to resolve the merge conflicts.

Can we consider to move this PR forward?

TylerHelmuth: If the CRDs are already installed then I'm not clear why extraManifests does not work

boqu: You are right, I think I didn't fully understand the purpose of extraManifests. With VPA CRDs pre-installed, we can create VPA object with extraManifests.

However, if we want to use the in-place resizing with VPA, we need to set `spec.resizePolicy`. `spec.resizePolicy` is a field of container spec. I think we still need to update the template to support `spec.resizePolicy`. Please correct me if I'm wrong. Thanks!

TylerHelmuth: I am ok adding support in the json schema for setting that value. Can you scope the PR done to that change?

boqu: I scoped the PR down to only the json schema.

TylerHelmuth: @boqu I think I misunderstood. If we have to make changes to the template that means we're adding features that depend on a CRD which we really want to avoid.

How are other helm charts, like the prometheus chart, handling this?

boqu: @TylerHelmuth  I had a quick look at prometheus helm-charts. It seems they don't support native VPA. See https://github.com/prometheus-community/helm-charts/issues/4711. 

I understand we try to avoid a dependency on a CRD. But in this PR, we have scoped it down to only exposing spec.containers[*].resizePolicy, which is a field of the core Kubernetes API. We don't rely on any additional CRDs(e.g. VPA CRDs). To use the VPA, it's completely up to users to take care all required VPA CRDs and resources.

TylerHelmuth: > which is a field of the core Kubernetes API

Perfect!

TylerHelmuth (approved):

Issue #2068, opened 2026-02-05
Update references to otlp and otlphttp exporters


codeboten:
The otlp and otlphttp exporters have been updated to follow the lower snake case naming convention (see https://github.com/open-telemetry/opentelemetry-collector/issues/14396). As a result, example configuration has been updated to match the new convention.

Note that the old names will continue to work for some time, but are officially deprecated.

Related to https://github.com/open-telemetry/opentelemetry-collector/issues/14429 and https://github.com/open-telemetry/opentelemetry-collector/issues/14430

TylerHelmuth:
We'll need to think about how to handle this in a backwards compatible way. Any chart that has a default collector config that uses `otlp` or `otlphttp` exporters could have situations where end users depend on those strings for the helm overrides. If we change them to `otlp_grpc` and `otlp_http` directly than those overrides wont work.

LHozzan:
Hi there.
Same problem here (helm chart version 0.158.1).
Have you please some ETA, when you expecting to solving the problem?
Thank you!

TylerHelmuth:
We can use the same strategy we're using for the other component rename and normalize to the new names.

LHozzan:
> Have you please some ETA, when you expecting to solving the problem?

OK, so no ETA so far, right?

TylerHelmuth:
No eta at this time. The old names continue to work.

LHozzan:
> No eta at this time. The old names continue to work.

OK, thank you for feedback, then we are waiting for fix :).

szelese:
Hi! I’d like to work on this issue.

Based on the current rewriteDeprecatedComponentNames mechanism, I propose the following scope:

add exporter mappings for otlp -> otlp_grpc and otlphttp -> otlp_http
rewrite both exporter configuration keys and pipeline references, including named instances such as otlp/backend
add deprecation warnings and regression coverage
update the upgrading documentation, bump the chart version, and regenerate the examples

I couldn’t find an active PR implementing #2068. Would this scope be acceptable before I start?

TylerHelmuth:
@szelese go aheard

PR #2080, opened 2026-02-16
[demo] Leverage `set_semconv_span_name()` for better span and http operation names

cyrille-leclerc:
Use the Transform Processor `set_semconv_span_name()` function to workaround span metrics cardinality explosion caused by high-cardinality span names while offering meaningful span and http operation names.

Port on the helm chart deployment of:

* https://github.com/open-telemetry/opentelemetry-demo/pull/2942

### DIFF
--- a/.github/scripts/enhance-release-notes.sh
+++ b/.github/scripts/enhance-release-notes.sh
@@ -3,6 +3,9 @@
 
 set -euo pipefail
 
+# Get the repository from environment or default
+REPO="${GITHUB_REPOSITORY:-open-telemetry/opentelemetry-helm-charts}"
+
 # Function to update release notes for a chart
 update_release_notes() {
     local chart_name="$1"
@@ -16,7 +19,7 @@ update_release_notes() {
     enhanced_notes=$(.github/scripts/generate-release-notes.sh "$chart_name" "$chart_version" "$app_version" "charts/$chart_name/")
 
     # Get the current release notes
-    current_notes=$(gh release view "$release_tag" --json body --jq '.body' 2>/dev/null || echo "")
+    current_notes=$(gh release view "$release_tag" --repo "$REPO" --json body --jq '.body' 2>/dev/null || echo "")
 
     if [[ -z "$current_notes" ]]; then
         echo "ERROR: No existing release notes found for ${release_tag}. This indicates a problem with the chart-releaser process."
@@ -32,7 +35,7 @@ update_release_notes() {
 $enhanced_notes"
 
     # Update the release notes
-    gh release edit "$release_tag" --notes "$combined_notes"
+    gh release edit "$release_tag" --repo "$REPO" --notes "$combined_notes"
 }
 
 # Main function to enhance release notes for recent releases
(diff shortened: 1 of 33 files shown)


charts/opentelemetry-demo/values.yaml
```diff
@@ -967,17 +967,51 @@ opentelemetry-collector:
...
+              # 2. Then normalize span names; http server spans lacking http.route default to operations "GET", "POST", etc.
+
+              # FRONTEND SERVICE
+
+              # Workaround for Next.js high cardinality span name issue: https://github.com/vercel/next.js/issues/54694
+              - set(span.attributes["http.route"], "/api/cart") where
+                span.kind == SPAN_KIND_SERVER and
+                resource.attributes["service.name"] == "frontend" and
+                span.attributes["http.route"] == nil and
+                IsMatch(span.attributes["http.target"], "\\/api\\/cart")  # e.g. # /api/cart
```
TylerHelmuth: Does this syntax for sure work? I expected to see

```yaml
- |
  set(span.attributes["http.route"], "/api/cart") where
    span.kind == SPAN_KIND_SERVER and
    resource.attributes["service.name"] == "frontend" and
    span.attributes["http.route"] == nil and
    IsMatch(span.attributes["http.target"], "\\/api\\/cart")  # e.g. # /api/cart
```

charts/opentelemetry-demo/values.yaml
```diff
@@ -967,17 +967,51 @@ opentelemetry-collector:
         spike_limit_percentage: 25
       resourcedetection:
         detectors: [env, system]
-      # This processor is used to help limit high cardinality on next.js span names
-      # When this PR is merged (and released) we can remove this transform processor
-      # https://github.com/vercel/next.js/pull/64852
-      transform:
+      transform/sanitize_spans:
```
TylerHelmuth: I dont think we gain anything by renaming the processor. Some customers by be extending it so lets keep the original name to make this change as backwards-compatible as we can.

cyrille-leclerc: On the docker-compose deployment of the otel-demo, we accepted this backward incompatibility ([here](https://github.com/open-telemetry/opentelemetry-demo/pull/2942)) because we bundled it with the backward incompatibility of the otlp exporters `otlp_http` and `otlp_grpc`.
Note that I saw your concern about the backward incompatibility `otlp_http` and `otlp_grpc` https://github.com/open-telemetry/opentelemetry-helm-charts/issues/2068#issuecomment-3862538836 and your point makes sense but I didn't see a solution yet.

charts/opentelemetry-demo/values.yaml
cyrille-leclerc: The syntax does work with docker-compose, let me double check on K8s.

charts/opentelemetry-demo/values.yaml
cyrille-leclerc: initial processor name rollbacked

charts/opentelemetry-demo/values.yaml
cyrille-leclerc: Replaced by new syntax with better performances thanks to usage of a `conditions` block

charts/opentelemetry-demo/values.yaml
```diff
@@ -967,17 +967,28 @@ opentelemetry-collector:
         spike_limit_percentage: 25
       resourcedetection:
         detectors: [env, system]
-      # This processor is used to help limit high cardinality on next.js span names
-      # When this PR is merged (and released) we can remove this transform processor
-      # https://github.com/vercel/next.js/pull/64852
```
cyrille-leclerc, line 972: @TylerHelmuth please review improved processor block.

TylerHelmuth (approved):

PR #2081, opened 2026-02-16
feat: add trafficDistribution support to opentelemetry-kube-stack

kema-dev:
## Description

- Add support for passing `trafficDistribution` to [OpenTelemetryCollector.spec](https://github.com/open-telemetry/opentelemetry-operator/blob/main/docs/api/opentelemetrycollectors.md#opentelemetrycollectorspec)

## Possible improvements

- We could add `PreferSameNode` as value for the daemon (default) collector, but that would require a fairly recent Kubernetes version


jaronoff97 (commented):
please read the `CONTRIBUTING.md` file for instructions on how to make a change to the chart, then i'll happily review. thank you!

charts/opentelemetry-kube-stack/templates/collector.yaml
```diff
@@ -49,6 +49,9 @@ spec:
   {{- if $collector.priorityClassName }}
   priorityClassName: {{ $collector.priorityClassName }}
   {{- end }}
+  {{- if $collector.trafficDistribution }}
+  trafficDistribution: {{ $collector.trafficDistribution }}
```
TylerHelmuth, line 53: You'll need to update the json schema to be allowed to set `trafficDistribution` in the `.Values.collectors` section.

PR #2085, opened 2026-02-20
Bump operator version to 0.145.0

iblancasa:


Makefile
TylerHelmuth, line 1: Can you split these changes out into a separate PR

TylerHelmuth (approved):

Issue #2103, opened 2026-03-02
feat(opentelemetry-operator) Add minReadySeconds support for operator deployment


jinja2:
Currently there is no way to configure `minReadySeconds` on the operator deployment via Helm values. This is useful for users who want to add a readiness buffer during rolling updates to avoid routing traffic to pods that crash shortly after becoming ready.

marcschaeferger:
Can be closed as implemented by #2190. The Operator chart now exposes `minReadySeconds` and renders it on the Deployment.

PR #2104, opened 2026-03-02
Add minReadySeconds support for operator deployment

jinja2:
Tracking issue - #2103

- Adds a new top-level `minReadySeconds` value to configure the minimum number of seconds a newly created pod must be ready before it is considered available.

- Defaults to the k8s default value of 0 seconds, which preserves existing behavior. The field is only rendered when set to a positive value.


jaronoff97 (approved):
Can you rerun make generate-examples when you get a chance? Then I'll merge

jinja2:
Opened a new one - https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2190

Issue #1963, opened 2025-11-26
[collector] Is there a reason why only the deployment creates a PDB?


snuggie12:
My GKE console is recommending my `StatefulSets` have PDBs. It's true that the rest of our STS's have PDBs but I see `Deployment` is specifically singled out in the helm chart.

Is there a reason for this? Should an "or" be added?

TylerHelmuth:
@snuggie12 I believe our PodDisruptionBudget feature was added before we supported StatefulSets. We're open to updating that template to supporting Statefulsets.

PR #2113, opened 2026-03-12
feat(opentelemetry-collector): support PodDisruptionBudget for statefulset mode

nedecz:
Extends the PodDisruptionBudget template to support statefulset mode in addition to deployment mode.

Previously, enabling podDisruptionBudget had no effect when running the collector as a StatefulSet. The condition in `pdb.yaml` only matched `deployment` mode, which meant `statefulset` users could not protect their pods from voluntary disruptions.

Existing issue:
https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1963


TylerHelmuth (approved):

nedecz:
Hi, @TylerHelmuth 

There were some merge conflicts, I fixed them.

PR #2126, opened 2026-03-26
feat: add helm profiling preset

wehzzz:
## What does this PR do?
Adds a new `profiling preset` to the opentelemetry-collector Helm chart. When `presets.profiling.enabled: true`, the chart automatically configures:

- A `profiling receiver` and a `profiles pipeline` with `debug exporter` as fallback if no exporter configured
- A `tracefs` host volume mount (`/sys/kernel/tracing`, read-only)
- `hostPID: true` on the pod spec
- Security context (runAsUser: 0, runAsGroup: 0, privileged: true)
- 
When combined with the kubernetesAttributes preset, the chart also:
- Injects the `k8sattributes processor` into the `profiles pipeline` (same pattern as logs/metrics/traces)
- Adds `container.id` to the `pod_association` configuration - eBPF profilers identify workloads by container ID, not pod IP/UID. 

## Motivation
Deploying an eBPF profiler (e.g. `opentelemetry-collector-ebpf-profiler`) on Kubernetes currently requires significant boilerplate. This was discussed in https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/1072 where the community expressed interest in a preset to simplify this.

## Describe how you validated your changes
Two new CI values files:
- `ci/preset-profiling-values.yaml` - profiling preset only
- `ci/preset-profiling-kubernetesattributes-values.yaml` - profiling + kubernetesAttribute

## Additional notes
- The `container.id` pod_association injection is gated on `$config.service.pipelines.profiles` existence in `applyKubernetesAttributesConfig`, not on `profiling.enabled`. This means any profiles pipeline (preset or user-defined) benefits from the enrichment when `kubernetesAttributes` is enabled.

- `hostPID` is applied to all three workload templates (`DaemonSet`, `Deployment`, `StatefulSet`) for consistency.


charts/opentelemetry-collector/ci/preset-profiling-kubernetesattributes-values.yaml
```diff
@@ -0,0 +1,19 @@
...
+presets:
+  profiling:
+    enabled: true
+  kubernetesAttributes:
+    enabled: true
+
+resources:
+  limits:
+    cpu: 100m
+    memory: 200M
```
florianl, line 19: Was this memory limit tested in some way? Depending on the environment, my experiments usually require 400M to 500M.

florianl (approved):
Overall looks good to me - but I'm not an helm expert. Just a minor question around the set resource limits.

charts/opentelemetry-collector/templates/deployment.yaml
```diff
@@ -55,5 +55,5 @@ spec:
       {{- $podData := dict "Values" $podValues "configmapSuffix" "" "isAgent" false }}
       {{- include "opentelemetry-collector.pod" ($podData | mustMergeOverwrite (deepCopy .)) | nindent 6 }}
       hostNetwork: {{ .Values.hostNetwork }}
-      hostPID: {{ .Values.hostPID }}
+      hostPID: {{ or .Values.hostPID .Values.presets.profiling.enabled }}
```
nicolastakashi, line 58: What we expect if profiling.enabled: true and hostPID: ""?

wehzzz: Since the eBPF profiler requires `hostPID: true` to function, it makes sense that enabling the profiling preset forces this value to `true`, regardless of whether `hostPID` is left empty ("") in the user's configuration.

charts/opentelemetry-collector/ci/preset-profiling-kubernetesattributes-values.yaml
wehzzz: If I'm not mistaken, this limit is only used for the CI environment. It's not a default value injected by `kubernetesattributes` or the new `profiling` preset.

nicolastakashi (approved):

charts/opentelemetry-collector/ci/preset-profiling-values.yaml
TylerHelmuth, line 1: To keep the CI running fast as possible lets remove this test and depend on the other one. They seem similar enough that I dont think we gain much from this one.

charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -323,6 +329,22 @@ receivers:
...
+{{- $_ := set $config.service.pipelines.profiles "receivers" (append $config.service.pipelines.profiles.receivers "profiling" | uniq)  }}
+{{- if not $config.service.pipelines.profiles.exporters }}
+{{- $_ := set $config.service.pipelines.profiles "exporters" (list "debug") }}
+{{- end }}
+{{- $config | toYaml }}
+{{- end }}
+
+{{- define "opentelemetry-collector.profilingConfig" -}}
+receivers:
+  profiling: {}
```
TylerHelmuth, line 343: We need to put this receiver in the [k8s distro](https://github.com/open-telemetry/opentelemetry-collector-releases/blob/main/distributions/otelcol-k8s/manifest.yaml) before we can merge this preset as it is the distro we expect users to use.

If the receiver is not ready to be put into that distro then I question if it is ready for a preset.

Gandem: @TylerHelmuth That's a fair question, but I think the profiling receiver is in a different category than other receivers in the chart. The eBPF profiler needs to run as `privileged` with `hostPID`. If we put that receiver in the k8s distro, users enabling profiling would be granting those privileges to the same collector that handles their metrics, traces, and logs. That's a security tradeoff we probably don't want users to make.

The receiver ships as its own distro (`otelcol-ebpf-profiler`) specifically for this reason. The preset is meant to make it easier to deploy that dedicated workload through the helm chart (e.g. in a dedicated separate daemonset or deployment), not to add profiling to the general-purpose collector. So I'd argue the receiver not being in the k8s distro is by design rather than a readiness concern.

We're happy to join a SIG meeting or schedule a sync if it'd be easier to talk through this!

TylerHelmuth: Is there anything we can do to help users not accidentally enable this preset when it wont work?

wehzzz: We came up with two options to handle this:
1. We could add a check on the `image` field and error out if the configured image is `otelcol-k8s`. This would not cover users using different/custom distributions.
2. We could alternatively add specific image fields directly to the preset, as follows:
```yaml
presets:
  profiling:
    enabled: false
    image:
      repository: ""
      pullPolicy: IfNotPresent
      tag: ""
      digest: ""
```
Then, we could do the following:
- Error out if the image is not set at the preset level.
- Error out if the image is otelcol-k8s.
- Otherwise, override the image config.

This is quite similar to the first option, but requiring the image to be specified under the preset ensures the user will look into the settings to understand why it errors out.

We are open to discussing these options or exploring something else if you have an idea!

wehzzz: @TylerHelmuth small bump on ^

TylerHelmuth: I think an error if the image is any of the community images that dont have the receiver is good.

wehzzz: Thanks! I added a check for this.

charts/opentelemetry-collector/templates/_pod.tpl
```diff
@@ -28,7 +28,11 @@ containers:
       - {{ . }}
       {{- end }}
     securityContext:
-      {{- if and (not (.Values.securityContext)) (.Values.presets.logsCollection.storeCheckpoints) }}
+      {{- if and (not (.Values.securityContext)) (.Values.presets.profiling.enabled) }}
+      runAsUser: 0
+      runAsGroup: 0
+      privileged: true
+      {{- else if and (not (.Values.securityContext)) (.Values.presets.logsCollection.storeCheckpoints) }}
```
TylerHelmuth, line 35: What happens if both these presets are in use?

wehzzz: `storeCheckpoints`'s requirements are a subset of `profiling`'s - both want `runAsUser`/`runAsGroup: 0`, and `storeCheckpoints` has no need for `privileged`. So when profiling is enabled, the `if`/`else if` renders the profiling branch, which already satisfies `storeCheckpoints` and needs `priviliged` for profiling. When `profiling` is disabled, only `runAsUser`/`runAsGroup: 0` is rendered - no `privileged`. No over-privileging in either case. 

While reviewing this, I've spotted two related gaps worth flagging - both stemming from the fact that the profiling branch is guarded by not `.Values.securityContext`:
- User-set `.Values.securityContext`: if the user defines any container `securityContext`, the profiling branch is skipped entirely and `privileged: true` is silently dropped. The eBPF profiler will then fail at runtime with no clear signal from the chart.
- User-set `podSecurityContext.runAsNonRoot: true`: inherited by the container, conflicts with the preset's `runAsUser: 0`. Kubelet blocks container start with container has runAsNonRoot and image will run as root 

Consistent with the distro-check we just added, we can handle these as follows in NOTES.txt:
- fail if `profiling.enabled` and `.Values.podSecurityContext.runAsNonRoot: true`.
- warn if `profiling.enabled` and `.Values.securityContext` is set, reminding the user they need to include either `privileged: true` or the required capabilities. We don't want to hard-fail here since a custom `securityContext` can be a legitimate setup.

wdyt?

TylerHelmuth: Ya that sounds good

TylerHelmuth (approved):

PR #2142, opened 2026-04-10
[CDS-2869] fix: add IsMap() guards to transform/kube-events processor

natnayr:
## Problem

The `transform/kube-events` preset uses OTTL `keep_keys` statements to filter Kubernetes event log fields:

```
keep_keys(body["object"], ["type", "eventTime", ...])
keep_keys(body["object"]["metadata"], ["creationTimestamp"])
keep_keys(body["object"]["regarding"], ["kind", "name", "namespace"])
```

These fail with `INVALID_ARGUMENT` when a Kubernetes event log body is a **plain string** rather than a JSON object. Some k8s controllers emit string-bodied events, causing the entire batch to fail and drop.

## Fix

Add inline `where IsMap()` guards to each `keep_keys` call:

```
keep_keys(body["object"], [...]) where IsMap(body["object"])
keep_keys(body["object"]["metadata"], ["creationTimestamp"]) where IsMap(body["object"]["metadata"])
keep_keys(body["object"]["regarding"], ["kind", "name", "namespace"]) where IsMap(body["object"]["regarding"])
```

String-bodied records pass through untouched. Map-bodied records are filtered as before.

**Why inline `where` and not the `conditions:` block?**  
The `conditions:` block syntax was introduced in a later collector version. Using `where IsMap()` inline is compatible across all supported collector versions.

## Testing

Locally reproduced with a 3-record OTLP test against the `otel-infrastructure-collector` chart:
- Record 1: string body → previously caused `INVALID_ARGUMENT`, now passes through unchanged ✓
- Record 2: map body with extra field → extra field correctly stripped by `keep_keys` ✓  
- Record 3: map body with valid keys only → all keys preserved ✓

CI will also validate via `validate-configs.sh` against `ci/preset-k8sevents-values.yaml`.

Jira: [CDS-2869](https://coralogix.atlassian.net/browse/CDS-2869)


natnayr:
Opened against wrong repo — this fix is for the Coralogix fork.

PR #2144, opened 2026-04-10
fix(operator): use toYaml for VPA controlledResources

LaikaN57:
## Summary
- `controlledResources` in the VPA template was rendered inline without `toYaml`, producing invalid YAML output (e.g. `controlledResources: [cpu memory]` instead of a proper YAML list)
- This caused the VPA controller to fail with:
  ```
  E0410 23:03:16.284497       1 types.go:131] "Cannot translate resource name" resourceName="cpu memory"
  ```
- Use `toYaml | nindent` to match the pattern used by the other fields (`recommenders`, `maxAllowed`, `minAllowed`) in the same template

### DIFF
--- a/charts/opentelemetry-operator/Chart.yaml
+++ b/charts/opentelemetry-operator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.109.1
+version: 0.109.2
 description: OpenTelemetry Operator Helm chart for Kubernetes
 type: application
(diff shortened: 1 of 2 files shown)


dmitryax:
@LaikaN57 Thanks for the fix. Please bump the chart's version

LaikaN57:
> @LaikaN57 Thanks for the fix. Please bump the chart's version

@dmitryax done ✅

dmitryax (approved):

PR #2145, opened 2026-04-11
feat(collector): add pod-level terminationGracePeriodSeconds to values

### LINKED ISSUE #2143: [collector] Add pod-level terminationGracePeriodSeconds to values
### What would you like to be added?

Add a `terminationGracePeriodSeconds` field to the pod spec in `_pod.tpl`, configurable via a top-level value in `values.yaml`.

The chart currently only supports `terminationGracePeriodSeconds` on liveness/startup probes, not on the pod spec itself. The pod always uses the Kubernetes default of 30 seconds.

### Why is this needed?

On SIGTERM, the collector gracefully shuts down by draining its in-memory exporter queues (receivers stop first, then processors, then exporters drain). The collector has no internal configurable timeout for this drain (open-telemetry/opentelemetry-collector-contrib#34054 was closed as not planned), so it drains until either the queue is empty or the process is killed externally.

With the default sending queue capacity of 1,000 items and 10 consumers, draining can take over 30 seconds depending on backend latency (e.g., 1,000 items / 10 consumers × 5s exporter timeout = 500s worst case). When the Kubernetes grace period expires before the drain completes, the pod is SIGKILLed and queued data is lost. The only way to give the collector more drain time is through the Kubernetes pod-level `terminationGracePeriodSeconds`, which this chart does not currently expose.

### Suggested change

In `values.yaml`:
```yaml
# terminationGracePeriodSeconds: 30
```

In `_pod.tpl`, at the pod spec level (not inside a container or probe):
```yaml
{{- if .Values.terminationGracePeriodSeconds }}
terminationGracePeriodSeconds: {{ .Values.terminationGracePeriodSeconds }}
{{- end }}
```

And add it to `values.schema.json`.

Happy to open a PR for this if the approach looks good.

younsl:
## Summary

The collector chart currently does not expose a pod-level terminationGracePeriodSeconds field. When the collector receives SIGTERM, it gracefully drains its in-memory exporter queues before shutting down. If the drain takes longer than the Kubernetes default grace period of 30 seconds, the pod is forcibly killed and queued telemetry data is lost.

This change adds terminationGracePeriodSeconds as a configurable value in the collector chart. The default is set to 30 seconds to match the existing Kubernetes behavior, and users can increase it as needed to give the collector enough time to flush queued data.

Resolves https://github.com/open-telemetry/opentelemetry-helm-charts/issues/2143

## Changes

The terminationGracePeriodSeconds field is added to values.yaml with a default of 30. The pod template in _pod.tpl conditionally renders this field at the pod spec level, following the same pattern used by priorityClassName and runtimeClassName. The field is also added to values.schema.json as an integer type. All rendered examples are regenerated via make generate-examples.

For more details on pod termination behavior, see the Kubernetes documentation on [pod lifecycle](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination) and [pod API spec](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#lifecycle).

<details>
<summary>Test results</summary>

All three workload modes (deployment, daemonset, statefulset) were tested with helm template. With the default value, terminationGracePeriodSeconds renders as 30 in the pod spec. When overridden to 120, the custom value is correctly applied across all modes.

Helm lint was run against all 24 CI values files and all passed without errors.

</details>

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.151.0
+version: 0.152.0
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-collector/examples/alternate-config/rendered/clusterrole.yaml
+++ b/charts/opentelemetry-collector/examples/alternate-config/rendered/clusterrole.yaml
@@ -5,7 +5,7 @@ kind: ClusterRole
 metadata:
   name: example-opentelemetry-collector
   labels:
-    helm.sh/chart: opentelemetry-collector-0.151.0
+    helm.sh/chart: opentelemetry-collector-0.152.0
     app.kubernetes.io/name: opentelemetry-collector
     app.kubernetes.io/instance: example
     app.kubernetes.io/version: "0.150.1"
--- a/charts/opentelemetry-collector/examples/alternate-config/rendered/clusterrolebinding.yaml
+++ b/charts/opentelemetry-collector/examples/alternate-config/rendered/clusterrolebinding.yaml
@@ -5,7 +5,7 @@ kind: ClusterRoleBinding
 metadata:
   name: example-opentelemetry-collector
   labels:
-    helm.sh/chart: opentelemetry-collector-0.151.0
+    helm.sh/chart: opentelemetry-collector-0.152.0
     app.kubernetes.io/name: opentelemetry-collector
     app.kubernetes.io/instance: example
     app.kubernetes.io/version: "0.150.1"
--- a/charts/opentelemetry-collector/examples/alternate-config/rendered/configmap.yaml
+++ b/charts/opentelemetry-collector/examples/alternate-config/rendered/configmap.yaml
@@ -6,7 +6,7 @@ metadata:
   name: example-opentelemetry-collector
   namespace: default
   labels:
-    helm.sh/chart: opentelemetry-collector-0.151.0
+    helm.sh/chart: opentelemetry-collector-0.152.0
     app.kubernetes.io/name: opentelemetry-collector
     app.kubernetes.io/instance: example
     app.kubernetes.io/version: "0.150.1"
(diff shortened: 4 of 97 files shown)


charts/opentelemetry-collector/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.150.0
+version: 0.151.0
```
dmitryax: @TylerHelmuth I see we bumped this version to be out of sync with the image in https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2050. Was that unexpected or we don't match them?

younsl: I also had some confusion about whether this should be a minor bump or a patch bump when submitting this PR. Happy to adjust the version once the versioning policy is clarified.

TylerHelmuth: Any matching between the chart version and appVersion are coincidental as they represent 2 different things. The helm chart value should change as the chart changes, independent of the appVersion. Typically we bump the `minor` version for appVersion bumps, big features, or breaking changes. We bump the patch version for minor things or bug fixes. Since this PR adds a new feature to the chart I'm ok bumping the minor version.

TylerHelmuth (approved):

younsl:
Ready to merge, thanks!

TylerHelmuth:
@younsl apologies, I just merge a PR for the collector chart that update the appVersion and now there are conflicts.

younsl:
@TylerHelmuth no worries, thanks for the heads up. Rebased onto latest main and resolved the conflicts:

- Chart version: 0.151.0 → 0.152.0 (since 0.151.0 was consumed by #2156)
- appVersion: kept 0.150.1 from #2156
- Regenerated all rendered examples via make generate-examples so the helm.sh/chart labels and image tags reflect the new versions
- terminationGracePeriodSeconds feature changes preserved

PR is MERGEABLE now and CI is re-running. PTAL when you have a moment.

younsl:
@TylerHelmuth Ready to merge, thanks!

PR #2157, opened 2026-04-19
feat(collector): add `resourceDetection` preset for parity with the kube-stack helm chart

cyrille-leclerc:
Here is the PR description:

---

### Goal

Implement the `presets.resourceDetection` feature in the **opentelemetry-collector** chart, bringing it to parity with the equivalent preset already present in the `opentelemetry-kube-stack` chart (see otel-kube-stack PR [here](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1981)).

### Implementation

A new `presets.resourceDetection` block is introduced with a top-level `enabled` flag (default: **`false`** for backward compatibility). When set to `true`, the chart injects a `resourcedetection/env` processor into all configured pipelines (logs, metrics, traces).

#### Supported detectors

| Detector | Enabled by default (when `resourceDetection.enabled: true`) | Purpose |
|---|---|---|
| `env` | ✅ yes | Reads resource attributes from environment variables |
| `k8snode` | ✅ yes | Detects node-level attributes from the Kubernetes API |
| `eks` | ❌ no | Detects EKS cluster name |
| `aks` | ❌ no | Detects AKS cluster name |
| `gcp` | ❌ no | Detects GCP/GKE cluster name |

Each detector can be individually overridden regardless of the top-level `enabled` state.

#### Sample configuration

```yaml
presets:
  # ...

  # Configures the collector to detect resource attributes adding a resourcedetection processor named `resourcedetection/env` to the pipelines.
  # Adds the resourcedetection/env processor to all pipelines.
  # Each detector can be enabled individually. Base detectors `env` and `k8snode` are enabled by default when `resourceDetection/enabled` is `true`. 
  # Cloud provider specific detectors (`eks`, `aks`, `gcp`)
  resourceDetection:
    enabled: false
    env:
      enabled: true
    k8snode:
      enabled: true
    eks:
      enabled: false
    aks:
      enabled: false
    gcp:
      enabled: false
```

### Follow-up

A subsequent PR will add `presets.resourceDetection` to the **opentelemetry-kube-stack** chart, including the `enabled` flag and the `env`/`k8snode` sub-configs. The only difference from this chart will be that `presets.resourceDetection.enabled` defaults to **`true`** there, since the kube-stack chart already ships `resourcedetection/env` (with `env` + `k8snode`) as part of its default daemon collector configuration.


cyrille-leclerc:
@TylerHelmuth @jaronoff97 can you please review? It should be straightforward as it's mostly feature parity with the kube-stack-helm-chart feature you reviewed a few months ago.

charts/opentelemetry-collector/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.153.0
+version: 0.153.1
```
TylerHelmuth: Lets bump the minor version for this feature.

charts/opentelemetry-collector/ci/preset-resourcedetection-values.yaml
```diff
@@ -0,0 +1,13 @@
...
+image:
+  repository: "ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector-k8s"
+
+command:
+  name: "otelcol-k8s"
+
+presets:
+  resourceDetection:
+    enabled: true
+    eks:
```
TylerHelmuth:
```suggestion
    env:
```

cyrille-leclerc: `env` is activated by default when `resourceDetection/enabled: true`.

I was interested in activating a cloud provider detector to verify that the combination of detectors: `env`, `k8snode`, and the cloud detector (eg `aks`, `eks`, or `gcp`) was successful.
Shall I only do this check in `examples/daemonset-resource-detection` ?

povilasv: Looks like CI is passing on this one https://github.com/open-telemetry/opentelemetry-helm-charts/actions/runs/25698764437/job/75453698318?pr=2157#step:4:3317

But I am not really sure it make sense to test this in ci since we run in kind cluster. Maybe better k8snode?

TylerHelmuth: ya it wont make sense to enable any of the cloud providers

TylerHelmuth (approved):

charts/opentelemetry-collector/Chart.yaml
cyrille-leclerc: done

cyrille-leclerc:
@TylerHelmuth FYI the build is fixed

PR #2160, opened 2026-04-22
fix(otel-collector): default k8snode resource detection on on-prem K8s [CDS-2925]

natnayr:
## Summary

Restores Coralogix Infra Catalog support on on-prem Kubernetes (e.g. EKSA) when users set `provider: on-prem`. Without this, the resource catalog cannot resolve pod-to-node relationships because the chart was emitting no \`k8s.node.*\` resource attributes for on-prem deployments.

Per Slack thread [CDS-2925](https://coralogix-dev.slack.com/archives/C042QPPJM0W/p1776762571953659) (EKSA customer report).

## Background

PR #411 (v0.129.2) made the resource detection processors provider-aware. For \`provider: on-prem\` it set the cloud detector list to empty in two places, which:

- Skipped the entire \`resourcedetection/resource_catalog\` processor (gated by \`if gt (len $catalogDetectors) 0\`), so the \`logs/resource_catalog\` pipeline emitted nothing useful for the Infra Catalog.
- Left \`resourcedetection/env\` with only \`[system, env]\`, so \`k8s.node.*\` attributes never landed on logs/metrics/traces either.

\`K8S_NODE_NAME\` is already injected via the Downward API in [\`templates/_pod.tpl\`](https://github.com/coralogix/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/templates/_pod.tpl) for non-ECS / non-standalone / non-macOS distributions, so the \`k8snode\` detector works out of the box without further plumbing.

## Changes

\`charts/opentelemetry-collector/templates/_config.tpl\`:

- \`opentelemetry-collector.kubernetesResourcesConfig\`: for \`provider: on-prem\` on Kubernetes, default \`$catalogDetectors\` to \`[k8snode]\` (was \`[]\`). Add a \`k8snode: { node_from_env_var: K8S_NODE_NAME }\` block under \`resourcedetection/resource_catalog\` when that detector is in use.
- \`opentelemetry-collector.resourceDetectionConfig\`: introduce \`$defaultEnvDetectors\` — \`[system, env]\` by default, \`[env, k8snode, system]\` for on-prem K8s. Add a sibling \`k8snode:\` config block under \`resourcedetection/env\` when that detector is in the list.

\`charts/opentelemetry-collector/values.yaml\`:

- Drop the hardcoded default \`presets.resourceDetection.detectors.env: [system, env]\` so the template default applies (including the on-prem K8s default). Documented behavior and the \`# env: []\` opt-out in comments.

\`charts/opentelemetry-collector/examples/on-prem-k8s/\`:

- New example covering \`distribution: ""\` + \`provider: on-prem\` + \`kubernetesResources\` + \`kubernetesAttributes\` + \`resourceDetection\`. Locks in the new defaults via rendered fixtures.

\`Chart.yaml\` / \`CHANGELOG.md\`:

- Bump chart from 0.130.12 → 0.130.13 with a CHANGELOG entry noting the fix and the v0.129.2 regression.

All other touched files are re-rendered fixtures from \`make generate-examples\` (chart version label bumps; the on-prem-k8s example is the only behavior change).

## Backward compatibility

- Users who set \`presets.resourceDetection.detectors.env\` or \`.cloud\` explicitly retain full control. Defaults only change when those values are unset.
- For non-K8s on-prem (\`distribution: standalone\` / \`ecs\` / \`macos\`) behavior is unchanged.

## Test plan

- [x] \`helm lint charts/opentelemetry-collector\` — passes
- [x] \`make check-examples CHARTS=opentelemetry-collector\` — all 60 examples diff-clean against \`make generate-examples\` output
- [x] \`./charts/opentelemetry-collector/validate-configs.sh\` — \`on-prem-k8s\` validates; only failures are pre-existing/environmental (\`standalone-systemd\` missing receiver in bundled binary, \`standalone-windows\` macOS sandbox \`hostmetrics\` issue)
- [x] On-prem-k8s rendered output verified to contain:
  - \`processors.resourcedetection/resource_catalog.detectors == [k8snode]\` with \`k8snode.node_from_env_var: K8S_NODE_NAME\`
  - \`processors.resourcedetection/env.detectors == [env, k8snode, system]\` with sibling \`k8snode:\` block
  - \`logs/resource_catalog\` pipeline includes \`resourcedetection/resource_catalog\`
  - \`K8S_NODE_NAME\` env var injected on the deployment container

## Out of scope

- No change in \`telemetry-shippers/otel-integration\`. The bot will pick up \`0.130.13\` on the next scheduled run via \`bump-otel-collector-version.sh\`.

Made with [Cursor](https://cursor.com)


natnayr:
Wrong target repo - reopening against the Coralogix fork.

PR #2170, opened 2026-04-30
Fix VPA updateMode YAML parsing by adding quotes

DaStru:
The VPA templates render updateMode without quotes, causing it to be interpreted as a boolean in YAML 1.1. This breaks deployments on clusters with strict VPA webhook validation (like GKE) with the error:

`admission webhook "gke-vpa.k8s.io" denied the request: json: cannot unmarshal bool into Go struct field PodUpdatePolicy.spec.updatePolicy.updateMode of type v1.UpdateMode`                              
                                                                                                                                                                                                        
The issue occurs when users set `updateMode: "Off"` in values - it gets rendered as unquoted `updateMode: Off`, which YAML interprets as the boolean `false`. The VPA API expects a string enum value.
                                                                                                                                                                                                    
This fix adds the `| quote` Helm filter to ensure the value is always rendered as a properly quoted string.

### DIFF
--- a/charts/opentelemetry-operator/Chart.yaml
+++ b/charts/opentelemetry-operator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-operator
-version: 0.112.0
+version: 0.112.1
 description: OpenTelemetry Operator Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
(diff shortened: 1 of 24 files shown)


TylerHelmuth:
please bump the helm chart's patch version and run `make generate-examples CHARTS=opentelemetry-operator`

DaStru:
Done 👍

TylerHelmuth (approved):

PR #2172, opened 2026-04-30
docs(operator): add ArgoCD ignoreDifferences workaround for cert recreation

### LINKED ISSUE #2168: opentelemetry-operator generated certificate is always recreated
I am using the `opentelemetry-operator` helm chart with argo-cd.
In order to generate the certificate with the helm chart I configured the following
```
admissionWebhooks:
  certManager:
    enabled: false
  autoGenerateCert:
    enabled: true
    recreate: false
```
Which works the first apply, but is always shown as a diff in argo-cd, and therefore will be recreated in the next apply.

tmohanvamsi:
## Summary
When using the opentelemetry-operator chart with ArgoCD, the webhook certificate secret shows a diff on every sync. This happens because ArgoCD uses `helm template` (dry-run mode) which does not support the Helm `lookup` function.

## Root Cause
ArgoCD renders Helm charts using dry-run mode. The `lookup` function always returns empty in this mode, causing a new certificate to be generated on every sync — even when `autoGenerateCert.recreate: false` is set.

## Fix
Added a new **'Using with ArgoCD'** section to the README documenting the recommended `ignoreDifferences` configuration to prevent certificate diffs on every ArgoCD sync.

## Testing
Documentation-only change. Verified YAML examples are correct.

## Related Issue
Fixes #2168


tmohanvamsi:
Updated the PR based on feedback from @ben-dov on the issue.

The `helm.sh/resource-policy: keep` approach does not fully solve 
the problem since the cert is regenerated at `helm template` time 
regardless of annotations.

The updated fix takes a documentation approach instead — added a 
new 'Using with ArgoCD' section to the README explaining how to 
configure `ignoreDifferences` in the ArgoCD Application to ignore 
cert diffs on every sync.

This is the ArgoCD-native recommended approach for handling 
resources that change on every Helm render.

tmohanvamsi:
Updated the README with the complete solution based on feedback from @vands56 on the issue.

The fix now includes:
1. Setting `secretAnnotations` to null out the helm hook annotations and add `helm.sh/resource-policy: keep`
2. Configuring ArgoCD `ignoreDifferences` with `RespectIgnoreDifferences=true`

The previous version only covered `ignoreDifferences` which alone is not sufficient.

cc @ben-dov @vands56 — would appreciate confirmation this works in your environment!

vands56:
Hi @tmohanvamsi,

for me it works for sure. Would just add that if you set "keep" flag by condition in [operator-webhook.yaml](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2172/changes#diff-8dfa10d09328363cdd12981018c7b1e8bc27d545fe07c274122506e7057de4ed), then you dont need it in secretAnnotations section explicitly.
And would be great to have a bash one liner or copy/paste script to verify caBundle status on both webhooks and CRDs and sync those to the latest state of generated ca.crt. On clean run you don't need it, but if certs were regenerated after you added ignoreDifferences or if cert was deleted and re-created, then it will be handy

Something like:
```
#!/bin/bash

NS="opentelemetry-operator-system"
MWC_NAME=$(kubectl get mutatingwebhookconfiguration -o name | grep "opentelemetry-operator-mutation" | head -n 1)
VWC_NAME=$(kubectl get validatingwebhookconfiguration -o name | grep "opentelemetry-operator-validation" | head -n 1)
CRD_NAME=$(kubectl get crd -o name | grep "opentelemetrycollectors.opentelemetry.io")

CA=$(kubectl get <secret-name> -n $NS -o jsonpath='{.data.ca\.crt}')

kubectl patch $MWC_NAME --type='json' -p="[{\"op\": \"replace\", \"path\": \"/webhooks/0/clientConfig/caBundle\", \"value\":\"$CA\"}]"

kubectl patch $VWC_NAME --type='json' -p="[{\"op\": \"replace\", \"path\": \"/webhooks/0/clientConfig/caBundle\", \"value\":\"$CA\"}]"

kubectl patch $CRD_NAME --type='json' -p="[{\"op\": \"replace\", \"path\": \"/spec/conversion/webhook/clientConfig/caBundle\", \"value\":\"$CA\"}]"

```

tmohanvamsi:
Updated the PR again based on feedback from @vands56:

1. Removed `helm.sh/resource-policy: keep` from `secretAnnotations` — it is already handled conditionally in `operator-webhook.yaml`
2. Added optional bash script to resync caBundle on webhooks and CRDs after cert regeneration (adapted from @vands56 community contribution)

Thanks @vands56 for the detailed feedback and the script!

Allex1:
@tmohanvamsi can you please follow the [instructions](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/CONTRIBUTING.md) so we can attempt to merge this ?

marcschaeferger:
@tmohanvamsi Are you still working on this?

PR #2194, opened 2026-05-12
feat(target-allocator): add namespaceOverride value

zeningc:
## Summary

- Adds a top-level `namespaceOverride` value to the `opentelemetry-target-allocator` chart, mirroring the existing parameter in the `opentelemetry-collector` chart.
- When set, `helper.namespace` uses it in place of `.Release.Namespace`, so every namespaced resource the chart renders (`ServiceAccount`, `ConfigMap`, `Service`, `Deployment` / `StatefulSet`, and the `ClusterRoleBinding` subject namespace) lands in the configured namespace.
- Default value is `""`, which preserves current behavior (`.Release.Namespace` is used). No existing example renders change beyond the chart version label.

## Motivation

The Target Allocator chart is increasingly consumed as a Helm subchart (alongside `opentelemetry-collector` or as part of larger umbrella charts). In that scenario the parent chart often deploys into a release namespace that differs from where the TA's resources should actually live (for example, the parent installs into the user's app namespace, but the operator-style components — including the TA — belong in a dedicated `observability` / `monitoring` namespace).

Today the only way to relocate the TA's resources is to install it as its own release. That breaks the "single Helm install" UX. Adding `namespaceOverride` makes the umbrella-chart use case work out of the box:

```yaml
# Parent values.yaml
opentelemetry-target-allocator:
  namespaceOverride: observability
```

This is the same pattern `opentelemetry-collector` already exposes (`namespaceOverride: ""` at the top level of its values.yaml), so behavior is consistent across the two charts when they are composed together.

## Changes

- `values.yaml`: new `namespaceOverride: ""` field with comment.
- `values.schema.json`: new `namespaceOverride` string property.
- `templates/_helpers.tpl`: `helper.namespace` now resolves to `.Values.namespaceOverride` when non-empty, else `.Release.Namespace`.
- `Chart.yaml`: minor version bump `0.127.3` → `0.128.0`.
- `examples/*/rendered/*`: regenerated via `make generate-examples CHARTS=opentelemetry-target-allocator` — the only diff is the chart-version label.

## Test plan

- [x] `make generate-examples CHARTS=opentelemetry-target-allocator` regenerates cleanly.
- [x] `make check-examples CHARTS=opentelemetry-target-allocator` passes for all 7 examples (provide-custom-volumes, statefulset-with-pvc, consistent-hashing, existing-configmap, existing-service-account, collector-fleet, prometheus-scrape-config).
- [x] `helm template t charts/opentelemetry-target-allocator --namespace default --set namespaceOverride=observability` — every resource's `metadata.namespace` and the `ClusterRoleBinding` subject namespace render as `observability`. `collector_namespace` in the TA ConfigMap is intentionally left bound to `.Release.Namespace` (it controls where the TA *looks for collectors*, not where its own resources are deployed — overriding it would be a behavior change unrelated to this PR).
- [x] `helm template t charts/opentelemetry-target-allocator --namespace mynamespace` (no override) — unchanged: every resource renders into `mynamespace`.
- [x] Verified as a subchart: created a parent chart with `opentelemetry-target-allocator` as a dependency and set `<alias>.namespaceOverride: observability` in the parent's `values.yaml`; all TA resources rendered into `observability` while the parent's release namespace remained `default`.

### DIFF
--- a/charts/opentelemetry-target-allocator/Chart.yaml
+++ b/charts/opentelemetry-target-allocator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-target-allocator
-version: 0.127.3
+version: 0.127.4
 description: OpenTelemetry Target Allocator Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-target-allocator/values.yaml
+++ b/charts/opentelemetry-target-allocator/values.yaml
@@ -15,6 +15,10 @@ mode: "deployment"
 nameOverride: ""
 # fullnameOverride completely replaces the generated name.
 fullnameOverride: ""
+# Specify which namespace should be used to deploy the resources into.
+# Useful when running this chart as a subchart whose resources need to live
+# in a namespace different from the parent's release namespace.
+namespaceOverride: ""
 
 # Number of replicas for the target allocator
 replicaCount: 1
--- a/charts/opentelemetry-target-allocator/values.yaml
+++ b/charts/opentelemetry-target-allocator/values.yaml
@@ -16,8 +16,6 @@ nameOverride: ""
 # fullnameOverride completely replaces the generated name.
 fullnameOverride: ""
 # Specify which namespace should be used to deploy the resources into.
-# Useful when running this chart as a subchart whose resources need to live
-# in a namespace different from the parent's release namespace.
 namespaceOverride: ""
 
 # Number of replicas for the target allocator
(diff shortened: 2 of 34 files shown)


charts/opentelemetry-target-allocator/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-target-allocator
-version: 0.127.3
+version: 0.128.0
```
povilasv:
```suggestion
version: 0.127.4
```

zeningc: Thanks for the review! Updated to `0.127.4`.

charts/opentelemetry-target-allocator/values.yaml
```diff
@@ -15,6 +15,10 @@ mode: "deployment"
 nameOverride: ""
 # fullnameOverride completely replaces the generated name.
 fullnameOverride: ""
+# Specify which namespace should be used to deploy the resources into.
+# Useful when running this chart as a subchart whose resources need to live
+# in a namespace different from the parent's release namespace.
```
TylerHelmuth:
```suggestion
```

TylerHelmuth (approved):

PR #2196, opened 2026-05-12
[collector] Add kubernetesObjects preset for k8sobjects receiver

kangyili:
Currently, if users want to use k8sobjects to collect Kubernetes manifests as logs, they need to manually list all resources and add the required cluster roles, which is tedious.

This PR simplifies that process by adding a new preset that configures the k8sobjects receiver to collect Kubernetes objects as logs, with grouped resource categories and configurable collection modes.

The preset supports:
- Resource groups: core, rbac, storage,
  networking, autoscaling, policy, apiExtensions
- Full compatibility with existing kubernetesEvents preset
  (objects lists are merged when both are enabled)


kangyili:
I’ll bump the chart and regenerate the example once this PR is approved to avoid unnecessary merge conflicts.

charts/opentelemetry-collector/values.yaml
```diff
@@ -72,6 +72,50 @@ presets:
...
+  # Adds the k8sobjects receiver to the logs pipeline
+  # and adds the necessary rules to ClusterRole.
+  # Best used with mode = deployment or statefulset.
+  # Compatible with kubernetesEvents preset - objects lists are merged when both are enabled.
+  kubernetesObjects:
+    enabled: false
+    # Collection mode: "pull", "watch", or "pullWatch"
+    mode: pull
+    # Interval for pull mode collection. If not set, the k8sobjects receiver default is used.
+    pullInterval: ""
```
TylerHelmuth: We prefer presets to be incredibly opinionated. For this reason we have a general rule (broken only by logsCollection for legacy reasons) that no part of the preset should take user input; it should all be on/off toggles. In this way the user doesn;t have to know anything about the component's configuration in order to get started.

The bellow toggles are good. We need to take a strong stance on the `mode` and the `pullInterval` and we should remove the `extraResources` option.

kangyili: Thanks for the feedback! I’ve addressed the comment https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2196/changes/57c528126cea1e69fad8fa5e135c8d3e3c38f9e3

kangyili: I was a bit torn about `pullInterval`.

In the end, I removed the per-object interval from the preset entirely and now just rely on the k8sobjects receiver’s default (1h). https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2196/changes/49bdc00126d721d8a94c2707d4b3b7d1f853ce91

The main reason is that if users want to use the preset but not the `pullInterval` value defined by the preset, they currently have to manually redefine every object just to override that single value, which feels pretty painful:
```
k8s_objects:                                                                                                                                                                                                          
    objects:
      - name: pods
        mode: pull
        interval: 15m
      - name: nodes                                                                                                                                                                                                       
        mode: pull
        interval: 15m                                                                                                                                                                                                     
      - name: deployments                                                                                                                                                                                               
        mode: pull
        interval: 15m
.......

```

To make this easier for advanced users, I opened an PR that adds a top-level interval field to the k8sobjects receiver. https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48452 
Once merged, users will be able to override the interval with a single `config.receivers.k8sobjects.interval` field, while still benefiting from the preset simplifying the rest of the configuration.
```
presets:
  kubernetesObjects:
    enabled: true

config:
  receivers:
    k8sobjects:
      interval: 3m #override default value defined by the preset
```

TylerHelmuth: Ya default lists are annoying in helm. I think relying on the component's default makes sense for the preset. We can move this work forward independent of your component feature add.

charts/opentelemetry-collector/values.yaml
```diff
@@ -72,6 +72,36 @@ presets:
   # See https://opentelemetry.io/docs/kubernetes/collector/components/#kubernetes-objects-receiver for details on the receiver.
   kubernetesEvents:
     enabled: false
+  # Collects Kubernetes objects via the k8sobjects receiver in pull and watch modes.
```
TylerHelmuth: This needs updated. Also lets add a comment explain how often things will get pulled.

charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -495,6 +498,143 @@ receivers:
           - "DELETED"
 {{- end }}
 
+{{- define "opentelemetry-collector.applyKubernetesObjectsConfig" -}}
+{{- /* Generate the new objects list from preset configuration */}}
```
TylerHelmuth: Please remove the comments (I know AI likes to add them all but it just makes the tpl file hard to read when interspersed)

charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -495,6 +498,143 @@ receivers:
...
+{{- $existingObjects := list }}
+{{- if .config.receivers }}
+{{- if index .config.receivers "k8sobjects" }}
+{{- if (index .config.receivers "k8sobjects").objects }}
+{{- $existingObjects = (index .config.receivers "k8sobjects").objects }}
+{{- end }}
+{{- end }}
+{{- end }}
+{{- /* Combine new and existing objects */}}
+{{- $allObjects := concat $newObjects $existingObjects }}
```
TylerHelmuth, line 529: Lets not worry about merging object lists yet. I think we could consider this in the future, but it is a lot of complexity for a preset. We typically want the helm chart to behave in expected ways and the default behavior would be for a `config.receivers.k8sobjects.opbjects` list to overwrite any preset values. Thats definitely annoying sometimes, but predictable, and we want to start with that predictability.

charts/opentelemetry-collector/values.yaml
kangyili: I added 1 hour as the default interval in the comment https://github.com/kangyili/opentelemetry-helm-charts/blob/b602b7a8afd9f2b4767dcb7d09ecc18cf498d2f7/charts/opentelemetry-collector/values.yaml#L75

> pull and watch modes.

This part is correct because this preset enables both pull and watch modes for resources, so user can have periodic reconciliation + real time k8s manifest updates.

https://github.com/kangyili/opentelemetry-helm-charts/blob/b602b7a8afd9f2b4767dcb7d09ecc18cf498d2f7/charts/opentelemetry-collector/templates/_config.tpl#L525-L530

charts/opentelemetry-collector/templates/_config.tpl
kangyili: Unfortunately, we have to keep the merge logic because `kubernetesEvents` also modifies the k8sobjects receiver. https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/templates/_config.tpl#L493-L502  Otherwise this new preset would conflict with the `kubernetesEvents` preset.

TylerHelmuth: Hmmmmm. I wonder if the kubernetesEvents present needs switch to the actual `k8s_events` receiver now that we've decided to keep it. Or maybe we should deprecate it in favor of the k8sobjects preset.

For now, can you experiment with merging only when both presets are enabled? We have precedence for special merge logic when a preset is enabled with the k8sattributes presets.

kangyili: > For now, can you experiment with merging only when both presets are enabled?

Yes: https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2196/commits/2ce4795ecbd791f976f2227eab9280277557e585

I think it actually makes the code more complex than merging unconditionally. 🤔 

Also, with this change, if users enable `kubernetesObjects` preset and want to collect additional custom resources via `config.receivers.k8sobjects.objects`, they would have to relist all the resources already defined by the preset, because we replaces arrays instead of appending to them.

So unconditional merging feels more extensible to me. But I’m happy to go with the overwrite logic if the community prefers it.

TylerHelmuth: Ok lets keep the original merging logic. I'll work on a plan for the future of the kuberentesEvents receiver separately.

kangyili: Sounds good, thanks for the help! I’ve reverted the last commit to keep the original merging logic.

charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -501,6 +504,123 @@ receivers:
...
+{{- $config := mustMergeOverwrite (dict "service" (dict "pipelines" (dict "logs" (dict "receivers" list)))) $objectsYaml .config }}
+{{- $_ := set (index $config.receivers "k8sobjects") "objects" $allObjects }}
+{{- $_ := set $config.service.pipelines.logs "receivers" (append $config.service.pipelines.logs.receivers "k8sobjects" | uniq) }}
+{{- $config | toYaml }}
+{{- end }}
+
+{{- define "opentelemetry-collector.kubernetesObjectsConfig" -}}
+{{- $preset := .Values.presets.kubernetesObjects -}}
+receivers:
+  k8sobjects:
```
TylerHelmuth, line 546: Does this component support leader election? If so we should automatically configure leader election when in daemonset mode. The cluster metrics and kuberentes events preset support this.

kangyili: Nice catch. I followed the cluster metrics pattern. https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2196/changes/78c3ceada72598936b9d320d6b4d97c336f40fac

TylerHelmuth (commented):
Please add a new test for this preset in the `ci` folder

charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -501,6 +507,141 @@ receivers:
...
+    k8s_leader_elector: k8s_leader_elector/{{ .electorName }}
+    {{- end }}
+    objects:
+{{- if $preset.core.enabled }}
+{{- range list "namespaces" "pods" "nodes" "services" "serviceaccounts" }}
+      - name: {{ . }}
+        mode: pull
+      - name: {{ . }}
+        mode: watch
+{{- end }}
```
TylerHelmuth, line 558: What does it mean to pull and watch at the same time? Wont that product a lot of duplicate data?

kangyili: The reasoning was mentioned here: https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2196#discussion_r3260650272

The default pull interval is 1 hour because pull mode is mainly there for periodic reconciliation. But with such a long interval, we can miss a lot of k8s resource state changes, for example deletion. So we also enable watch mode and it lets us receive k8s resources updates in real time, while pull mode still acts as a periodic reconciliation mechanism to resync the full state.

kangyili: That was also why I originally introduced this option so the preset could support different user use cases.
https://github.com/kangyili/opentelemetry-helm-charts/blob/9bf4249386ecc585665b20687051c37fc37428f2/charts/opentelemetry-collector/values.yaml#L82-L83

But we later removed it because we want the preset to become opinionated.

TylerHelmuth: I've always used this component pretty preciously when it came to what to watch and what to pull. I've always used `watch` for events (out of scope for this PR) and watch for everything else. As a user of this component, are you typically adding a watch to objects like `pods` to try and get a realtime stream of that object?

I'm guessing we could get a lot of value from making the preset Pull all the objects and decreasing the interval to something like 5 or 10 minutes. How would that align with how you use the receiver?

TylerHelmuth: @povilasv do you have opinions here?

kangyili: I mainly use watch for deletion events. As for deletions, the final state of the resource can be missed no matter how small we make the pull interval. and in my case I use pull mode mainly to re-list resources every 3 minutes.

If that works for you, I can use `exclude_watch_type` to exclude all other event types except `DELETED` in watch https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/k8sobjectsreceiver/README.md?plain=1#L62 So we will have less traffic.

Regarding the pull interval, I think we can let it falls back to the default value. I opened a [PR](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48452) in the collector to let users override the default value while still using the preset in helm chart. 
Otherwise, if we explicitly set a specific interval in the preset without exposing a way to override it, users who need a different value would have to re-list all resources just to change that one setting, which is annoying.

kangyili: Or we could make pull the default behavior and add an option like `include_watch: false` in the preset. Users who need both pull and watch could then enable it.

TylerHelmuth: >Otherwise, if we explicitly set a specific interval in the preset without exposing a way to override it, users who need a different value would have to re-list all resources just to change that one setting, which is annoying.

No matter what interval we choose to have the preset use this will be a problem with the current receiver. We should have the preset use the interval that is generally most useful for the most number of users. If that's the default 1 hour, thats fine with me. 

I like the idea of an additional `watch: false` option at the top-level or for all the different sections. Thats simple configuration that could empower users but keeps the default preset from producing duplicate data.

kangyili: Sounds good.  I think we can start with the top-level option to keep preset simple https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2196/changes/491406eb345526c16ca1340ca77c89801217532f

> or for all the different sections

This can be easily added later if needed

TylerHelmuth (approved):

kangyili:
Thanks for the approval. Feel free to merge it @TylerHelmuth

TylerHelmuth:
@kangyili sorry I merged something else first. please bump the chart version again and regenerate the examples

povilasv (approved):

povilasv:
Would be nice to have some example with this enabled

kangyili:
Sure, the example is added https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2196/changes/f3a3cea71c7d47f7144c84e75fd510ab5f9a1fe6

PR #2200, opened 2026-05-14
[chore]: update stale action to allow running manually

TylerHelmuth:


.github/workflows/stale.yaml
```diff
@@ -22,3 +23,4 @@ jobs:
           days-before-issue-stale: 730
           days-before-pr-close: 14
           days-before-issue-close: 30
+          operations-per-run: 1000
```
TylerHelmuth, line 26: We were hitting operation number limits before which results in a run not being able to check everything. We ran into this in Collector Contrib, the easiest solution is to let the job do more. Contrib has a limit of 6000, 1000 should be enough for this repo.

povilasv (approved):

PR #2203, opened 2026-05-18
[kube-stack] add prometheus option to `presets/annotationDiscovery/metrics`

cyrille-leclerc:
## Summary

Adds a new `presets.annotationDiscovery.metrics.prometheus.enabled` toggle to the `opentelemetry-kube-stack` chart. When enabled, the `receiver_creator/metrics` is configured with a `prometheus_simple/prometheus_annotations` receiver that discovers and scrapes pods carrying classic Prometheus annotations:

- `prometheus.io/scrape=true` (required to opt in)
- `prometheus.io/port` (defaults to `9090`)
- `prometheus.io/path` (defaults to `/metrics`)

The discovery rule excludes the opentelemetry-collector pods (label `app.kubernetes.io/component=opentelemetry-collector`) to avoid self-scrape loops.

This complements the existing OpenTelemetry-native annotation discovery (`io.opentelemetry.discovery.metrics/*`) by also supporting workloads that already expose metrics via the classic Prometheus annotation convention.

It's disabled by default for backward compatibility.

UX

```yaml
    presets:
      annotationDiscovery:
        metrics:
          enabled: true
          prometheus:
            enabled: true
        logs:
          enabled: ...
     #...
```
## Changes

- `values.yaml`: new `prometheus.enabled: false` sub-option under `presets.annotationDiscovery.metrics` in both `defaultCRConfig.presets` and `collectors.daemon.presets`
- `values.schema.json`: schema entry for the new option
- `templates/_config.tpl`: emit the `prometheus_simple/prometheus_annotations` receiver in `receiver_creator/metrics` when the toggle is on
- `Chart.yaml`: bump version `0.14.13` -> `0.14.14`
- `examples/daemonset-annotation-discovery-metrics`: enable the new option in the example and regenerate rendered manifests

## Test plan

- [ ] `make generate-examples CHARTS=opentelemetry-kube-stack` succeeds with no diff
- [ ] `helm template` with `prometheus.enabled: true` emits the `prometheus_simple/prometheus_annotations` receiver under `receiver_creator/metrics`
- [ ] `helm template` with `prometheus.enabled: false` (default) produces no diff vs. previous behavior (only chart-label version change)
- [ ] CI lint (`ct lint`) passes
- [ ] Manual smoke test in a kind cluster: a pod with `prometheus.io/scrape=true` is discovered and scraped; pods without the annotation are not; opentelemetry-collector pods are not self-scraped

### DIFF
--- a/charts/opentelemetry-kube-stack/Chart.yaml
+++ b/charts/opentelemetry-kube-stack/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-kube-stack
-version: 0.14.13
+version: 0.14.14
 description: |
   OpenTelemetry Quickstart chart for Kubernetes.
   Installs an operator and collector for an easy way to get started with Kubernetes observability.
--- a/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/bridge.yaml
+++ b/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/bridge.yaml
@@ -5,7 +5,7 @@ kind: OpAMPBridge
 metadata:
   name: example
   labels:
-    helm.sh/chart: opentelemetry-kube-stack-0.14.13
+    helm.sh/chart: opentelemetry-kube-stack-0.14.14
     app.kubernetes.io/version: "0.144.0"
     app.kubernetes.io/managed-by: Helm
     release: "example"    
--- a/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/collector.yaml
+++ b/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/collector.yaml
@@ -6,7 +6,7 @@ metadata:
   name: example-daemon
   namespace: default
   labels:
-    helm.sh/chart: opentelemetry-kube-stack-0.14.13
+    helm.sh/chart: opentelemetry-kube-stack-0.14.14
     app.kubernetes.io/version: "0.144.0"
     app.kubernetes.io/managed-by: Helm
     release: "example"        
--- a/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/hooks.yaml
+++ b/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/hooks.yaml
@@ -61,4 +61,4 @@ spec:
           - "delete"
           - "instrumentations,opampbridges,opentelemetrycollectors"
           - "-l"
-          - "helm.sh/chart=opentelemetry-kube-stack-0.14.13"
+          - "helm.sh/chart=opentelemetry-kube-stack-0.14.14"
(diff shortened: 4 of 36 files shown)


jaronoff97:
@cyrille-leclerc IIRC we already have this enabled for the daemon collector in the prometheus receiver via the daemon_scrape_configs, i don't think we need to add this into the receiver creator, i would prefer we just default to the prometheus receiver scrape config.

cyrille-leclerc:
While `daemon_scrape_configs.yaml` has been a great way to bridge OTel with Prometheus-based Kubernetes infrastructure monitoring, I think an `annotationDiscovery` preset is a better long-term direction. A few reasons:

**Limitations of `daemon_scrape_configs.yaml`:**

- **Generates false-positive warnings** when the Prometheus Node Exporter isn't deployed. For example:
  ```
  2026-05-18T15:48:07.347Z    warn    internal/transaction.go:127    Failed to scrape Prometheus endpoint    {"resource": {"host.name": "k3d-k3s-default-server-0", "k8s.namespace.name": "opentelemetry-operator-system", "k8s.node.ip": "172.18.0.3", "k8s.node.name": "k3d-k3s-default-server-0", "k8s.pod.ip": "10.42.0.10", "k8s.pod.name": "opentelemetry-stack-daemon-collector-vrxht", "service.instance.id": "3e49ef16-0189-4046-9ce5-965f20188cee", "service.name": "otelcol-contrib", "service.version": "0.152.0"}, "otelcol.component.id": "prometheus", "otelcol.component.kind": "receiver", "otelcol.signal": "metrics", "scrape_timestamp": 1779119287307, "target_labels": "{__name__=\"up\", instance=\"172.18.0.3:9100\", job=\"node-exporter\"}"}
  ```
- **Hard to grasp for OTel-native users.** Its 175 lines of Prometheus YAML lean heavily on Prometheus concepts (`job`, `kubernetes_sd_configs`, `relabel_configs`) that are increasingly unfamiliar to practitioners coming to OTel without a Prometheus background.
- **All-or-nothing granularity** — you opt into Kubernetes Prometheus annotation scraping *and* Node Exporter scraping together, with no way to enable just one.
- **Scoped to the kube-stack chart only**, and unlikely to land in the OpenTelemetry Collector Helm chart as-is.

**Why the Receiver Creator is a better fit:**

- It's an **OpenTelemetry-native** alternative to Prometheus service discovery.
- We can **scope it narrowly** to just scraping pods with Prometheus annotations, without pulling in Node Exporter support and its associated complexity.
- It can be exposed as a preset in **both** the kube-stack chart *and* the standalone collector chart, giving users a consistent experience across deployments.

That's the motivation behind proposing this preset.

WDYT?

jaronoff97:
Please read the comment again. I am suggesting we can just default a prometheus receiver based on the user preferences in the preset. There is no need to use the receiver creator here, it just adds complexity.

cyrille-leclerc:
Replaced by:
- https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2206

cyrille-leclerc:
replaced by:
- https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2206

PR #2206, opened 2026-05-19
feat(kube-stack): add presets.prometheus.{nodeExporter,cadvisor,podAnnotations}

cyrille-leclerc:
## Summary

Adds a new top-level `presets.prometheus.*` namespace to the kube-stack chart — a daemonset-local replacement for `scrape_configs_file: daemon_scrape_configs.yaml`. Three independently-toggleable presets, each adding a dedicated named instance of the prometheus receiver to the daemonset collector's metrics pipeline:

* **`prometheus.nodeExporter`** — scrapes a Prometheus Node Exporter on the local node via `static_configs` targeting `${OTEL_K8S_NODE_IP}:9100`.
* **`prometheus.cadvisor`** — scrapes cAdvisor from the local kubelet at `https://${OTEL_K8S_NODE_IP}:10250/metrics/cadvisor` with bearer-token auth. Drops the high-cardinality / low-value series by default (same drop list as `daemon_scrape_configs.yaml`).
* **`prometheus.podAnnotations`** — scrapes pods on the local node carrying classic `prometheus.io/scrape=true` annotations via `kubernetes_sd_configs` with a node field selector.

Sharded by design: each daemonset pod scrapes only its own node, no central Prometheus bottleneck.

## UX

```yaml
collectors:
  daemon:
    scrape_configs_file: ""   # required when enabling any prometheus/* preset
    presets:
      prometheus:
        nodeExporter:
          enabled: true
        cadvisor:
          enabled: true
        podAnnotations:
          enabled: true
```

Per-preset knobs: `scrapeInterval` (default 30s), `scrapeTimeout` (default 10s), plus `port` for `nodeExporter` / `cadvisor` (`podAnnotations` reads port per-pod from the `prometheus.io/port` annotation).

## Time-series alignment with the `daemon_scrape_configs.yaml` jobs

All three presets produce the same identifying labels as `daemon_scrape_configs.yaml` — `job` / `instance` / Kubernetes label sets, so existing queries, dashboards, alerts, and recording rules relying on these labels keep working unchanged:

| Preset | Maps to `daemon_scrape_configs.yaml` job | Compatibility |
|---|---|---|
| `nodeExporter` | `node-exporter` | Drop-in + bonus `node` label |
| `cadvisor` | `kubelet` (`/metrics/cadvisor`) | Drop-in — `job_name: kubelet` kept for KPS + `daemon_scrape_configs.yaml` parity |
| `podAnnotations` | `kubernetes-pods` | Drop-in + bonus `node` label + collector self-scrape filter |

## Documentation

* `README.md`: new subsection `presets.prometheus.*` under the existing Kube-Prometheus-Stack compatability heading. Covers the constraints, discovery model, per-preset label set, alignment with `daemon_scrape_configs.yaml`, and remaining duplicate-scrape risks (e.g. the prometheus-node-exporter subchart's own ServiceMonitor).
* `values.yaml`: each preset documented inline; constraints called out at the umbrella block.
* `values.schema.json`: full schema entries with descriptions; `scrape_configs_file` description notes the mutual exclusivity.

## Constraints (chart-enforced)

Template-render assertions fail with a clear error if either is violated:

* **Mutually exclusive with `scrape_configs_file`** — the presets are a replacement, not an addition. The default `daemon_scrape_configs.yaml` already has node-exporter, kubelet/cAdvisor, and pod-annotation jobs; running both would cause duplicate scrapes.
* **Require `mode: daemonset`** — the scrape configs use `${OTEL_K8S_NODE_IP}` / `${OTEL_K8S_NODE_NAME}` and node field selectors that only make sense on daemonset collectors.

Errors name the specific presets enabled, e.g. `presets.prometheus.{nodeExporter, cadvisor} are a replacement for scrape_configs_file...`.

## Why a standalone Prometheus receiver rather than receiver_creator?

The standalone Prometheus receiver reuses the native Prometheus discovery & scrape code paths and config, so users can lift-and-shift existing Prometheus jobs and relabel chains. The receiver_creator path is still available via the existing `presets.annotationDiscovery.metrics.enabled`.

## Test plan

- [x] `make generate-examples CHARTS=opentelemetry-kube-stack` produces no further diff.
- [x] `helm lint charts/opentelemetry-kube-stack` clean.
- [x] New example `examples/daemonset-prometheus-presets/` demonstrates enabling all three presets together.
- [x] Validation: `helm template` with `scrape_configs_file` set + a preset enabled fails with the expected error naming the enabled presets.
- [x] Validation: `helm template` with `mode: deployment` + a preset enabled fails with the expected error.
- [x] Render examples and confirm each preset's scrape job has the expected `job_name` (`node-exporter` / `kubelet` / `kubernetes-pods`), `scheme`, `metrics_path`, relabel rules (including `node` label injection), and `metric_relabel_configs`.
- [x] Deploy to a k3d cluster and verify metrics flow through each receiver.


cyrille-leclerc:
@dashpole FYI, as I suggested in https://github.com/open-telemetry/opentelemetry-specification/pull/4956#discussion_r3262174108, we could evolve the `prometheus/addPrometheusLabels` config in this PR so that it not only avoids generating labels such as pod, namespace, etc., but also maps the Prometheus job and instance labels to prometheus.job and prometheus.instance.

This would allow `service.name` and `service.instance.id` to remain unset in the PRometheus receiver layer, so that the `k8sattributes` processor can generate the desired values based on Kubernetes metadata, matching the guidance from SemConv: Specify resource attributes using Kubernetes annotations.
We could temporarily implement this with a transform processor waiting to add the capability in the PRometheus receiver.

cyrille-leclerc:
@jaronoff97 I hope this meets your expectations. The code change is quite small.

cyrille-leclerc:
FYI the build failure looks odd, `make check-examples` on github actions is unhappy with a few line breaks in generated files when there are no problems on my laptop. I have to investigate more.

dashpole:
The latest iteration of the spec PR just keeps job/instance as-is.  Hopefully that simplifies things.

cyrille-leclerc:
@jaronoff97 the build is fixed could you please have a look?

charts/opentelemetry-kube-stack/templates/_config.tpl
```diff
@@ -492,3 +505,178 @@ gcp:
...
+{{- $prom := .collector.presets.prometheus }}
+{{- $enabled := list }}
+{{- if $prom.nodeExporter.enabled }}{{- $enabled = append $enabled "nodeExporter" }}{{- end }}
+{{- if $prom.cadvisor.enabled }}{{- $enabled = append $enabled "cadvisor" }}{{- end }}
+{{- if $prom.podAnnotations.enabled }}{{- $enabled = append $enabled "podAnnotations" }}{{- end }}
+{{- if $enabled }}
+{{- $collectorName := .collector.suffix | default "unnamed" }}
+{{- $enabledList := join ", " $enabled }}
+{{- if .collector.scrape_configs_file }}
+{{- fail (printf "collector %q: presets.prometheus.{%s} are a replacement for `scrape_configs_file` (the chart default is `daemon_scrape_configs.yaml`). Both are configured: scrape_configs_file=%q. Choose one — either keep `scrape_configs_file` and disable the listed presets, or keep the presets and set `scrape_configs_file: \"\"`." $collectorName $enabledList .collector.scrape_configs_file) }}
```
jaronoff97, line 530: though we could probably do some fancy helm things here, i think it's fine to error here.

charts/opentelemetry-kube-stack/README.md
```diff
@@ -34,6 +34,89 @@ This chart provides functionality to port an existing scrape configuration from
 > [!NOTE]
 > This chart aims to provide compatibility for scrape targets from the kube-prometheus-stack chart. This chart is not responsible for applying Prometheus Rules, Alertmanager, or a Prometheus instance.
 
+#### `presets.prometheus.*` vs. kube-prometheus-stack ServiceMonitors
+
+The `presets.prometheus.*` family (`nodeExporter`, `cadvisor`, `podAnnotations`) adds named instances of the prometheus receiver to the daemonset collector's metrics pipeline. They are a DaemonSet-local alternative to the kube-prometheus-stack (KPS) `ServiceMonitor`-based approach and a **replacement** for the `daemon_scrape_configs.yaml` scrape file. The two approaches produce **similar but not identical** time series.
```
jaronoff97: Why are we mentioning service monitors here? We're only replacing the need for the daemon_scrape_configs.yaml no? IMO there are many users who will want want to use the presets you're adding _and_ service monitors which are a very standard prometheus pattern.

charts/opentelemetry-kube-stack/README.md
```diff
@@ -34,6 +34,89 @@ This chart provides functionality to port an existing scrape configuration from
...
+
+##### Constraints (chart-enforced)
```
jaronoff97: can we put all of this new content below into a dropdown? Or at least we should put this in the issue/PR description and remove it from here, this is quite verbose.

charts/opentelemetry-kube-stack/README.md
```diff
@@ -34,6 +34,89 @@ This chart provides functionality to port an existing scrape configuration from
...
+
+##### Other duplicate-scrape risks
+
+* **`prometheus-node-exporter` subchart**: when installed via the top-level `nodeExporter.enabled: true` flag, the subchart creates its own `ServiceMonitor` by default. If a target allocator picks it up, node-exporter is scraped twice — once locally by every daemonset collector and once via the ServiceMonitor through the cluster collector.
+
+##### When to use which
+
+* **Use the presets** if you want a sharded, DaemonSet-native pipeline with no central Prometheus bottleneck and your dashboards rely primarily on `job` / `instance` / `node` (and, for `podAnnotations`, the standard Kubernetes label set).
+* **Use the target allocator + ServiceMonitors** (set `targetAllocator.enabled: true` instead) if you need full KPS label parity (including `service` / `endpoint` / `container`), want CRD-driven dynamic configuration, or are porting existing KPS dashboards/recording rules unchanged.
```
jaronoff97: There are users who will be doing both, and the presets are not set to be mutually exclusive so this inclusion is incorrect and will confuse users.

charts/opentelemetry-kube-stack/README.md
```diff
@@ -34,6 +34,89 @@ This chart provides functionality to port an existing scrape configuration from
...
+
+The `nodeExporter` preset drops them deliberately; this is a feature, not a parity gap. `cadvisor` and `podAnnotations` series, by contrast, *are* per-workload and so emit `namespace`/`pod`/`container` meaningfully.
+
+##### Backward compatibility with `daemon_scrape_configs.yaml`
+
+Each preset replaces a specific scrape job in `daemon_scrape_configs.yaml`. All three are drop-in compatible — `job` / `instance` / Kubernetes label set are preserved, so existing queries, dashboards, alerts, and recording rules keep working. The remaining differences are either additive (a new `node` label) or default-value tweaks called out per preset below.
+
+###### `nodeExporter` ↔ `node-exporter` job
+
+**Drop-in compatible.** Same scrape target (`${OTEL_K8S_NODE_IP}:9100`), same `job=node-exporter` label, same `scrape_interval: 30s`. The preset additionally emits a `node` label (the legacy job didn't) — purely additive, won't break queries.
```
jaronoff97: this will objectively break queries as the label _is_ changing no?

charts/opentelemetry-kube-stack/README.md
```diff
@@ -34,6 +34,89 @@ This chart provides functionality to port an existing scrape configuration from
...
+
+The `nodeExporter` preset drops them deliberately; this is a feature, not a parity gap. `cadvisor` and `podAnnotations` series, by contrast, *are* per-workload and so emit `namespace`/`pod`/`container` meaningfully.
+
+##### Backward compatibility with `daemon_scrape_configs.yaml`
+
+Each preset replaces a specific scrape job in `daemon_scrape_configs.yaml`. All three are drop-in compatible — `job` / `instance` / Kubernetes label set are preserved, so existing queries, dashboards, alerts, and recording rules keep working. The remaining differences are either additive (a new `node` label) or default-value tweaks called out per preset below.
```
jaronoff97: Unless we provide a rewrite path with a transform processor this is decidedly NOT backwards compatible.

jaronoff97 (changes requested):
The docs are incorrect and quite misleading. Please go through them again and correct the errors. The content of the PR's actual helm changes looks good however.

charts/opentelemetry-kube-stack/templates/_config.tpl
cyrille-leclerc: I preferred to be strict to start with as side effects may be hard to anticipate.

charts/opentelemetry-kube-stack/README.md
cyrille-leclerc: Section removed. I just keep a reference to ServiceMonitor for the nodeExporter in the section "risk of duplicates"

charts/opentelemetry-kube-stack/README.md
cyrille-leclerc: Thanks for the suggestion. The details are now hidden by default using collapsable sections.

charts/opentelemetry-kube-stack/README.md
cyrille-leclerc: Section removed, I just kept "Duplicate-scrape risks" that mentions challenges when mixing the scrappings.

charts/opentelemetry-kube-stack/README.md
```diff
@@ -43,12 +134,6 @@ collector.image.tag: 0.103.1
 bridge.image.tag: 0.103.0
 ```
 
-### scrape_configs_file Details
```
cyrille-leclerc, line 46: FYI section moved above, next to "`presets.prometheus.*` presets"

cyrille-leclerc:
Thanks for your feedback @jaronoff97. The PR is ready for another review.

I have cleaned up the docs following your recommendations.
I kept the documentation of the differences between metrics captured by `daemon_scrape_configs.yaml` and by `presets.prometheus.*` but I stop talking of compatibility.

jaronoff97 (approved):
much better, thank you very much!

jaronoff97:
@dashpole or @TylerHelmuth mind adding your review in here when you get a chance?

charts/opentelemetry-kube-stack/README.md
```diff
@@ -34,6 +34,97 @@ This chart provides functionality to port an existing scrape configuration from
 > [!NOTE]
 > This chart aims to provide compatibility for scrape targets from the kube-prometheus-stack chart. This chart is not responsible for applying Prometheus Rules, Alertmanager, or a Prometheus instance.
 
+#### `presets.prometheus.*` presets
+
+The `presets.prometheus.*` family (`nodeExporter`, `cadvisor`, `podAnnotations`) configure the OpenTelemetry Collectors to scrape popular Prometheus Kubernetes metrics:
+
+* `presets.prometheus.nodeExporter`: Kubernetes node metrics exposed with the [prometheus-node-exporter](https://github.com/prometheus-community/helm-charts/tree/main/charts/prometheus-node-exporter),
+* `presets.prometheus.cadvisor`: [cAdvisor](https://github.com/google/cadvisor) metrics exposed with the [Kube State Metrics](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-state-metrics) (aka KSM)
```
dashpole: What does it mean for cadvisor metrics to be "exposed with the KSM"?

dashpole:
Looks correct overall to me.  I don't have much experience with helm, though

charts/opentelemetry-kube-stack/README.md
cyrille-leclerc:
```suggestion
* `presets.prometheus.cadvisor`: [cAdvisor](https://github.com/google/cadvisor) metrics exposed with the [Prometheus Kube State Metrics Helm Chart](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-state-metrics) (aka KSM Helm Chart) 
```

@dashpole To clarify, I was referring to the Prometheus Community's KSM Helm Chart, which is the setup I tested against. Does my proposed change make more sense in that context?

dashpole: Why would we need to use the KSM chart for cAdvisor metrics?

cyrille-leclerc: I'm not deeply familiar with this area 👼  I was mostly following the pattern used by the OTel Kube Stack Helm Chart here:

https://github.com/open-telemetry/opentelemetry-helm-charts/blob/c7dc0383503231e8001cdc6d4f5a2d0a6e3b0853/charts/opentelemetry-kube-stack/Chart.yaml#L31-L34

dashpole: I took a deeper look.  You aren't actually using KSM in any capacity from what I can tell, so I would just say something like "cAdvisor metrics scraped from the kubelet".

cyrille-leclerc:
```suggestion
* `presets.prometheus.cadvisor`: [cAdvisor](https://github.com/google/cadvisor) metrics scraped from the kubelet,
```

Thank you so much for your thorough review!

PR #2215, opened 2026-05-21
ci: Add CI support for Helm v4

### LINKED ISSUE #2212: [chore] Add CI support for Helm v4
We've started getting issues opened for different annoyances with how the charts work with helm v4:

- https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1966
- https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2152#issuecomment-4498058212

It's time we add some explicit CI support/testing for the latest helm version in addition to helm v3, which we should continue to support for a while.

vanshiz:
## Summary
This PR adds the helm-version strategy matrix to all CI test workflows so that it tests both versions  v3  and v4 (v4.0.3)

## Changes
- setup action — bumped default from v3.9.4 to v3.14.4
- All 7 workflow files — added strategy.matrix with helm-version: [v3.14.4, v4.0.3] and pass ${{ matrix.helm-version }} to the setup step

## Why
Now that Helm v4 is out, users are hitting issues with the charts . This way we catch regressions against both major versions in CI instead of finding out after release.

This Fixes #2212


.github/actions/setup/action.yaml
```diff
@@ -8,7 +8,7 @@ inputs:
   helm-version:
     description: "The helm version to be used when setting up"
     required: false
-    default: "v3.9.4"
+    default: "v3.14.4"
```
TylerHelmuth: We shouldn't need to change the default.

.github/workflows/collector-test.yaml
```diff
@@ -13,6 +13,11 @@ permissions:
 jobs:
   collector-test:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        helm-version:
+          - "v3.14.4"
```
TylerHelmuth: We should keep testing the collector helm chart at 3.9 as long as it keeps working

.github/workflows/ebpf-test.yaml
```diff
@@ -13,6 +13,11 @@ permissions:
 jobs:
   ebpf-test:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        helm-version:
+          - "v3.14.4"
```
TylerHelmuth: Same here

.github/workflows/kube-stack-test.yaml
```diff
@@ -13,6 +13,11 @@ permissions:
 jobs:
   opentelemetry-kube-stack-test:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        helm-version:
+          - "v3.14.4"
```
TylerHelmuth:
```suggestion
          - "v3.11.3"
```

.github/workflows/operator-test.yaml
```diff
@@ -13,6 +13,11 @@ permissions:
 jobs:
   operator-test:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        helm-version:
+          - "v3.14.4"
```
TylerHelmuth: Same here

.github/workflows/operator-test.yaml
```diff
@@ -79,6 +85,11 @@ jobs:
 
   operator-test-no-certmanager:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        helm-version:
+          - "v3.14.4"
```
TylerHelmuth: Same here

.github/workflows/ta-test.yaml
```diff
@@ -13,6 +13,11 @@ permissions:
 jobs:
   ta-test:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        helm-version:
+          - "v3.14.4"
```
TylerHelmuth: Same here

.github/actions/setup/action.yaml
vanshiz: Done, reverted to v3.9.4

.github/workflows/collector-test.yaml
vanshiz: Done, reverted to v3.9.4

.github/workflows/ebpf-test.yaml
vanshiz: Done

.github/workflows/kube-stack-test.yaml
vanshiz: Changed to v3.11.3

.github/workflows/operator-test.yaml
vanshiz: Done, reverted to v3.9.4

.github/workflows/operator-test.yaml
vanshiz: Done

.github/workflows/ta-test.yaml
vanshiz: Done

TylerHelmuth (approved):

TylerHelmuth:
@vanshiz in order for this to pass CI we need to do the strategy that Collector Contrib uses with all of its matrix jobs in `build-and-test`. The matrix jobs should be run in their own step, and then we need an additional step, named `lint-test`, that passes or fails based on the matrix. This lets us avoid changing our branch merge rules when the matrix changes. See https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/08fa7b253d480f5f0dd145f9aa2f4c379f2b3114/.github/workflows/build-and-test.yml#L76-L91

vanshiz:
Hey @TylerHelmuth, are the changes correct now?

PR #2218, opened 2026-05-22
fix(target-allocator): use Downward API for OTELCOL_NAMESPACE so namespaceOverride works

sfc-gh-zeningchen:
## Problem

The target-allocator subchart sets the pod's `OTELCOL_NAMESPACE` env from `.Release.Namespace` (or the optional `targetAllocator.config.collector_namespace` value):

```yaml
# charts/opentelemetry-target-allocator/templates/_pod.tpl
env:
  - name: OTELCOL_NAMESPACE
    value: {{ .Values.targetAllocator.config.collector_namespace | default .Release.Namespace }}
```

When the chart is consumed as a subchart and installed into a different namespace via `namespaceOverride` (added in #2194), every other resource (Deployment, Service, ConfigMap, ServiceAccount, ClusterRoleBinding subject) honours the override via `helper.namespace`, but the pod itself is still told to watch `.Release.Namespace`.

## What this causes

`OTELCOL_NAMESPACE` is read into `Config.CollectorNamespace` (`cmd/otel-allocator/internal/config/config.go`) and used to scope the TA's collector pod informer. When that env points at a namespace where no collectors are running, the watcher returns zero pods regardless of `collector_selector`. As a result, on installs that use `namespaceOverride`:

- `opentelemetry_allocator_collectors_discovered` and `opentelemetry_allocator_collectors_allocatable` stay at 0
- `/jobs/<job>/targets?collector_id=<pod>` returns `[]` for every collector
- targets get discovered but never assigned

Setting `targetAllocator.config.collector_namespace` to the same value as `namespaceOverride` works around it, but nothing in the chart docs flags the dependency, so it's easy to miss.

## Fix

Source the env from the Downward API instead of templating a literal:

```yaml
env:
  - name: OTELCOL_NAMESPACE
    {{- with .Values.targetAllocator.config.collector_namespace }}
    value: {{ . }}
    {{- else }}
    valueFrom:
      fieldRef:
        fieldPath: metadata.namespace
    {{- end }}
```

A pod always knows what namespace it landed in, regardless of install method (`helm install`, subchart, `helm template | kustomize`, plain `kubectl apply -n <ns>`, or ArgoCD with destination override). `targetAllocator.config.collector_namespace` is preserved as an explicit escape hatch that wins over the fieldRef, so deployments that intentionally watch a different namespace from their own are unaffected.

This matches the precedent already set by `open-telemetry/opentelemetry-operator`:
- TA reconciler: `internal/manifests/targetallocator/container.go`
- Standalone manifest: `config/target-allocator/deployment.yaml`

## Why #2194 missed it

#2194 only touched `_helpers.tpl` to make `helper.namespace` honour `namespaceOverride`. The `_pod.tpl` env block was not modified, and the rendered-example fixtures don't exercise `namespaceOverride` and render with `--namespace default`, so the buggy line happens to produce the right value in CI. It only surfaces on installs with a non-default `.Release.Namespace` or `namespaceOverride`.

## Changes

- `charts/opentelemetry-target-allocator/templates/_pod.tpl`: switch `OTELCOL_NAMESPACE` to `valueFrom.fieldRef: metadata.namespace`, keeping `collector_namespace` as an explicit override.
- `charts/opentelemetry-target-allocator/Chart.yaml`: bump to `0.128.1`.
- Regenerated examples via `make generate-examples CHARTS=opentelemetry-target-allocator`.

## Verification

- `make check-examples CHARTS=opentelemetry-target-allocator` passes.
- `helm lint charts/opentelemetry-target-allocator` clean.
- `helm template t charts/opentelemetry-target-allocator --namespace default --set namespaceOverride=foo` now renders the fieldRef form.
- `helm template t charts/opentelemetry-target-allocator --set targetAllocator.config.collector_namespace=watch-this-ns` still renders the explicit value (escape hatch preserved).

### DIFF
--- a/charts/opentelemetry-target-allocator/Chart.yaml
+++ b/charts/opentelemetry-target-allocator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-target-allocator
-version: 0.128.0
+version: 0.128.1
 description: OpenTelemetry Target Allocator Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-target-allocator/examples/collector-fleet/rendered/clusterRole.yaml
+++ b/charts/opentelemetry-target-allocator/examples/collector-fleet/rendered/clusterRole.yaml
@@ -6,7 +6,7 @@ metadata:
   namespace: default
   name: example-opentelemetry-target-allocator-ta-clusterRole
   labels:
-    helm.sh/chart: opentelemetry-target-allocator-0.151.0
+    helm.sh/chart: opentelemetry-target-allocator-0.128.1
     app.kubernetes.io/managed-by: Helm
     app.kubernetes.io/part-of: opentelemetry-target-allocator
     app.kubernetes.io/version: "0.151.0"
--- a/charts/opentelemetry-target-allocator/examples/collector-fleet/rendered/clusterRoleBinding.yaml
+++ b/charts/opentelemetry-target-allocator/examples/collector-fleet/rendered/clusterRoleBinding.yaml
@@ -5,7 +5,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: example-opentelemetry-target-allocator-ta-clusterRoleBinding
   labels:
-    helm.sh/chart: opentelemetry-target-allocator-0.151.0
+    helm.sh/chart: opentelemetry-target-allocator-0.128.1
     app.kubernetes.io/managed-by: Helm
     app.kubernetes.io/part-of: opentelemetry-target-allocator
     app.kubernetes.io/version: "0.151.0"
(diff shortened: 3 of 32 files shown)


atoulme:
Change looks ok but the version change seems to be problematic somehow. Please check the version numbers.

sfc-gh-zeningchen:
> Change looks ok but the version change seems to be problematic somehow. Please check the version numbers.

@atoulme Thanks for the review. The chart was at version: 0.128.0 before this PR; this is a non-breaking bug fix so I bumped to 0.128.1. It looks like #2199 initially set the chart version to 0.151.0 (matching the new TA appVersion) and rendered the examples from that, then corrected version to 0.128.0 but didn't regenerate the examples. 
The regen in this PR brings the label back in sync with Chart.yaml.

atoulme (approved):

TylerHelmuth (approved):

Issue #2183, opened 2026-05-06
k8sattributes renamed to k8s_attributes in  opentelemetry-collector 0.146.0


mlec1:
In the release 0.146.0 https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.146.0, the `k8sattributes` processor was deprecated and renamed `k8s_attributes` https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45894.

I think this should be changed in the helm chart too. 
https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-helm-charts%20k8sattributes&type=code

I will try to come with a Merge Request for it.

TylerHelmuth:
Its easy enough to update the preset to use the new name, but the issue is that a lot of people extend the preset capability, so their config will include `k8sattributes` and it wont merge properly anymore. This is gonna be kinda hard to solve.

A few quick ideas I have are:
1. Just do the breaking change and make everyone sad. Put some stuff in NOTES.txt to prevent installs/upgrades if the image is an upstream image with the change and the config includes `k8sattributes`.
2. Start warning whenever we see `k8sattributes` in a config. In some future release do a breaking change.
3. if a config has `k8sattributes`, rewrite the config to be `k8s_attributes`.

mlec1:
I was thinking to do a breaking change. 
How would you warn people like you mentioned point 2 ??

TylerHelmuth:
Something in NOTES.txt to print out a warning whenever `k8sattributes` is found in the config.

skoenig:
I would like to work on this one.

povilasv:
Maybe we can just do this:
>if a config has k8sattributes, rewrite the config to be k8s_attributes.

and some warning that this is deprecated and we automatically rewrite? 


I kinda hate doing this breaking change :/

skoenig:
I'd be fine with not doing the actual breaking change (removing support for `k8sattributes`).
Here's what I'm planning to do: introduce a generic helper for deprecation warnings that could also be used for the other renamings. Then automatically rewrite user config referencing the old name to the new canonical name, so existing user config does not break.

The actual deprecation could be done separately after a grace period, perhaps something like 2-3 months.

TylerHelmuth:
A rename would work. To make it so that the eventual image version change is less painful we could also handle renaming named configs, like `k8sattributes/custom_name`

skoenig:
[...]

PR #2220, opened 2026-05-22
Add useCanonicalNames toggle for upstream-renamed preset components

kangyili:
https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45339 has been renaming several components from their camelCase aliases to snake_case canonical names:
* hostmetrics: host_metrics
* kubeletstats: kubelet_stats
* filelog: file_log
* k8sattributes: k8s_attributes
* k8sobjects: k8s_objects.

The chart's presets still inject these under the deprecated aliases. So a user who enables a preset and references the canonical name in their own config,  expecting it to override or extend the preset:
```
  presets:
    hostMetrics:
      enabled: true
  config:
    receivers:
      host_metrics:
        collection_interval: 30s
```
  …ends up with two separate receiver instances in the rendered collector config:
```
  receivers:
    host_metrics:          # user's override
      collection_interval: 30s
    hostmetrics:           # injected by preset
      collection_interval: 10s
  service:
    pipelines:
      metrics:
        receivers:
        - host_metrics
        - otlp
        - hostmetrics      # both names in the pipeline
```

Both receivers run,  telemetry is silently doubled. The user's "override" doesn't override anything because the names don't match.

This PR adds an opt-in toggle:

```
  presets:
    useCanonicalNames: true   # default false
    hostMetrics:
      enabled: true
```

Default false and existing users unaffected. With true, the affected presets inject under the canonical name, so .Values.config overrides line up and emit a single instance instead of two.

### DIFF
--- a/charts/opentelemetry-collector/ci/preset-canonical-names-values.yaml
+++ b/charts/opentelemetry-collector/ci/preset-canonical-names-values.yaml
@@ -1,0 +1,17 @@
+mode: daemonset
+image:
+  repository: ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector-k8s
+presets:
+  useCanonicalNames: true
+  hostMetrics:
+    enabled: true
+  kubeletMetrics:
+    enabled: true
+  logsCollection:
+    enabled: true
+  kubernetesAttributes:
+    enabled: true
+  kubernetesEvents:
+    enabled: true
+  kubernetesObjects:
+    enabled: true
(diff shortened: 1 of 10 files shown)


TylerHelmuth:
We have an issue tracking this FYI: https://github.com/open-telemetry/opentelemetry-helm-charts/issues/2183. I'd like to scope this done to one PR per preset.

kangyili:
Thanks, I missed that issue. I see there’s another [PR](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2225) to fix it. if that’s the preferred way, I can close this one

PR #2221, opened 2026-05-23
New chart for auto-instrumentation

atoulme:
Not ready for merge yet - this is a chart that takes auto-instrumentation from the operator and makes it work without CRDs.

This will be ready to merge once operator 0.152.0 is released.


atoulme:
Closing per #2223

PR #2222, opened 2026-05-23
feat(kubernetesEvents): add useK8sEventsReceiver feature flag

### LINKED ISSUE #2209: [opentelemetry-collector]: Update kuberenetsEvents preset to use the k8s_events receiver
## Description

The Collector SIG decided not to deprecate the k8s_events receiver (https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/24242). As a result, the k8s_events receiver has become the proper semconv-compliant component for kubernetes events in OpenTelemetry. We need to transition the preset to use the component.

My initial thoughts for the transition are:
- Add a k8s objects receiver preset: https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2196
- Add an additional option to that preset (off by default), to collect k8s events like the `kubernetesEvents` preset does today
- Add a root-level feature flag ([like we did for memory_ballast extension removal](https://github.com/open-telemetry/opentelemetry-helm-charts/issues/891)) that controls the switch, defaulting to false. When true, the `kubernetesEvents` preset uses the `k8s_event` receiver.
- After a few releases, switch the flag to be true by default.
- After a few releases, remove the flag.

If users prefer the k8sobject receiver's collection of this data, they can always enable the option in the k8sobjects preset.

vanshiz:
Closes #2209
## Changes
- Added `useK8sEventsReceiver` root-level feature flag (default: `false`)
- When `true`, the `kubernetesEvents` preset uses the `k8s_events` receiver 
  instead of `k8sobjects`
- Updated RBAC in `clusterrole.yaml` to use correct API group per receiver
- Added `events.enabled` option to `kubernetesObjects` preset for users 
  who prefer collecting events via k8sobjects
- Added CI test for the new flag
- Updated `values.schema.json`
## Migration path
Per the plan in #2209, this flag defaults to `false`. It will default to 
`true` in a future release, then be removed entirely.

### DIFF
--- a/charts/opentelemetry-collector/values.yaml
+++ b/charts/opentelemetry-collector/values.yaml
@@ -106,6 +106,9 @@ presets:
     # API extensions resources: customresourcedefinitions
     apiExtensions:
       enabled: true
+
+    events:
+      enabled: false
   # Configures the Kubernetes Cluster Receiver to collect cluster-level metrics.
   # Adds the k8s_cluster receiver to the metrics pipeline
   # and adds the necessary rules to ClusterRole.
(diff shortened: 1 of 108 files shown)


charts/opentelemetry-collector/values.yaml
```diff
@@ -455,6 +458,11 @@ ports:
 # It is HIGHLY recommend to enable this setting and set a value for resources.limits.memory.
 useGOMEMLIMIT: true
 
+# When enabled, the kubernetesEvents preset uses the dedicated k8s_events receiver
+# instead of the k8sobjects receiver. This flag defaults to false and will default
+# to true in a future release before being removed.
+useK8sEventsReceiver: false
```
TylerHelmuth, line 464: Lets nest this under the kubernetesEvents preset.

PR #2234, opened 2026-05-27
ci: enable helm chart provenance and GPG signing

### LINKED ISSUE #2167: Helm verify Provenance/Integrity
I have a feature request for this repo.

Enable helm verify to work.

Would it be possible to sign the helm charts to confirm their authenticity.
https://helm.sh/docs/topics/provenance/

I would like to build a automation with Terraform that uses opentelemetry with the helm provider and since this will be in deep automation I would prefer to check the authenticity of the chart the automation is using.

suyogkumawat:
## Description
This PR enables Helm chart provenance (`helm verify`) by integrating GPG signing into the release workflow. 

Added a step to prepare the GPG keyring from repository secrets and passed the required `CR_SIGN`, `CR_KEY`, `CR_KEYRING`, and `CR_PASSPHRASE_FILE` variables to the `chart-releaser-action`.

Fixes #2167


TylerHelmuth:
@suyogkumawat thanks for opening this PR. I just learned OTel does support Environment secrets, so I've opened a PR to create an environment for us to use instead of repo-level secrets. I'll let you know what that is in place, as you'll need to make changes to this PR to reference the environment secrets instead.

TylerHelmuth:
@suyogkumawat thank you so much for being patient. I have finally added the `release` environment. It now contains:

 - `GPG_PRIVATE_KEY`
 - `GPG_PASSPHRASE`
 - `GPG_KEY_ID`
 - `GPG_KEY_NAME`

suyogkumawat:
Thanks @TylerHelmuth  — really appreciate you setting up the `release` environment! 

I've pushed an update that incorporates your feedback:

* Added `environment: release` to the release job so `GPG_PRIVATE_KEY`, `GPG_PASSPHRASE`, `GPG_KEY_ID`, and `GPG_KEY_NAME` are sourced from the GitHub Environment rather than repository-level secrets.
* Hardened the GPG setup while I was in there:

  * Secrets are passed via the step's `env:` block instead of being referenced directly in the script.
  * GPG operations use an isolated `GNUPGHOME` under `RUNNER_TEMP`.
  * The exported keyring is scoped to `GPG_KEY_ID` and used only for chart signing.

Are there any approval requirements or branch protection rules configured on the `release` environment that I should be aware of before the first signed release runs on `main`?

Let me know if you'd like any other adjustments. Thanks again!

.github/workflows/release.yaml
```diff
@@ -1,82 +1,95 @@
...
+          mkdir -p -m 700 "${GNUPGHOME}"
 
+          printf '%s\n' "${GPG_PRIVATE_KEY}" | gpg --batch --import
+          gpg --batch --export-secret-keys "${GPG_KEY_ID}" > "${keyring}"
+          printf '%s' "${GPG_PASSPHRASE}" > "${passphrase_file}"
+
+          chmod 600 "${keyring}" "${passphrase_file}"
       - name: Run chart-releaser
-        uses: helm/chart-releaser-action@cae68fefc6b5f367a0275617c9f83181ba54714f # v1.7.0
+        uses: helm/chart-releaser-action@cae68fefc6b5f367a0275617c9f83181ba54714f
```
povilasv: lets keep the comment with version

povilasv: and in rest of places please :)

.github/workflows/release.yaml
```diff
@@ -1,82 +1,95 @@
 name: Release Charts
-
```
TylerHelmuth, line 2: Can you add these newlines back? They help make the file easy to read

.github/workflows/release.yaml
```diff
@@ -1,82 +1,95 @@
...
       - name: Add dependent repositories
         run: |
           helm repo add open-telemetry https://open-telemetry.github.io/opentelemetry-helm-charts
           helm repo add prometheus https://prometheus-community.github.io/helm-charts
           helm repo add grafana https://grafana-community.github.io/helm-charts
           helm repo add jaeger https://jaegertracing.github.io/helm-charts
           helm repo add opensearch https://opensearch-project.github.io/helm-charts
+      # Import the GPG key into an isolated GnuPG home, then export only the
+      # requested key into a binary keyring file for chart-releaser to be used.
```
TylerHelmuth:
```suggestion
```

.github/workflows/release.yaml
```diff
@@ -1,82 +1,95 @@
...
+      - name: Prepare GPG keyring
```
TylerHelmuth: Is it possible for us to instead duplicate the logic of the prometheus community charts? https://github.com/prometheus-community/helm-charts/blob/22cdf40b432a99426de14b1e235a2350e123265d/.github/workflows/release.yaml#L44-L67. I'd feel more confident that it is going to work first try

.github/workflows/release.yaml
suyogkumawat: Updated with the versions

.github/workflows/release.yaml
suyogkumawat: Sure, used similar configuration setup for GPG

.github/workflows/release.yaml
suyogkumawat: Updated, Please check now

suyogkumawat:
Hi @TylerHelmuth @povilasv , Thanks for your valuable feedback. Updated the release file. Please review and let me know for any further changes.

.github/workflows/release.yaml
```diff
@@ -39,6 +40,29 @@ jobs:
...
+          echo "pinentry-mode loopback" >> "$HOME/.gnupg/gpg.conf"
+          echo "allow-loopback-pinentry" > "$HOME/.gnupg/gpg-agent.conf"
+          echo "max-cache-ttl 86400" >> "$HOME/.gnupg/gpg-agent.conf"
+          echo "default-cache-ttl 86400" >> "$HOME/.gnupg/gpg-agent.conf"
+          gpgconf --kill gpg-agent
+          gpgconf --launch gpg-agent
+          gpg --batch --yes --passphrase-fd 0 --import "$GPG_KEY_PATH" <<< "$GPG_PASSPHRASE"
+          gpg --batch --yes --export "$GPG_KEY_ID" > ~/.gnupg/pubring.gpg
+          gpg --batch --yes --passphrase-fd 0 --export-secret-keys "$GPG_KEY_ID" > ~/.gnupg/secring.gpg <<< "$GPG_PASSPHRASE"
+          cat > "$GPG_PASSPHRASE_FILE" <<< "$GPG_PASSPHRASE"
```
TylerHelmuth, line 58: The prom community charts do `mkdir "${{ runner.temp }}/.gnupg/"` before these `gpg --batch` command. Do we need that? https://github.com/prometheus-community/helm-charts/blob/22cdf40b432a99426de14b1e235a2350e123265d/.github/workflows/release.yaml#L58

.github/workflows/release.yaml
```diff
@@ -39,6 +40,29 @@ jobs:
...
+        env:
+          GPG_KEY_ID: "${{ secrets.GPG_KEY_ID }}"
```
TylerHelmuth: GPG_KEY_ID is a variable in the `release` environment.

.github/workflows/release.yaml
```diff
@@ -15,6 +15,7 @@ jobs:
       packages: write # needed for ghcr access
```
TylerHelmuth, line 15: the prom community chart has `id-token: write` permissions. Can you research if we'll need that? https://github.com/prometheus-community/helm-charts/blob/22cdf40b432a99426de14b1e235a2350e123265d/.github/workflows/release.yaml#L15

.github/workflows/release.yaml
```diff
@@ -47,6 +71,10 @@ jobs:
         env:
           CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
           CR_GENERATE_RELEASE_NOTES: true
+          CR_SIGN: true
+          CR_KEY: "${{ secrets.GPG_KEY_NAME }}"
```
TylerHelmuth: GPG_KEY_NAME is a variable in the `release` environment.

.github/workflows/release.yaml
```diff
@@ -81,4 +109,4 @@ jobs:
         run: .github/scripts/enhance-release-notes.sh
         env:
           GH_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
-          GITHUB_REPOSITORY: "${{ github.repository }}"
+          GITHUB_REPOSITORY: "${{ github.repository }}"
```
TylerHelmuth:
```suggestion
          GITHUB_REPOSITORY: "${{ github.repository }}"

```

.github/workflows/release.yaml
suyogkumawat: Thanks for clarification, I will update it.

.github/workflows/release.yaml
suyogkumawat: Hi @TylerHelmuth ,
GnuPG uses $HOME/.gnupg (created via mkdir -p "$HOME/.gnupg"), and all imports/exports operate against ~/.gnupg/*. GNUPGHOME is never set to ${{ runner.temp }}/.gnupg, so that directory is never referenced. As a result, I left it out since it doesn't appear to be functionally required.

.github/workflows/release.yaml
suyogkumawat: Updated

.github/workflows/release.yaml
suyogkumawat: id-token: write permission is required when a workflow needs an OIDC token. This workflow performs GPG-based provenance signing using the configured private key and generates a .prov file directly, so no OIDC token is requested as part of the signing flow.

Given that, I added the permission for future if we use keyless cosign signing for OCI artifacts.

.github/workflows/release.yaml
suyogkumawat: Added new line

suyogkumawat:
Thanks @TylerHelmuth , Updated the code, please review it.

.github/workflows/release.yaml
suyogkumawat: Added new line

suyogkumawat (commented):
Added newline

.github/workflows/release.yaml
TylerHelmuth: @suyogkumawat if we're not using it today lets leave it out. Thank you for explaining why it is used for in the prom charts.

suyogkumawat: Sure, I will remove it right away.

suyogkumawat:
@TylerHelmuth , Please review it now.

povilasv (approved):

TylerHelmuth (approved):
🤞

PR #2237, opened 2026-05-28
[chore]: update all CI to helm 4+

TylerHelmuth:
Helm 3 and helm 4 render different, and it is causing problems in CI and when users render examples. This updates all CI to helm 4

### DIFF
--- a/charts/opentelemetry-demo/values.yaml
+++ b/charts/opentelemetry-demo/values.yaml
@@ -953,13 +953,13 @@ opentelemetry-collector:
         tls:
           insecure: true
         sending_queue:
-          batch:
+          batch: {}
       otlphttp/prometheus:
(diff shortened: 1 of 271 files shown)


jaronoff97 (approved):

charts/opentelemetry-demo/values.yaml
```diff
@@ -953,13 +953,13 @@ opentelemetry-collector:
...
         sending_queue:
-          batch:
+          batch: {}
       otlphttp/prometheus:
         endpoint: http://prometheus:9090/api/v1/otlp
         tls:
           insecure: true
         sending_queue:
-          batch:
+          batch: {}
```
TylerHelmuth, line 962: The collector treats this the same and setting this prevents 4.0.x and 4.2.x from rendering differently

jaronoff97 (approved):

PR #2238, opened 2026-05-30
Modify the deprecated fields of the configmap for the automatic configuration of logsCollection and kubernetesAttributes

rouize:
Modify the deprecated fields of the configmap for the automatic configuration of logsCollection and kubernetesAttributes.
Including k8s_attributes and file_log.


povilasv:
We are woking on a different approach https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2242 so this is not needed

TylerHelmuth:
@rouize we have https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2242 to handle the k8s_attributes rename, but feel free to open a PR for file_log following the same strategy

PR #2239, opened 2026-05-31
feat(operator): Add support for the `operator.clusterobservability` feature gate

### LINKED ISSUE #2240: [opentelemetry-operator]: Add support for the `operator.clusterobservability` feature gate
Currently, it is not possible to enable the `ClusterObservability` CRD and its respective feature gate for the operator through the helm chart.

olipinski:
## Overview

This PR adds support for the `operator.clusterobservability` and its associated CRD.

It also adds the necessary RBAC permissions.

All features enabled by this PR are conditional on the feature gate being enabled. If any of these permissions should be enabled by default, I am happy to make the changes.

## Linked issues
Closes #2240


charts/opentelemetry-operator/conf/crds/crd-opentelemetry.io_clusterobservabilities.yaml
TylerHelmuth, line 1: The CRDs are managed automatically by our make file. Is this a new unreleased CRD?

olipinski: The CRD is available in the operator, but relatively new: https://github.com/open-telemetry/opentelemetry-operator/blob/8c88634d35066518894bfc4f2ff9abf7e60debde/config/crd/bases/opentelemetry.io_clusterobservabilities.yaml .

Since it is not the same bundle path as the other CRDs, should this one be added to the Makefile? Or should it not be added at all to the chart yet?

TylerHelmuth: Where does it live in https://github.com/open-telemetry/opentelemetry-operator/tree/main/bundle ?

olipinski: As far as I can tell it is not referenced in the bundle. However, it does seem to work as intended with this PR. I have the `ClusterObservability` deployed locally, and it does report metrics back.

If the CRD should be in the bundle, then I am happy to shelve this PR until it is released properly.

TylerHelmuth: @olipinski I'll reach out to the Operator maintainers for clarification.

frzifus: The new CRD+Controller are highly experimental. If we add it to the helm chart, we should make that super clear. We may add/remove fields, rename the CRD, change permission on the operator, add other CRDs like the `ClusterObservabilityPatch`, or maybe deprecate the experiment.

TylerHelmuth: It would be good if the helm chart and support people trying the experiment as long as we can explain the risks. @olipinski we'd need to be able to easily update the operator version in the future and bring along and changes to this CRD, so we're gonna need it managed via the make target somehow.

olipinski: Happy to implement the make target management @TylerHelmuth!

Before I start, should we keep the `ClusterObservability` CRD as optional, so that it is only be deployed when the feature gate is enabled? This is the current behaviour, but keeping this will require bigger modifications to the Makefile, to handle inserting the conditional.

As a quick draft, this seems to work to get the CRDs from the repo. So the main thing left would be to implement handling the feature gate behaviour.

```bash
$(call get-crd,./charts/opentelemetry-operator/conf/crds/crd-opentelemetry.io_clusterobservabilities.yaml,https://raw.githubusercontent.com/open-telemetry/opentelemetry-operator/v$(OPERATOR_APP_VERSION)/config/crd/bases/opentelemetry.io_clusterobservabilities.yaml)
```

frzifus: I am not super familiar with the helm chart, but from a user perspective I would recommend to add a `installCRDClusterObservability=false` flag that needs to be manually enabled.

Btw not sure if that helps as a reference, but I am running the ClusterObservability CR on my homelab:
https://github.com/frzifus/hmlb/blob/e7146dddc612987167a52aa0c72c29bd29edfd6d/clusters/homelab/core/observability/operator.yaml#L39-L71

olipinski: Thanks @frzifus, @TylerHelmuth! I have added the updating of the CRD to the Makefile.

This way the CRD update will also render the conditional block, so the CRD **will not be deployed** unless the feature gate is turned on. I think this should cover the `installCRDClusterObservability=false` flag use-case.

Looking at @frzifus setup for the permissions, would it also make sense/be cleaner to separate the ClusterRole RBAC into a single conditional block, rather than co-opting the blocks where permissions matched already?

charts/opentelemetry-operator/conf/crds/crd-opentelemetry.io_clusterobservabilities.yaml
```diff
@@ -0,0 +1,183 @@
+{{- if .Values.crds.create }}
+{{- if or
+  (get .Values.manager.featureGatesMap "operator.clusterobservability")
+  (and
+    (ne .Values.manager.featureGates "")
+    (regexMatch "\\+?operator\\.clusterobservability" .Values.manager.featureGates)
+  )
```
TylerHelmuth: Im not sure what this extra condition is. I was expecting that checking the value returned from `get .Values.manager.featureGatesMap "operator.clusterobservability"` as the boolean `true` would be enough

TylerHelmuth (commented):
@olipinski left a question about the condition we're using otherwise things look good.

charts/opentelemetry-operator/conf/crds/crd-opentelemetry.io_clusterobservabilities.yaml
olipinski: This second condition is for compatibility with the other (deprecated) way of setting feature gates with `FeatureGates` in `values.yaml` (as far as I understand):

https://github.com/open-telemetry/opentelemetry-helm-charts/blob/b7326a18fba6df36c0ec81e9e62909395d209b8a/charts/opentelemetry-operator/values.yaml#L97-L105

 Since the old way uses a string of feature gates it needed a regex match to see if it was defined in this string. If this deprecated way shouldn't be supported, we could remove this additional check.

TylerHelmuth: Ah yes, I totally forgot we had that. Thanks for clarifying.

TylerHelmuth:
@olipinski checkout the conflicts and then I think we're good

olipinski:
@TylerHelmuth Thanks! Rebased onto main and bumped the chart version. Should be no more conflicts.

TylerHelmuth:
@olipinski sorry again, the examples need regenerated

olipinski:
@TylerHelmuth no worries! updated the examples.

TylerHelmuth:
~~@olipinski please remove the ebpf example updates~~ nevermind

TylerHelmuth (approved):

TylerHelmuth:
@olipinski I didn't merge any other operator commits recently, looks like you didn't have the latest before updating.

olipinski:
@TylerHelmuth it's https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2271 that conflicted with this. I have now removed the deprecated handling as well.

PR #2246, opened 2026-06-04
[opentelemetry-collector]: add Gateway API HTTPRoute support

somaz94:
Adds optional Gateway API `HTTPRoute` support to the `opentelemetry-collector` chart, as an alternative to `Ingress` for exposing the collector's HTTP-based receivers.

The new `httproute` block is disabled by default, so existing users are unaffected. It mirrors the existing `ingress` conventions (labels, namespace, `tpl`-ed annotations) and exposes the standard `HTTPRoute` fields:

- `enabled` (default `false`)
- `apiVersion` (defaults to `gateway.networking.k8s.io/v1`)
- `annotations`, `parentRefs`, `hostnames`
- `rules` — `matches` + `backendRefs` (+ optional `filters`); `backendRefs[].name` defaults to the collector service when omitted, matching the ingress backend convention

`values.schema.json` is extended with the matching `httproute` schema (the chart enforces `additionalProperties: false`). Version bumped `0.158.0` -> `0.159.0` and rendered examples regenerated.

Validated locally:

- `helm lint` and `ct lint --charts charts/opentelemetry-collector` (all `ci/*` value sets)
- `values.schema.json` rejects unknown keys under `httproute`
- `helm template` for: disabled (no output), enabled with `parentRefs`/`hostnames`/`rules`, `backendRefs` name auto-defaulting, and `apiVersion` override

`httproute.enabled: true` is intentionally not added to `ci/` because the collector test job's kind cluster does not install the Gateway API CRDs (only the operator test job does), so `ct install` would fail to create the resource. The disabled default keeps lint coverage.


somaz94:
Friendly ping — this has been open for about a week. CLA is signed and CI is green. Happy to address any feedback whenever a maintainer has a moment to take a look. Thanks!

charts/opentelemetry-collector/templates/httproute.yaml
```diff
@@ -0,0 +1,47 @@
...
+  labels:
+    {{- include "opentelemetry-collector.labels" . | nindent 4 }}
+    {{- include "opentelemetry-collector.component" . | nindent 4 }}
+  {{- if $route.annotations }}
+  annotations:
+    {{- range $key, $value := $route.annotations }}
+    {{- printf "%s: %s" $key (tpl $value $ | quote) | nindent 4 }}
+    {{- end }}
+  {{- end }}
+spec:
```
TylerHelmuth, line 17: I wonder if it would be better to expose in the values.yaml a `.Values.httproute.spec` section and let the user configure the whole thing and we use it directly.

somaz94:
Good idea. A passthrough `.Values.httproute.spec` keeps the chart from mirroring every Gateway API field, and it lines up with how the collector config is already handled here.

The one bit I'd keep is defaulting the backendRef to the collector service, so users don't have to hand-write the generated service name and port (that's the main thing the current template does for them). So: render `.Values.httproute.spec` as-is, but fill in backendRefs from the chart's service when they aren't set. Does that split work for you, or would you rather keep it a pure passthrough?

TylerHelmuth:
I missed that you're setting a default backend. In that case managing it like you have it is best.

somaz94:
Thanks for confirming the approach — I kept the default backendRef behavior as you blessed.

The `lint-test` failure was `ct lint`'s version-bump check (the chart files changed without a `Chart.yaml` bump). I bumped the chart version `0.159.0 → 0.160.0` (the new HTTPRoute template is a feature). `helm lint` is green locally, so CI should go green now. No README or rendered-example changes were needed.

TylerHelmuth (approved):

PR #2258, opened 2026-06-11
[opentelemetry-collector]: honor schedulerName in daemonset and statefulset modes

somaz94:
The `schedulerName` value is documented in values.yaml and wired into the Deployment pod spec, but it was never applied in `daemonset` or `statefulset` mode — setting it there had no effect.

This mirrors the existing deployment.yaml block into daemonset.yaml and statefulset.yaml so a custom scheduler is honored in all three modes.

Verified with `helm template --set schedulerName=...` for each mode (field now renders; absent by default), plus `ct lint` and `make check-examples`.

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.159.0
+version: 0.159.1
 description: OpenTelemetry Collector Helm chart for Kubernetes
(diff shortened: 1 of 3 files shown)


TylerHelmuth:
@somaz94 these changes look right. Please handle merge conflicts.

somaz94:
@TylerHelmuth rebased onto `main` and resolved the conflicts — the PR is mergeable again. Ready for re-review, thanks!

TylerHelmuth (approved):

PR #2273, opened 2026-06-19
[opentelemetry-collector] Force string type for telemetry resource attributes

### LINKED ISSUE #2250: [opentelemetry-collector] Pod fails to start when node name is digits only

```yaml
resource:
  k8s.node.name: 'null'
  host.name: 'null'
```

kshitizj03:
#### Description
The collector resolves `${env:...}` placeholders after parsing the config and infers the type of the substituted value. When a node's name is all digits, `host.name` and `k8s.node.name` get coerced to integers, so the collector fails to start with:
```service::telemetry::resource: legacy resource attribute "host.name" must be string or null```
This renders `service.telemetry.resource` attribute values with an explicit `!!str` YAML tag so the collector always treats them as strings, regardless of the node name. Non-string values (e.g. `null`) are left untouched, so the existing `null` workaround still works.

Bumped the chart version and regenerated the examples.
#### Link to tracking issue
Fixes #2250 

#### Authorship

### DIFF
--- a/charts/opentelemetry-collector/templates/_config.tpl
+++ b/charts/opentelemetry-collector/templates/_config.tpl
@@ -86,6 +86,24 @@ logs:
 {{- end }}
 
 {{/*
+Serialize the collector config to YAML, forcing service.telemetry.resource
+attribute values to be strings. This prevents the collector from coercing
+numeric-looking values (such as an all-digit node name) to integers, which
+would fail validation for string-typed attributes like host.name.
+*/}}
+{{- define "opentelemetry-collector.serializedConfig" -}}
+{{- $config := .config }}
+{{- $resource := dig "service" "telemetry" "resource" dict $config }}
+{{- range $key, $value := $resource }}
+{{- if and (kindIs "string" $value) (ne $value "") }}
+{{- $_ := set $resource $key (printf "!!str %s" $value) }}
+{{- end }}
+{{- end }}
+{{- $rendered := tpl (toYaml $config) .ctx }}
+{{- regexReplaceAll "'(!!str [^']*)'" $rendered "${1}" }}
+{{- end }}
+
+{{/*
 Build config file for daemonset OpenTelemetry Collector
 */}}
 {{- define "opentelemetry-collector.daemonsetConfig" -}}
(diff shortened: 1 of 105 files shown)

### CODE REVIEW

File: charts/opentelemetry-collector/templates/_config.tpl [hunk 1]
```diff
@@ -85,6 +85,24 @@ logs:
 {{- end }}
 {{- end }}
 
+{{/*
+Serialize the collector config to YAML, forcing service.telemetry.resource
+attribute values to be strings. This prevents the collector from coercing
+numeric-looking values (such as an all-digit node name) to integers, which
+would fail validation for string-typed attributes like host.name.
+*/}}
+{{- define "opentelemetry-collector.serializedConfig" -}}
```
TylerHelmuth, line 94: This seems like a bit of a brute force. Is there a default value that is causing this issue that we can wrap in `""`? Or is the issue from your values.yaml?

kshitizj03: It's actually our default `values.yaml`, not just my config:

```yaml
service:
  telemetry:
    resource:
      k8s.node.name: "${env:OTEL_K8S_NODE_NAME}"
      host.name: "${env:OTEL_K8S_NODE_NAME}"
```

so digit-only node names break out of the box. Wrapping in `""` won't fix it: the collector expands `${env:...}` after parsing and re-infers the type, so only the inline `!!str` tag forces a string. And we can't put `!!str` directly in `values.yaml` because `toYaml` quotes it into inert text, which is why the helper re-tags the `resource` values and strips those quotes. The regex only matches the `'!!str ...'` strings this loop adds, so nothing else is touched.

Happy to narrow it to just `host.name`/`k8s.node.name` if the generic pass feels too broad.

TylerHelmuth: In that case I think we can get by with passing all the values through `quote` right?

kshitizj03: Thanks! I actually tested this against collector 0.154/0.155 before changing anything. `quote` renders `host.name: "${env:OTEL_K8S_NODE_NAME}"`, but YAML strips the quotes at parse time, so the collector still sees a bare `${env:...}` and re-infers the type after expansion. A digit-only node name still fails with `must be string`. Turns out the unquoted `!!str` tag I had in this PR doesn't hold up either, for the same reason (the tag is consumed before env resolution).

The only thing that reliably forces a string without polluting the value is making it a multi-reference value, so the collector interpolates instead of type-casting. Appending an empty expansion does it:

```yaml
host.name: "${env:OTEL_K8S_NODE_NAME}${env:OTEL_FORCE_STRING:-}"
```

Verified end to end: it renders through `toYaml`, passes `otelcol validate`, resolves to a clean `12345`, and preserves leading zeros like `007`. No template helper needed, it lives entirely in values.yaml.

One thing I wanted to check with you: I'm using an unset sentinel (`OTEL_FORCE_STRING`) for the empty half. Would you prefer I define it as an explicit empty env var in the pod spec instead, so it can't collide with anything in the environment? Happy either way. I've already reverted the helper and switched values.yaml over locally, just need to regenerate the examples once you're good with the direction.

kshitizj03:
Makes sense. #2283 migrates to the supported `resource.attributes` array, which fixes this and also clears the deprecation warning, so it's the better fix. Closing this in favor of it. Thanks all!

PR #2278, opened 2026-06-23
fix(OBI): Fixed network metrics config

Orenico10:
#### Description
1. The chart now has all the stuff mentioned in [opentelemetry.io](https://opentelemetry.io/docs/zero-code/obi/distributed-traces/#kubernetes-configuration) as needed (NET_ADMIN when not privileged & mount /sys/fs/cgroup when network feature is in use).

2. The config now allows for templating (tpl) similarly to how it is in the otel collector chart.

#### Authorship

### DIFF
--- a/charts/opentelemetry-ebpf-instrumentation/values.yaml
+++ b/charts/opentelemetry-ebpf-instrumentation/values.yaml
@@ -195,10 +195,8 @@ config:
   ## If empty, default configuration below is used.
   name: ""
   # -- default value of Opentelemetry eBPF Instrumentation configuration
-  # Supports Go template expressions (e.g. {{ .Release.Name }}). BREAKING: values that contain
-  # literal {{ }} are now interpreted as template expressions. Escape them as {{` {{ }} `}}.
-  # Do not use non-idempotent functions (e.g. randAlphaNum) — config.data is rendered
-  # separately for the ConfigMap and the DaemonSet checksum/config annotation.
+  # Supports templating. To escape existing instances of {{ }}, use {{` <original content> `}}.
+  # For example, {{ REDACTED_EMAIL }} becomes {{` {{ REDACTED_EMAIL }} `}}.
   data:
     # profile_port: 6060
     # open_port: 8443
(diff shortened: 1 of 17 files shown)


charts/opentelemetry-ebpf-instrumentation/values.yaml
```diff
@@ -221,8 +225,9 @@ config:
         k8s_src_owner_name:
           not_match: '{kube*,*jaeger-agent*,*prometheus*,*promtail*,*grafana-agent*}'
     ## to enable network metrics
-    # network:
-    #   enable: true
+    # metrics:
```
povilasv: why did this move under metrics?

Orenico10: Initially I wanted to update the networks detection to use the new config (as it looks like in the devdocs under OBI's repo) which is like this:
```yaml
metrics:
  features:
    - network
 ```
instead of 
```yaml
network:
  ...
```
but decided to revert this part because I'm not sure how mature the new schema is, I missed this when reverting.

Should be fixed now.

NimrodAvni78 (approved):
LGTM

charts/opentelemetry-ebpf-instrumentation/values.yaml
```diff
@@ -195,6 +195,10 @@ config:
   ## If empty, default configuration below is used.
   name: ""
   # -- default value of Opentelemetry eBPF Instrumentation configuration
+  # Supports Go template expressions (e.g. {{ .Release.Name }}). BREAKING: values that contain
+  # literal {{ }} are now interpreted as template expressions. Escape them as {{` {{ }} `}}.
+  # Do not use non-idempotent functions (e.g. randAlphaNum) — config.data is rendered
+  # separately for the ConfigMap and the DaemonSet checksum/config annotation.
```
TylerHelmuth: Can you add an UPDATING.md file and record this breaking change there? See our other charts as an example.

Then update this comment to match the templating warning of another charts like collector's config.

Orenico10: Done :)

TylerHelmuth (approved):

PR #2280, opened 2026-06-24
[opentelemetry-operator]: add a CI check that feature gates stay in sync

### LINKED ISSUE #2248: [opentelemetry-operator]: Keep chart in sync with the operator's available feature gates
A couple times we've seen issue like https://github.com/open-telemetry/opentelemetry-helm-charts/issues/2240 where new operator version adds/removes a features gate and the corresponding PR to bump the operator helm chart doesn't add/drop support for the new gate.

I'm not sure exactly how we'd handle this. Open to ideas.

thc1006:
#### Description

`manager.featureGatesMap` is a hand-maintained allowlist with `additionalProperties: false`, so operator version bumps can add or remove feature gates without the Helm chart schema being updated. This PR fixes the current drift by adding the missing `operator.collector.usedefaulttelemetryshape` gate from operator `v0.153.0`.

It also adds `make check-operator-feature-gates` and wires it into the operator CI next to `check-operator-crds`. The check reads the alpha/beta gates registered in `pkg/featuregate/featuregate.go` for the chart `appVersion`, compares them with the schema keys under `manager.featureGatesMap`, and fails CI if the two lists drift or if no gates can be parsed.

The implementation follows the existing CRD check pattern and keeps the scope limited to the chart schema sync path; it does not change runtime behavior or add RBAC for individual gates.

#### Link to tracking issue

Fixes #2248

#### Authorship


Makefile
```diff
@@ -87,6 +87,10 @@ check-operator-crds:
 		exit 1; \
 	fi; \
 
+.PHONY: check-operator-feature-gates
+check-operator-feature-gates:
+	.github/scripts/check-operator-feature-gates.sh
```
TylerHelmuth: Instead of adding new scripts, can we add 2 new targets in the makefile, `check-operator-feature-gates` and `update-operator-featuregates`? Preferably these would both be implemented directly in the make file.

.github/scripts/test-check-operator-feature-gates.sh
TylerHelmuth, line 1: Is this needed?

charts/opentelemetry-operator/values.schema.json
```diff
@@ -625,42 +625,44 @@
           "title": "The featureGates Schema",
           "additionalProperties": false,
           "properties": {
-            "operator.clusterobservability": {
+            "operator.golang.flags": {
               "type": "boolean",
               "default": false,
-              "title": "Whether to enable the ClusterObservability controller",
```
TylerHelmuth, line 631: No need to remove these titles

Makefile
```diff
@@ -106,3 +129,11 @@ define get-clusterobservability-crd
 @sed -i '1a{{- if get .Values.manager.featureGatesMap "operator.clusterobservability" }}' $(1)
 @echo '{{- end }}\n{{- end }}' >> $(1)
 endef
+
+define gen-operator-feature-gates
```
TylerHelmuth: I realized this would be first introduction of generating the schema file. That adds some lint concerns I don't want to take on yet. 

Lets not gen the file yet. Instead, the check target should fail if the schema feature gates does not match the operator's feature gates. It should print in its error the feature gates that need to be manually added to  the schema.

charts/opentelemetry-operator/values.schema.json
thc1006: Reverted. Kept all the existing titles, the only schema change now is the added `operator.collector.usedefaulttelemetryshape` entry (with its own title).

Makefile
thc1006: Makes sense. Dropped the schema generation (removed the update target and the gen macro). `check-operator-feature-gates` now just compares the operator's alpha/beta feature gates against the schema keys and fails when they differ, printing which gates to add or remove, so the schema stays hand-maintained.

TylerHelmuth (approved):

PR #2279, opened 2026-06-24
deprecate k8snode detector in favor of k8s_api in resourceDetection preset

kangyili:
#### Description

Handle the deprecation of the `k8snode` resource detector in the `resourcedetection` processor in favor of `k8s_api`, introduced in v0.154.0.

- Add `k8s_api` as the new default detector in presets.resourceDetection, replacing `k8snode`
- Both detectors are supported simultaneously — `k8snode` takes precedence if explicitly set, preventing users from accidentally running both
- Add `k8s_api` to values.schema.json; keep `k8snode` in the schema so existing values.yaml files continue to validate
- Add namespaces/kube-system: get ClusterRole permission required by `k8s_api` for resolving `k8s.cluster.uid`
- Emit a [DEPRECATION] warning in NOTES.txt when `k8snode` is explicitly enabled
- Document the breaking change and migration path in `UPGRADING.md`

**Who is not impacted:**
- Users with `presets.resourceDetection.enabled: false `— no change
- Users who explicitly set `k8snode.enabled: true` — still get `k8snode`, with a deprecation warning

**Who is impacted:**
- Users relying on default settings with `presets.resourceDetection.enabled: true` — silently switch from `k8snode` to `k8s_api`. Safe on image >= 0.154.0, breaks if pinned to an older image (see `UPGRADING.md`)

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-helm-charts/issues/2268

#### Authorship

- [x] I, a human, wrote this pull request description myself.

PR #2282, opened 2026-06-25
fix(kube-stack): update generated component names

gimsesu:
#### Description

Update generated OpenTelemetry component type names in the opentelemetry-kube-stack presets to match current lower_snake_case names:

- `k8sattributes` -> `k8s_attributes` from collector-contrib [v0.146.0](https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.146.0)
- `filelog` -> `file_log` from collector-contrib [v0.149.0](https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.149.0)
- `hostmetrics` -> `host_metrics` from collector-contrib [v0.151.0](https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.151.0)
- `k8sobjects` -> `k8s_objects` from collector-contrib v0.151.0
- `kubeletstats` -> `kubelet_stats` from collector-contrib [v0.152.0](https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.152.0)
- `resourcedetection` -> `resource_detection` from collector-contrib [v0.153.0
](https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.153.0)

#### Link to tracking issue
Fixes

#### Authorship


cyrille-leclerc:
Aligning to the new names makes sense and is asked by users due to the warning log messages produced by the old deprecated names.
One concern though: does this introduce a silent breaking change for users who override preset-generated config via Helm value merges?

For example, a user overriding the `collection_interval` of the `hostmetrics` receiver today might have:

```yaml
collectors:
  daemon:
    enabled: true
    presets:
      hostMetrics:
        enabled: true
      ...
    config:
      receivers:
        hostmetrics:
          collection_interval: 1m
```

After this change, the preset generates a `host_metrics` receiver, but the user's override still references `hostmetrics` — so it would create a second, disconnected receiver entry rather than merging with the preset. No error, no warning, just silently wrong behavior.

Worth calling out in the PR description, or addressing with a migration note / deprecation warning if the chart can surface one.

@TylerHelmuth from what I remember, you started looking at this question. WDYT?

JaredTan95 (approved):

TylerHelmuth:
@cyrille-leclerc correct. We need a graceful transition to the new names. We'll have to support both names for some time and have both configure the same underlying behavior. We should have checks that ensure both a new name and old name for the same feature are not used at the same time. And we should remove old names from the values.yaml (I believe this is safe since all the presets are opt in and they should be providing any default config).

See https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2279 as an example

gimsesu:
@cyrille-leclerc @TylerHelmuth 

I updated the PR to address the backward compatibility concern:

1. Deprecated component names in user-provided collector config are now normalized to the new names during Helm rendering.
2. If deprecated component names are used, the chart now prints deprecation warnings through `NOTES.txt`.

The chart also fails rendering when both old and new names for the same component are configured at the same time.

charts/opentelemetry-kube-stack/values.yaml
```diff
@@ -686,7 +686,7 @@ collectors:
             http:
               endpoint: 0.0.0.0:4318
       processors:
-        resourcedetection/env:
+        resource_detection/env:
```
TylerHelmuth: Changing this would be a breaking change. We have to keep the old name for now

TylerHelmuth: That should be safe because the normalization will handle converting the name

gimsesu:
@TylerHelmuth Reverted the shipped collector defaults in values.yaml back to the old component names, so existing users overriding those blocks with the old spelling keep rendering.

  Also relaxed the normalization guard from a hard fail to new-wins: when both the old and new name for a component are present (dict key or pipeline reference), the new name is kept and
  the old one is dropped. This is needed because Helm deep-merges user values onto chart defaults — with defaults on the old name, a user overriding with the new name would otherwise
  produce both keys and hit the guard. new-wins makes overrides safe in either direction.

charts/opentelemetry-kube-stack/values.yaml
TylerHelmuth, line 1: like the opentelemetry-collector chart has lets add a root boolean field in the values.yaml to turn this feature on and off. It can have a default value of true.

gimsesu: @TylerHelmuth Thanks for the suggestion. I added a root-level boolean field, `rewriteDeprecatedComponentNames`, following the pattern used in the `opentelemetry-collector` chart. It defaults to `true`.

charts/opentelemetry-kube-stack/values.yaml
```diff
@@ -8,6 +8,12 @@ namespaceOverride: ""
 # better integration.
 clusterName: "unknown_k8s_cluster"
 
+# Rewrite deprecated component names (e.g. k8sattributes -> k8s_attributes
+# processor, hostmetrics -> host_metrics receiver) in the generated config.
+# Set to false if using older Collector images that do not recognize the new
+# names.
+rewriteDeprecatedComponentNames: true
```
TylerHelmuth, line 15: Please add an entry to `UPGRADING.md` explaining this new default and how to disable it. See https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/UPGRADING.md as an example.

gimsesu: @TylerHelmuth Added charts/opentelemetry-kube-stack/UPGRADING.md so users can refer to it. Thanks!

charts/opentelemetry-kube-stack/UPGRADING.md
```diff
@@ -0,0 +1,20 @@
+# Upgrade guidelines
+
+These upgrade guidelines only contain instructions for version upgrades which require manual modifications on the user's side.
+If the version you want to upgrade to is not listed here, then there is nothing to do for you.
+Just upgrade and enjoy.
+
```
TylerHelmuth:
```suggestion
```

TylerHelmuth (approved):

PR #2283, opened 2026-06-26
feat(opentelemetry-collector): migrate default service.telemetry.resource to attributes array

### LINKED ISSUE #2250: [opentelemetry-collector] Pod fails to start when node name is digits only

### LINKED ISSUE #2255: Deprecating warning for legacy `service.telemetry.resource` inline map format
Hi there.

In upgrading process (v0.152.0 -> 0.153.0) we discovered, that there was a application version change and from this helm chart version the OTel pods started generating the warning:

`
2026-06-09T08:01:36.524Z    warn    otelconftelemetry/logger.go:133    Using legacy service.telemetry.resource inline map format; prefer service.telemetry.resource.attributes (array of maps with `name` and `value` keys)    {"resource": {"host.name": "REDACTED", "k8s.namespace.name": "REDACTED", "k8s.node.ip": "REDACTED", "k8s.node.name": "REDACTED", "k8s.pod.ip": "REDACTED", "k8s.pod.name": "REDACTED", "service.instance.id": "REDACTED", "service.name": "REDACTED", "service.version": "0.153.0"}, "legacy_resource_attributes": ["host.name", "k8s.namespace.name", "k8s.node.ip", "k8s.node.name", "k8s.pod.ip", "k8s.pod.name"]}
`

It is not harmful, the OTel working fine without any problems, but we afraid, that with new application version, this can be problem. It is tedious, when we saw the warning on all start / restart, especially on production environment.

Unfortunately, the settings are assembled inside the helm chart logic, so I dont see straight way, how to fix / patch it on user side.

I would like to kindly ask you to fix the problem. I am sure, that this can be benefit for all OTel helm chart users.

Thank you!

nicknikolakakis:
#### Description

The collector chart's default `config.service.telemetry.resource` uses the legacy inline-map form. Since Collector appVersion 0.153.0 this form is deprecated, so the Collector logs a deprecation warning on every startup.

This migrates the default to the supported `resource.attributes` array form (`- name: ... value: ...`), which clears the warning. Because each attribute's `type` defaults to `string`, the values are explicitly string-typed, so this also removes the all-numeric node-name int-coercion problem tracked in #2250 without needing `!!str` tags.

Changes:
- `values.yaml`: `service.telemetry.resource` map -> `resource.attributes` array.
- `templates/_config.tpl`: build `with_resource_constant_labels.included` from `resource.attributes[].name`; the list is still alpha-sorted, so rendered output for that block is unchanged.
- `Chart.yaml`: 0.159.0 -> 0.160.0.
- `UPGRADING.md`: note for users who override `service.telemetry.resource` with the legacy map (it cannot be combined with `attributes`).
- Regenerated examples via `make generate-examples`. One example (`old-metrics-url-values`) that overrode the resource block was migrated to the array form too.

This supersedes #2273, which addressed #2250 on the same `values.yaml` block via `!!str` tags.

#### Link to tracking issue
Fixes #2255
Closes #2250

### DIFF
--- a/charts/opentelemetry-collector/values.yaml
+++ b/charts/opentelemetry-collector/values.yaml
@@ -251,12 +251,19 @@ config:
   service:
     telemetry:
       resource:
-        k8s.namespace.name: "${env:OTEL_K8S_NAMESPACE}"
-        k8s.node.name: "${env:OTEL_K8S_NODE_NAME}"
-        k8s.node.ip: "${env:OTEL_K8S_NODE_IP}"
-        k8s.pod.name: "${env:OTEL_K8S_POD_NAME}"
-        k8s.pod.ip: "${env:OTEL_K8S_POD_IP}"
-        host.name: "${env:OTEL_K8S_NODE_NAME}"
+        attributes:
+          - name: k8s.namespace.name
+            value: "${env:OTEL_K8S_NAMESPACE}"
+          - name: k8s.node.name
+            value: "${env:OTEL_K8S_NODE_NAME}"
+          - name: k8s.node.ip
+            value: "${env:OTEL_K8S_NODE_IP}"
+          - name: k8s.pod.name
+            value: "${env:OTEL_K8S_POD_NAME}"
+          - name: k8s.pod.ip
+            value: "${env:OTEL_K8S_POD_IP}"
+          - name: host.name
+            value: "${env:OTEL_K8S_NODE_NAME}"
       metrics:
         readers:
           - pull:
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.159.1
+version: 0.160.0
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-collector/UPGRADING.md
+++ b/charts/opentelemetry-collector/UPGRADING.md
@@ -4,6 +4,24 @@ These upgrade guidelines only contain instructions for version upgrades which re
 If the version you want to upgrade to is not listed here, then there is nothing to do for you.
 Just upgrade and enjoy.
 
+## 0.159.1 to 0.160.0
+
+The default `config.service.telemetry.resource` now uses the Collector's `resource.attributes` array format instead of the deprecated inline map. This clears the deprecation warning the Collector logs on startup (appVersion 0.153.0+).
+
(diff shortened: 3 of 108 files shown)


charts/opentelemetry-collector/values.yaml
```diff
@@ -251,12 +251,19 @@ config:
   service:
     telemetry:
       resource:
-        k8s.namespace.name: "${env:OTEL_K8S_NAMESPACE}"
-        k8s.node.name: "${env:OTEL_K8S_NODE_NAME}"
-        k8s.node.ip: "${env:OTEL_K8S_NODE_IP}"
-        k8s.pod.name: "${env:OTEL_K8S_POD_NAME}"
-        k8s.pod.ip: "${env:OTEL_K8S_POD_IP}"
-        host.name: "${env:OTEL_K8S_NODE_NAME}"
+        attributes:
```
TylerHelmuth, line 254: I believe you have implemented this breaking change correctly, but I am worried about handling it as a breaking change. Like we did for `service.telemetry.metrics.address`, it would be nice to help transition users.

Is it possible to detect if `.Values.config.service.telemetry.resource.attributes` is a map or a list? If so we could:

- update the default value to be a list like you have for all new users
- detect existing users who are using the field as a map and convert those values to a list.

charts/opentelemetry-collector/values.yaml
```diff
@@ -251,12 +251,19 @@ config:
...
+          - name: k8s.namespace.name
```
TylerHelmuth, line 255: switching to a list has the annoying side effect that users have to re-add all these default values when they want to add their own attributes. This is constantly a pain and we've actually dont work throughout these charts to switch to maps where we can.

I wonder if we need to allow configuring these attributes not in `.Values.config` but somewhere else as a map still and then merge them in. @open-telemetry/helm-approvers what do you think?

TylerHelmuth:
@nicknikolakakis FYI I create https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2284 to help simplify our situation.

charts/opentelemetry-collector/values.yaml
jaronoff97: @TylerHelmuth yes, we could use [helm / spring kind functions ](https://helm.sh/docs/chart_template_guide/function_list/#kind-functions)for this purpose. i think this would be fine and we should also ensure that we do eventually deprecate the old path for users.

charts/opentelemetry-collector/values.yaml
jaronoff97: I think we could have a set of defaults in a map to easily override and a list for user provided attrs. lists are quite annoying unfortunately :/

TylerHelmuth: If we do that I dont think we'd want to put it in `.Values.config` since for the lifetime of the chart that has been copy-pasteable collector config.

PR #2287, opened 2026-06-30
[opentelemetry-collector] Add unprivileged and additionalCapabilities setting to profiling preset

theomagellan:
#### Description

The helm charts with the profiling preset run a collector with `privileged: true` which could make this instance a
high risk target for pod compromise.

The profiler can run unprivileged with a small set of capabilities instead of full privilege.
Today, the only way to run unprivileged is to write a complete `securityContext` block.

This PR replaces `privileged: true` with a minimal capability set based on the [current profiler's minimum kernel version `5.10`](https://github.com/open-telemetry/opentelemetry-ebpf-profiler#supported-linux-kernel-version).

##### Capabilities

- `BPF`, `PERFMON`: load eBPF programs / perf_event_open
- `SYS_PTRACE`, `DAC_READ_SEARCH`, `CHECKPOINT_RESTORE`: read process memory and open memory-mapped files
- `SYSLOG`: kernel ring buffer
- `IPC_LOCK`, `SYS_RESOURCE`: locked memory for eBPF maps / raise rlimits

When the `profiling` preset is enabled and no custom `securityContext` is set, the following values are applied:

```yaml
runAsUser: 0
runAsGroup: 0
privileged: false
allowPrivilegeEscalation: false
seccompProfile:
  type: Unconfined
appArmorProfile: # set as annotation on Kubernetes < 1.30
  type: Unconfined
capabilities:
  drop: [ALL]
  add: [BPF, PERFMON, SYS_PTRACE, SYS_RESOURCE, DAC_READ_SEARCH, SYSLOG, CHECKPOINT_RESTORE, IPC_LOCK]
```
For collectors running additional components that could require any other capabilities, they can still write a fully custom `securityContext`

##### Validating

The Datadog eBPF profiler distribution has been running unprivileged under this capability set for some time across different node sizes and kernel versions. The distribution uses the following components:
- `k8s_attributes` configured by the `profiling` and `kubernetesAttributes` (with `extractAllPodLabels`) templates
- `otlp_http` exporter
- `resource_detection` with the `system` detector
- `prometheus` receiver for collector self-telemetry metrics
- custom processor to retrieve specific Datadog node hostnames

The set was also tested using `opentelemetry-collector-ebpf-profiler 0.155.0` on ubuntu 24.04.4 and 26.04 (arm64/amd64) and with all the available components in the image individually.
Only the `batch` component did not work with error `telemetry type is not supported`, so unrelated to this.

Happy to discuss if the direction doesn't sound right or if there are more validation steps to take!

#### Authorship

### DIFF
--- a/charts/opentelemetry-collector/templates/_pod.tpl
+++ b/charts/opentelemetry-collector/templates/_pod.tpl
@@ -28,18 +28,14 @@ containers:
       - {{ . }}
       {{- end }}
     securityContext:
-      {{- if and (not (.Values.securityContext)) (.Values.presets.profiling.enabled) (.Values.presets.profiling.unprivileged) }}
+      {{- if and (not (.Values.securityContext)) (.Values.presets.profiling.enabled) }}
       runAsUser: 0
       runAsGroup: 0
       privileged: false
       allowPrivilegeEscalation: false
       capabilities:
         drop: [ALL]
-        add: {{ concat (list "BPF" "PERFMON" "SYS_PTRACE" "SYS_RESOURCE" "DAC_READ_SEARCH" "SYSLOG" "CHECKPOINT_RESTORE" "IPC_LOCK") .Values.presets.profiling.additionalCapabilities | toJson }}
-      {{- else if and (not (.Values.securityContext)) (.Values.presets.profiling.enabled) }}
-      runAsUser: 0
-      runAsGroup: 0
-      privileged: true
+        add: ["BPF", "PERFMON", "SYS_PTRACE", "SYS_RESOURCE", "DAC_READ_SEARCH", "SYSLOG", "CHECKPOINT_RESTORE", "IPC_LOCK"]
       {{- else if and (not (.Values.securityContext)) (.Values.presets.logsCollection.storeCheckpoints) }}
       runAsUser: 0
       runAsGroup: 0
--- a/charts/opentelemetry-collector/templates/_pod.tpl
+++ b/charts/opentelemetry-collector/templates/_pod.tpl
@@ -28,7 +28,15 @@ containers:
       - {{ . }}
       {{- end }}
     securityContext:
-      {{- if and (not (.Values.securityContext)) (.Values.presets.profiling.enabled) }}
+      {{- if and (not (.Values.securityContext)) (.Values.presets.profiling.enabled) (.Values.presets.profiling.unprivileged) }}
+      runAsUser: 0
+      runAsGroup: 0
+      privileged: false
+      allowPrivilegeEscalation: false
+      capabilities:
+        drop: [ALL]
+        add: {{ concat (list "BPF" "PERFMON" "SYS_PTRACE" "SYS_RESOURCE" "DAC_READ_SEARCH" "SYSLOG" "CHECKPOINT_RESTORE" "IPC_LOCK") .Values.presets.profiling.additionalCapabilities | toJson }}
+      {{- else if and (not (.Values.securityContext)) (.Values.presets.profiling.enabled) }}
       runAsUser: 0
       runAsGroup: 0
       privileged: true
--- a/charts/opentelemetry-collector/templates/_pod.tpl
+++ b/charts/opentelemetry-collector/templates/_pod.tpl
@@ -39,6 +39,8 @@ containers:
       appArmorProfile:
         type: Unconfined
       {{- end }}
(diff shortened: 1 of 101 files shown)


charts/opentelemetry-collector/values.yaml
```diff
@@ -138,6 +138,10 @@ presets:
   # See https://github.com/open-telemetry/opentelemetry-collector-releases/tree/main/distributions/otelcol-ebpf-profiler for more details.
   profiling:
     enabled: false
+    # Run unprivileged with a minimal capability set instead of privileged: true.
+    unprivileged: false
+    # Extra capabilities to add when unprivileged, for additional components in the pipeline.
+    additionalCapabilities: []
```
TylerHelmuth: We typically do not allow presets configurations other than bools since the presets should be very very opinionated. Lets remove this.

charts/opentelemetry-collector/values.yaml
```diff
@@ -138,6 +138,10 @@ presets:
   # See https://github.com/open-telemetry/opentelemetry-collector-releases/tree/main/distributions/otelcol-ebpf-profiler for more details.
   profiling:
     enabled: false
+    # Run unprivileged with a minimal capability set instead of privileged: true.
+    unprivileged: false
```
TylerHelmuth: If the component can run unprivileged than I'd opt to have the existing `enabled` flag do it. If someone wants to use the component with privilege they can write a securityContext instead of using the preset.

theomagellan: Sure! I addressed your comments in [1f1c08f](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2287/commits/1f1c08fbccc884445f65533c7b6419dee0e8805f), thanks!

theomagellan:
Hello @TylerHelmuth 👋 
Gentle bump on this PR, since your last review I have addressed your comments making the unprivileged settings the preset defaults, removed the `additionalCapabilities` knob, and also made the preset run under `unconfined` apparmor and seccomp profiles.

Let me know if there is anything I can do to help make the review easier :)

charts/opentelemetry-collector/templates/_pod.tpl
```diff
@@ -31,7 +31,17 @@ containers:
...
+      allowPrivilegeEscalation: false
+      seccompProfile:
+        type: Unconfined
+      {{- if semverCompare ">= 1.30-0" .Capabilities.KubeVersion.Version }}
+      appArmorProfile:
+        type: Unconfined
+      {{- end }}
+      capabilities:
+        drop: [ALL]
+        add: ["BPF", "PERFMON", "SYS_PTRACE", "SYS_RESOURCE", "DAC_READ_SEARCH", "SYSLOG", "CHECKPOINT_RESTORE", "IPC_LOCK"]
```
TylerHelmuth, line 46: The implementation looks good, but can you confirm how you're deciding these are the min-required permissions to use the profiling preset?

theomagellan: I narrowed down to these capabilities after iterating on unprivileged deployments and looking at profiles for processes with different configurations (process root owned, user owned, compiled, interpreted, etc.).
After your question, I re-tested the set while removing one capability at a time to see if I could trim it down and I couldn't, they're all needed.

On the other hand this allowed me to note down behavior when each is not present!

| Capability | Failure mode |
|---|---|
| `BPF` | Hard crash: `map create: operation not permitted` |
| `PERFMON` | Hard crash: `load program: operation not permitted` |
| `SYS_PTRACE` | Silent degradation: kernel-only frames for all workloads |
| `SYS_RESOURCE` | Hard crash: `failed to adjust rlimit: operation not permitted` |
| `DAC_READ_SEARCH` | Silent degradation: kernel-only frames for non-root processes |
| `SYSLOG` | Hard crash: `unable to read kallsyms addresses - check capabilities` |
| `CHECKPOINT_RESTORE` | Silent degradation: kernel-only frames for all workloads |
| `IPC_LOCK` | Not re-tested here, but observed crash on nodes with 64+ cores |

These tests were ran under the same conditions as described in the PR description.
Let me know if there are any more tests you would like!

TylerHelmuth:
@theomagellan I'm back from PTO and ready to merge this. Please handle the conflicts.

TylerHelmuth (approved):

PR #2299, opened 2026-07-10
feat(opentelemetry-operator): support custom annotations on CRDs

### LINKED ISSUE #1713: Add  helm.sh/resource-policy: keep in CRDs

somaz94:
#### Description
Add an opt-in crds.annotations value so the operator CRDs can carry helm.sh/resource-policy: keep and survive helm uninstall. Disabled by default.

#### Link to tracking issue
Fixes #1713

#### Authorship


charts/opentelemetry-operator/values.yaml
```diff
@@ -349,6 +349,10 @@ admissionWebhooks:
 ## See https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/UPGRADING.md#0560-to-0570 for more details.
 crds:
   create: true
+  # Annotations to add to all CRDs managed by this chart.
+  # For example, set "helm.sh/resource-policy: keep" to retain the cluster-scoped
+  # CRDs when the chart is uninstalled.
+  annotations: {}
```
TylerHelmuth, line 354: Adding annotations to the CRDs managed by this charts is ok, but I'm not sure of the implications of not uninstalling them and then upgrading the chart. I think its best not to promote that annotation explicitly as its behavior we've not tested.

somaz94: Good point — agreed. I've dropped the `helm.sh/resource-policy: keep` example so the chart no longer promotes the uninstall-then-upgrade behavior we haven't tested. The generic `crds.annotations` feature stays, but the `values.yaml` comment, the README note, and the `values.schema.json` example now use a neutral annotation (`example.com/managed-by: opentelemetry-operator-chart`, matching the CI scenario). Anyone who wants `resource-policy: keep` can still set it themselves; the chart just doesn't suggest it.

TylerHelmuth (approved):

TylerHelmuth:
@somaz94 I merged another operator PR first. Please re-bump the chart version and regenerate the examples.

povilasv (approved):

somaz94:
Done in `88ba1105` — re-bumped the chart to `0.120.1` and re-ran `make generate-examples`. Thanks for the heads-up on the version collision after the other operator PR landed.

somaz94:
Two approvals here and CI is green. Flagging one practical thing rather than just nudging:

The chart version was already re-bumped once after another operator PR landed ahead of this one (`88ba1105`, 07-20). Right now this branch is at `0.120.1` against `main` at `0.120.0`, so there's no collision — but the same round trip happens again if another operator PR bumps first. Mentioning it only in case merging as-is is easier than another re-bump; happy to redo it if not.

TylerHelmuth:
@somaz94 I messed up and came back from PTO and merged another operator PR first. Sorry for the churn, you'll need to bump the chart version again

somaz94:
No worries at all — bumped to `0.120.2` and re-ran `make generate-examples` in `7882569c`. The diff is 35 files but purely the version string (53 insertions / 53 deletions, every changed line is `0.120.1` → `0.120.2`); the CRD annotations change itself is untouched.

Branch was already merged with `main`, so it's current apart from #2320.

charts/opentelemetry-operator/README.md
```diff
@@ -40,6 +40,8 @@ _See [helm repo](https://helm.sh/docs/helm/helm_repo/) for command documentation
 
 > [!NOTE]
 > This Chart uses templated CRDs, and therefore does not support `--skip-crds`. Use `crds.create=false` instead if you do not want the chart to install the OpenTelemetry Operator's CRDs.
+>
+> To add annotations to all CRDs managed by this chart, set `crds.annotations`.
```
TylerHelmuth:
```suggestion
```

TylerHelmuth (approved):

PR #2306, opened 2026-07-11
[opentelemetry-demo]: add HTTPRoute support as Gateway API alternative to Ingress

ryanfaircloth:
#### Description
This PR adds support for HTTPRoute as the natural path to replace ingress for access to the astroshop demo app.

#### Link to tracking issue
Fixes

#### Authorship


charts/opentelemetry-demo/examples/public-hosted-httproute/values.yaml
TylerHelmuth, line 1: I think it is ok to not add a new example for this feature.

charts/opentelemetry-demo/templates/_objects.tpl
```diff
@@ -308,3 +308,65 @@ spec:
...
 {{- end}}
 {{- end}}
+
+{{/*
+Demo component HTTPRoute template
+*/}}
+{{- define "otel-demo.httproute" }}
+{{- $hasRoute := false}}
+{{- if .httpRoute }}
+{{- if .httpRoute.enabled }}
```
TylerHelmuth, line 318: These can be simplified with `{{- if (.httpRoute).enabled }}`

charts/opentelemetry-demo/examples/public-hosted-httproute/values.yaml
ryanfaircloth: Should I remove it u did add one

TylerHelmuth: Ya please remove it. We didn't add any new examples in #2246

PR #2317, opened 2026-07-16
fix(kube-stack): grant leader election lease RBAC

### LINKED ISSUE #2311: ClusterRole coordination.k8s.io rules enablement depends on daemonset mode
In scope: charts/opentelemetry-kube-stack
Version: >= 0.18.0

## Problem statement

After upgrade to [opentelemetry-kube-stack-0.18.0](https://github.com/open-telemetry/opentelemetry-helm-charts/releases/tag/opentelemetry-kube-stack-0.18.0), we noticed that leader election failed.

After investigation, we noticed that the following rule was missing in the ClusterRole managed by opentelemetry-kube-stack:
```yaml
    - apiGroups:
      - coordination.k8s.io
      resources:
      - leases
      verbs:
      - get
      - list
      - update
 ```

## Investigation

We found that [change](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2266/changes#diff-41eeb9e0f2db6ef4cfb9f94ece00404f3a8f6a02be7c556b63c24761de42253aR170-R204) introduced in #2266 makes `$useLeaderElection` entirely dependent on `daemonset` mode of collector.

In our case, we have several collectors, including `gateway` collector that is **not** a daemonset but is responsible for collecting cluster metrics. This [line](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2266/changes#diff-41eeb9e0f2db6ef4cfb9f94ece00404f3a8f6a02be7c556b63c24761de42253aR187) makes condition based whether collector is in `daemonset` mode, which is not the case for our collector, our gateway runs in `deployment` mode and actual `daemon` collector does not have `clusterMetrics` enabled.

```
    daemon:
      enabled: true
      suffix: agent
      mode: daemonset # <---
      # ...
      presets:
        # ...
        clusterMetrics:
          enabled: false # <---
    gateway:
      enabled: true
      suffix: gateway
      mode: deployment # <---
      presets:
        clusterMetrics:
          enabled: true # <---
      # ...
```

In this setup, `$useLeaderElection` will evaluate to `false`, and rule for `coordination.k8s.io` API group does not get appended.

## Workaround

Setting missing rule through `.Values.clusterRole.rules` helps:
```yaml
  clusterRole:
    rules:
    - apiGroups:
      - coordination.k8s.io
      resources:
      - leases
      verbs:
      - get
      - list
      - update
```

## Questions

1. Is the described behaviour expected?
2. Are there plans to support this kind of non-standard setup, or should we rely on `.Values.clusterRole.rules`?

ihopenre-eng:
#### Description

The `clusterMetrics` and `kubernetesEvents` presets enable the `k8s_leader_elector` extension even when the collector is deployed in `deployment` mode. However, the ClusterRole generation logic in `_helpers.tpl` only set `$useLeaderElection` when the collector mode was `daemonset`. As a result, the rendered collector configuration enabled leader election while the rendered ClusterRole omitted the required `coordination.k8s.io/leases` RBAC permissions for non-daemonset collectors. This made the leader election configuration and RBAC generation inconsistent, preventing deployment-based collectors from creating and renewing Kubernetes Lease objects.

This change removes the `mode == daemonset` restriction from the `$useLeaderElection` evaluation for the `clusterMetrics` and `kubernetesEvents` presets. When either preset is enabled and `disableLeaderElection` is `false`, the chart now generates the required lease RBAC permissions regardless of the collector deployment mode, aligning RBAC generation with the generated collector configuration. The existing `disableLeaderElection` behavior is preserved, and the `kubernetesObjects` preset remains unchanged because it only uses leader election for daemonset collectors. The kube-stack chart version is bumped from `0.19.1` to `0.19.2`, and the checked-in rendered examples are regenerated accordingly.

Validation was performed with:

* `helm lint charts/opentelemetry-kube-stack`
* `make check-examples CHARTS=opentelemetry-kube-stack`

The rendered output was also verified for the following deployment scenarios:

* `deployment + clusterMetrics.enabled=true` → lease RBAC is generated.
* `deployment + clusterMetrics.enabled=true + clusterMetrics.disableLeaderElection=true` → lease RBAC is not generated.
* `deployment + kubernetesEvents.enabled=true` → lease RBAC is generated.
* `deployment + kubernetesObjects.enabled=true` → lease RBAC is not generated.

#### Link to tracking issue

Fixes #2311

#### Authorship

* [O] I, a human, wrote this pull request description myself.

### DIFF
--- a/charts/opentelemetry-kube-stack/templates/_helpers.tpl
+++ b/charts/opentelemetry-kube-stack/templates/_helpers.tpl
@@ -184,12 +184,12 @@ Optionally include the RBAC for the k8sCluster receiver
 {{- $clusterMetricsEnabled = (any $clusterMetricsEnabled (dig "config" "receivers" "k8s_cluster" false $collector)) }}
 {{- if (dig "presets" "clusterMetrics" "enabled" false $collector) }}
 {{- $clusterMetricsEnabled = true }}
-{{- $useLeaderElection = (any $useLeaderElection (and (eq $collector.mode "daemonset") (not (dig "presets" "clusterMetrics" "disableLeaderElection" false $collector)))) }}
+{{- $useLeaderElection = (any $useLeaderElection (not (dig "presets" "clusterMetrics" "disableLeaderElection" false $collector))) }}
 {{- end }}
 {{- $eventsEnabled = (any $eventsEnabled (dig "config" "receivers" "k8s_cluster" false $collector)) }}
 {{- if (dig "presets" "kubernetesEvents" "enabled" false $collector) }}
 {{- $eventsEnabled = true }}
-{{- $useLeaderElection = (any $useLeaderElection (and (eq $collector.mode "daemonset") (not (dig "presets" "kubernetesEvents" "disableLeaderElection" false $collector)))) }}
+{{- $useLeaderElection = (any $useLeaderElection (not (dig "presets" "kubernetesEvents" "disableLeaderElection" false $collector))) }}
 {{- end }}
 {{- if (dig "presets" "kubernetesObjects" "enabled" false $collector) }}
 {{- $kubernetesObjectsEnabled = true }}
--- a/charts/opentelemetry-kube-stack/Chart.yaml
+++ b/charts/opentelemetry-kube-stack/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-kube-stack
-version: 0.19.1
+version: 0.19.2
 description: |
   OpenTelemetry Quickstart chart for Kubernetes.
   Installs an operator and collector for an easy way to get started with Kubernetes observability.
--- a/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/bridge.yaml
+++ b/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/bridge.yaml
@@ -5,7 +5,7 @@ kind: OpAMPBridge
 metadata:
   name: example
   labels:
-    helm.sh/chart: opentelemetry-kube-stack-0.19.1
+    helm.sh/chart: opentelemetry-kube-stack-0.19.2
     app.kubernetes.io/version: "0.154.0"
     app.kubernetes.io/managed-by: Helm
     release: "example"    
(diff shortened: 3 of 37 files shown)


charts/opentelemetry-kube-stack/templates/_helpers.tpl
```diff
@@ -184,12 +184,12 @@ Optionally include the RBAC for the k8sCluster receiver
 {{- $clusterMetricsEnabled = (any $clusterMetricsEnabled (dig "config" "receivers" "k8s_cluster" false $collector)) }}
 {{- if (dig "presets" "clusterMetrics" "enabled" false $collector) }}
 {{- $clusterMetricsEnabled = true }}
-{{- $useLeaderElection = (any $useLeaderElection (and (eq $collector.mode "daemonset") (not (dig "presets" "clusterMetrics" "disableLeaderElection" false $collector)))) }}
+{{- $useLeaderElection = (any $useLeaderElection (not (dig "presets" "clusterMetrics" "disableLeaderElection" false $collector))) }}
```
TylerHelmuth, line 187: I just realized that since this is the first place `$useLeaderElection` is set/used, we can remove the above `{{- $useLeaderElection := false }}` and on this line do `{{- $useLeaderElection = (not (dig "presets" "clusterMetrics" "disableLeaderElection" false $collector)) }}`

ihopenre-eng: I think we should keep the initial `{{- $useLeaderElection := false }}`.

  `$useLeaderElection` is used as an aggregate flag across multiple preset branches in this helper, not just
  `clusterMetrics`. It is also updated later for `kubernetesEvents` and `kubernetesObjects`, so replacing the
  initialization with a direct assignment in the first branch would stop it from behaving like an accumulator.

  For example, the current structure relies on sequentially folding multiple preset conditions into the same
  `$useLeaderElection` value. If we turn the first branch into the effective initialization, we lose that clearer
  accumulation pattern for other presets evaluated below this block.

  The intent of this change is only to align the RBAC-side predicate with the existing config-generation logic for
  `clusterMetrics` and `kubernetesEvents`, so deployment-mode collectors get lease RBAC whenever leader election is
  actually enabled.

TylerHelmuth (approved):

PR #2318, opened 2026-07-16
[opentelemetry-operator] Support custom annotations on CRDs via crds.…

### LINKED ISSUE #1713: Add  helm.sh/resource-policy: keep in CRDs

shekhargit1912:
This PR adds a new crds.annotations value to the opentelemetry-operator chart. By default it is empty ({}), so nothing changes for existing users.
The purpose is to let users set annotations on the CRDs — for example helm.sh/resource-policy: keep — so the CRDs are not deleted when the chart is uninstalled.
The CRD files under conf/crds/ are generated by the Makefile, so I added the annotation template to the get-crd steps and regenerated the files. This keeps the check-operator-crds CI check passing. I also bumped the chart version to 0.121.0, added the new value to values.schema.json, and regenerated the examples.
Tested: default rendering is unchanged; with the annotation set it appears on all CRDs; helm lint passes.
Link to tracking issue
Fixes #1713

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -118,6 +118,7 @@ define get-crd
 @sed -i 's/opentelemetry-operator-system/{{ template "opentelemetry-operator.namespace" . }}/g' $(1)
 @sed -i 's/opentelemetry-operator-webhook-service/{{ template "opentelemetry-operator.fullname" . }}-webhook/g' $(1)
 @sed -i '1s/^/{{- if .Values.crds.create }}\n/' $(1)
+@sed -i '0,/^  annotations:/s//  annotations:\n{{- with .Values.crds.annotations }}\n{{ toYaml . | indent 4 }}\n{{- end }}/' $(1)
 @sed -i 's#\(.*\)path: /convert#&\n\1port: {{ .Values.admissionWebhooks.servicePort }}#' $(1)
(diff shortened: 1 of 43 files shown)


TylerHelmuth:
@shekhargit1912 https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2299 is already opened for this issue. As it was opened first I plan on merging it. If the user is not able to complete the PR (unlikely) I'll come back to this one.

PR #2329, opened 2026-07-25
feat(opentelemetry-collector): add useK8sEventsReceiver to kubernetesEvents preset

nicknikolakakis:
#### Description

Implements steps 2 and 3 of the transition plan in #2209.

Adds `presets.kubernetesEvents.useK8sEventsReceiver` (default `false`). When enabled, the `kubernetesEvents` preset emits a `k8s_events` receiver instead of the `k8sobjects` receiver, and the generated ClusterRole grants `events` in the core (`""`) API group instead of `events.k8s.io`, which is what that receiver actually reads.

Also adds `presets.kubernetesObjects.events.enabled` (default `false`), so users who prefer collecting events through the `k8sobjects` receiver can keep doing that alongside the other Kubernetes objects.

The flag is nested under the preset rather than root-level, following the review feedback on #2222 ("Lets nest this under the kubernetesEvents preset"). That PR went stale before it was updated. Per the plan in the issue, the flag defaults to `false` now, would default to `true` in a future release, and would be removed after that.

Both defaults are off, so the generated config and RBAC are unchanged for existing users. I verified this by rendering `ci/preset-k8sevents-values.yaml`, `ci/preset-kubernetesobjects-mode-deployment-values.yaml` and `ci/preset-kubernetesobjects-mode-daemonset-values.yaml` against `main` and diffing: the only difference is the `helm.sh/chart` version label.

Added `ci/preset-k8sevents-receiver-values.yaml` so `ct install` exercises the flag-on path. Chart bumped to 0.166.0 and examples regenerated.

One thing I left alone deliberately: enabling `presets.kubernetesEvents` (with the flag off) together with `presets.kubernetesObjects.events` produces two identical `events` entries in the same `k8sobjects` receiver, so events get collected twice. I did not add validation for it because I would rather have your call on whether that should fail, warn, or be silently deduplicated.

#### Link to tracking issue

Ref #2209

Not using a closing keyword, since steps 4 and 5 of the plan (flipping the default, then removing the flag) are still outstanding.

#### Authorship

### DIFF
--- a/charts/opentelemetry-collector/values.yaml
+++ b/charts/opentelemetry-collector/values.yaml
@@ -72,6 +72,8 @@ presets:
   # See https://opentelemetry.io/docs/kubernetes/collector/components/#kubernetes-objects-receiver for details on the receiver.
   kubernetesEvents:
     enabled: false
+    # When enabled, the preset uses the k8s_events receiver instead of k8sobjects. Will default to true in a future release.
+    useK8sEventsReceiver: false
   # Collects Kubernetes objects via the k8sobjects receiver in pull mode (default interval: 1h).
   # Can be used with mode = deployment, statefulset, or daemonset.
   # Compatible with kubernetesEvents preset. Extra resources can be added via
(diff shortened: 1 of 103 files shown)


charts/opentelemetry-collector/values.yaml
```diff
@@ -106,6 +108,9 @@ presets:
     # API extensions resources: customresourcedefinitions
     apiExtensions:
       enabled: true
+    # Kubernetes events, collected in watch mode. Equivalent to the kubernetesEvents preset.
```
TylerHelmuth, line 111: This wont be completely equivalent to the k8sevents receiver because that receiver does some attribute manipulation based on the events it received. 
```suggestion
    # Kubernetes events, collected in watch mode.
```

TylerHelmuth (approved):
Looks good. Please handle merge conflicts

TylerHelmuth (approved):

Issue #2338, opened 2026-08-03
[opentelemetry-kube-stack] Add top-level resourceAttributes to set default resource attributes across collectors and Instrumentation


cyrille-leclerc:
**Description:**

**Is your feature request related to a problem?**

There's no single value to set a resource attribute (e.g. `[deployment.environment.name](http://deployment.environment.name/)`, `team`, or `k8s.cluster.name` if the latter can't be inferred by a resource detection processor) across both the collectors this chart deploys and the Instrumentation CR it manages. Today this requires duplicating config in two places:
- `instrumentation.resource.resourceAttributes` (exists, applies to auto-instrumented workloads only)
- manually adding a `resource` processor to each collector's `config` or to `defaultCRConfig`

Other observability Helm charts/operators solve this with a single field: Datadog's `datadog.tags`, New Relic's `global.customAttributes`, Dynatrace's DynaKube-level `metadataEnrichment`. This chart is the natural place to do the same for OTel, since it's the only chart spanning both the collector and the operator/Instrumentation CRD.

**Describe the solution you'd like**

Add a top-level `resourceAttributes: {}` map, following the existing convention of plain top-level fields (`clusterName`, `extraEnvs`) rather than a Helm `global:` block — the collector/Instrumentation resources are rendered by this chart's own templates, not a subchart, so `global` semantics don't apply here.

```yaml
# Default resource attributes applied to every collector pipeline and to
# auto-instrumented workloads via the Instrumentation CR. Keys should follow
# OTel semantic conventions where one exists (e.g. [deployment.environment.name](http://deployment.environment.name/)).
resourceAttributes: {}
# [deployment.environment.name](http://deployment.environment.name/): production
# team: fraud
# k8s.cluster.name: john_doe_dev_workstation
```

Proposed wiring:
1. **Instrumentation:** merge as defaults into `instrumentation.resource.resourceAttributes`, with the existing per-field value overriding on key conflict.
2. **Collectors:** template into a new `resource/global` processor, inserted into `defaultCRConfig`'s pipelines after `resourcedetection/env` (same position/precedence as the existing `resource/hostname` processor), so explicit values win over auto-detected ones.
3. **Per-collector override:** since `defaultCRConfig` already overlays per-entry under `collectors.<name>` (see how `presets` and `resourceDetection` work today), placing `resourceAttributes` inside that overlay path gets per-collector override for free.

[...]

PR #2339, opened 2026-08-04
feat(kube-stack): set env var `OTEL_K8S_CLUSTER_NAME`

cyrille-leclerc:
#### Description

set env var `OTEL_K8S_CLUSTER_NAME` in collector containers when `clusterName` is set in `values.yaml`.

This helps using the `clusterName` value in otelcol processors, for example to set the `k8s.cluster.name` resource attribute on telemetry.

#### Authorship

### DIFF
--- a/charts/opentelemetry-kube-stack/templates/collector.yaml
+++ b/charts/opentelemetry-kube-stack/templates/collector.yaml
@@ -191,6 +191,10 @@ spec:
       fieldRef:
         apiVersion: v1
         fieldPath: status.podIP
+  {{- with $.Values.clusterName }}
+  - name: OTEL_K8S_CLUSTER_NAME
+    value: {{ . | quote }}
+  {{- end }}
   - name: OTEL_RESOURCE_ATTRIBUTES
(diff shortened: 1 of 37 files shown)


charts/opentelemetry-kube-stack/templates/collector.yaml
```diff
@@ -191,6 +191,10 @@ spec:
       fieldRef:
         apiVersion: v1
         fieldPath: status.podIP
+  {{- with $.Values.clusterName }}
+  - name: OTEL_K8S_CLUSTER_NAME
+    value: {{ . | quote }}
+  {{- end }}
```
jaronoff97, line 197: @cyrille-leclerc we set this right below in the otel resource attributes value, do we need both?

cyrille-leclerc: If we want to define an otel col "resource" processor to set k8s.cluster.name, a dedicated env var is much easier to consume. 
Using the resource detection processor with the env detector is challenging because many resource attributes it will try to inject are undesired (k8s.pod.* , even k8s.node.* for cluster scope collectors).
An alternate that came to my mind is to enable in the chart to define static resource attributes but it's a much bigger change as explained in https://github.com/open-telemetry/opentelemetry-helm-charts/issues/2338

Does it make sense?

jaronoff97: hm i think #2338 is probably a better way longer term for this. For trying to set this in resource detection processors can the user not just hardcode the value in the configuration? Or is it preferred to pull it via environment?

jaronoff97: i guess i don't want to set a precedent of putting some values in environment variables and others in the resource attrs variable. Realistically we should be using declarative configuration too.

PR #2341, opened 2026-08-06
fix(opentelemetry-operator): make cert-manager test address configurable

### LINKED ISSUE #2323: Tests for cert-manager should at least allow specifying the address
We have cert-manager deployed in a different namespace than `cert-manager` and this test just can't work -- and the address is hard-coded in the latest versions of _all_ the charts.

E.g.

https://github.com/open-telemetry/opentelemetry-helm-charts/blob/opentelemetry-kube-stack-0.19.2/charts/opentelemetry-operator/templates/tests/test-certmanager-connection.yaml#L18

iamhaseebn:
#### Description

The cert-manager connection test in the OpenTelemetry Operator chart hard-codes the webhook address `cert-manager-webhook.cert-manager`. As a result, `helm test` fails when cert-manager uses a different release name or namespace.

This PR adds `testFramework.certManagerAddress` and uses it in the existing cert-manager connection test. The default remains `cert-manager-webhook.cert-manager`, so existing installations require no changes. Keeping the setting under the existing `testFramework` section also makes its test-only purpose explicit.

Changes:

- Add `testFramework.certManagerAddress` to the chart values and schema.
- Use the configured address in the existing cert-manager connection test.
- Add a CI values case for a non-default address.
- Bump the operator chart version from `0.120.2` to `0.120.3` and regenerate the rendered examples.

Verification: chart-testing lint with all nine operator CI values cases, Helm 4.0.5 lint, schema validation, example generation and operator example checks, and default and custom rendering all pass. Default rendering is unchanged apart from the expected `helm.sh/chart` version labels.

#### Link to tracking issue

Fixes #2323

#### Authorship

### DIFF
--- a/charts/opentelemetry-operator/ci/cert-manager-address-values.yaml
+++ b/charts/opentelemetry-operator/ci/cert-manager-address-values.yaml
@@ -1,2 +1,7 @@
 testFramework:
-  certManagerAddress: cert-manager-webhook.cert-manager.svc
+  certManager:
+    env:
+      - name: CERT_MANAGER_CLUSTERIP
+        value: "cert-manager-webhook.cert-manager.svc"
+      - name: CERT_MANAGER_PORT
+        value: "443"
--- a/charts/opentelemetry-operator/ci/cert-manager-address-values.yaml
+++ b/charts/opentelemetry-operator/ci/cert-manager-address-values.yaml
@@ -1,7 +1,0 @@
-testFramework:
-  certManager:
-    env:
-      - name: CERT_MANAGER_CLUSTERIP
-        value: "cert-manager-webhook.cert-manager.svc"
-      - name: CERT_MANAGER_PORT
-        value: "443"
--- a/charts/opentelemetry-operator/ci/cert-manager-address-values.yaml
+++ b/charts/opentelemetry-operator/ci/cert-manager-address-values.yaml
@@ -1,0 +1,2 @@
+testFramework:
+  certManagerAddress: cert-manager-webhook.cert-manager.svc
--- a/charts/opentelemetry-operator/templates/tests/test-certmanager-connection.yaml
+++ b/charts/opentelemetry-operator/templates/tests/test-certmanager-connection.yaml
@@ -14,10 +14,7 @@ spec:
     - name: wget
       image: "{{ .Values.testFramework.image.repository }}:{{ .Values.testFramework.image.tag }}"
       env:
-        - name: CERT_MANAGER_CLUSTERIP
-          value: {{ .Values.testFramework.certManagerAddress | quote }}
-        - name: CERT_MANAGER_PORT
-          value: "443"
+{{ toYaml .Values.testFramework.certManager.env | indent 8 }}
       command:
         - sh
         - -c
--- a/charts/opentelemetry-operator/templates/tests/test-certmanager-connection.yaml
+++ b/charts/opentelemetry-operator/templates/tests/test-certmanager-connection.yaml
@@ -14,7 +14,7 @@ spec:
     - name: wget
       image: "{{ .Values.testFramework.image.repository }}:{{ .Values.testFramework.image.tag }}"
       env:
-{{ toYaml .Values.testFramework.certManager.env | indent 8 }}
+        {{- toYaml .Values.testFramework.certManager.env | nindent 8 }}
       command:
         - sh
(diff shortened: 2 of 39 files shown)


charts/opentelemetry-operator/values.yaml
```diff
@@ -297,6 +297,8 @@ admissionWebhooks:
   ## certManager must be enabled. If enabled, always takes precedence over options 2 and 3.
   certManager:
     enabled: true
+    ## Address used by the Helm test hook to verify the cert-manager webhook.
+    testConnectionAddress: cert-manager-webhook.cert-manager
```
TylerHelmuth: I really dont want to have to expose a test-only feature. Is there any other way to do this? What do other helm charts in other communities to do configure tests?

iamhaseebn: Good point. Other charts generally test resources created by their own release rather than external dependencies.

The operator chart already tests its webhook service, and the deployment cannot become ready unless cert-manager creates the serving certificate Secret. I updated the PR to remove the direct cert-manager webhook probe instead of adding a test-only value.

This keeps the existing coverage while removing the namespace and release-name coupling. All chart lint, CI values, schema, example, and rendering checks pass.

The kube-stack chart currently depends on operator chart `0.119.0`, so it will need a follow-up dependency bump after `0.120.3` is published.

TylerHelmuth: @iamhaseebn I missed that we already have a `testFramework` section in our values.yaml. Add the configuration there so we can keep the test.

iamhaseebn: Updated as requested. The existing test is retained, and its webhook address is now configured through `testFramework.certManagerAddress`. The custom CI values case and schema are restored, and chart-testing, all nine operator values cases, example checks, and default and custom rendering pass.

charts/opentelemetry-operator/values.yaml
```diff
@@ -398,6 +398,8 @@ securityContext:
 automountServiceAccountToken: true
 
 testFramework:
+  ## Address used by the Helm test hook to verify the cert-manager webhook.
+  certManagerAddress: cert-manager-webhook.cert-manager
```
TylerHelmuth: In case there are other things we need to configure in the future lets make this be something like:

```yaml
certManager:
  env:
    - name: CERT_MANAGER_CLUSTERIP
      value: "cert-manager-webhook.cert-manager"
    - name: CERT_MANAGER_PORT
      value: "443"
```

And then use that to set the env vars in the container in the test

iamhaseebn: Thanks - switched to the nested `testFramework.certManager.env` list and wired the test pod's `env` from it (default `CERT_MANAGER_CLUSTERIP` / `CERT_MANAGER_PORT`), done in 012caa22.

charts/opentelemetry-operator/ci/cert-manager-address-values.yaml
TylerHelmuth, line 1: We dont need a new CI scenario for this

charts/opentelemetry-operator/templates/tests/test-certmanager-connection.yaml
```diff
@@ -14,10 +14,7 @@ spec:
     - name: wget
       image: "{{ .Values.testFramework.image.repository }}:{{ .Values.testFramework.image.tag }}"
       env:
-        - name: CERT_MANAGER_CLUSTERIP
-          value: "cert-manager-webhook.cert-manager"
-        - name: CERT_MANAGER_PORT
-          value: "443"
+{{ toYaml .Values.testFramework.certManager.env | indent 8 }}
```
TylerHelmuth:
```suggestion
        {{- toYaml .Values.testFramework.certManager.env | indent 8 }}
```

charts/opentelemetry-operator/ci/cert-manager-address-values.yaml
iamhaseebn: Updated - thanks!

charts/opentelemetry-operator/templates/tests/test-certmanager-connection.yaml
iamhaseebn: Updated - thanks!

TylerHelmuth (approved):

Issue #1824, opened 2025-09-04
Remove references to Collector's batch processor in opentelemetry-helm-charts


mx-psi:
There are a bunch of references: https://github.com/search?q=org%3Aopen-telemetry+%2Fbatch%3A%2F+%28language%3AYAML+OR+language%3AMarkdown%29+repo%3Aopen-telemetry%2Fopentelemetry-helm-charts+NOT+path%3A%2Frendered%2F&type=code (excluding 'rendered' paths)

I am not sure how easy/desirable it would be to remove the batch processor also from the default configuration?

The idea is to replace as follows:

<table>
  <tr>
<td>

**Before:**

```yaml
receivers:
  foo:

processors:
  batch:

exporters:
  foo:

service:
  pipelines:
    metrics:
      receivers: [foo]
      processors: [batch]
      exporters: [foo]
```

</td>

<td>

**After:**

```yaml
receivers:
  foo:



exporters:
  foo:
    sending_queue:
      batch:

service:
  pipelines:
    metrics:
      receivers: [foo]
      exporters: [foo]

```

</td>
  </tr>
</table>

TylerHelmuth:
We can transition to the default exporters we have using the default exporterhelper batch setup.

dmitryax:
@TylerHelmuth I'm not sure I get why do we need to be so careful here. The default config isn't something that supposed to be used by default. It has debug exporter which isn't even production component. So I expect the pipelines to be overridden by the end users. So why can't we just remove the batch processor from the default pipelines with instructions on how to enable it on exporters? (enabling it on debug exporter doesn't make sense)

TylerHelmuth:
We could, but its a pretty big breaking change. The problem I see today is that it is really easy to be using the default batchexporter with a customer exporter. If we remove it and leave users with instructions a lot of users will need to go update their configured exporters.

We have started doing a lot more config manipulation as part of handling the component renames tho. I think instead of my original plan we could (in a single PR):

- remove the default batchprocessor
- add logic to automatically configure the batching settings in any exporter we see in the .Values that does not already have it configured.
- add a toggle to disable this functionality

This of course assumes that every exporter can have batching configure. What do you think?

dmitryax:
Do we have any precedence of manipulating user-provided config? I don't think it's something that users expect?

[...]

PR #2343, opened 2026-08-08
feat(opentelemetry-collector): render telemetry resource attributes as a list

### LINKED ISSUE #2250: [opentelemetry-collector] Pod fails to start when node name is digits only

### LINKED ISSUE #2255: Deprecating warning for legacy `service.telemetry.resource` inline map format

nicknikolakakis:
#### Description

The collector chart ships `service.telemetry.resource` as an inline map. The Collector deprecated that format in v0.123.0 and logs a warning at every start.

This replaces #2283, which took the straight migration route. Per the review discussion there with @TylerHelmuth and @jaronoff97, the defaults now live in a new top-level `telemetryResourceAttributes` map, outside `.Values.config`, and the chart renders them into `config.service.telemetry.resource.attributes`. Keeping the input as a map means a single attribute can be overridden without re-declaring the defaults, which was the main objection to switching to a list.

Existing overrides do not break. A map left under `config.service.telemetry.resource` is detected with `kindIs` and folded into the rendered list, and it still wins over the defaults. A list supplied at `config.service.telemetry.resource.attributes` is passed through and appended.

Values are rendered as strings, so an all-digits node name no longer produces a config the collector rejects at startup.

Behaviour covered by `helm template`:

| Input | Result |
| --- | --- |
| defaults | list of `{name, value}`, alpha-sorted |
| map override under `config.service.telemetry.resource` | merged in, wins over defaults |
| numeric value | rendered quoted |
| `null` value | attribute dropped |
| explicit list at `resource.attributes` | passed through, appended |

`values.schema.json` gained the new key because the chart sets `additionalProperties: false`. Chart bumped to 0.170.0, `UPGRADING.md` entry added, examples regenerated (`make check-examples`: 22 passed, 0 failed).

#### Link to tracking issue
Fixes #2255
Closes #2250

#### Authorship

- [ ] I, a human, wrote this pull request description myself.


charts/opentelemetry-collector/values.yaml
```diff
@@ -199,6 +199,19 @@ internalTelemetryViaOTLP:
     # overrides internalTelemetryViaOTLP.headers for logs
     headers: []
 
+# Resource attributes attached to the collector's own internal telemetry.
+# Rendered into config.service.telemetry.resource.attributes, which is the
+# format the Collector expects since v0.123.0. Kept as a map so a single
+# attribute can be overridden without re-declaring the defaults; set a value to
+# null to drop that attribute.
```
TylerHelmuth, line 206: The `null` bug came back in helm v4.2.0 so we'll want to have an explicit on/off capability so that attributes can be disabled when this chart is used in a subchart. Maybe something like:

```
telemetry:
  defaultResourceAttributes:
    k8s.namespace.name: 
      enabled: true
      value: "${env:OTEL_K8S_NAMESPACE}"
  additionalResourceAttributes: {}
  # foo: bar
``` 

@open-telemetry/helm-approvers please review.

jaronoff97:
Do we have an example showing the dual/three states these can be in now?

charts/opentelemetry-collector/values.yaml
```diff
@@ -204,6 +204,36 @@ internalTelemetryViaOTLP:
     # overrides internalTelemetryViaOTLP.headers for logs
     headers: []
 
+# Resource attributes attached to the collector's own internal telemetry.
+# Rendered into config.service.telemetry.resource.attributes, which is the
+# format the Collector expects since v0.123.0.
+telemetry:
+  # Attributes the chart sets by default. Set enabled to false to drop one.
+  # An explicit flag is used rather than a null value because null does not
+  # reliably delete a subchart default (helm/helm#31943).
```
TylerHelmuth:
```suggestion
```

charts/opentelemetry-collector/values.yaml
```diff
@@ -204,6 +204,36 @@ internalTelemetryViaOTLP:
     # overrides internalTelemetryViaOTLP.headers for logs
     headers: []
 
+# Resource attributes attached to the collector's own internal telemetry.
+# Rendered into config.service.telemetry.resource.attributes, which is the
+# format the Collector expects since v0.123.0.
```
TylerHelmuth:
```suggestion
# Rendered into config.service.telemetry.resource.attributes.
```

charts/opentelemetry-collector/examples/telemetry-resource-attributes/values.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+telemetry:
+  defaultResourceAttributes:
+    host.name:
+      enabled: false
+    k8s.pod.ip:
+      enabled: false
+    k8s.namespace.name:
+      value: "${env:MY_NAMESPACE_OVERRIDE}"
+  additionalResourceAttributes:
```
TylerHelmuth: What happens if you set an additional attribute that matches a default attribute?

```yaml
additionalResourceAttributes:
  k8s.namespace.name: "foo"
```

charts/opentelemetry-collector/values.yaml
```diff
@@ -204,6 +204,36 @@ internalTelemetryViaOTLP:
...
+telemetry:
+  # Attributes the chart sets by default. Set enabled to false to drop one.
+  # An explicit flag is used rather than a null value because null does not
+  # reliably delete a subchart default (helm/helm#31943).
+  defaultResourceAttributes:
```
TylerHelmuth: In anticipation of doing more with `telemetry` in the future what do you think of nesting this one additional layer to keep the grouping explicit?

```yaml
telemetry:
  resourceAttributes:
    default:
    extra:
```

TylerHelmuth:
@nicknikolakakis once you handle the merge conflicts I'll be ready to merge

charts/opentelemetry-collector/templates/_config.tpl
```diff
@@ -40,11 +40,65 @@ logs:
...
+{{- $service := $config.service | default dict }}
+{{- $telemetry := $service.telemetry | default dict }}
+{{- $resource := include "opentelemetry-collector.telemetryResource" . | fromYaml }}
+{{- if $resource }}
+{{- $_ := set $telemetry "resource" $resource }}
+{{- else }}
+{{- $_ := unset $telemetry "resource" }}
+{{- end }}
+{{- $_ := set $service "telemetry" $telemetry }}
+{{- $_ := set $config "service" $service }}
```
TylerHelmuth, line 110: Can you run a test where the user is explicitly setting  a list of resource attributes via `.Values.config.service.telemetry.resource`? I am curious how the new merging logic interacts with that use case.

marcschaeferger:
@TylerHelmuth Conflicts are solved, could you pls take a look.

JaredTan95 (commented):
LGTM

but you need to resolve conflicts again.

charts/opentelemetry-collector/templates/_config.tpl
TylerHelmuth: @nicknikolakakis were you able to run this test, what was the outcome?

TylerHelmuth (changes requested):
Blocking while we have discussions about the future of `.Values.config` over in https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1824

PR #2363, opened 2026-08-21
fix(opentelemetry-collector): preserve file_log overrides in logsCollection

### LINKED ISSUE #2321: [opentelemetry-collector] presets.logsCollection is using deprecated filelog name instead of file_log
Currently if `presets.logsCollection.enabled = true` then at startup helm injects a `filelog` receiver. If you have also explicitly added a `file_log` receiver to your templatel this will cause an orphaned `filelog` receiver to be created.

We can test this with `helm template test ...` for example, if our helm template contains:

```yaml
daemon:
   presets:
      logsCollection:
         enabled: true
   receivers:
      file_log:
```

We render the template with `helm template test ...`:

```bash
helm template test open-telemetry/opentelemetry-kube-stack --values deploy/helm/edot-collector/kube-stack/values.yaml 2>/dev/null | grep -E "^\s+(filelog|file_log):"
      file_log:
      filelog:
```

You can see that two file logs receivers are rendered.

A workaround for this is to use the deprecated name `filelog` receiver:

```yaml
daemon:
   presets:
      logsCollection:
         enabled: true
   receivers:
      filelog:
```

And if we render the template again:

```bash
helm template test open-telemetry/opentelemetry-kube-stack --values deploy/helm/edot-collector/kube-stack/values.yaml 2>/dev/null | grep -E "^\s+(filelog|file_log):"
      filelog:
```

However, the issue with this is now a warning msg is logged regarding using a deprecated name `filelog`.

Can we update the `filelog` name in [opentelemetry-collector/templates/_config.tmpl](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/templates/_config.tpl#L267) to its canonical name `file_log`?

marcschaeferger:
#### Description

Fixes a regression introduced by #2340 where `presets.logsCollection` can silently overwrite an explicitly configured `file_log` receiver when `rewriteDeprecatedComponentNames` is enabled.

The preset now selects `file_log` or `filelog` based on `rewriteDeprecatedComponentNames` before merging user configuration. This preserves user-provided `exclude` and `operators` values while keeping the legacy behavior unchanged when rewriting is disabled.

Regression coverage was added for canonical `file_log` overrides and the legacy receiver-name path. The chart version is bumped to `0.170.1` and examples are regenerated.

#### Link to tracking issue

Fixes https://github.com/open-telemetry/opentelemetry-helm-charts/issues/2321

#### Authorship

---
<a href="https://gitkraken.dev/review/github/open-telemetry/opentelemetry-helm-charts/pull/2363?source=pr_review_chip">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://gitkraken.dev/images/figures/gitkraken-review-badge-dark.svg">
    Open with GitKraken
  </picture>
</a>


marcschaeferger:
Intentionally no Git Tag as there might be other Features for the `opentelemetry-collector-0.170.1` Chart Release.

charts/opentelemetry-collector/examples/README.md
```diff
@@ -6,6 +6,8 @@ Here is a collection of common configurations for the OpenTelemetry collector. E
 - [Deployment only](deployment-only)
 - [Daemonset and deployment](daemonset-and-deployment)
 - [Log collection, including collector logs](daemonset-collector-logs)
+- [Log collection with a custom `file_log` receiver configuration](daemonset-logs-collection-custom-receiver)
+- [Log collection using the deprecated `filelog` receiver name](daemonset-logs-collection-legacy-name)
```
TylerHelmuth: Since the situation of support both names is temporary we can exclude new examples.

petewall (approved):
Thanks for fixing this!

marcschaeferger:
ok, both new example directories are gone and examples/README.md is back to its original state.

To keep the regression from returning silently, I moved the assertion into the existing daemonset-collector-logs example instead of adding a new one.

TylerHelmuth (approved):

PR #2373, opened 2026-08-27
feat(opentelemetry-demo): add BROWSER_GENERATOR_VUS env var to load-generator

### LINKED ISSUE #3813: Load-generator browser scenario shares one browser context across all iterations breaks RUM agents session tracking
browserScenario() in src/load-generator/script.js hardcodes vus: 1 for the browser scenario, independent of the LOAD_GENERATOR_VUS env var that controls the HTTP scenario's scale. This means only a single simulated browser session ever runs, no matter how much load-generator throughput or replica count is scaled up for the rest of the demo — there's no way to generate more than one concurrent browser/RUM session.

This makes it hard to load-test or exercise a browser-based RUM/analytics integration at any real scale: the browser side of the load generator is effectively fixed at "one user, forever," while every other part of the traffic generator scales normally.

What is the expected behavior?

The browser scenario's VU count should be independently configurable via an env var (following the same pattern LOAD_GENERATOR_VUS already uses for the HTTP scenario), so it can be scaled up separately from HTTP load.

What is the actual behavior?

vus: 1 is hardcoded in the browser scenario config; there is no way to run more than one concurrent browser session without editing script.js directly.

Reproduce

Run the demo with K6_BROWSER_ENABLED=true.
Try to increase concurrent browser/RUM traffic via any existing env var (LOAD_GENERATOR_VUS only affects httpScenario).
Observe that the browser scenario always runs exactly one VU regardless of configuration.
Proposed fix: add a BROWSER_GENERATOR_VUS env var (defaulting to 1 to preserve current behavior) and use it for the browser scenario's vus, mirroring how LOAD_GENERATOR_VUS already works for httpScenario: vus: parseInt(__ENV.BROWSER_GENERATOR_VUS || '1'). Branch: https://github.com/dpacheconr/opentelemetry-demo/tree/fix/k6-browser-vus-configurable

---
Assisted-by: Claude Sonnet 5

dpacheconr:
#### Description
Adds a `BROWSER_GENERATOR_VUS` env var to the load-generator's `values.yaml` env list, letting the k6 browser/RUM scenario scale independently of `LOAD_GENERATOR_VUS`. Default `"1"`, matching the script's own fallback, so chart behavior is unchanged until overridden.

**Depends on open-telemetry/opentelemetry-demo#3903 (must merge first)** — that PR adds the `BROWSER_GENERATOR_VUS` support in `script.js`; without it this env var is a no-op.

#### Link to tracking issue
Fixes open-telemetry/opentelemetry-demo#3813

#### Authorship

---
Assisted-by: Claude Sonnet 5

### DIFF
--- a/charts/opentelemetry-demo/values.yaml
+++ b/charts/opentelemetry-demo/values.yaml
@@ -638,6 +638,8 @@ components:
     env:
       - name: LOAD_GENERATOR_VUS
         value: "5"
+      - name: BROWSER_GENERATOR_VUS
+        value: "1"
       - name: K6_TARGET_URL
         value: http://frontend-proxy:8080
       - name: K6_BROWSER_ENABLED


TylerHelmuth:
Closing for now since the demo PR is closed.

PR #2374, opened 2026-08-27
Add an interval setting to the k8sobjects preset

agagniere:
#### Description

Allows setting an interval to the pull-mode entries of the `k8s_objects` receiver via the `kubernetesObjects` preset.

Currently we cannot use the preset if we want a non-default pull interval. And the default interval is 1h.

This PR adds a single setting:

```yaml
presets:
  kubernetesObjects:
    enabled: true
    watch: true
    interval: 30s # <- instead of 1h
```

#### Authorship

### DIFF
--- a/charts/opentelemetry-collector/values.yaml
+++ b/charts/opentelemetry-collector/values.yaml
@@ -84,6 +84,8 @@ presets:
     # disableLeaderElection: false
     # When enabled, watch mode is added alongside pull to stream real-time changes.
     watch: false
+    # Sets the polling interval for objects collected in pull mode (e.g. "30m"). Defaults to the receiver's default (1h) when unset.
+    interval: ""
     # Core Kubernetes workload resources: pods, nodes, namespaces, services, serviceaccounts, deployments, replicasets, daemonsets, statefulsets, jobs, cronjobs
     core:
       enabled: true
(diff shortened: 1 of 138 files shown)


charts/opentelemetry-collector/values.yaml
```diff
@@ -84,6 +84,8 @@ presets:
     # disableLeaderElection: false
     # When enabled, watch mode is added alongside pull to stream real-time changes.
     watch: false
+    # Sets the polling interval for objects collected in pull mode (e.g. "30m"). Defaults to the receiver's default (1h) when unset.
+    interval: ""
```
TylerHelmuth, line 88: We purposefully make the presets very very opinionated and we do not want to expose component config through them. In general presets should be all boolean configuration options. 

The best way to achieve this same outcome while using the preset is to combine the preset with config:

```yaml
presets:
  kubernetesObjects:
    enabled: true
config:
  receivers:
    k8sobjects:
      interval: 30m
```

TylerHelmuth (changes requested):
Blocking because we shouldn't expose component config in a preset. Instead we could support meaningful duration presets.

charts/opentelemetry-collector/values.yaml
agagniere: Good solution.
My previous attempt to set only certain values in the config and use presets at the same time resulted in the config getting overridden by the preset's config, which is why I had given up on it.
But in this case the field is unset by the preset, so it works nicely 👍 

I'll just have to be on the lookout for when the chart will update the receiver name from `k8sobjects` to `k8s_objects`

agagniere:
There is a way to customize this setting while also benefiting from the preset

PR #2377, opened 2026-08-28
fix(opentelemetry-kube-stack): allow instrumentation.java.extensions in values.schema.json

sakshitposting-irl:
#### Description

The Instrumentation CRD supports `spec.java.extensions` for loading Java agent extensions via an array of `{image, dir}` objects. However, the `opentelemetry-kube-stack` values schema did not include this field while setting `additionalProperties: false` on the Java schema. As a result, Helm rejected valid configurations during schema validation before rendering any manifests.

This PR adds an `extensions` property to the Java schema definition that mirrors the Instrumentation CRD:

* `extensions` is an array.
* Each item requires `image` and `dir`, both strings.
* Item objects disallow additional properties, consistent with the rest of the schema.

This is a backward-compatible change. Existing values files continue to validate unchanged, while users can now configure a field already supported by the OpenTelemetry Operator.

As required for chart changes, this PR also:

* Bumps the chart version from `0.20.5` to `0.20.6`.
* Regenerates the rendered example manifests, updating only the `helm.sh/chart` label to the new chart version.

#### Validation

* Reproduced the original Helm schema validation failure when `instrumentation.java.extensions` was present.
* Verified that the same values file now renders successfully and the generated `Instrumentation` resource contains `spec.java.extensions`.
* Confirmed invalid configurations are still rejected (missing required fields, incorrect types, unknown properties, and incorrect array shape).
* Verified the default `values.yaml` continues to validate.
* Ran `make check-examples CHARTS=opentelemetry-kube-stack` successfully (`11/11` passing).

#### Link to tracking issue

N/A

#### Authorship

* [x] I, a human, wrote this pull request description myself.

### DIFF
--- a/charts/opentelemetry-kube-stack/Chart.yaml
+++ b/charts/opentelemetry-kube-stack/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-kube-stack
-version: 0.20.5
+version: 0.20.6
 description: |
   OpenTelemetry Quickstart chart for Kubernetes.
   Installs an operator and collector for an easy way to get started with Kubernetes observability.
--- a/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/bridge.yaml
+++ b/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/bridge.yaml
@@ -5,7 +5,7 @@ kind: OpAMPBridge
 metadata:
   name: example
   labels:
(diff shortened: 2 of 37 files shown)


jaronoff97 (approved):

sakshitposting-irl:
Hi @jaronoff97 
Just nudging for the lint-test, so we can close this PR
Thanks

jaronoff97 (approved):

jaronoff97:
@sakshitposting-irl thank you for your contribution 🙇

PR #2380, opened 2026-08-30
fix(schema): remove required apiVersion value

baprx:
#### Description
I've just updated the values schema, I wasn't sure if a Chart bump was required, I can revert it if you prefer.
`.Values.apiVersion` is set to a default value so the JSON schema should not require it to be set in user configurations.

#### Authorship

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.172.0
+version: 0.172.1
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-collector/examples/alternate-config/rendered/clusterrole.yaml
+++ b/charts/opentelemetry-collector/examples/alternate-config/rendered/clusterrole.yaml
@@ -5,7 +5,7 @@ kind: ClusterRole
 metadata:
   name: example-opentelemetry-collector
   labels:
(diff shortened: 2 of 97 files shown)


JaredTan95:
why we need to remove `apiversion` from schema?

baprx:
> why we need to remove `apiversion` from schema?

It's not removed it's still [here](https://github.com/baprx/opentelemetry-helm-charts/blob/5c593c857f99ce142b33e156a2207e44b5b3c943/charts/opentelemetry-collector/values.schema.json#L43-L46). I removed it from the required fields because the field is an override and if we're fine with using the [default value](https://github.com/baprx/opentelemetry-helm-charts/blob/5c593c857f99ce142b33e156a2207e44b5b3c943/charts/opentelemetry-collector/values.yaml#L12) because of this required field we still have to provide it.

TylerHelmuth:
@baprx it would be dangerous to remove `apiVersion` from the required values because a value is necessary to successfully create the deployment/daemonset/statefulset objects. It has a default value tho so the only way to fail the json schema to to null it. Failing in that scenario is good because the resulting render wouldnt be installable.

What issue are you running into with it being required?

baprx:
@TylerHelmuth the issue is not blocking on my side, just a bit annoying since the schema validation flags the whole file as errored in my IDE because of this field not being explicitly configured but I can ignore it you don't think merging this is a good idea.

I thought this would be an acceptable change since explicitly setting it to null would likely not happen or lead to an error anyway (either erroring because of the schema or because missing to create the deployment/daemonset/statefulset objects).

TylerHelmuth:
We like to rely on the json schema whenever possible for validation so keep apiVersion required is necessary.

PR #2385, opened 2026-09-03
fix(opentelemetry-demo): add missing agent/load-generator env vars

dpacheconr:
## Description

`agent` and `load-generator` components are missing env vars needed to
actually run once deployed via this chart — found while testing
open-telemetry/opentelemetry-demo#3933, but independent of it:

- `agent`: no explicit `FLAGD_HOST`/`FLAGD_PORT`. Any code in `agent` that
  reads `FLAGD_PORT` and parses it as an int (e.g. an OpenFeature flagd
  provider) breaks, because Kubernetes auto-injects a legacy
  `FLAGD_PORT=tcp://<cluster-ip>:8013` env var for any Service named `flagd`
  in the namespace, and nothing in the chart currently shadows it with the
  plain numeric value other flagd-consuming components (e.g.
  `recommendation`) already set explicitly.
- `load-generator`: only sets k6-oriented env (`K6_TARGET_URL`,
  `K6_OTEL_*`, ...), but the currently-shipped `src/load-generator` is
  Locust-based and needs `LOCUST_HOST`/`LOCUST_HEADLESS`/`LOCUST_AUTOSTART`/
  `LOCUST_USERS` to actually generate traffic instead of sitting idle.

## Testing

Deployed this chart locally against minikube with both fixes; confirmed
`agent` no longer crashes on startup, and `load-generator` runs headless and
sends traffic per `LOCUST_USERS`.

## Related

- open-telemetry/opentelemetry-demo#3933 (issue)
- open-telemetry/opentelemetry-demo#3934 (adds the `aiSlowResponse`/
  `aiRunawayAgent` flags this env var fix was found while testing)


JaredTan95 (approved):

TylerHelmuth:
@dpacheconr I merged in another demo change first, place bump the chart version again and regenerate the examples.

charts/opentelemetry-demo/examples/bring-your-own-observability/rendered/component.yaml
```diff
@@ -839,6 +843,10 @@ spec:
               value: "0.85"
             - name: LLM_TLS_VERIFY
               value: "True"
+            - name: FLAGD_HOST
+              value: flagd
+            - name: FLAGD_PORT
+              value: "8013"
```
TylerHelmuth: This is a duplicate in the same list.

TylerHelmuth (commented):
@dpacheconr your CI is failing bc the env vars are added to the same list twice

charts/opentelemetry-demo/examples/bring-your-own-observability/rendered/component.yaml
dpacheconr: thank you, now resolved

PR #2393, opened 2026-09-07
 feat(kube-stack): add top-level resourceAttributes

### LINKED ISSUE #2338: [opentelemetry-kube-stack] Add top-level resourceAttributes to set default resource attributes across collectors and Instrumentation

kangyili:
#### Description

Adds a top-level `resourceAttributes: {}` map to the `opentelemetry-kube-stack`
chart, providing a single place to set default resource attributes (e.g.
`deployment.environment.name`, `k8s.cluster.name`) that apply to both the collectors this
chart deploys and the `Instrumentation` CR it manages. Previously, users had
to duplicate the same attributes in two unrelated places: manually as a
`resource` processor in each collector's config, and again in
`instrumentation.resource.resourceAttributes`.

Wiring:

- **Collectors:** the map is rendered as a `resource/global` processor with
  `action: upsert` and appended to every existing pipeline (traces, metrics,
  logs). `upsert` semantics ensure explicit values here take precedence over
  anything auto-detected earlier in the pipeline (e.g. by
  `resourcedetection`).
- **Instrumentation CR:** the map is merged as defaults into
  `instrumentation.resource.resourceAttributes`; any per-key value already set
  in the instrumentation section wins on conflict, preserving existing
  behavior for users who already configure it there.
- **Per-collector override:** `collectors.<name>.resourceAttributes` is also
  accepted and overlays the top-level map (per-collector keys win) so a
  specific collector pool can extend or override the defaults.

#### Link to tracking issue
Fixes #2338

#### Authorship

### DIFF
--- a/charts/opentelemetry-kube-stack/Chart.yaml
+++ b/charts/opentelemetry-kube-stack/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-kube-stack
-version: 0.20.6
+version: 0.20.7
 description: |
   OpenTelemetry Quickstart chart for Kubernetes.
   Installs an operator and collector for an easy way to get started with Kubernetes observability.
--- a/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/bridge.yaml
+++ b/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/bridge.yaml
@@ -5,7 +5,7 @@ kind: OpAMPBridge
 metadata:
   name: example
   labels:
-    helm.sh/chart: opentelemetry-kube-stack-0.20.6
+    helm.sh/chart: opentelemetry-kube-stack-0.20.7
     app.kubernetes.io/version: "0.154.0"
     app.kubernetes.io/managed-by: Helm
     release: "example"    
--- a/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/collector.yaml
+++ b/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/collector.yaml
@@ -6,7 +6,7 @@ metadata:
   name: example-daemon
   namespace: default
   labels:
-    helm.sh/chart: opentelemetry-kube-stack-0.20.6
+    helm.sh/chart: opentelemetry-kube-stack-0.20.7
     app.kubernetes.io/version: "0.154.0"
     app.kubernetes.io/managed-by: Helm
     release: "example"        
--- a/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/hooks.yaml
+++ b/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/hooks.yaml
@@ -61,4 +61,4 @@ spec:
           - "delete"
           - "instrumentations,opampbridges,opentelemetrycollectors"
           - "-l"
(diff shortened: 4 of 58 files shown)


kangyili:
cc @cyrille-leclerc

jaronoff97 (approved):

jaronoff97:
thank you for your contribution! :bow:

PR #2394, opened 2026-09-08
[opentelemetry-kube-stack] new inheritDefaultCRConfig option

wehzzz:
#### Description

We want to port the ebpf profiler effort introduced in https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2287 and https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2126 and  to the `kube-stack `chart.

The profiling receiver runs as a dedicated distribution (`otelcol-ebpf-profiler`), not as part of the general-purpose k8s distro. As discussed in https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2126#discussion_r3065602137, the eBPF profiler needs hostPID and elevated capabilities, so bundling it with the collector that handles metrics, traces and logs would grant those privileges to that same collector. The dedicated distribution also ships a subset of the components.

`kube-stack` currently merges `defaultCRConfig` into every collector, and that config is written for the general purpose distro. Deploying a profiler collector today crashes it at startup, on components it inherited but cannot build.

This PR adds `collectors.<name>.inheritDefaultCRConfig` (default `true`). When `false`, the collector no longer inherits `config`, `presets`, `scrape_configs_file` and `targetAllocator` from `defaultCRConfig`. It keeps the other defaults (`mode`, `image`, `env`, `resources`, `clusterRoleBinding`, ...) and defines its own config. The profiling preset itself lands in a follow-up PR stacked on this one.

Rest of the diff is groundwork for that option:
 - merge logic now lives in one helper, `opentelemetry-kube-stack.mergeCollector`.
 - presets can now be absent from a collector, so preset lookups use dig with the documented defaults instead of direct field access
 - config can be absent too, so the `k8s_cluster` receiver detection read it through `config | default dict`

#### Authorship

### DIFF
--- a/charts/opentelemetry-kube-stack/Chart.yaml
+++ b/charts/opentelemetry-kube-stack/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-kube-stack
-version: 0.20.6
+version: 0.20.7
 description: |
   OpenTelemetry Quickstart chart for Kubernetes.
   Installs an operator and collector for an easy way to get started with Kubernetes observability.
--- a/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/bridge.yaml
+++ b/charts/opentelemetry-kube-stack/examples/cloud-demo/rendered/bridge.yaml
@@ -5,7 +5,7 @@ kind: OpAMPBridge
 metadata:
   name: example
   labels:
-    helm.sh/chart: opentelemetry-kube-stack-0.20.6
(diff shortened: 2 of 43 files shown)


jaronoff97 (approved):
change makes sense to me and i appreciate the use of `dig` to help clean this up as well. Will there be a follow up to add in a profiler example? I think it would be good to be able to point users to one.

wehzzz:
Yes, absolutely. I'll open a follow-up PR to add the profiling preset along with an example.

PR #2396, opened 2026-09-09
[opentelemetry-kube-stack] add profiling support

wehzzz:
#### Description

This is the next step in porting the profiling preset and security context required for the eBPF profiler, following open-telemetry/opentelemetry-helm-charts#2394.
This PR mirrors the work done in #2287 and #2126 to add the eBPF profiler to the opentelemetry-collector chart.

Added an example to illustrate how to define a profiler config alongside another daemon config.

#### Link to tracking issue
Fixes

#### Authorship


jaronoff97 (commented):
@wehzzz thank you for the updates here! I jsut merged a PR for OKS, mind updating the patch version here?

wehzzz:
It should be all set with commit b30737d @jaronoff97 ! Thanks!

jaronoff97 (approved):
I think this looks good, but i would appreciate another review from someone in the ebpf / profiler group to confirm that. Mind tagging in a couple more people to confirm?

wehzzz:
> I think this looks good, but i would appreciate another review from someone in the ebpf / profiler group to confirm that. Mind tagging in a couple more people to confirm?

cc @florianl @gandem if you have time for a quick review! These are essentially the same changes you reviewed in #2126.

(also cc @cyrille-leclerc)

florianl (approved):
Not a Helm expert here 😅  but the changes look solid to me!

charts/opentelemetry-kube-stack/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-kube-stack
-version: 0.20.8
+version: 0.20.9
```
TylerHelmuth: This is a big feature add, lets bump the minor version

charts/opentelemetry-kube-stack/templates/_config.tpl
```diff
@@ -196,6 +200,22 @@ receivers:
...
+{{- $_ := set $config.service.pipelines.profiles "exporters" (list "debug") }}
+{{- end }}
+{{- $config | toYaml }}
+{{- end }}
+
+{{- define "opentelemetry-kube-stack.collector.profilingConfig" -}}
+receivers:
+  profiling: {}
+exporters:
+  debug: {}
```
TylerHelmuth, line 216: Why does the profiling preset need to enable a debug exporter?

wehzzz: If no debug exporter is defined, the profiles pipeline won't start. If the user defines custom exporters, they take precedence, but the only drawback is that we still declare the debug exporter in that case.

I left it as-is in this PR to match the exact behavior implemented in https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2126.

If we want to revisit this, I suggest handling it in a separate PR for both charts, as it feels out of scope here.

TylerHelmuth (approved):

PR #2402, opened 2026-09-11
fix(opentelemetry-collector): Hash only the ConfigMap data in checksum/config

DrFaust92:
#### Description

`checksum/config` hashes the entire rendered ConfigMap, and the manifest's `metadata.labels` include `helm.sh/chart`. That value changes on every chart version bump, so the annotation changed on every upgrade and rolled the collector pods even when the configuration was identical — in daemonset mode, that's every node in the cluster.

This adds an `opentelemetry-collector.configChecksum` helper that hashes only the `data`/`stringData` of the rendered ConfigMap, and uses it in `opentelemetry-collector.configTemplateChecksumAnnotation` for all four branches (daemonset, deployment, statefulset, and `configMap.existingPath`). A pure version bump no longer changes the checksum; any real config or preset change still does. Upgrading to 0.173.1 causes one final restart as the annotation value changes.

The argo-cd chart fixed the same issue in argoproj/argo-helm#4044, and the loki chart uses the same data-only hashing in `loki.configMapOrSecretContentHash`.

#### Link to tracking issue
Fixes

#### Authorship

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: opentelemetry-collector
-version: 0.173.0
+version: 0.173.1
 description: OpenTelemetry Collector Helm chart for Kubernetes
 type: application
 home: https://opentelemetry.io/
--- a/charts/opentelemetry-collector/examples/alternate-config/rendered/clusterrole.yaml
+++ b/charts/opentelemetry-collector/examples/alternate-config/rendered/clusterrole.yaml
@@ -5,7 +5,7 @@ kind: ClusterRole
 metadata:
   name: example-opentelemetry-collector
   labels:
-    helm.sh/chart: opentelemetry-collector-0.173.0
+    helm.sh/chart: opentelemetry-collector-0.173.1
     app.kubernetes.io/name: opentelemetry-collector
     app.kubernetes.io/instance: example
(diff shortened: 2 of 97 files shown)


TylerHelmuth:
This feels right, but I'm curious what is happening during a `helm upgrade` and the chart version is bumped but the pod doesn't restart. Is it to stay on the latest versions (even if they contain changes you're not using)? If the chart's appVersion bumps but the configmap is unchanges and you do a `helm upgrade` will it still roll the pod so you get the latest collector version?

DrFaust92:
The annotation's job is to restart pods when the config content changes, and that still happens.

A chart-version-only bump touches nothing inside `spec.template`. Rendering 0.173.1 vs 0.174.0 with identical values gives a 3-line diff, all `helm.sh/chart` on object-level metadata (ServiceAccount, ConfigMap, DaemonSet), and the pod template starts below all of them — so the DaemonSet is updated in place and the pods keep running.

An appVersion bump still rolls the pods: the default image tag comes from `.Chart.AppVersion`, so `image:` inside the pod spec changes and Kubernetes recreates them to pick up the new collector. The checksum annotation was never what drove that.

The motivation is that we run the collector as a DaemonSet on every node and take chart bumps from Renovate. Today every one of those rolls the whole fleet even when neither the config nor the collector image changed.

TylerHelmuth (approved):

PR #2403, opened 2026-09-13
feat(operator): decouple admission webhook selectors

### LINKED ISSUE #2146: [opentelemetry-operator] Decouple the namespaceSelector and objectSelector for admissionWebhooks
Now the 3 kinds of admissionWebhooks share the same set of namespaceSelector and objectSelector, see https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook.yaml.

This may not work well in some cases. For example: 

* the o11y admins create instrumentation CR  in `infra` namespace, which is shared by the workloads of the entire cluster
* the o11y admins want to prevent pods in the `infra`, `argocd`, and some other namespaces from being mutated by the operator, so they set the namespaceSelector for the admissionWebhooks. However, this namespaceSelector also updates the instrumentation webhook, which is not  wanted by the admin.

szelese:
#### Description

Decouples the `namespaceSelector` and `objectSelector` settings for the admission webhook types:

- Pods
- Instrumentation
- OpenTelemetryCollector

Per-webhook selectors are merged with the common top-level selectors. When keys conflict, the more specific per-webhook value takes precedence. Empty per-webhook selectors preserve backward-compatible fallback behavior.

## Tests

- `pre-commit run --all-files`
- `helm lint` with cert-manager enabled and disabled
- `helm template` with per-webhook selectors
- `ct lint --target-branch main --charts charts/opentelemetry-operator`

#### Link to tracking issue
Fixes #2146 

#### Authorship


charts/opentelemetry-operator/values.yaml
```diff
@@ -282,6 +282,20 @@ admissionWebhooks:
   ## - Fail, the injection is fail-close. If the webhook pod is not ready, pods cannot be created.
   pods:
     failurePolicy: Ignore
+    namespaceSelector: {}
+    objectSelector: {}
+
+  ## Selectors for Instrumentation webhooks.
+  ## Empty maps inherit the top-level selectors.
```
TylerHelmuth: Lets specifically say admissionWebhooks.pods instead of top-level

charts/opentelemetry-operator/values.yaml
```diff
@@ -282,6 +282,20 @@ admissionWebhooks:
...
+  instrumentation:
+    namespaceSelector: {}
+    objectSelector: {}
+
+  ## Selectors for OpenTelemetryCollector webhooks.
+  ## Empty maps inherit the top-level selectors.
```
TylerHelmuth: Lets specifically say admissionWebhooks.pods instead of top-level

TylerHelmuth (commented):
1 nit otherwise looks good

charts/opentelemetry-operator/values.yaml
szelese: Updated both, thanks!

charts/opentelemetry-operator/values.yaml
TylerHelmuth: Add a comment that specifies what all these apply to.

charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook-with-cert-manager.yaml
```diff
@@ -19,13 +19,13 @@ webhooks:
         port: {{ .Values.admissionWebhooks.servicePort }}
     failurePolicy: {{ .Values.admissionWebhooks.failurePolicy }}
     name: minstrumentation.kb.io
-    {{- if .Values.admissionWebhooks.namespaceSelector }}
+    {{- if or .Values.admissionWebhooks.instrumentation.namespaceSelector .Values.admissionWebhooks.namespaceSelector }}
     namespaceSelector:
-    {{- toYaml .Values.admissionWebhooks.namespaceSelector | nindent 6 }}
+    {{- toYaml (default .Values.admissionWebhooks.namespaceSelector .Values.admissionWebhooks.instrumentation.namespaceSelector) | nindent 6 }}
```
TylerHelmuth: Do we need to pick one or the other? It would be nice to support `.Values.admissionWebhooks.namespaceSelector` and `.Values.admissionWebhooks.instrumentation.namespaceSelector` at the same time. This way common selectors for all objects can be configured in one place and then only instrumentation-specific selectors need to go in `.Values.admissionWebhooks.instrumentation.namespaceSelector`.

TylerHelmuth (dismissed):

charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook-with-cert-manager.yaml
szelese: Yes, both can be used together; per-webhook selectors override the top-level defaults.

TylerHelmuth: I mean is it possible to take values from both instead of overriding?

szelese: Yes, we can merge them. The top-level selector would provide common constraints, while the per-webhook selector would add to them. We would just need to define how conflicting values should be handled.

TylerHelmuth: Lets have the more specific sections take precedence.

szelese: Implemented as discussed — more specific selectors now take precedence when merged.

Allex1 (approved):

TylerHelmuth (approved):

PR #2404, opened 2026-09-14
[chore]: remove duplicate dependency builds when checking examples

TylerHelmuth:
#### Description

Move the `helm dependency build` call to be per chart instead of per example. This'll hopefully speed up CI a little.

#### Authorship


Allex1 (approved):
pls fix typo in the commit message

PR #2411, opened 2026-09-16
feat(opentelemetry-kube-stack): Add secret-based auth to control-plane ServiceMonitors

### LINKED ISSUE #2408: [opentelemetry-kube-stack] Built-in ServiceMonitors use deprecated file-based credentials, breaking under deny-by-default collectors
If denyFSAccessThroughSMs https://github.com/open-telemetry/opentelemetry-operator/blob/main/docs/api/targetallocators.md#targetallocatorspecprometheuscr is set true on a targetallocator spec.prometheuscr and a ServiceMonitor like [kube-api-server](https://github.com/cabrinha/opentelemetry-helm-charts/blob/main/charts/opentelemetry-kube-stack/examples/prometheus-otel/rendered/exporters/kube-api-server/servicemonitor.yaml#L17C5-L17C73) specifies a file like  `bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token` then the target allocator drops the scrape config like so:

`
[pod/x-targetallocator-54695b6d9ck45/ta-container] {"level":"warn","ts":"2026-09-15T18:32:45Z","logger":"setup.prometheus-cr-watcher","caller":"watcher/promOperator.go:687","msg":"dropping scrape config that references arbitrary file path","job":"serviceMonitor/x/x-apiserver/0","reason":"authorization.credentials_file: /var/run/secrets/kubernetes.io/serviceaccount/token"}
`

See the kube-prometheus-stack equivalent issue https://github.com/prometheus-community/helm-charts/issues/7230 (which the title of this issue is copied from). The issue was fixed with the merge of https://github.com/prometheus-community/helm-charts/pull/7238 last week. So fixing this issue is a matter of making the same change here.

I'll prepare a PR. (I'm not sure if there are other ways the exporters here have diverged from kube-prometheus-stack's in the last ~2 years but if so I'll leave those out of scope)

wongma7:
#### Description

Default the control-plane ServiceMonitors, i.e. those in https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-kube-stack/templates/exporters, to use secret-based auth so that they work with targetallocators whose spec.prometheuscr.denyFSAccessThroughSMs are set true. This depends on secrets existing in the first place so they can be created by the new `kubernetesServiceMonitors.authorization` fields: setting kubernetesServiceMonitors.authorization.create true results in serviceaccounts + corresponding secrets being created in the default & kube-system namespaces so that the servicemonitors can reference them in their spec.endpoints[x].authorization. 

```
  authorization:
    ## Create a dedicated service account and long-lived service account token Secret used only
    ## to authenticate control-plane metrics scrapes.
    create: true
    ## Override the generated service account or Secret names.
    serviceAccountName: ""
    secretName: ""
```

This PR is based almost entirely on the kube-prometheus-stack change https://github.com/prometheus-community/helm-charts/pull/7238, credit goes to [jkroepke](https://github.com/jkroepke). (Since the ServiceMonitors originate from there and the opentelemetry-kube-stack values.yaml states `This configuration sections allows for a direct replacement of the kube-prometheus-stack chart where the collector scrapes the same metrics as the default prometheus installation.` https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-kube-stack/values.yaml#L1119  I took it as a goal to mirror the kube-prometheus-stack change as closely as possible.) However, this PR necessarily diverges from the kube-prometheus-stack change since they can rely on a 'prometheus' serviceaccount but we can't so we have to create a new serviceaccount.

(Alternative approach (but still breaking): instead of creating a new serviceaccount, we could leverage/reuse the serviceaccount that the operator would create for a collector. But this would require the installer to specify that collector and couple the servicemonitor with that collector's serviceaccount. So I rejected this alternative: if the servicemonitor must reference a particular serviceaccount secret, it's easier/safer to just create the secret ourselves in the chart. A side benefit of this is we can scope the serviceaccount's clusterrole+clusterrolebinding down to getting /metrics)

##### Testing
I deployed opentelemetry-kube-stack with kubeApiServer, kubeControllerManager, coreDns, kubEtcd, kubesSheduler, and kubeProxy enabled: true and
```
  kubernetesServiceMonitors:
    enabled: true
    authorization:
      create: true
```
I deployed a collector with this config
```
  targetAllocator:
    enabled: true
    replicas: 1
    allocationStrategy: consistent-hashing
    mtls:
      enabled: true
      useCertManager: true
    prometheusCR:
      enabled: true
      denyFSAccessThroughSMs: true
```

and the collector successfully scrapes all of them

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-helm-charts/issues/2408

#### Authorship


TylerHelmuth:
I merged an older PR first, please handle the merge conflicts.

wongma7:
rebased :+1:

charts/opentelemetry-kube-stack/examples/prometheus-otel/rendered/collector.yaml
```diff
@@ -93,6 +93,8 @@ spec:
     allocationStrategy: per-node
     enabled: true
     image: ghcr.io/open-telemetry/opentelemetry-operator/target-allocator:main
+    mtls:
```
marcschaeferger: default enabled?

charts/opentelemetry-kube-stack/examples/prometheus-otel/values.yaml
```diff
@@ -11,6 +11,8 @@ collectors:
       enabled: true
       image: ghcr.io/open-telemetry/opentelemetry-operator/target-allocator:main
       allocationStrategy: per-node
+      mtls:
```
marcschaeferger, line 14: I think this should be made explicit.

useCertManager defaults to true, so enabling Target Allocator mTLS here makes the TA mTLS setup depend on cert-manager. The kube-stack itself can currently be installed without cert-manager by using the Helm-generated webhook certificates, but that configuration would no longer be sufficient for this example.

The PR description also says this was tested explicitly with useCertManager: true, while the example relies on the CRD default.

Could we either set useCertManager: true explicitly here and document the requirement, or make the example work with the supported non-cert-manager setup as well?

charts/opentelemetry-kube-stack/values.schema.json
```diff
@@ -3899,13 +3899,16 @@
     },
     "TLSConfig": {
       "type": "object",
-      "additionalProperties": false,
+      "additionalProperties": true,
```
marcschaeferger, line 3902: I’m concerned about making this TLS configuration open-ended via additionalProperties: true.

Besides weakening typo validation, this also allows fields outside SafeTLSConfig to pass the values schema. In particular, ServiceMonitor endpoints support the full TLSConfig, including caFile, certFile, and keyFile.

Since this PR specifically moves these ServiceMonitors away from filesystem references so they remain usable with denyFSAccessThroughSMs: true, I think the values schema should model SafeTLSConfig explicitly and prevent the file-based fields from being reintroduced through this passthrough.

Could we keep this strict and define the supported ca, cert, keySecret, serverName, insecureSkipVerify, minVersion, and maxVersion fields instead?

charts/opentelemetry-kube-stack/values.schema.json
```diff
@@ -4147,17 +4150,22 @@
         "additionalLabels": {
           "$ref": "#/$defs/AdditionalLabels"
         },
-        "https": {
+        "authorization": {
           "anyOf": [
             {
-              "type": "boolean"
+              "type": "object",
+              "additionalProperties": true
```
marcschaeferger, line 4157: Same concern here: allowing arbitrary properties means the values schema no longer validates whether an authorization object is actually a valid SafeAuthorization.

Could we add a reusable schema definition for SafeAuthorization / SecretKeySelector and reference that here instead?

I would prefer keeping schema validation strict so configuration errors are caught during helm install / helm upgrade rather than after rendering.

charts/opentelemetry-kube-stack/templates/exporters/kubernetes-metrics-authorization.yaml
```diff
@@ -0,0 +1,58 @@
+{{- if and .Values.kubernetesServiceMonitors.enabled .Values.kubernetesServiceMonitors.authorization.create }}
+{{- $root := . }}
+{{- $namespaces := list (include "opentelemetry-kube-stack.namespace" .) }}
+{{- if .Values.kubernetesServiceMonitors.ignoreNamespaceSelectors }}
```
marcschaeferger, line 4: There are two namespace-related details.

When ignoreNamespaceSelectors is enabled, these chart-managed ServiceMonitors are created in default and kube-system, but $namespaces still starts with the release namespace. For a release installed into another namespace, this therefore creates an additional long-lived ServiceAccount token and binds that ServiceAccount to the /metrics ClusterRole even though none of these ServiceMonitors reference that credential.

Also, the Target Allocator can initially resolve Secrets outside prometheusCR.secretNamespaces through a direct API GET, but secretNamespaces controls which Secret namespaces are watched for changes and defaults to the Target Allocator’s own namespace. With the credentials living in default / kube-system, updates to those Secrets may therefore not trigger configuration regeneration unless those namespaces are watched.

Could we limit credential creation to the namespaces that actually contain these ServiceMonitors and either configure or document the corresponding secretNamespaces requirements for the Target Allocator?

charts/opentelemetry-kube-stack/examples/prometheus-otel/values.yaml
```diff
@@ -11,6 +11,8 @@ collectors:
       enabled: true
       image: ghcr.io/open-telemetry/opentelemetry-operator/target-allocator:main
       allocationStrategy: per-node
```
marcschaeferger, line 13: The PR description says this was tested with allocationStrategy: consistent-hashing, while this example still uses per-node.

The Target Allocator documentation explicitly notes that per-node leaves targets without a Node assignment, including control-plane components, unassigned unless a fallback strategy is configured.

Since this example enables the control-plane ServiceMonitors, should we either use consistent-hashing here as in the tested configuration or configure a fallback strategy for per-node?

charts/opentelemetry-kube-stack/values.yaml
```diff
@@ -1125,14 +1125,24 @@ opAMPBridge:
...
 
 ## Flag to disable all the kubernetes component scrapers
 ##
 kubernetesServiceMonitors:
   enabled: false
   ignoreNamespaceSelectors: false
+  authorization:
+    ## Create a dedicated service account and long-lived service account token Secret used only
+    ## to authenticate Kubernetes component metrics scrapes.
+    create: true
```
marcschaeferger, line 1142: With create: false, the component defaults still reference the chart-generated token Secret name, while this template no longer creates that Secret.

Could we fail rendering when the chart-generated default authorization is still being used while create is false, similar to the safeguard in prometheus-community/helm-charts#7238?

A user who manages the credential externally could still provide their own authorization, set it to null, or explicitly configure the expected Secret. This would mainly prevent accidentally rendering ServiceMonitors that reference a credential the chart is no longer managing.

charts/opentelemetry-kube-stack/values.yaml
```diff
@@ -1141,7 +1151,16 @@ kubeApiServer:
   tlsConfig:
     serverName: kubernetes
     insecureSkipVerify: false
+    ca:
+      configMap:
```
marcschaeferger, line 1155: The new ca.configMap references appear to require an RBAC change for the Target Allocator.

During ServiceMonitor selection, the Prometheus Operator config generator resolves tlsConfig references through AddTLSConfig, which reads the referenced ConfigMap from the ServiceMonitor namespace.

The kube-stack ClusterRole currently grants the Collector/Target Allocator access to Secrets, but I don’t see configmaps in its core resource permissions. The Target Allocator documentation also lists get on ConfigMaps as required RBAC for this case.

Unless I’m missing another binding that provides this permission, won’t ServiceMonitors using the new kube-root-ca.crt reference fail validation with a forbidden ConfigMap GET?

Could we add the required ConfigMap RBAC, ideally only get if that is sufficient?

charts/opentelemetry-kube-stack/values.schema.json
```diff
@@ -4147,17 +4150,22 @@
...
             },
             {
               "type": "null"
             }
           ]
         },
-        "insecureSkipVerify": {
+        "tlsConfig": {
```
marcschaeferger, line 4164: I think KubeServiceMonitor.tlsConfig needs the same strictness as the kube-apiserver TLS configuration.

With additionalProperties: true, values such as caFile, certFile, or keyFile are accepted by the chart and rendered directly into the ServiceMonitor. Those are valid fields of the ServiceMonitor endpoint’s full TLSConfig, but they are exactly the filesystem references that denyFSAccessThroughSMs: true rejects.

Could this reference BE a strict reusable SafeTLSConfig schema instead?

charts/opentelemetry-kube-stack/templates/exporters/kubernetes-metrics-authorization.yaml
marcschaeferger: When ignoreNamespaceSelectors is true, all of these chart-managed ServiceMonitors are created either in default or kube-system, but $namespaces still starts with the release namespace.

That means a release installed into another namespace creates an additional long-lived ServiceAccount token there and binds that ServiceAccount to the /metrics ClusterRole even though none of these ServiceMonitors reference that credential.

Is there a reason we need the additional release-namespace credential? If not, could we limit token creation to the namespaces that actually contain the ServiceMonitors?

marcschaeferger:
Waiting for Maintainer Feedback, if they agree

charts/opentelemetry-kube-stack/examples/prometheus-otel/values.yaml
wongma7: yes, I've set it true and updated the README to explain that cert-manager is required for mtls between target allocator & collector. It occurred to me while updating it that we don't need kube-dns, kube-etcd, kube-proxy to have any authorization to be configured, & we don't have to copy/cargo-cult that from kube-prometheus-stack, so I nulled them all and updated the example renders accordingly . They might in the future have configuration to serve metrics over authenticated endpoint but there's no point in configuring them to use the bearer token now when it's unnecessary.

charts/opentelemetry-kube-stack/templates/exporters/kubernetes-metrics-authorization.yaml
wongma7: updated to remove release namespace when ignoreNamespaceSelectors is enabled  and set secretNamespaces https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2411/changes/f1247b97b6ac3f2e4edad9340616ea9422363285#diff-41eeb9e0f2db6ef4cfb9f94ece00404f3a8f6a02be7c556b63c24761de42253aR501

charts/opentelemetry-kube-stack/values.schema.json
wongma7: ACK , I agree, additionalProperties is a bit of a copout, I am still working on the schema changes but addressed other comments in the meantime

charts/opentelemetry-kube-stack/values.yaml
wongma7: added and verified that `get` is sufficient. (in my initial manual testing I had an existing permissive role+rolebinding which gave a false negative, I retested)

charts/opentelemetry-kube-stack/values.yaml
wongma7: added a fail in the helper herre https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2411/changes/f1247b97b6ac3f2e4edad9340616ea9422363285#diff-41eeb9e0f2db6ef4cfb9f94ece00404f3a8f6a02be7c556b63c24761de42253aR501

charts/opentelemetry-kube-stack/examples/prometheus-otel/values.yaml
wongma7: you & docs are right, I configured a fallback strategy here now https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2411/commits/7bb0a183af5f560d23e7d331916f946b5c63de29

charts/opentelemetry-kube-stack/values.schema.json
wongma7: added here https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2411/changes/c549ad5e318497506d8cec98328c152aaf40fe44#diff-b9c6094076d90b22e4611968a13f2837d837496558a253bb441a253a23d2fb6aR3944

charts/opentelemetry-kube-stack/values.schema.json
wongma7: added here, https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2411/changes#diff-b9c6094076d90b22e4611968a13f2837d837496558a253bb441a253a23d2fb6aR3944 agree to be strict and use SafeTLSConfig even if technically the CR could accept TLSConfig https://prometheus-operator.dev/docs/api-reference/api/#monitoring.coreos.com/v1.Endpoint

charts/opentelemetry-kube-stack/values.schema.json
wongma7: updated to reuse SafeTLSConfig https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2411/changes#diff-b9c6094076d90b22e4611968a13f2837d837496558a253bb441a253a23d2fb6aR4228

PR #2412, opened 2026-09-16
Adopt Config v2 defaults for OBI chart

### LINKED ISSUE #2410: [opentelemetry-ebpf-instrumentation] Chart doesn't support v2 config
### Versions

**Chart version:** 0.13.1
**OBI (app) version:** 0.13.0

## Description

When setting `.Values.config.data` to a Config v2-formatted config, OBI fails to start because of a `discovery` block injected by [`_helpers.tpl`](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/4c5662db2db7b349b473cc7ceaf2cd1f868df437/charts/opentelemetry-ebpf-instrumentation/templates/_helpers.tpl#L154-L159), which uses a Config v1 field (`discovery`) that Config v2's stricter validation rejects as an unknown key.

## Cause 1 - the `discovery` injection has no opt-out under Config v2

`.Values.preset` defaults to `application`. When `preset == "application"` and `.Values.config.data.discovery` is unset, the helper injects:

```yaml
discovery:
  instrument:
  - k8s_namespace: '*'
  exclude_instrument:
  - exe_path: '{*ebpf-instrument*,*otelcol*}'
```

This is Config v1 syntax and has no Config v2 equivalent (v2 replaces this with `capture.policy`/`capture.rules`). Since `discovery` is v1-only, there's no value you can set under that key that both satisfies the `if not .Values.config.data.discovery` check and passes v2 validation — the field itself shouldn't exist in a v2 config at all.

The only way to skip the injection is to set `preset` to something other than `application`, but `preset` is constrained by `values.schema.json` to a fixed set of options (`application`/`network`), so it can't be nulled, emptied, or set to a no-op value. And switching to `preset: network` isn't a real workaround — it changes unrelated behavior (`hostNetwork: true`, `dnsPolicy`) and, per the same helper, has its own conditional injection for `config.data.network` that likely has the same v1/v2 mismatch.

More broadly, `preset` being a mutually-exclusive single-select is a limitation on its own: there's no way to get the combined defaults of both `application` and `network` presets at once, since picking one forecloses the other's injected block. Given how differently they affect deployment (host networking, DNS policy, config injection), this probably deserves its own design discussion beyond the v2 bug.

## Cause 2 - the chart's default `config.data` in `values.yaml` is also Config v1-shaped

The chart's own [default `config.data`](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/c62d35c6c687ea1d682308e833c28ad7e5f2244d/charts/opentelemetry-ebpf-instrumentation/values.yaml#L211) is written in Config v1 format. Unlike the injected `discovery` block, these defaults *can* be suppressed (by nulling each default key in an overriding values file), but that's an easy trap: unless a user explicitly nulls out every default key, the v1-shaped defaults merge into a v2 config and cause the same class of validation failure.

## Suggested fix

- Make the chart aware of which config schema version is in use (e.g. via an explicit `config.version` value, or by detecting v2-shaped keys already present in `config.data`), and skip both the `discovery` injection and the v1-shaped `values.yaml` defaults when v2 is in use.
- Longer term, reconsider `preset` as a single mutually-exclusive enum — allowing it to be unset/`none`, or allowing multiple presets to combine, would avoid forcing users into a specific config schema version as a side effect of an unrelated feature toggle.

cc @MrAlias

MrAlias:
- Make chart-managed configurations use OBI Config v2 while preserving automatic compatibility for existing Config v1 overrides.
- Derive Kubernetes resources from the effective schema, document the migration contract, and release the behavior as chart version 0.14.0.

Fixes #2410

#### Authorship


charts/opentelemetry-ebpf-instrumentation/UPGRADING.md
```diff
@@ -7,3 +7,24 @@ Just upgrade and enjoy.
 ## 0.9.3 to 0.9.4
 
 The `tpl` function has been added to `.Values.config.data`. If you are currently using any `{{ }}` syntax in `.Values.config.data` it will now be rendered. To escape existing instances of `{{ }}`, use ``` {{` <original content> `}} ```. For example, `{{ REDACTED_EMAIL }}` becomes ``` {{` {{ REDACTED_EMAIL }} `}} ```.
+
+## 0.13.1 to 0.14.0
```
TylerHelmuth, line 11: Add this to the top so the latest breaking change is first.

charts/opentelemetry-ebpf-instrumentation/templates/_helpers.tpl
```diff
@@ -140,21 +140,267 @@ app.kubernetes.io/instance: {{ .Release.Name }}
...
+            kubernetes:
+              namespace_glob:
+                - '*'
 {{- end }}
-{{- if eq .Values.preset "application" }}
-{{- if not .Values.config.data.discovery }}
-{{- $discovery := dict "instrument" (list (dict "k8s_namespace" "*")) "exclude_instrument" (list (dict "exe_path" "{*ebpf-instrument*,*otelcol*}")) }}
-{{- $_ := set $config "discovery" $discovery }}
+
+{{/* Remove keys explicitly set to null in user overrides. */}}
```
TylerHelmuth, line 288: Is this a workaround for the helm subchart null bug?

charts/opentelemetry-ebpf-instrumentation/UPGRADING.md
MrAlias: Moved the 0.13.1 to 0.14.0 section above the older 0.9.3 to 0.9.4 note in 29b0948d so the newest breaking change appears first.

charts/opentelemetry-ebpf-instrumentation/templates/_helpers.tpl
MrAlias: No, this is not specific to subcharts. The v1 and v2 defaults are generated inside the template, so Helm's values coalescing cannot remove those generated keys. After `mergeOverwrite`, `obi.removeNulls` recursively unsets keys the user explicitly set to `null`. The ConfigMap tests cover top-level and nested v2 defaults as well as v1 defaults.

MrAlias:
CI failure:

> Save error occurred:  could not download https://github.com/prometheus-community/helm-charts/releases/download/prometheus-29.30.2/prometheus-29.30.2.tgz: failed to fetch https://github.com/prometheus-community/helm-charts/releases/download/prometheus-29.30.2/prometheus-29.30.2.tgz : 504 Gateway Time-out

TylerHelmuth (approved):

PR #2415, opened 2026-09-17
fix(opentelemetry-operator): bind the metrics server on all interfaces

### LINKED ISSUE #2414: opentelemetry-operator: metrics server binds 0.0.0.0 with no value to override it, so it is unreachable on IPv6 clusters
**Chart:** opentelemetry-operator 0.123.0 (appVersion 0.159.0)

The manager's metrics bind address is hardcoded to IPv4 in
`templates/deployment.yaml`:

```
- --metrics-addr=0.0.0.0:{{ .Values.manager.ports.metricsPort }}
```

`values.yaml` only exposes the port, so on an IPv6 or dual-stack cluster the
metrics server ends up listening where nothing can reach it, and there is no
way to fix it from values.

It took me a while to spot, because nothing complains. `--health-probe-addr`
is `:8081` and already dual-stack, so the pod stays Ready. The only symptom is
a scrape target that never comes up, timing out rather than returning a 401 --
which is what told me nothing was listening at all:

```
Get "https://[2a01:...:63c9]:8443/metrics": net/http: request canceled
while waiting for connection (Client.Timeout exceeded while awaiting headers)
```

So with `manager.serviceMonitor.enabled: true` you silently get no operator
metrics.

I think the fix is to just drop the host, not to add a value for it. The two
flags right next to it already do that (`--health-probe-addr=:{{ ... }}`,
`--pprof-addr=:{{ ... }}`), the operator's own default is `":8443"`
(`internal/config/config.go`), and its kustomize deployment passes no
`--metrics-addr` at all -- the chart is the only place that narrows it. Looks
like a leftover from #253, when metrics was plain HTTP behind kube-rbac-proxy.

An empty host is a superset of `0.0.0.0` rather than a swap of one family for
the other, so nobody loses anything: Go binds the IPv6 wildcard with
`IPV6_V6ONLY` off and accepts both families, and on a host with no IPv6 it
falls back to `AF_INET`. `[::]` would not -- that one is explicitly AF_INET6
and fails to listen there.

Happy to open a PR -- I have the one-line fix ready.

Lillecarl:
#### Description

The manager's metrics bind address is hardcoded to IPv4, and `values.yaml`
only exposes the port:

```
- --metrics-addr=0.0.0.0:{{ .Values.manager.ports.metricsPort }}
```

On an IPv6 or dual-stack cluster the metrics server then listens where nothing
can reach it, and nothing says so -- `--health-probe-addr` is `:8081`, so the
pod stays Ready, and the scrape target just times out instead of returning a
401.

This drops the host rather than adding a value for it. The two flags beside it
already do that, the operator's own default is `":8443"`
(`internal/config/config.go`), and its kustomize deployment passes no
`--metrics-addr` at all -- the chart is the only place that narrows it.

An empty host is a superset of `0.0.0.0`, not a swap of one family for the
other: Go binds the IPv6 wildcard with `IPV6_V6ONLY` off and accepts both, and
on a host with no IPv6 it falls back to `AF_INET`. `[::]` would not.

Verified on a dual-stack cluster: the endpoint now answers 401 in 0.4 ms over
IPv6 and the ServiceMonitor target reports up.

Second commit is the chart bump and `make generate-examples`, kept separate so
the fix stays one line.

#### Link to tracking issue
Fixes #2414

#### Authorship

- [ ] I, a human, wrote this pull request description myself.
- [x] I, an LLM, wrote this pull request description myself.


TylerHelmuth:
>  I, an LLM, wrote this pull request description myself.

[Your PR is failing to meet our communication guidelines](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/AGENTS.md).

PR #2418, opened 2026-09-19
fix(operator): remove redundant schema requirements

### LINKED ISSUE #1852: Remove `required` from `values.schema.json`
You have a lot of `required` in your `values.schema.json`:

https://github.com/open-telemetry/opentelemetry-helm-charts/blob/9ef9caaaf70a25c3b1d3c39421f4e88653d0334b/charts/opentelemetry-operator/values.schema.json#L7-L26

This can be seen as a nice-to-have for extra validation, but it makes it more difficult to use the schema file in editors.

If I create a new `values.yaml` file with just this:

```yaml
# yaml-language-server: $schema=https://github.com/open-telemetry/opentelemetry-helm-charts/raw/refs/heads/main/charts/opentelemetry-operator/values.schema.json

enabled: true
```

Then my editor screams at me that I'm missing the other fields:

Image

Editors (using the <https://github.com/redhat-developer/yaml-language-server> LSP) are not able to understand that the values are merged with the chart's `values.yaml`. This creates poor user experience.

Helm does understand this, so any `helm` commands work perfectly. It's just editors that don't.

When it comes to writing values schema files for Helm charts, is is generally an anti-pattern to use the `required` keyword because of this, as your Helm chart's templates will try to use these fields anyways. So removing the `required` keywords does not really hurt the validation done.

There's one edge-case where `required` does make sense, and that's inside arrays. Because when Helm merges arrays it overrides the entire array. So in that case there's no "prior knowledge" that the editor/LSP needs to know about.

As a general rule, I advise not using `required` at all, and instead enforce these rules where it's really necessary in your chart templates using the [`required`](https://helm.sh/docs/chart_template_guide/function_list/#required) template function.

szelese:
#### Description

- Remove object-level `required` constraints from the operator chart schema where values are provided by chart defaults.
- Retain array-item requirements for `metricsEndpoints.port` and `testFramework.certManager.env.name`.
- Bump the chart version from `0.123.0` to `0.123.1`.
- Regenerate the operator examples.

#### Link to tracking issue
Fixes #1852 

#### Validation

- [x] `helm lint charts/opentelemetry-operator`
- [x] `helm template` with default and minimal values
- [x] `make check-examples CHARTS=opentelemetry-operator`
- [x] `pre-commit run --all-files`
- [x] `git diff --check`

#### Authorship


JaredTan95 (approved):
thanks

charts/opentelemetry-operator/values.schema.json
```diff
@@ -4,25 +4,6 @@
...
-    "crds",
-    "role",
-    "clusterRole",
-    "hostNetwork",
-    "priorityClassName",
-    "securityContext",
-    "testFramework",
-    "additionalLabels",
-    "automountServiceAccountToken"
-  ],
```
TylerHelmuth, line 25: I believe all of these have default values in the values.yaml, so json schema should be content. If a user nulled one of these values during an install I believe the render would fail.

What problem are they causing?

szelese: Thanks Tyler! I verified this. Minimal override values pass Helm lint/template, but the root-level required entries cause false-positive missing-property warnings in editors. I kept required for the fields that fail rendering when set to null, removed it for the rest, and all local/CI checks pass.

TylerHelmuth: So the issue is in an editor? With the current state of this PR if someone used the operator as a subchart and nulled `manager.image` they'd get an unhelpful error message. The json schema automatically handles these situations so we don't need a bunch of use cases in the NOTES.txt.

TylerHelmuth: What IDE is struggling with the json schema? This is the second PR I've seen that has tried to remove required fields because of their IDE

szelese: Thanks, Tyler. That makes sense. I audited the null-value cases and restored required only where null values break rendering. helm lint now catches them, and all checks pass.

szelese: The issue was with the Red Hat YAML extension for VS Code (redhat.vscode-yaml).

TylerHelmuth: Strange I don't see any errors with that extension in my vscode locally

charts/opentelemetry-operator/values.schema.json
```diff
@@ -1471,22 +1340,8 @@
       "default": {},
       "title": "The admissionWebhooks Schema",
       "required": [
-        "create",
-        "failurePolicy",
-        "secretName",
         "pods",
-        "namePrefix",
```
TylerHelmuth, line 1478: This value also cannot be excluded. I believe it is the last required field

szelese: Good catch. I restored admissionWebhooks.namePrefix. All checks pass.

TylerHelmuth (approved):

szelese:
Thank you for your guidance, Tyler.

